Issue #7, opened 2016-12-03
env: node\r: No such file or directory


pacuna:
I'm getting this error on MacOs sierra. I found some similar issues with other packages and apparently it has to do with line ending format?

TomatoMus:
I can solve it in the following way.
```
$vim /your/oni/path/oni

:set ff=unix
:w
```

pacuna:
I solved by building it myself.

extr0py:
Thanks @TomatoMus  and @pacuna  for the details!

Ya, looks related to the issue with npm/npm#12371. I've been doing the publish from a Windows machine, which hits this issue. Hopefully the fix to convert the line endings prior to publish will address this.

Cloning and building will work fine, since git auto-converts the line endings when you check out. Glad that works ok :)

PR #8, opened 2016-12-03
Add 'prepublish' script in package.json, to convert to EOF line endings

extr0py:
Fix for #7 in future publishes. I usually publish from a windows machine, and it looks like the windows-style line endings cause problems. This will set the line endings to just LF prior to publishing.


extr0py:
Will republish in a bit after a few more fixes

PR #25, opened 2016-12-04
Stricter tsc settings, add editorconfig, add tslint, etc.

freebroccolo:
Changes introduced by this PR:

* stricter typescript settings (2.0)
* add editorconfig file
* add tslint file
* fix all tslint issues
* update dependencies
* update .gitignore for osx, etc.

For the stricter typescript settings, I've enabled strict `null` checking, disallowed implicit `any`, etc. I tried to update the code to account for this without changing it too much. In several cases, that meant just using a cast for now because some algorithms will need to be updated to account for the possibility of `null` values.


extr0py:
Thanks for doing this cleanup work!

extr0py:
Will be really helpful as people contribute to help keep a more consistent standard, and catch more errors at compile time

PR #26, opened 2016-12-04
Added the Oni logo to the README

bert88sta:
The file is named Oni_2 because it is 2000x2000 in size.


extr0py:
Thanks for the PR! Would it be possible to shrink the image a bit? I'm thinking maybe 128x128 or 256x256. It looks great but a bit overpowering at the moment ;)

bert88sta:
Haha no problem, I didn't know what size would be appropriate. I'm going to include a 128x128, as well as one in a .icns file so that mac will handle it properly if you decide to add it as the icon of the program.

PR #27, opened 2016-12-04
Added Oni Logo

bert88sta:
Inside, there is an Oni logo png, 256x256, and a .icns (apple format if you want to change the actual runtime icon to it) as well as a 2000x2000 full res logo for other people to polish up


extr0py:
Awesome, thanks @bert88sta!

PR #37, opened 2016-12-06
Added default font configuration for Linux.

JakubJecminek:
Hello, I've added default font setting for linux. I chose DejaVu Sans Mono as it is usually already installed on most of the linux distributions. I've tested this solution on my machine which is running Fedora Workstation 25.


extr0py:
Awesome, this looks great! Thanks for adding this, I had no idea which font would be a good default on Linux.

Issue #36, opened 2016-12-06
Cannot figure out how to install on OSX


meain:
Need a bit of help installing it on macOS. I installed the npm package but it won't start.
What do you mean by `ensure the 'nvim.exe' is available`? I do have neovim installed via homebrew.

bert88sta:
Uh, just mean you have to have a neovim executable available on MacOS. What error do you get when you type 'oni' in the shell?

tbodt:
I've run into the fact that the oni executable has Windows (CRLF) line endings, which screws up the shebang line and gives the error `env: node\r: no such file or directory`. I opened it in Neovim, did `:set fileformat=unix` and saved, and that was fixed.

extr0py:
Bummer - thought this was fixed in the last npmjs update. I'll revisit it, it seems like publishing from a Windows machine is problematic.  Thanks for the details

We might need a .gitattributes file to have git ignore the line ending conversion for this file.

Gajit:
I'm having similar problems, the install looks like it's erroring out at 
> npm run install:oni-plugin-typescript.

extr0py:
@Gajit - might be #19 also. I'll take a look at that one tomorrow, looks like quite a few people are hitting that

extr0py:
I published a new version to NPM - 0.1.2.

I believe the line-endings issue should be fixed now. I downloaded the tarball here:
https://registry.npmjs.org/oni-vim/-/oni-vim-0.1.2.tgz

And validated that the line endings are LF in the cli/oni file. Hopefully that means users won't need the :set fileformat=unix workaround. Let me know though if that is still showing up!

Gajit:
Ok I was able to npm install now but when Oni starts up it just gives a blank screen. Console errors with a Failed to load resources on
> file:///usr/local/lib/node_modules/oni-vim/dist/browser/bundle.js Failed to load resource: net::ERR_FILE_NOT_FOUND

Issue #19, opened 2016-12-04
npm run install:oni-plugin-typescript fails


olleharstedt:
0 info it worked if it ends with ok
1 verbose cli [ '/usr/bin/nodejs',
1 verbose cli   '/usr/bin/npm',
1 verbose cli   'run',
1 verbose cli   'install:oni-plugin-typescript' ]
2 info using npm@3.5.2
3 info using node@v4.2.6
4 verbose config Skipping project config: /home/olle/.npmrc. (matches userconfig)
5 verbose stack Error: ENOENT: no such file or directory, open '/home/olle/package.json'
5 verbose stack     at Error (native)
6 verbose cwd /home/olle
7 error Linux 4.4.0-47-generic
8 error argv "/usr/bin/nodejs" "/usr/bin/npm" "run" "install:oni-plugin-typescript"
9 error node v4.2.6
10 error npm  v3.5.2
11 error path /home/olle/package.json
12 error code ENOENT
13 error errno -2
14 error syscall open
15 error enoent ENOENT: no such file or directory, open '/home/olle/package.json'
16 error enoent ENOENT: no such file or directory, open '/home/olle/package.json'
16 error enoent This is most likely not a problem with npm itself
16 error enoent and is related to npm not being able to find a file.
17 verbose exit [ -2, true ]

olleharstedt:
I get this error when running `npm install -g oni-vim`

extr0py:
Interesting, I haven't seen this - and it looks like the npm install step is working okay on TravisCI (which is Linux). 

If you clone & build the repo, does it work?

olleharstedt:
Yes, I could build it manually.

erahhal:
I get the same problem on Ubuntu 16.04, but even using sudo:

`sudo npm install -g oni-vim`

For some reason a global install is trying to write to ~/.npm in my home folder, with the following error:

`npm ERR! Error: EACCES: permission denied, mkdir '/home/erahhal/.npm/oni-plugin-typescript'`
 
There's definitely something wrong here.

hxegon:
Getting the same error as @erahhal. A little more info: it's trying to open $HOME/package.json when I try and run ```sudo npm run install:oni-plugin-typescript```.

jetz:
Same problem on Ubuntu 16.04. ` Missing write access to /usr/local/lib/node_modules ...`

[...]

PR #40, opened 2016-12-07
Move dependencies up to root so that postinstall task is no longer needed

extr0py:
Should address #19, or at least allow the npm install to get further on Linux.

#36, the line-ending issue, may still be present though


extr0py:
Still need to upload this to npm for npm install -g to (hopefully) work better

Issue #16, opened 2016-12-04
init.vim not sourced


extr0py:
Currently, the default init.vim is overwritten, to set the runtimepaths to pick up the plugins that are needed for Oni to function correctly.

However, this means settings aren't pulled in for custom init.vim

Need to source the user's init.vim, if it exists, once the app has started. Seems like we might want to add this in (NeovimInstance.ts)[https://github.com/extr0py/oni/blob/master/browser/src/NeovimInstance.ts] once the process has been spawned, we can source the init.vim if it exists.

darkwater:
Couldn't the extra runtimepaths be added through a command line option? `--cmd <command>` is executed before processing any vimrc, so this looks like a good way to set rtp.

darkwater:
Applying http://novaember.com/s/8f9453/ig79Ko.txt will append ONI's runtimepaths to the standard, but I haven't tested for any side-effects. (And of course the code is ugly because I just changed some things to make it work.)

 I also noticed a rather long startup time, not sure what that's caused by.

extr0py:
Thanks for the code snippet! I ported it over to a PR - #18. That's a great suggestion, I didn't know about the --cmd option, would be more robust than dynamically generating the init.vim.

I'm seeing a problem too though with the PR - it looks like Neovim is hitting an error that isn't getting rendered, so you need a keypress to activate the UI. (This might explain the long startup time in your case, too). Unfortunately, I'm not sure what that error is, so need to dig in a little more and see what's happening before bringing that PR in. Sometimes, if there's a script error on load, it doesn't make it to the UI - seems related to this issue in Neovim core: https://github.com/neovim/neovim/issues/3901

extr0py:
I believe this should be fixed by PR #18 - try out the `oni.useDefaultConfig` option and let me know.

Thanks!

PR #43, opened 2016-12-07
Extr0py/16 use init vim

extr0py:
- WIP for #16 

- Use runtime paths instead of generating init.vim
- PENDING: Add option to use own config instead of defaults


extr0py:
@darkwater - I think this is getting closer now. Your suggestion to use runtime paths instead of overriding init.vim is implemented, so I think that should be helpful. 

In addition, there's also a configuration option 'oni.useDefaultConfig' which defaults to true, but since the chosen defaults might conflict, this prevents those from being added to the runtimepath.

extr0py:
Remaining work is to fix outstanding build issues and add an entry in the readme / faq documenting this option.

jordwalke:
Looking forward to this! I've curated a very specific [vim config](https://github.com/jordwalke/VimBox) to create a modern editor feel and I'm looking forward to trying it with `oni`.

jordwalke:
I like the idea of either using everything out of the box, or a completely from-scratch config. The odds of my plugins conflicting with the built in ones is so great, that it's probably better to just have them be separate (at least to start with).
When I try this pull request, I get an error in the console (missing `init_template.vim` or something).

extr0py:
@jordwalke - the config looks really cool! 

> I like the idea of either using everything out of the box, or a completely from-scratch config. 

Yes, same here. I want to make sure the out-of-box experience is on point, and that the first time you launch the editor it _just feels good_. The first time I started VIM it felt like it was a wild beast that I would need to tame or something.... whereas installing and launching Atom was a very pleasant experience. Hopefully that makes drawing new people to modal editing easier...

And then of course it's important for VIM veterans to be able to use their finely-honed and crafted config with no friction, and still experience some of the UI benefits.

extr0py:
Make sure to set `oni.useDefaultConfig` to false in your config.json to try this out :) 

If you see that the screen 'sticks' and waits for keypress on launch, that's an indication that something went wrong while loading init.vim - unfortunately Neovim doesn't bubble up the errors on launch, at the moment. Keep me posted if you see any issues like that.

jordwalke:
Yes I like the idea of an out of box awesome experience even if people quickly disable it to get their own config. It sets the bar and shows people "yes you can create an amazing, modern experience with Vim".  They will poke around into how the out of box experience was achieved and start moving the best parts to their own config.

PR #45, opened 2016-12-08
Vertical Scrolling Is Here! (v2)

bert88sta:
Vertical scrolling with modifier keys is handled. I will do more work on horizontal later, but this is an essential part of an editor.

### DIFF
--- a/browser/src/Input/Mouse.ts
+++ b/browser/src/Input/Mouse.ts
@@ -81,7 +81,7 @@ export class Mouse extends EventEmitter {
         })
     }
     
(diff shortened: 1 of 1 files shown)


bert88sta:
I see extr0py is working on a scrolling optimisation branch, which is good. scrolling causes some matrix-effect level artifacting. I guess it's a new "feature"

extr0py:
Awesome, thanks for the PR @bert88sta ! Looks great!

Ya, regarding the artifacting - it's a 'feature' that the canvas is incrementally rendered, but ideally render would be fast enough that the 'incremental' isn't as necessary :)

Issue #52, opened 2016-12-08
QuickInfo: QuickInfo UI follows the cursor down to command line


extr0py:
The QuickInfo UI follows the cursor to the command line.
![quickinfo-bug](https://cloud.githubusercontent.com/assets/13532591/20998375/2abf83d0-bcc2-11e6-855e-e254ca4c3001.gif)

Seems like, if we go to command mode, we should just clear the quick info UI. The challenge here is to figure out how we detect that case to remove it. We have similar logic in browser\src\index.tsx to close it when the mode changes.

extr0py:
This fix will be enabled once the 'cmdline' mode is enabled downstream in neovim. Thanks @bert88sta !

PR #53, opened 2016-12-08
Added UI closing on commandline mode 

bert88sta:
This only works in the latest version of neovim RPC. which was updated as of 8 days ago. I don't know how long it will be until that code makes its way into an update, but I could add a temporary fix in as well if you would like. I've set it to close autocompletions on the command line, but that could be changed if there is a need for vim-completions

#52

### DIFF
--- a/browser/src/index.tsx
+++ b/browser/src/index.tsx
@@ -136,6 +136,10 @@ const start = (args: string[]) => {
             UI.hideSignatureHelp()
         } else if (newMode === "insert") {
             UI.hideQuickInfo()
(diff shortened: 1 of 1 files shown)


extr0py:
Cool, that's great! If they have that mode added and it will show up in an update, that's cool - not worth us spending a ton of time on a workaround.

I think this PR:
https://github.com/neovim/neovim/pull/5686

Will help a lot with some of theses issues in the future, too... Can't wait.

extr0py:
Thanks for looking at this, @bert88sta !

Issue #50, opened 2016-12-08
Upgrade to TypeScript 2.1


extr0py:
TypeScript 2.1 has some awesome functionality:
https://blogs.msdn.microsoft.com/typescript/2016/12/07/announcing-typescript-2-1/

- Async/await
- Spread operator

etc

PR #65, opened 2016-12-09
Bump ts version

extr0py:
Fix for #50


extr0py:
Still need to take advantage of some of the new features :)

PR #66, opened 2016-12-09
Extr0py/high dpi

### LINKED ISSUE #21: Text rendering blurry on retina.
I noticed that text rendering is blurry on retina screens.

extr0py:
Fix #21 

Text rendering was getting upscaled on high-DPI devices, because the devicePixelRatio wasn't taken into account during rendering. Both the primary canvas and the RenderCache (the one that actually draws the characters, so that they can be re-used as images later for perf) needed to be updated to take this into account. 

The canvasContext also has a concept of a backingStorePixelRatio, but it seems like this might be deprecated: http://stackoverflow.com/questions/24332639/why-context2d-backingstorepixelratio-deprecated (or at least always 1 in Electron), so I did not include it.

One case this won't handle is if the user has multiple displays connected - a high-DPI and a standard-DPI - and moves the window between them. We can track that as a separate issue if it comes up.


extr0py:
Still looks fine on my low DPI monitor too

PR #69, opened 2016-12-09
modify icons to have transparent background

keforbes:
The Oni_Full_Res.png image had a transparent background but the resized images (Oni_128.png and Oni_256.png) did not.  So I modified Oni_128.png and Oni_256.png to have a transparent background like Oni_Full_Res.png.  This gets rid of the ugly white box around the application icon when Oni is running.  I also created Oni_64.png in case we want it.


extr0py:
Nice! Thanks @keforbes , this will look much nicer!

PR #71, opened 2016-12-10
Brought Coordinates back into Mouse, also fixed spacing.

bert88sta:


extr0py:
Thanks for fixing the spacing issue, @bert88sta !

Issue #49, opened 2016-12-08
Feature - Preview: Markdown preview


bryphe:
A feature that would be really helpful for me (in other words, a feature that makes me jump to VSCode...) is live markdown preview.

A few things need to be considered for this:
1) What is the entry point?
2) How do we implement it?

In terms of implementation, it should be relatively straightforward to expose the UI to a plugin as a 'layer'. We could potentially render this 'layer' over a buffer. This way, we could open a separate buffer of filetype=preview, and just render an electron window of the rendered markdown. This allows us to use the window navigation present in VIM. 

Another option would be to just manage the window outside of VIM. The challenge with this is that the window navigation needs to be coordinated with Neovim.

bert88sta:
Are there tools out there that have general case document rendering? Doing it live for markdown might be a bit harder, but having pdf viewers available (I use LaTex for everything),  as well as other types of text that gets converted into some graphic filetype would be extremely useful

extr0py:
Actually, markdown rendering is really easy because there are some great javascript libraries that handle this!

I built a prototype for this in a previous project: https://github.com/extr0py/vim-electrify-markdown-preview

This library in particular is great : https://github.com/chjj/marked and allows options for github flavored markdown. (I also included another library called Mermaid, which allows diagram rendering from text).

The main use case for this feature for me is so that I can see my changes live as I change README.MD/ROADMAP.MD, otherwise I need to round-trip through a pull request to see my changes (or open VSCode), but of those options feel sloppy to me :)

I think LaTex might be a bit harder from a cross-platform perspective, just because I believe the build tools are more of a pain to install in Windows. It would be awesome to have that as another plugin too though - I know a common case for Emacs is to draft and show text in LaTex immediately.

bert88sta:
Not exactly what you're looking for, but if we can integrate a pdf viewer that supports auto-reloading, [this](https://github.com/donRaphaco/neotex) seems pretty sweet

bert88sta:
Well, as per your suggestion, I'm diving in headfirst. Is there an easy way to get an idea for the structure of the project besides combing source? I just want a hierarchy of what's attached to what and how things are subdivided with windows etc.

I think if we've got individual windows (buffers technically, i suppose) we could split one of those or something else, otherwise this will be a real pain

extr0py:
Awesome! Glad you're diving in :)

In terms of this - my hope was we could reuse the existing javascript / typescript plugin infrastructure for this. So basically it would work as follows:
[...]

Issue #61, opened 2016-12-08
Scroll improvements.


bert88sta:
First, the scroll bar jitters from left to right

https://gyazo.com/e886941e230b901f7bb06fd8209d0651

Second, I've tried implementing some math to determine if the scroll is in a different buffer, but it gets laggy. Basically in Mouse.ts I just created variables for the last cell that the mouse event registered to, and then I compared them. Is there a better way to implement this? I like being able to scroll without clicking, but it's not going to be a dealbreaker

extr0py:
> First, the scroll bar jitters from left to right

Unfortunately there's some bugs in `OverlayManager.ts` in terms of how the window position is calculated. Sometimes, it gets off a bit. I think we could use some of the API members in the `msgpack-rpc` API to get the open windows - there's methods to get all the windows, and then their position and size - that would be a much more reliable way than the current strategy. It actually impacts everything that is an _overlay_ - you can even see it in the last animated gif in the readme for live evaluation :( As well as the error messages.

We can also always get the currentBuffer / currentWindow from those same msgpack-rpc APIs (They live in `_neovim` on NeovimInstance - there aren't good typings for it, but you can see some of the members in the debugger -> `window.neovim._neovim` gives some details). Fixing the OverlayManager to give more reliable window dimensions will help across the board, and I anticipate several other pieces of functionality being built on top of it later too (overlaying git blame / git changes, etc)

extr0py:
Regarding needing the current buffer for scrolling without clicking - I think if you just send the scroll commands as input, it should 'just work' as Neovim will apply those to the currently active buffer. Does that not work?

bert88sta:
@extr0py for some reason I removed the coords from the emit, just went and fixed it. Is there anything else you want done it Mouse before a PR comes your way?

extr0py:
@bert88sta -Hmm, I don't think there is any more mouse functionality I had in mind in the short-term. Eventually for some of the enhanced UI components like auto-complete, quick-open, etc it might be helpful to add some click handling, but not really needed for now and it's separate from Mouse.ts - that would be handled in the components.

extr0py:
Jitteriness is fixed by #75 - if there any issues we want to track, feel free to log them as separate items!

PR #75, opened 2016-12-10
Extr0py/overlay improvements

extr0py:
WIP to address a bunch of issues with the overlays:
- Fix bugs where the overlays shift during changes (you can see this even in the LiveReload gif on the main page)
- Decouple the window state management from the actual UI state (emit events like window-new, window-dimensions-changed, window-close)

The long-term goal is to be able to implement #49 via a plugin, which would be a great proof of concept, and would open the door to lots of interesting UI-driven plugins.


extr0py:
This fixes the jitteriness mentioned in #61, because we now have a more reliable way to calculate the window dimensions that doesn't depend on the cursor position

PR #79, opened 2016-12-10
Update logic in main.js because mainWindow isn't stored anymore, the …

### LINKED ISSUE #77: Error when in a different workspace/desktop
On mac, you have multiple "desktops" to work on, think like workspaces in linux. Basically, if I'm one that does not have oni open and I click on it in the dock, I get this.

screen shot 2016-12-10 at 2 34 40 pm

extr0py:
This should fix #77


extr0py:
@bert88sta - hopefully this fixes the issue you were seeing!

Issue #59, opened 2016-12-08
Background suggestions?


extr0py:
I just tossed in this background:
```
    "prototype.editor.backgroundImageUrl": "http://cdn.wonderfulengineering.com/wp-content/uploads/2014/04/code-wallpaper-2.jpg",
```
in Config.ts.

Anyone have suggestions for a better background to use?

Long-term, I'd like to have some extensibility here - maybe there is some interesting code visualization we could put in the background - even stuff like this (JSCity)[https://github.com/aserg-ufmg/JSCity] might be cool, if it isn't too intrusive. Or some sort of animated document map... Or just sets of backgrounds that fade-in/fade-out periodically. Or no background at all :)

bert88sta:
As long as it's  configurable item,  starting with the current image (or none) is honestly fine. Most people will appreciate that it's minimal,  and If they want more they can install a plug-in or choose a different wallpaper.

pmurias:
Get rid of the background ASAP, I thought it was a bug and made a screenshot to report and looked into the issue tracker avoid duplicates.
Have a plain black background as a default.
Text on text SUCKS HARD.

pmurias:
In terms of suggestions, having an option for a plain white background makes sense.

extr0py:
@pmurias - thanks for the feedback. I'll keep an empty background for now (so it's just the colorscheme's bgcolor), and then the README.md documents how to change the background image if it is important for someone.

pmurias:
Sorry for being rude.

extr0py:
No worries @pmurias , I imagine there are others that felt the same way too :) Will close this out

PR #82, opened 2016-12-10
Extr0py/59 background suggestions

extr0py:
- Remove default background


extr0py:
Fix #59 - remove background

Issue #68, opened 2016-12-09
Project-specific settings


bryphe:
Often when switching between projects, there are some project-specific settings that are useful to preserver in that project's context. For example, I might want to use tabs for indentation in one project, but spaces in for indendation in another. 

This has always been somewhat tricky in VIM - there are some plugins like this:
https://github.com/embear/vim-localvimrc
https://github.com/tpope/vim-projectionist

But it's still not perfect.

In addition, with Oni, there are some additional IDE-capabilities that I'd like to eventually expose per-project:
- Launching the project (ie, F5 in Visual Studio / VSCode)
- Debugging the project (much later...)
- Project-specific settings
   - Whitespace
   - Find / Ctrl-P strategies
   - Test-file lookup (to easily switch back and forth from source to test, a la projectionist

The _Launch Project_ is actually the most important one at the moment - the scenario I have in mind is that I can toss up some language extension examples on github, and then someone can clone them, open in Oni, press F5, and be ready to test it out in a sandbox instance. That will make it really easy for people to get started and understand the API.

My proposal is to follow VSCode's format:
- .oni folder at root of project
   - settings.json - settings overrides (whitespace, configuration settings)
   - launch.json - instructions on 'launch' tasks
 
There are a few other things VSCode has, like tasks.json, but I believe those could be deferred for now

cyansprite:
This is really interesting...
I was thinking about this a lot today, I think something that would be helpful would be template.json files... you can create?

Example for Unity I always want to look for .cs usually and sometimes shaders, so if I have a template I can find just .cs files, and so when I make a new projcet, open in Oni I can do a task generate .oni from template or something, and bam.

It would pretty much be a specialized version of what normally would happen just with those standards already in place.

Also, it would be nice if ```.oni/*``` was added automatically to my .gitignore when I ran this template task ( at least have an option to do so in the general case? )

Next, how do we know when to generate these .oni workspaces... and do we tie these into bookmarks? So when we open oni, and look at bookmarks, should we append workspaces we already made and treat them as bookmarks but with a workspace icon?

With this, also, would be a good point to cache files, as well as having this we could figure out how we should implement includes and excludes, which would lead to faster searching for files, lines, and find and replacing across files.

owais:
I think vscode handles this pretty well the "user settings" and "workspace settings" concept. User settings are stored globally in user's config dir where as workspace settings are stored at the root of the open directory/project. Workspace settings always override user settings. It works out pretty well.

[...]

PR #87, opened 2016-12-12
Extr0py/tasks

extr0py:
WIP to add the concept of 'tasks' to the IDE
- Tasks are accessed by Control+Shift+P <C-P> by default
- In this initial work, they would be included via .oni/launch.json as well as all the npm tasks. Later, this would be extended
- Includes a `oni.tasks.launch` command to launch the first launch task specified in .oni/launch.json, or bring up multiple choices.
- Later, would include commands provided by plugin
- Also later, the npm tasks would be factored out and provided by a plugin, as opposed to being baked in the ONI core

This also starts the work for per-project settings in #68


extr0py:
Remaining work:
1) Clean up TSLint errors
2) Test on Linux

Issue #47, opened 2016-12-08
QuickOpen: Ctrl-P / QuickOpen menu only works with git


extr0py:
Right now, the QuickOpen menu only works with git. The code for finding files looks like this:

> oni/browser/src/Services/QuickOpen.ts

        const trackedFiles = execSync("git ls-files").toString("utf8").split("\n")
        const untrackedFiles = execSync("git ls-files --others --exclude-standard").toString("utf8").split("\n")

It would be helpful to have, at the very least, a fallback strategy to use a node module for recursively traversing the file structure. In addition, having the command line be a user-configurable option would be really helpful - so people can use the preferred search tool of their choice (ag, grep, etc).

extr0py:
Fixed by #97

Issue #92, opened 2016-12-12
Document current keyboard shortcuts


keforbes:
The main project page has images showing fuzzy finder, quick info, code completion, and line evaluation working in Oni.  Please document the keyboard shortcuts for activating these features.

extr0py:
Great point, I just sort of through those up there and if people don't know where to look or where the support lives... they are probably pretty disappointed when they open the app :) I'll start calling them out in the docs shortly

extr0py:
We have most of the keyboard shortcuts documented now, and are in the process of moving some of our documentation to the gitbook / wiki.

PR #97, opened 2016-12-14
Initial work to split out strategy for QuickOpen

extr0py:
Start of work for #47 to split out QuickOpen and add non-git strategies

Will also add keyboard shortcuts to documentation, as part of #92


extr0py:
Fix #47

PR #104, opened 2016-12-15
Fix #91, autocompletion after returning to insert mode

### LINKED ISSUE #91: UI - Autocompletion: If you go to normal mode, and back to insert mode, completion duplicates part of word
1) Open 'NeovimInstance.ts'
2) Start typing `window.set` in empty line, wait for completion to kick in
3) Press Control+c <C-c> to go to normal mode
4) Press I to go back to insert mode
5) Accept completion for `setTimeout`

Expected: Should see `window.setTimeout`
Actual: See `window.setTimeout`

autocompletion bug 2

This is due to the hacky logic we use for resolving the completion in `index.tsx`:
```
        if (UI.areCompletionsVisible()) {

            if (key === "<enter>") {
                // Put a dummy character in front so it removes the word,
                // but not a '.' if the completion comes directly after
                instance.input("a<c-w>" + UI.getSelectedCompletion())

                UI.hideCompletions()
                return
```

Need to do something more robust, like take into account the base and replace based on the coordinates of the base + current cursor

keforbes:
In #91, you typed part of a word, exited insert mode, then re-entered insert mode and typed more characters.  Your "insert autocompletion" code was doing a `<C-w>` to delete the partial word you'd typed but that only deletes to the beginning of the characters entered since starting insert mode.

My solution is slightly less of a hack, but still a hack.  Move the cursor one character to the left, get the word under the cursor, and move the cursor back to the right.  This gets the "word" regardless of what was typed since starting insert mode.  I then remove this "word" prefix from the autocompletion string and insert what's left.  So the code makes the following assumptions:
1. There is a previous character to move the cursor under.  Given that you have to type at least one character to get the autocompletion pop-up to appear, I think this is a safe assumption.
2. The first part of the "word" you typed will be found at the beginning of the autocompletion string.  Today, this is true (it even has to be case-sensitive) but if we were to implement fuzzy-matching or case-insensitive autocompletion, it would fail.

I'm creating this Pull Request now mostly to request for comments.  The only other solution I can think of would be: exit insert mode, perform `db`, enter insert mode, insert unmodified autocompletion string.  Changing modes has side-effects though (like the `<C-w>` example above, and the small-delete register) and I wanted to minimize impact.  Thoughts?

### DIFF
--- a/browser/src/index.tsx
+++ b/browser/src/index.tsx
@@ -202,9 +202,19 @@ const start = (args: string[]) => {
         if (UI.areCompletionsVisible()) {
 
             if (key === "<enter>") {
-                // Put a dummy character in front so it removes the word,
-                // but not a '.' if the completion comes directly after
-                instance.input("a<c-w>" + UI.getSelectedCompletion())
+                let completion = UI.getSelectedCompletion() || ''
+
+                //move one character left so the cursor is "within" the word
+                //(we wouldn't be displaying completions if there wasn't at least one character)
+                instance.input('<left>')
+                //get current word under cursor
+                instance.eval<string>("expand('<cword>')")
+                    .then((word) => {
+                        //move back to where we were
+                        instance.input('<right>')
+                        //remove the first instance of the word under the cursor
+                        instance.input(completion.replace(word,''))
+                    })
 
                 UI.hideCompletions()
                 return


keforbes:
Still playing a bit with this... It looks like Vim has `i_<C-Left>` which can go back one word while in insert mode.  This would tell us the starting position of the word we need to replace.  Unfortunately, `i_<C-Left>` doesn't work today in Oni.  It isn't handling that key combo correctly.

extr0py:
Thanks for looking at this, @keforbes !

I like the idea you have using <cword> to grab the base - and I think it will generalize to other types of autocompletion (like, integrating with neovim's external popupmenu UI). Prior to your change, I was thinking that we might have to do something besides vim commands: Like call getline() directly on the buffer, replace the word based on the completion base and the cursor position with the completion candidate, and then call setline(). 

There might be some cases where <cword> still doesn't work 100% (like if there is a hyphenated token, ie 'neovim-actions'), but it's much better than the logic today, and I'm not sure if those cases would even come up in practice. I'll take this PR since it makes completion much better, and if we see cases where the expand(<cword>) + replace logic isn't enough, we can revisit at that time with a new issue.

Really appreciate your help!

Issue #9, opened 2016-12-03
UI - Integrate 'popupmenu_external' with completion


extr0py:
Came about from discussion with  #6 - there are some really cool UI integration points being added here:
https://github.com/neovim/neovim/pull/5686

Initially, this PR has already been merged and is available - so integrating with the external popupmenu would be a good starting point.

bert88sta:
So I changed the args in guiattach from 
(80,40,true)
 to
 (80,40, {"rgb" : true, "popupmenu_external" : true})

as was suggested by the rpc doc you linked. What's weird is that the screen doesn't draw **at all.**

 It still works, (as in connects to neovim) I edited and saved a file with no screen 💯 

I'm not super familiar with the hierarchy of this repo and am still exploring, but if you have any guidance let me know; I'd like to learn :)

extr0py:
Thanks for looking at this, @bert88sta !

Ah ya, unfortunately there's a bug with neovim/node-client that causes an issue with this.

Try starting from this branch: https://github.com/extr0py/oni/tree/extr0py/external-popup-menu

This commit specifically has the workaround I needed: https://github.com/extr0py/oni/commit/2d3e535eff2146fea529bae6a6e002c8b21c8ec1

bert88sta:
I don't know much about the principles of Typescript, i.e. what needs encapsulation and whatnot, so here's my question.

the popupmenu needs access to UI for completion and menu access. Do we move these popupmenu signals from "actions" to a diffferent type of event? it seems like you put all the other redraw options under actions, so I went with that, but short of bringing UI into Screen.ts there doesn't seem another way to do it.

bert88sta:
Actually, looking at it I'll probably have to make it a different type of event so vim and external plugins can communicate.

bert88sta:
@extr0py: do we want to add a dispatch method to UI? As neovim goes, the level of UI control we have is only going to increase, and more of the on("action") (which I assume you want to maintain since all of the "action"s are used for redraw) are going to make calls to UI for popups. If not, just give me a general overview of what you'd think is a good way to implement.

bert88sta:
@extr0py 

After reading some crash course node/electron/flux/redux materials, I realize now that the last comment made 0 sense in terms of design. I do actually have an important question for you though:

[...]

PR #108, opened 2016-12-17
Integrate external popupmenu

extr0py:
Looking at #9


extr0py:
With @keforbes recent fix for autocompletion, this integration works a lot better, since we don't really need to take into account the base position (which we don't get from popupmenu_external)

extr0py:
Adding some documentation around current completion commands for #92

extr0py:
Fix #9

Issue #60, opened 2016-12-08
Hook up react-hot-loader for development


extr0py:
Right now, `npm run start` sets up a webpack live-reload server. `react-hot-loader` isn't fully integrated though.... This would have some really nice development scenarios. 

For example, imagine working on the Cursor component (which will be in React after #46) and seeing your changes live as you work on that code :) Would be really cool, and helpful for all the UI functionality we're adding to have that instant feedback. 

In theory everything should be hooked up - but I think there are some quirks with the way I'm bringing in the hot-reload webpack bundle.

extr0py:
Fixed by #109

PR #109, opened 2016-12-17
Initial refactoring to move cursor to react component

### LINKED ISSUE #46: Refactor 'Cursor' to React component
Right now, the `Cursor` is implemented using pure DOM manipulation. It would be much better to implement this in the React layer - that would give us the flexibility to more easily add some interesting capabilities like animation to the cursor.

This should be a relatively straightforward refactor, although the following will be needed:
- Mode will need to be included in the redux store
- Cursor x,y is already present

It'd be nice to have some (optional through config) animations for the cursor. An example is sometimes when I have co-workers watching me in the UI, it's hard for their eyes to follow where the cursor is moving.

### LINKED ISSUE #48: QuickOpen: Cursor renders on top of quickopen
VIM cursor renders on top of QuickOpen:
image

#46 could help with this

extr0py:
Fix #46 
Fix #48


extr0py:
Also addresses #60 - brings in react-hot-loader

PR #111, opened 2016-12-18
Make the ONI main process single-instance

extr0py:
Make Oni run in a single instance, so that it's easy to switch between browser windows


extr0py:
Still need to do the following:
- Add a command to switch between instances
- Send it to main process via IPC
- Add logic in main.js to focus window

extr0py:
Enables switching between active running instances via <C-PageUp> and <C-PageDown> - this is useful for some of my workflows

extr0py:
Also adds a '--verbose' flag which is recognized by the main process, which does some logging to the console

Issue #112, opened 2016-12-19
PageUp/PageDown don't work on Linux


cjb:
Thanks for oni!

Pressing Page Up/Page Down just inserts the string "PageUp" or "PageDown" into my documents on Linux.  (Also, while scrolling with the mouse wheel, looks like the bottom line of the display gets corrupted/overwritten, see below.)

Otherwise looks to be working well on Linux!

![screenshot from 2016-12-19 14-27-21](https://cloud.githubusercontent.com/assets/21217/21326121/538b5418-c5f7-11e6-8a37-9b3176b8a14e.png)

extr0py:
Cheers @cjb ! Looks great. It was previously much uglier on Linux before @JakubJecminek fixed the fonts. :)

The <PageUp>/<PageDown> issue should be fixed, and I have a PR to fix the bottom line issue out now. I'll publish a new NPM package today too with those fixes.

Thanks for trying it out and reporting those issues!

extr0py:
Just pushed 0.1.5 to NPM - should be fixed now.

PR #113, opened 2016-12-20
Fix boundary cases in scrolling 

extr0py:
Far right & very bottom line were not being re-rendered on scroll - missed boundary case.

Need to get some unit tests around here soon.


extr0py:
Fix #112

Issue #106, opened 2016-12-16
[NERDTree] "Press ? for help" appear in the wrong buffer


Ilphrin:
Hi! First, let me thank you a lot for your awesome GUI, even if it is still in development/beta, it looks really cool (I tried first Nyaovim, and was a bit disappointed x) ).

Now, when I open neovim, I have automatically NERDTree opened, but there is a little graphic issue when using Oni:
![untitled](https://cloud.githubusercontent.com/assets/7887258/21263250/031a303e-c396-11e6-8eae-94c1c7f96b33.jpg)

When I write on the buffer, it replaces these letters, so it really looks like a graphical issue as it is not in the buffer. I have updated to the latest npm package for oni, and for neovim and NERDTree too
(By the way, I am on Linux Mint 18)

Here is my .vimrc if you need https://github.com/Ilphrin/.vim/blob/master/vimrc

extr0py:
Hi @Ilphrin ! Thanks for the kind words!

Yes, this definitely looks like a rendering bug. I believe I hit this on Windows too on startup - it seems like the beginning of the first line is always problematic. It might be we're missing or not handling one of the neovim msgpack-RPC actions correctly.

Ilphrin:
You're welcome! When i'll have time, i'll try to give some help on Oni, this really deserves more hands =D

BTW, when i run Ctrl+L it is refreshing the page and the glitch doesn't appear anymore, as my NERDTree is has a plugin to be launched in every tab and on startup, maybe it is because the browsers draws too soon the content? (Just suppositions, I don't know a thing about what's is happening!)

extr0py:
Awesome, would be great to have the help! :)

I just checked out this issue... It looks like it was a bug that was caused by a couple of contributing factors:
1) On startup, we would tell Neovim that we had a fixed size screen (80 cols x 40 rows), and then we would always resize to the proper size based on the font afterwards
2) During resize, Neovim sends a `CLEAR` action via the msgpack-rpc API. For that action, the cursor position wasn't being reset - so after clearing, it would just start rendering the first line whereever the cursor had been previously. So that was the root problem, and easy to fix.  

Should be addressed now by PR #115

PR #115, opened 2016-12-20
Initial fix for render issue

extr0py:
Turns out we are not handling the CLEAR action from NeoVim correctly. For CLEAR, the cursor should be reset back to 0, 0.

When we resize the screen (which, in the current flow, always happens since there is a default size set), the cursor position does not get reset - so that means the first text that is drawn is almost always in the wrong place.


extr0py:
Fix #106

extr0py:
In addition to the core fix of resetting the cursor position, this also implements a couple of improvements:
1) Do not always resize on initial render - we have enough information to know the proper dimensions when we call nvim_ui_attach, so we should use those instead of the baked [80, 40] value.
2) Do not resize if the rows,cols values are unchanged, by caching them on NeovimInstance

PR #116, opened 2016-12-20
a couple minor fixes to README.md

keforbes:
The `npm install -g oni-vim` and `oni` lines were appearing on the same line.

Mention that Linux also needs a pre-compiled binary.

Don't refer to the binary as `nvim.exe` when talking about OSX and Linux.


extr0py:
Thanks for the fixes!

Issue #56, opened 2016-12-08
Plugin - TSLint


extr0py:
Now that we have TSLint integrated in the project, it'd be really nice to have a TSLint plugin to show the 'error markers' for lint errors.

We'd want to add another plugin:
1) oni-plugin-tslint
2) Enable it for .ts files
3) Listen to the 'save' event
4) Run tslint in a separate process
5) Report the results back via error markers (we should use a color besides red - yellow might be a good one). It would also be nice to customize the gutter error icon

PR #121, opened 2016-12-23
fix (almost) all tslint errors

keforbes:
I installed the tslint plugin for my IDE and found there were a bunch of tslint errors in Oni.  I decided to go through and fix them all.  Well, almost.  All of the UI components have a `require("./xxx.less")` line which tslint doesn't like but I don't know how to change a `less` file to an import statement so I left those alone.

If you don't actually care about tslint errors or you haven't really configured the `tslint.json` to your liking then you can delete this Pull Request.  I just figured it was an easy fix.

### DIFF
--- a/browser/src/UI/Icon.tsx
+++ b/browser/src/UI/Icon.tsx
@@ -13,7 +13,7 @@ export enum IconSize {
     TwoX,
     ThreeX,
     FourX,
-    FiveX
+    FiveX,
 }
 
 export interface IconProps {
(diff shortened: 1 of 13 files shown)


extr0py:
Thanks for making these @keforbes , appreciate it! Hopefully once #56  is here these will be discoverable inside ONI as well. The change looks good overall, but it seems like there is a compile error in the typescript plugin - I'll approve it and merge once that's fixed.

extr0py:
Looks like the `object-key-literal-quotes` rule is causing problems in tslint, which was just fixed recently:
https://github.com/palantir/tslint/issues/1916

I'll ignore the rule for now until we can pull the fix in

PR #127, opened 2016-12-29
Fix #126 - Support multibyte characters in rendering

### LINKED ISSUE #126: Multibyte language support: oni doesn't display complete charactors
The Chinese characters are twice as wide as English characters.   In the default setting, the neovim can correctly display the characters in terminals. But oni display only the left half of characters. And the right half is blank.

extr0py:
Initial fix for #126 - allow multibyte characters in rendering

### DIFF
--- a/browser/src/DeltaRegionTracker.ts
+++ b/browser/src/DeltaRegionTracker.ts
@@ -61,7 +61,7 @@ export class IncrementalDeltaRegionTracker implements IDeltaRegionTracker {
         this._cells.push({
             x,
             y,
-            force
+            force,
         })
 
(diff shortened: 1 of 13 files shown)


extr0py:
Renders full character. Still some remaining work:
1) Cursor should resize to cover multibyte characters
2) When highlighting, the character sometimes gets overwritten. I believe this is because part of the cell is getting 'cleared' after it is being drawn

extr0py:
Cursor pixel width looks good now, and refactored so wcwidth isn't called everywhere. There's some additional work to properly clear characters now

PR #132, opened 2016-12-31
Fix #56 - TSLint plugin

### LINKED ISSUE #56: Plugin - TSLint

extr0py:
- Add tslint plugin, that is active whenever a `tslint.json` is fond
- Fix issues in error markers / scroll bar 
   - Add dynamic coloring
   - Set up sources to not overwrite each other


extr0py:
Fixes #56

extr0py:
Adds tslint warnings as yellow:
image

And serves as a demo of how other linters could be hooked up

PR #136, opened 2016-12-31
Extr0py/126.2/multibyte characters scrolling

extr0py:
More multi-width character fixes. This addresses the C-E issue but disables some of the render optimizations.

Remaining work:
- Still a bug with afterimages when pressing "o" or inserting a blank line, such that lines with multi-width characters get shifted down
- Need to check and see if the file open case works as expected


extr0py:
Fixes for #126 (part 2) in progress

lasypig:
How is this PR going?

extr0py:
Sorry @lasypig , it fell off my radar a bit. I'll try to revisit it soon. It is high on my priority list!

extr0py:
Revisiting this in #290, the fix will be a little different now that we have DOM rendering.

PR #139, opened 2017-01-02
Fix #134 - TSLint errors don't go away once they are all fixed

### LINKED ISSUE #134: TSLint Plugin: errors don't go away

extr0py:
Couple fixes here:
1) When saving a file, only run tslint on that file
2) Clear errors that been addressed (didn't show up in the last full-project tslint)
3) Refactor some of the common code out - there is still duplication between doLintForFile/doLintForProject, but this can be improved later.


extr0py:
Fix #134

PR #140, opened 2017-01-02
Extr0py/138/dom rendering

extr0py:
Set up a DOM rendering strategy for ONI. 

Few remaining issues:
1) Performance regression - this re-renders the entire DOM for every change. Need to improve performance (use delta rendering to only re-render divs that were changed)
2) Some highlighting regressions


kybaeus:
This is awesome! I'll give it a go shortly.

extr0py:
@kybaeus , thanks for trying it out! Glad it worked.

I just started working on the delta rendering. For the really performance sensitive pieces (like the rendering), I'm planning on not using React, so that we can have more flexibility in tuning it.  My first render strategy, before the canvas, was using React, and there was some performance overhead that was difficult to workaround.

I definitely prefer writing code in React, so for other things like the the autocompletion/popup menu/etc, the rendering isn't the bottleneck and it makes sense to leverage it.

With the latest changes, the performance seems better, but there is one issue with the delta rendering - sometimes it will split `<span>` and break the ligature rendering. 

So there is the following outstanding work left:
- Fix that split span case, I'm planning on adding to the logic adding a character directly to a span in cases where it makes sense (ie, merging with a previous span)
- Add unit tests for some of the delta / merging / coalescing logic
- Remove the `force` flag on the delta region tracker, since it was really just a hack
- Split out the tokenrenderer items to separate files

extr0py:
More work:
- Move console.log -> Log, only log if not debug
- Modify Performance to only log if debug
- Fix tests after splitting apart files
- Fix bug with coalescing spans - should not coalesce whitespace / non whitespace
- Fix issue with multibyte characters (make isWhitespace return correct value)
- Scrolling - performance work

trusktr:
Have you tried rendering by running `setState` inside `requestAnimationFrame`?

PR #147, opened 2017-01-06
Changed the default menu of electron

Rafi993:
Changed the default menu of electron so that it does not look generic.


package.json
```diff
@@ -42,6 +42,7 @@
   "license": "MIT",
   "dependencies": {
     "electron": "^1.4.12",
+    "electron-default-menu": "^1.0.0",
```
extr0py: Might need to be merged with a recent change to the package.json to clamp the version numbers (this helps predictability of the build, so that newer versions of libraries don't break us). Can you please merge that and then change the version here to "1.0.0" instead of "^1.0.0"?

main.js
```diff
@@ -50,6 +52,44 @@ function createWindow(commandLineArguments, workingDirectory) {
...
+   menu[2].submenu.splice(2,1)
+   menu[4].submenu = [
+       {
+        label: 'Learn more',
+        click: (item, focusedWindow) => {
+        shell.openExternal('https://github.com/extr0py/oni#introduction');
+        }
+      },
+        {
+        label: 'Issues',
```
extr0py: Thanks for adding these options, that's awesome! This will give us a solid foundation to add more menu items down the road as well.

main.js
```diff
@@ -50,6 +52,44 @@ function createWindow(commandLineArguments, workingDirectory) {
 
     // Create the browser window.
     let mainWindow = new BrowserWindow({ width: 800, height: 600, icon: path.join(__dirname, "images", "Oni_128.png") })
+    let menu = defaultMenu(app, shell);
+
+  menu.unshift ({
+    label: 'Oni',
```
extr0py: I tested this on Windows and it looks a bit unnatural - I think we might want `File` on Windows, and `Oni` on OSX/Linux

extr0py:
👍 

Thanks for making this change, @Rafi993 ! It's really cool to have the direct links to Github (as opposed to the Electron github), and gives us a solid foundation to build on to add more menu options later.

I added a few comments on the files - can you please take a look at those and see how you feel? The main concern I had was that the `Oni` item looks unnatural on Windows, so making that specific to OSX / Linux would help a lot.

main.js
Rafi993: File menu is good. May be we can add more broader menu set in the future.

Rafi993:
I did replace oni menu with File but I think it could use more menu items within it than just quit.

extr0py:
Makes sense. Long-term I'm thinking of adding additional items to the `File` menu, something like VSCode has:
image

Specifically, make it really easy to get to `init.vim` and `config.js` along with any other configuration (key bindings, snippets, etc). In addition, the save / new / open commands would be helpful for someone new to  vim/neovim. But those can all come later - it will be much easier to add this functionality now with the work you're doing with this PR.

Rafi993:
That is good to here @extr0py

Rafi993:
@extr0py noticed that there was small tslint error in the pull request I gave. The new commit should fix that

extr0py:
Looks great! Thanks for the contribution, @Rafi993 !

PR #148, opened 2017-01-06
Replacing switch with Object Literals

Rafi993:
I replaced some switch statements with object literals. Even though i does not add any features it makes the code more cleaner.


extr0py:
Cool, I like this approach, it's much more concise - thanks for making the change, @Rafi993 !

It looks like there are a few tslint issues in AutoCompletion.tsx - once those are addressed I'll merge in this PR.

extr0py:
Looks great - thanks @Rafi993 !

Issue #146, opened 2017-01-05
Oni process doesn't go to background at start


xangelux:
Version 0.1.6 (the one installed by NPM).
Issue: When oni is started, it doesn't release the console to the prompt
expected behavior: being able to use the terminal window after starting Oni
Additional info: When started from terminal, if the terminal is closed, Oni is closed with it

extr0py:
👍 👍 

Thanks for logging this, @xangelux ! Agreed, this is annoying, would be nice to have a fix for it. If anyone is up for looking at this, it's definitely a problem with the way we are starting the process in `cli\oni`. 

This is some good info: http://stackoverflow.com/questions/12871740/how-to-detach-a-spawned-child-process-in-a-node-js-script

A few things we likely need to do here:
1) Set the `detached` flag to true
2) Stop piping stdin/stdout/stderr

Rafi993:
[This](https://github.com/electron/electron/issues/2208) seemed promising

Rafi993:
I gave a small[ pull request ](https://github.com/extr0py/oni/pull/150)as given there hope it solves the issue

extr0py:
Thanks @Rafi993 ! Works great on Linux. There is still a problem in Windows - it does immediately exit the terminal process, but it also spins up a separate Node console. I'll open a separate PR to address the windows issue.

city41:
I am on Ubuntu 17.10 using bash and I still find Oni needs to be remained attached to the shell. If I launch with `oni .`, then it doesn't seem to detach. If I ctrl-c in the terminal, Oni quits. If I launch with `oni . &`, Oni never quite comes up right. I can only get it working if I then `fg`, which then has the net result as just launching with `oni .`.

I just installed Oni this morning, it is version 0.3.1 according to the about box.

SamHH:
I experience this issue on Arch with the Alacritty terminal, but not on macOS with iTerm2.

PR #150, opened 2017-01-07
Oni as background process

Rafi993:
I made little change to solve issue https://github.com/extr0py/oni/issues/146. It works smootly on Linux.

### DIFF
--- a/cli/oni
+++ b/cli/oni
@@ -2,7 +2,15 @@
 
 var path = require('path')
 var spawn = require('child_process').spawn
-const os = require("os");
+const os = require("os"),
+      fs = require("fs");
+
+let processOptions = {}
+let log = require('path').join(__dirname, 'out.log')
+let out = fs.openSync(log, 'a')
+let err = fs.openSync(log, 'a')
+processOptions.detached = true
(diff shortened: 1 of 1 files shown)


cli/oni
```diff
@@ -2,7 +2,15 @@
...
+      fs = require("fs");
+
+let processOptions = {}
+let log = require('path').join(__dirname, 'out.log')
+let out = fs.openSync(log, 'a')
+let err = fs.openSync(log, 'a')
+processOptions.detached = true
+processOptions.stdio = [ 'ignore', out, err ]
 
 // Check for Windows, assume macOS/Windows otherwise
```
kopischke: Unrelated to this PR, but: this comment cannot possibly be correct.

Rafi993: sorry pasted it by mistake. ill remove it

extr0py:
Thanks @Rafi993 ! I'll bring this in for Linux. On Windows, there is still an issue where a node prompt pops up when the app is started. I think we need to use spawn with `cmd.exe /c` on Windows, I'll address that in another PR for #146 on top of this.

PR #151, opened 2017-01-07
Update typescript plugin path

multikatt:
It's been moved


extr0py:
Thanks for updating this!

PR #152, opened 2017-01-07
Fix #146 for Windows as well

### LINKED ISSUE #146: Oni process doesn't go to background at start

extr0py:
- Fix for launching Oni in Windows, so that command window doesn't pop up


extr0py:
Fix #146

extr0py:
Verified Linux still works correctly

PR #158, opened 2017-01-12
Prototype: Show errors in quickfix list

extr0py:


browser/src/neovim/QuickFix.ts
```diff
@@ -0,0 +1,29 @@
...
+export class QuickFixList implements IQuickFixList {
+    private _neovimInstance: INeovimInstance
+
+    constructor(neovimInstance: INeovimInstance) {
+        this._neovimInstance = neovimInstance
+    }
+
+    public setqflist(list: IQuickListEntry[], title: string, action?: Action): Q.Promise<void> {
+        action = action || " "
+        return this._neovimInstance.command(`call setqflist(${JSON.stringify(list)}, "${title}", "${action}")`)
```
justinmk: nvim_call_function could be used instead of nvim_command

extr0py: Thanks for the suggestion! Will make the change

PR #164, opened 2017-01-14
Displaying all entries in autocompletion

Rafi993:
All entries is now displayed in autocompletion rather than just first 10 entries


extr0py:
Looks good! Thanks for the contribution, @Rafi993 ! I opened #166 to track some other scroll issues

PR #169, opened 2017-01-17
Fix #156, add new default background

### LINKED ISSUE #156: Background-image is null by default
If you look at the Developer Tools in Oni, you'll see an error at startup:
```
GET file:///<path>/oni/null net::ERR_FILE_NOT_FOUND
```

This is because the background-image is defaulting to null
```
<div class="background-image" style="background-image: url("null"); background-size: cover;"></div>
```

The image is defined here
```
const backgroundImageUrl = Config.getValue<string>("prototype.editor.backgroundImageUrl")
backgroundImageElement.style.backgroundImage = "url(" + backgroundImageUrl + ")"
```

So it looks like `prototype.editor.backgroundImageUrl` isn't defined by default.  I'm filing this as a defect because I'm not sure what we want to do in the default case.

keforbes:
I played with a couple repeating patterns but I always came back to this one.  Smaller repeating tiles sometimes made text appear blurry; this larger pattern doesn't have that problem.  I also think it's subtle enough to not be distracting, but that's subjective.

If you don't like this background then you can close this Pull Request.  No one ever commented when I offered suggestions so I decided to just go ahead and commit this one.

new-background


extr0py:
Hey @keforbes ! Sorry I just saw I missed your comment on #156 

Thanks for going through and finding a good background to use. The background looks great to me.

Just a couple of minor comments on the code-

index.html
```diff
@@ -25,8 +25,8 @@
   </head>
   <body>
     <div id="background">
-        <div class="background-image"></div>
         <div class="background-cover"></div>
+        <div class="background-image"></div>
```
extr0py: Could we keep the order as-is? And then set the `prototype.editor.backgroundOpacity` to a value like `0.8`? The concern I have is that this color might not work well with all color schemes (although it looks great with the default one). If we flip the order back, and use opacity, it will still be present but pretty subtle, and will blend with other colorschemes as well

keforbes:
Done.

extr0py:
Awesome, thanks @keforbes !

PR #175, opened 2017-01-19
Update window title with current filename

keforbes:
When a file is loaded, update the window title with the current filename similar to GVIM's format.

Note that at startup, the `set_title` event may fire before we've set the event listener, causing the title to be `ONI` until a different file is loaded.  I don't know how to query the `title` property after connecting to neovim to ensure we have the initial value so I only call `:set title` after all event listeners are set.  This is a bit of a hack but it means we'll never miss that first event.

### DIFF
--- a/browser/src/NeovimInstance.ts
+++ b/browser/src/NeovimInstance.ts
@@ -327,6 +327,8 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                 ))
             } else if (command === "resize") {
                 this.emit("action", Actions.resize(a[0][0], a[0][1]))
+            } else if (command === "set_title") {
+                this.emit("set-title", a[0][0])
             } else if (command === "eol_clear") {
                 this.emit("action", Actions.clearToEndOfLine())
             } else if (command === "clear") {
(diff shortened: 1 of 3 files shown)


justinmk:
Should this be driven by the 'titlestring' option ?

keforbes:
> Should this be driven by the 'titlestring' option ?
Well of course there's an existing option for that. 🙄   Yes, I should probably be using it.  However, that property won't update the window for this Electron app which is communicating with the headless neovim.  Is there an event that fires when the titlestring changes so we can propagate it to our window title?

justinmk:
Yes `OptionSet` event can listen to changed options. Note that 'titlestring' holds am expression so the option itself wouldn't change often. But the expression could be eval'd by the front-end via nvim_eval() as needed.

keforbes:
Great, thanks!

keforbes:
Thanks for your help @justinmk!  The code is much simpler now.

We still have the race condition where the event might fire before the listener is set but that's our problem.  It only happens if you open a file with `oni <filename>`.  It might be worthwhile to query the `title` property during startup but I'm not sure how to do that so I'll leave it for later.

extr0py:
Thanks @keforbes and @justinmk , this looks great!

PR #190, opened 2017-02-01
add attribution for default background image

keforbes:
The default background I committed earlier is available via Creative Commons but that requires attribution.  I'm not sure what format you want `ASSETS.md` to be in so I just took a guess.


extr0py:
Thanks for doing this!

PR #194, opened 2017-02-03
Replace edit menu with one that matches gvim

keforbes:
The default Edit menu provided by Electron doesn't communicate with NeoVim so the options don't actually work.  This Pull Request replaces the Edit menu with the options found in GVim.

Unfortunately, Electron won't allow for accelerators (key shortcuts) to be lower-case so it refuses to display things like `ggVG` as an accelerator.  I think we'd have to create a custom display of the accelerators to do that.  For now, I'm just not displaying any accelerators.

Also, the `Redo` operation should be `^R` or `<C-r>` but I can't figure out how to get that stupid thing to work.  So everything in this Pull Request works except `Redo`.  I decided to just create this Pull Request now in the hopes that someone might be able to figure out how to get the `Redo` command to work.  Besides, the default Edit menu is completely useless so at least this Edit menu is *mostly* functional. :smile:

### DIFF
--- a/browser/src/index.tsx
+++ b/browser/src/index.tsx
@@ -292,6 +292,10 @@ const start = (args: string[]) => {
     window["neovim"] = instance // tslint:disable-line no-string-literal
 
     UI.init()
+
+    ipcRenderer.on("menu-item-click", (_evt, message) => {
+        instance.command("normal! " + message)
+    })
 }
 
 ipcRenderer.on("init", (_evt, message) => {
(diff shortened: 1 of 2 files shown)


extr0py:
Great contribution, @keforbes ! I really like this.

I could see lots of other ways to extend this as well - like creating new buffers, save, etc - all this would help a lot in terms of bringing non-vim users to ONI as well.

And we can track the `Redo` command in a separate issue - as you mentioned, this is a big step forward from what we have today :)

PR #197, opened 2017-02-06
	modified:   browser/src/index.tsx

### LINKED ISSUE #196: Redo in edit menu not working
@keforbes did some great work in enabling our 'Edit' menu in the toolbar - adding cut, copy, paste, put before/after, etc.

There is a strange issue with `Redo` though - for some reason, `normal! <C-r>` doesn't seem to work as expected. Not sure if this is a neovim bug or something we are doing wrong. It would be interesting to see how vim/gvim behave in that case.

It looks like the `:redo` command works, so we could potentially expand our interface for passing messages to support both `execute` and `normal!` actions (really, the `execute` is a superset anyway) - we'd just have to update our existing messages to match.

bert88sta:
This should fix #196

### DIFF
--- a/browser/src/index.tsx
+++ b/browser/src/index.tsx
@@ -294,7 +294,7 @@ const start = (args: string[]) => {
     UI.init()
 
     ipcRenderer.on("menu-item-click", (_evt, message) => {
-        instance.command("normal! " + message)
+        instance.command("exec \":normal! " + message + "\"")
     })
 }
 
--- a/main.js
+++ b/main.js
@@ -78,7 +78,7 @@ function createWindow(commandLineArguments, workingDirectory) {
        {
            label: 'Redo',
            click: (item, focusedWindow) => {
-               mainWindow.webContents.send("menu-item-click", "\<C-r>")
+               mainWindow.webContents.send("menu-item-click", "\\<C-r>")
            }
        },
        {


extr0py:
Nice, thanks for the fix, @bert88sta ! Looks great

keforbes:
This fix broke `Cut`, `Copy`, and `Paste`.  I'm guessing it's because those options use `"` in their command.

browser/src/index.tsx
```diff
@@ -294,7 +294,7 @@ const start = (args: string[]) => {
     UI.init()
 
     ipcRenderer.on("menu-item-click", (_evt, message) => {
-        instance.command("normal! " + message)
+        instance.command("exec \":normal! " + message + "\"")
```
justinmk: Instead of using command() and then doing the escaping here, one could use `nvim_feedkeys()` (which takes same options as VimL `feedkeys()`). Assuming that's mapped to `instance.feedkeys()`,

    instance.feedkeys(message, "n", false)

bert88sta:
Just wrote a fix,  will make a PR in a minute here.

browser/src/index.tsx
keforbes: @justinmk, I tried implementing this with `feedkeys` but `\<c-r>` still doesn't get interpreted.  Any idea why I can't send a Ctrl key to `feedkeys()`?

justinmk: Hmm, this works from the nvim command line:

    :call nvim_feedkeys("\<c-r>", 'n', v:false)

But of course that "escaped" keycode `"\<c-r>"` is a special thing that won't work outside of vim-land. Using the API from something like python or oni, one would need to send the raw ^R byte (however one's client represents "octal 22"):

    :call nvim_feedkeys("\022", 'n', v:false)

Yuck, that's actually worse than sending the literal string `"norm! "\<c-r>"` to nvim_command().

`nvim_input()` is much more convenient in this regard: `nvim_input('<c-r>')` does what you might expect. But it is raw user input, so one can't disable mappings. 

I should have realized this when I suggested feedkeys(), my mistake.

extr0py:
Good catch @keforbes , sorry I missed this when merging in

PR #198, opened 2017-02-06
This worked for me for #196 (Fix for the fix)

bert88sta:
If we want to move over to a feedkeys implementation as justinmk suggested, we could do that, but this will get them going in the mean time.

EDIT: Actually, as justinmk noted, feed keys is probably not the optimal method for using normal! programatically. This PR fixes the other menu items by properly escaping them.


extr0py:
Thanks @bert88sta for the quick fix!

PR #203, opened 2017-02-09
Add option to fullscreen on start

bert88sta:
I *think* I rebased right...

Other than that, this just adds editor.fullScreenOnStart as a config option. From the BrowserWindow documentation, it seems like setting fullscreen as a **parameter** will have affect behavior, but as a method it just changes it harmlessly


extr0py:
Thanks @bert88sta !

PR #208, opened 2017-02-11
Extr0py/205/incorporate react redux devtools

extr0py:
Installs the [redux devtools](https://github.com/zalmoxisus/redux-devtools-extension) when running via either `oni --debug` or via `npm run start`. This is really helpful for understanding and debugging the UI functionality, and offers things like time travel debugging or going back to past history of the app. It also helps in terms of understanding the flow of actions and how they change the app state.

### DIFF
--- a/browser/src/index.tsx
+++ b/browser/src/index.tsx
@@ -34,7 +34,7 @@ const start = (args: string[]) => {
 
     // Helper for debugging:
     window["UI"] = UI // tslint:disable-line no-string-literal
-    remote.BrowserWindow.getFocusedWindow().setFullScreen(Config.getValue<boolean>("editor.fullScreenOnStart"))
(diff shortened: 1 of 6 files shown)


browser/src/index.tsx
```diff
@@ -34,7 +34,7 @@ const start = (args: string[]) => {
 
     // Helper for debugging:
     window["UI"] = UI // tslint:disable-line no-string-literal
-    remote.BrowserWindow.getFocusedWindow().setFullScreen(Config.getValue<boolean>("editor.fullScreenOnStart"))
+    remote.getCurrentWindow().setFullScreen(Config.getValue<boolean>("editor.fullScreenOnStart"))
```
extr0py: FYI - @bert88sta  - just wanted to let you know I had to make this minor change. What happened was that, in some cases, like using `npm run start` if the debugger opened up, and the browser window wasn't the active window, this would crash because `getFocusedWindow` would return null. Minor detail though, just wanted to give you a heads up

PR #210, opened 2017-02-11
Fix #171 - Replace IPC

### LINKED ISSUE #171: Profile IPC layer / investigate replacements
Based on profiling, it looks like the IPC calls are very expensive, and they tend to be numerous. In addition, there is extra overhead because they are routed through the main process all the time. A couple of things to look at here:
- What is the `ping` roundtrip time from the renderer to a plugin via IPC?
- What's the synchronous time taken when executing an IPC call? Even though we use `send`, which is async, there still seems to be a `sendSync` call that takes a non-trivial amount of time as part of that.

It might be worth moving to an alternate method to communicate between the processes - a raw port or socket.io, and compare the `ping` time and synchronous overhead.

extr0py:
WIP to replace the IPC layer, which is very slow


extr0py:
Going to pursue an in-proc solution as opposed to the socket.io solution

Issue #193, opened 2017-02-03
Clicking outside of completions


bert88sta:
Clicking around in completions should either close or re-evaluate at the line, I'm not sure which. However, this is not what should happen

![git issue completion menu](https://cloud.githubusercontent.com/assets/8421641/22579939/1bb17490-e9a2-11e6-99e6-84c5ce8948e3.gif)

extr0py:
Definitely, this is not the right behavior! Thanks for tracking. I think we assume when we're in insert mode, if the cursor moves, we should still be displaying the completion... But that is certainly wrong in this case.

bert88sta:
I've noticed that completions in general are slow,  and as a someone who
types 120+ wpm,  the completions get "behind"  both in terms of completion
items and actual position of the menu

extr0py:
Agree, it's almost not usable in some cases - hopefully can get some time to look at the performance soon. That's the biggest pain point for me in using ONI for my daily editing tasks - the completion just breaks my flow at the moment.

bert88sta:
I know it's probably hard to do with a massive program,  but do we have a
way to probe and profile bits of code?

extr0py:
Thanks for the fix! Looks like this is addressed. I'll close this now.

PR #211, opened 2017-02-11
Bugfix: 193

bert88sta:
This generically fixes issue #193, with clicking around causing completions to stay visible and potentially outdated. We might want to add logic to re-evaluate if there's a new line, but I figured it would be better to wait until someone actually complains about it.


extr0py:
Looks good! Thanks for the fix, @bert88sta !

PR #212, opened 2017-02-11
[WIP] Fix for #171 - Replace IPC for plugins with in-proc messaging to improve performance

extr0py:
Fix for #171


extr0py:
Need to clean up tslint issues

extr0py:
Also need to re-enable tslint to be performant, now that it is in-proc

extr0py:
Tslint is now re-enabled - but discovered there is a regression where plugin events are being broadcasted to all plugins, all the time, due to all plugins sharing a single channel. Need to take a look at that.

extr0py:
Fixed remaining tslint & test issues - green now. Still need to fix the plugin performance regression though-

extr0py:
Things are working again, and performance is improved with this strategy because the `sendSync` calls for IPC were adding a lot of overhead. This also has the additional benefit that plugins are easier to debug with the same debug session. The downside is that slow plugins can bog down the editor, but this is an issue with VimScript today too anyway, and JavaScript/TypeScript has async out of the box.

PR #213, opened 2017-02-12
Always have a default value for the `activeWindow` property

extr0py:
Set up `activeWindowDimensions` property always have a value populated


extr0py:
@bert88sta  - FYI, I think once you pick this up you shouldn't get that `object can be null` error

PR #219, opened 2017-02-14
202 - Leverage Externalized cursorcolumn and cursorline

bert88sta:
I deleted my old fork because I don't understand git and screwed it all up.

This was made in a separate branch with slightly cleaner commmits. I don't know that this is necessarily the way you would have implemented it, so if you want any changes just let me know and I'll take care of them.


browser/src/UI/components/CursorColumn.tsx
```diff
@@ -0,0 +1,60 @@
...
+        const cursorStyle = {
+            position: "absolute",
+            left: this.props.x.toString() + "px", // Window Start
+            top: this.props.y.toString() + "px", // Same as cursor
+            width: width.toString() + "px", // Window width
+
+            height: this.props.height.toString() + "px", // Same as cursor
+            backgroundColor: this.props.color,
+            opacity: 0.2,
+            fontFamily,
```
extr0py: I think we can get rid of fontFamily and fontSize here, same for CursorLineRenderer. We could actually use the same `Renderer` class for both the CursorColumn/CursorLine, since there isn't actually anything different about them, except the way we figure out the height/width, which happens in the mapStateToProps

browser/src/UI/components/Cursorline.tsx
```diff
@@ -0,0 +1,59 @@
...
+// require("./Cursor.less") // tslint:disable-line no-var-requires
+
+class CursorLineRenderer extends React.Component<ICursorLineProps, void> {
+
+    public render(): null |  JSX.Element {
+        if (!this.props.visible) {
+            return null
+        }
+
+        const fontFamily = Config.getValue<string>("editor.fontFamily")
```
extr0py: We should be able to remove both of these getValue calls

browser/src/index.tsx
```diff
@@ -165,11 +178,25 @@ const start = (args: string[]) => {
         UI.setMode(newMode)
 
         if (newMode === "normal") {
+            if (cursorLine) { // TODO: Add "unhide" i.e. only show if previously visible
```
extr0py: The TODO makes sense. I'm thinking we could have an event like Config.on("changed", (newConfigKey, newConfigValue) => { ... }) where we could add this. This could happen if we reload the config file, or if we synchronize config values from Neovim, etc. No changes needed, just thinking out loud

extr0py:
Looks like you're a Redux/React master now, @bert88sta ! You should try out the hot reloading with `npm run start` or try out the redux dev tools with `ctrl-shift-i` - these actions seem like they would work well with the "time-travel" debugging. 

There are a couple of places we could look at removing duplication (CursorLineRenderer / CursorColumnRenderer are exactly the same - we could move that to a shared class or just reuse `CursorLineRenderer` for both). It seems like there are a few lint issues to clean up, but the overall change looks good. Thanks for making the change, this is great!

bert88sta:
@extr0py: 

Can I just pass in parameters in Root component, i.e

```
<... />
<CursorLine("line") />
<CursorLine("column") />
<... />
```

Typescript expects an identifier on those lines, do they need to be assigned to something first?

extr0py:
Yep, a few things you need to do:

First, update `mapStateToProps` to take an argument. So something like this:
```
interface ICursorLineProps {
   type: string; // An enum might be better here since we get type safety
}

const mapStateToProps = (state: State.IState, props: ICursorLineProps) => {
     return {
         x: type === "line" ? state.activeWindowDimensions.x : state.cursorPixelX
         y: type === "line" ? state.cursorPixelY : state.activeWindowDimensions.y
         ...
     }
 }
```

Note that there is actually already an `ICursorLineProps` so you might want to rename that `ICursorLineRendererProps`. We basically need two props - one for the rendered component (width, height, color, visible, etc), and the other for the _connected_ component - which is just the type (line/column).  This article has some great information about the difference between these two components: [Presentational and Container Components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.ba0gew8ec)

Second, pass in your argument - it would look like this:

```
<... />
<CursorLine type={"line"} />
<CursorLine type={"column"} />
<... />
```

And for bonus points, we could use an enum, so we'd get:
```
<... />
<CursorLine type={CursorLineType.Line} />
<CursorLine type={CursorLineType.Column} />
<... />
```

Hope that helps!

extr0py:
Woohoo! Green build is nice to see :) Let me know if you're all set with this and I'll merge it in. The change looks good. Looking forward to switching my config over to this.

bert88sta:
This is ready to go when you are, enjoy

extr0py:
Just tested it out - looks great, @bert88sta ! Nice work.

PR #221, opened 2017-02-14
Add entries to File and Window menus

keforbes:
Add options to the File and Window menus to be similar to GVim.  I know the goal for Oni isn't to create a GVim clone in Electron but I figure these menu entries are good starting points for us.  And again, the default menu options weren't very helpful so we aren't losing anything by adding these.

### DIFF
--- a/browser/src/index.tsx
+++ b/browser/src/index.tsx
@@ -297,8 +297,12 @@ const start = (args: string[]) => {
 
     UI.init()
 
-    ipcRenderer.on("menu-item-click", (_evt, message) => {
-        instance.command("exec \":normal! " + message + "\"")
+    ipcRenderer.on("menu-item-click", (_evt, message: string) => {
+        if (message.startsWith(":")) {
+            instance.command("exec \"" + message + "\"")
+        } else {
+            instance.command("exec \":normal! " + message + "\"")
+        }
     })
 }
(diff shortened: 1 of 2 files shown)


keforbes:
There's one thing I wanted to mention about this implementation.  GVim has a `Split File Explorer` option in the `Window` menu.  Somehow, GVim opens the file explorer in a split and forces its width to 20 columns.  If I just call `:Lexplore` it opens the file explorer in a split that takes up half the window.  So I set the menu option to `:Lexplore | vertical resize 30` to force its width to 30 columns.  I figured 20 columns was too small so I made it 30 but I'm not sure if this is the best way to implement that option.

extr0py:
This is really cool! One of the goals I have for the project, longer-term, is to lower the barrier to entry for modal editing - and having these menu options definitely helps. And even though this isn't a GVim clone, having basic parity with GVim makes sense - a barrier to entry I've heard for Windows devs in adopting Neovim is that there is no GVim-equivalent, so changes like this can help Oni fill that niche.

Thanks for calling out the `Split File Explorer` option - I tried it out and you're right, 20 columns would be too small - the default of 30 sounds like the right choice.

extr0py:
Merged - Thanks @keforbes !

Issue #216, opened 2017-02-12
Pressing `Ctrl-A` highlights everything


extr0py:
Pressing `Ctrl-A` highlights everything and puts the text into a bad state. The only fix is to click with the mouse somewhere to remove selection.

We need to fix this by either `preventDefault`ing the Ctrl-A keypress, or adding some CSS such that the text does not show up as selectable.

bert88sta:
I'm going to weigh in right now and say that CSS is the right answer, mainly because on OSX it's cmd - a which we don't handle but will still cause the same undesired effect. I know nothing about CSS but I'll see if I can throw something together

extr0py:
Yep, seems like it would be straightforward. I think we can just use this CSS style:
https://developer.mozilla.org/en-US/docs/Web/CSS/user-select

And add something like:

body, html {
user-select: none;
-webkit-user-select: none;
}

keforbes:
Wouldn't it be better to prevent `Ctrl+A` from performing any action at all?  The current behavior prevents `Ctrl+A` from being mapped to anything because Electron will perform the action first.  Similarly, `Ctrl+R` will perform Electron's `Reload` action rather than NeoVim's `Redo` command.

bert88sta:
We need to do a similar thing for Super key and prevent default on all the
bindings then.  They behave the same way but don't convert into ctrl-a.

extr0py:
Right, that's a good point. Seems like we might want to call `preventDefault` on all actions we handle by default.

extr0py:
Looks like this is fixed now :) Thanks @bert88sta !

PR #223, opened 2017-02-15
Prevented default on all modifier keys (temporary fix for 216)

bert88sta:
This is a temporary fix, but I think it's in the right direction for #216. Odds are we're going to only be whitelisting keys based on config, everything else should be blocked. This fixed command-a select for me and made ctrl-r work

Also this was an excuse to get rid of some gross whitespace

### DIFF
--- a/browser/src/Input/Keyboard.ts
+++ b/browser/src/Input/Keyboard.ts
@@ -19,9 +19,6 @@ export class Keyboard extends EventEmitter {
             // so we need to stop that.
             //
             // Later, the menu should be customized to fix this.
-            if (mappedKey === "<C-w>" || mappedKey === "<C-r>") {
-                evt.preventDefault()
-            }
         })
     }
 
@@ -38,10 +35,16 @@ export class Keyboard extends EventEmitter {
 
         if (evt.ctrlKey) {
             mappedKey = "C-" + vimKey + ""
+            evt.preventDefault()
         }
 
         if (evt.altKey) {
             mappedKey = "A-" + mappedKey
+            evt.preventDefault()
+        }
+
+        if (evt.metaKey) {
+            evt.preventDefault()
         }
 
         if (mappedKey.length > 1) {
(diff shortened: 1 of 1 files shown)


extr0py:
Great, thanks! The only downside that I can see is that I can't use `Control+Shift+I` to open the dev tools anymore - but it still works from the menubar, right?

bert88sta:
If you have any other white listed things I can put them in before this is
merged in,  just list them

extr0py:
That's the only one I can think of for now - thanks for whitelisting it

bert88sta:
@extr0py: Uhm, is there a different way to open it on mac? Which menu bar should it be under? I can't seem to open it with ctrl-shift-I, and that's after removing any of the code I put in originally.

extr0py:
Looks like it's `Cmd+Opt+I` on Mac - try that and see if they open. Sorry, the `Ctrl+Shift+I` is the windows shortcut.

bert88sta:
Actually as of right now I can't, and that's after removing all evt.preventDefault(). Is there a specific way I need to be building it?

keforbes:
> Which menu bar should it be under?

It's under View -> Toggle Developer Tools

bert88sta:
Perhaps this is another issue for another time, but I can't find it

screen shot 2017-02-17 at 8 43 16 am

keforbes:
Ahh, interesting, thanks for the screenshot.  We're defining the menus by telling Electron "generate your default menus" (which includes the default `View` menu) and then we override `menu[3]` with our `Window` items.  It looks like on Mac, `menu[3]` is the `View` menu, not `Window`.

It looks like there's the default `Electron` menu provided by Mac and then there's the `File` menu which on your system is also labeled `Electron`, throwing off the indexing and creating an off-by-one issue.

I wonder if all Mac users are seeing this behavior or if there's something screwy in your environment.  At the very least, I'd expect that second `Electron` menu to be labeled `Oni` since that's what the code is trying to do:

```
    let firstMenu = os.platform() == "win32" ? 'File' : 'Oni';
    menu.unshift({
         label: firstMenu,
```

bert88sta:
It's got to be something weird, because after leaving it open for a while and coming back to it, it's there... What I showed you was immediately after startup. I'd say ignore that for now, it might just be me. However, the issue I'm running into with this PR is getting the actual key combined with the modifier state. We might have to abuse key codes or something similar in order to catch key chords involving more than one modifier. Until then, the developer tools (for me at least) does work when invoked from the menu, and I can catch

extr0py:
I was able to repro this on Mac as well, it looks like there is some craziness there. I'm wondering if at this point, we should just scrap the default menu and use the custom built one we have.

This PR looks good though in the scope of fixing this bug. We can continue to iterate / improve the menu - it would be worth having a separate issue to track the Mac menu craziness. 

Thanks for the PR @bert88sta ! I'll merge it in.

PR #224, opened 2017-02-15
handle case where open/save dialog is cancelled

keforbes:
If you selected the `Save As` or any of the `Open` menu options and then hit `Cancel` on the corresponding file dialog, I was still sending the request to NeoVim with a filename of `undefined`.  This caused `Save As` to create a file named `undefined` rather than cancelling the operation.  I now check to see if a filename was selected before sending the request.


extr0py:
Good catch! Thanks for the fix!

PR #226, opened 2017-02-16
Tweak some menu labels

keforbes:
extr0py made a comment when I created the Window menu that these menus are good for people getting started with Vim.  That got me thinking about how some of these menu labels are really obtuse.  Even after executing some of them I couldn't tell the difference between some options.  So I decided to change the labels to be more explanatory.  Rather than "New" and "Split" it's "New Horizontal Split" and "Split File Horizontally".  Also, the menu was called "Window" when all of the operations affect splits.  So I changed it to "Split".  I think this makes it more intuitive.  `Split -> New Horizontal Split` makes much more sense to me than `Window -> New`.  I kept wondering if `Window -> New` was supposed to be creating a new window until I tried it in GVim and found that no, it really does just create an empty split.

Oh, I also added the `Move To` sub-menu which I missed last time.  So I guess this Pull Request isn't purely superficial.

### DIFF
--- a/main.js
+++ b/main.js
@@ -188,13 +188,13 @@ function createWindow(commandLineArguments, workingDirectory) {
            }
        },
        {
-           label: 'Put Before',
+           label: 'Paste Line Before',
            click: (item, focusedWindow) => {
                mainWindow.webContents.send("menu-item-click", "[p")
            }
        },
        {
-           label: 'Put After',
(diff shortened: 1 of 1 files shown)


extr0py:
Awesome, thanks for this @keforbes ! This definitely does make the intention of the menu items more clear. Will bring this in

Issue #225, opened 2017-02-16
Mixing settings


TalAmuyal:
I enabled the default configuration of Oni in addition to my own nvim conf.
As far as I understand, Oni's configuration take precedence.
I came to this conclusion, as only some of my configurations are enabled, while others aren't.
Am I right? and if so, how can I add a Vim configuration that take precedence over both?

Ideally, I would like to use Oni's default configuration as a starting point, and add my own / override specific behavior.

For example, currently I can't use Oni as it forces space indentation while the project I work on has tab indentation.

bert88sta:
In your config file for oni, have you set 

oni.loadInitVim: true

If you haven't, your init.vim is completely skipped over in terms of loading, and there might just be some happy overlap between yours and the default. If you're feeling brave, you can set both that setting and the default setting to true. I'll have to see the order the files are loaded in, as that would set the precedence for neovim.

TalAmuyal:
Yes, my `~/.oni/config.json`:

```
{
    "oni.useDefaultConfig": true,
    "oni.loadInitVim": true,
    "editor.completions.enabled": true,
    "editor.fullScreenOnStart": true
}
```

If `oni.useDefaultConfig` is set to `false`, the BG becomes black and some other coloring degrade, but Oni behaives as I expect nvim to.
If `oni.useDefaultConfig` is set to `true`, Oni behaves as Oni usually does (like the white-space approach described in the previous post), but some of my configurations are still in place (like YCM).

From looking a bit around the code I found that in `browser/src/NeovimInstance.ts` line 367 there is a function called `startNeovim` that starts Neovim and either lets it load files as usual or denies the loading of the configuration using the `-u` flag.

So I assume that the configurations that are Oni's default are loaded afterward (using my common sense only, so I might be wrong) and thus take precedence by overriding previous settings.

Is there a current method of cooperating with it?

extr0py:
Just added a long comment in #227 :) Thanks for all the details, it's very helpful!

Here's the TL;DR from the notes I put in the PR:
Does oni.useDefaultConfig set to false + setting the colorscheme in your init.vim give you the desired result? If so, I would lean towards that. And if that works, maybe we could update the documentation to be more clearer about the intended purpose / use of this setting?

extr0py:
Looks like this was addressed by #256, so closing this one. Feel free to open a new issue if there are still any concerns. Thanks!

PR #227, opened 2017-02-17
Support setting a prioritized vim-script in config.json

TalAmuyal:
Solves issue #225.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ A few interesting configuration options to set:
 - `oni.audio.bellUrl` - Set a custom sound effect for the `bell` (`:help bell`). The value should be an _absolute path_ to a supported audio file, such as a WAV file.
 - `oni.useDefaultConfig` - ONI comes with an opinionated default set of plugins for a predictable out-of-box experience. This will be great for newcomes to ONI or Vim, but for Vim/Neovim veterans, this will likely conflict. Set this to `false` to avoid loading the default config, and to load settings from `init.vim` instead (If this is false, it implies `oni.loadInitVim` is true)
 - `oni.loadInitVim` - This determines whether the user's `init.vim` is loaded. Use caution when setting this to `true` and setting `oni.useDefaultConfig` to true, as there could be conflicts with the default configuration.
+- `oni.prioritizedVimScript` - A path to a folder with scripts that will be loaded last and thus will override any other configuration. That can be useful if some of ONI's default configuration or `init.vim` are wished to be overridden. The simplest usage is to drop all your script files under `/my_path/plugin/` (assuming `oni.prioritizedVimScript` is set to `"/my_path"`).
 - `editor.fontSize` - Font size
 - `editor.fontFamily` - Font family
 - `prototype.editor.backgroundImageUrl` - specific a custom background image
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -33,6 +33,12 @@ const DefaultConfig: any = {
     // Set this to `true` to enable loading of init.vim.
     "oni.loadInitVim": false,
 
+    // In case both `useDefaultConfig` and `loadInitVim` are set to `true`, the user's init.vim has a lower
+    // priority in the case of a configuration conflict.
+    // Set this to a path of a vim script that will take precedence over both.
+    // For example, with `"oni.prioritizedVimScript": "~/my/special/place`, all the ~/my/special/place/plugin/*.vim" files will be loaded
+    "oni.prioritizedVimScript": "",
+
     // Sets the `popupmenu_external` option in Neovim
     // This will override the default UI to show a consistent popupmenu,
     // whether using Oni's completion mechanisms or VIMs
--- a/browser/src/Plugins/PluginManager.ts
+++ b/browser/src/Plugins/PluginManager.ts
@@ -53,6 +53,10 @@ export class PluginManager extends EventEmitter {
 
         this._rootPluginPaths.push(path.join(Config.getUserFolder(), "plugins"))
 
+        if (Config.hasValue("oni.prioritizedVimScript")) {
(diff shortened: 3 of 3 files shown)


keforbes:
Maybe I'm missing something, but is there a reason why we would ever want a user's vim scripts loaded **before** Oni's default set?  It seems to me we'd always want the user's custom scripts to be last.  I can understand having the config setting of "load init.vim or not" but I'd assume init.vim would always be loaded **after** the default config.

Again, it's possible I'm misunderstanding something here but if we now have settings for "scripts loaded before default set" and "scripts loaded after default set" then I think we should just throw away the "before" setting.

TalAmuyal:
Oni's default configurations are also loaded via the command arguments (--cmd set rtp+=...).

Maybe another option is to move the loading of the user's `init.vim` to `runtimepath` (rtp) and then control the order within it.

Though I think that a 3rd script has the additional benefit of custom Oni-specific configurations.

justinmk:
In my view init.vim should always be loaded, and the user may choose to opt-out of oni's default stuff it they want. That is, `.oni/config.json` should load after init.vim.

At least, that's the way we're planning for GUIs to work on the nvim side...

keforbes:
I'm inclined to agree.  If a user has an initi.vim defined, they want it loaded.  In my mind, the main use for the default Oni config is if a user doesn't have an initi.vim of their own.  Although, with that said, my init.vim is just a couple mappings so I still like getting Oni's "gui stuff" while keeping my mappings.

But @justinmk, why would that mean `.oni/config.json` should be loaded after init.vim?  Wouldn't we want init.vim to be loaded last so it has the final say on any values that are defined in both places?

justinmk:
> why would that mean .oni/config.json should be loaded after init.vim? Wouldn't we want init.vim to be loaded last so it has the final say on any values that are defined in both places?

The conventional way to provide "standard plugins" is to put them on the runtimepath. Then the user may set flags in vimrc (init.vim) to disable any plugins they don't want loaded.

Typically for mappings, a plugin should skip them by default if the user already defined something.

```vim
if mapcheck('foo', 'n') ==# ''
   ...
endif
```

Default options are more tricky. I guess they should be set before loading init.vim, until there is a better way for plugins to detect whether the user modified them. (It's possible, but awkward, using the `OptionSet` event)

TalAmuyal:
In order to avoid the collision, we could make the default configuration configurable.
That is, segmenting the `oni/vim/default/bundle/oni-vim-defaults/plugin/init.vim` file into chunks that can be enabled or disabled by the user in `config.json`.

What do you think?

extr0py:
Hey all, sorry I'm late to this thread! Great discussion and thanks for the PR.

This is really the intended use case for the `oni.useDefaultPlugins` flag - someone has their own `init.vim` that they want to bring and use rather than Oni's set of defaults. Turning this to `false` only disables the default config (stuff in `oni/vim/default`), but not Oni's UI integrations + javascript plugins, which live in `oni/vim/core`. 

There really isn't much to build on top of here:
`https://github.com/extr0py/oni/blob/master/vim/default/bundle/oni-vim-defaults/plugin/init.vim` -
as only the `completeopt` is really needed to make the popupmenu work in a predictable way for Oni... and given that, it probably belongs somewhere where it can't be turned off.

It looks like from the discussion in #225 that if `oni.useDefaultConfig` is `false`, things mostly work - except the colorscheme. If you set the colorscheme explicitly in your `init.vim` (you could do `colorscheme onedark` to get the default look), does that make Oni behave as you expect? We might have to move the `onedark` colorscheme to be in core instead of default, but if that works we might want to just go with that.

The loading sequence I'm picturing is:
- Load `init.vim` (default Neovim behavior)
- Load plugins in `runtimepath` (default Neovim behavior, but we also add to it to pick up some of Oni's plugins)
- Load the config in `config.js` (new Oni behavior - UI toggles as well as the loading behavior)

I'm thinking it's already a pretty complicated loading sequence as it stands, so I'd like to see if we could use the existing settings to make it work. 

So, TL:DR-
- Does `oni.useDefaultConfig` set to `false` + setting the `colorscheme` in your `init.vim` give you the desired result? If so, I would lean towards that. And if that works, maybe we could update the documentation to be more clearer about the intended purpose / use of this setting?

TalAmuyal:
I tried your suggestion:
Setting `oni.useDefaultConfig` to `false` doesn't quite do the trick, as that disables the loading of everything under `https://github.com/extr0py/oni/blob/master/vim/default/` (including the implementation of the color-schemes).

On the other hand, just removing the content of `https://github.com/extr0py/oni/blob/master/vim/default/bundle/oni-vim-defaults/plugin/init.vim` (except for `colorscheme onedark` and `set completeopt=longest,menu`) does yield the desired result.

How about moving the conflicting content (all except `colorscheme onedark` and `set completeopt=longest,menu`) to a different path for the `rtp` and make that disableable?

extr0py:
I'm thinking something along similiar lines - maybe we could do the following:
1) Move the `onedark` colorscheme from `default` to `core`
2) Move the `set completeopt=longest,menu` to the `init.vim` in `oni-plugin-interop`

Then, if you set `colorscheme onedark` in your own `init.vim`, and set the `oni.useDefaultPlugins` to `false` in `config.js`, I believe everything should work - and we won't need to change the config loaded pipeline or add additional complexity there. Let me know if there would be any issues with that - I'm happy to help put together a PR tomorrow if it seems like it would work for you (or if you'd like to make those changes, that'd be great too!)

extr0py:
If you grab that PR, and set `oni.useDefaultPlugins` to false + set `colorscheme onedark` in your init.vim, do things work as expected?

TalAmuyal:
Yes, as far as I can see they do.

extr0py:
Closing this PR since it looks like it was addressed by #256. Feel free to open a new issue if there is still something untracked here. Thanks!

PR #235, opened 2017-02-20
Add menu items for `Preferences`

extr0py:
Building on the work @keforbes and @Rafi993 did - I'm adding a `Preferences` submenu in file to make it easy to pull up the `config.js` (ONI config) or `init.vim` (Neovim config). 

These are also accessible via the command palette (Ctrl+Shift+P)

I also did a couple of minor refactorings:
- Moved the menu code to a separate js file, since it was getting pretty big
- Refactored the mainWindow.webContents.send to a helper method


browser/src/Config.ts
```diff
@@ -80,8 +80,8 @@ const DefaultPlatformConfig = Platform.isWindows() ? WindowsConfig : Platform.is
 
 Performance.mark("Config.load.start")
 
-const userConfigFile = path.join(getUserFolder(), "config.json")
-const userJsConfig = path.join(getUserFolder(), "config.js")
+export const userConfigFile = path.join(getUserFolder(), "config.json")
+export const userJsConfig = path.join(getUserFolder(), "config.js")
```
keforbes: Why do we have a `config.json` and a `config.js`?  Can we please come up with more unique filenames here?  I don't think the file extension should determine the difference between these two files.

Also, why is there a menu option to edit `config.js` but not one for editing `config.json`?  Maybe I'm missing something, I'm just confused.

extr0py: Not confused, totally valid point. `config.json` was added first, but `config.js` is like configuration v2... There were some things for my config that I wanted to use JavaScript for (for example, have a random background image on boot). We should remove `config.json`. It would be great to track in a separate issue and make sure that the documentation is up-to-date (there might still be references to `config.js`).

PR #238, opened 2017-02-20
229/buffer update -- Changed OniNotifyBufferUpdate

bert88sta:
Sorry for the dud PR from before.

This is a short and simple one, I just added some logic in init.vim to only send relevant parts of the buffer when updating


extr0py:
Thanks for looking at this, @bert88sta !

There's a few things I'm not sure how they will work with this PR:
1) Right now, when we receive the `buffer_update` event in UI (which happens in NeovimInstance.ts), it emits an event where the listeners expect all the buffer lines. We might have to add some context around which line numbers we are sending, in the case we send a subset. Right now we just send an array of lines, but maybe we should send something like { startLine, endLine, array of lines that changed }

2) It doesn't look like we're ever sending the full contents over. I'm thinking that, for the first change, we'd send over the entire file, but then, for incremental changes, we'd send over only the lines that changed, something like.

bert88sta:
@extr0py I've been using this for a little bit now and haven't managed to break anything, but I'll try to get into it and make the appropriate changes

extr0py:
The sort of issues you'd see here would be that completion or quick info might not work as expected. 

You might want to try setting a breakpoint in `vim\core\oni-plugin-typescript\src\index.ts` in the `Oni.on("buffer-update", ...)` section, around line 255 - and check out the buffer being passed in. It looks like what would happen here is that we'd only send a subset of the current file, but we'd be telling the language service provider that it is the whole file.

A scenario to try might be to open up `NeovimInstance.ts` and try typing `Neov` at the top and see if you get that completion picked up - I would guess it may not come up, because that wouldn't be in the buffer.

bert88sta:
@extr0py : I've definitely been able to find some of the glitchy cases here. I think this is going to have to go hand in hand with the adding of the Change event for TS Language Server.

extr0py:
Sounds good - thanks for the update

PR #239, opened 2017-02-21
Extr0py/add commands for plugins

extr0py:
- Add ability for plugins to add commands


extr0py:
Need to hook up filter, so that the right plugin is targetted for sending a command.

PR #243, opened 2017-02-22
Fix #241 - Remove `config.json` [BREAKING CHANGE]

### LINKED ISSUE #241: Remove `config.json` - just use `config.js`

extr0py:
Right now, our configuration story is extra-confusing because we support two types of Oni extensions - `config.js` and `config.json`. The `config.js` is more flexible, as it can be easily programmatically extended, so this is what we'll move forward with.

However, this is a __BREAKING CHANGE__ for anyone that has configuration in `config.json`.

The following steps will be needed:
- Rename `config.json` to `config.js`
- Change the top curly brace `{` to `module.exports = {`

Once that is done, Oni should work as expected

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "oni-vim",
-  "version": "0.1.8",
+  "version": "0.1.9",
   "description": "NeoVim front-end with IDE-style extensibility",
   "keywords": [
     "vim",
--- a/README.md
+++ b/README.md
@@ -159,11 +159,11 @@ Leave the cursor hovering over an identifier.
 
 ### Configuration
 
-> ONI is configurable via a 'config.json' located in $HOME/.oni
+> ONI is configurable via a 'config.js' located in $HOME/.oni
 
-Here's an example config.json:
+Here's an example config.js:
 ```
-{
+module.exports = {
     "oni.useDefaultConfig": true,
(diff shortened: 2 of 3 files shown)


extr0py:
FYI - @keforbes

keforbes:
Cool, thanks for cleaning this up!

keforbes:
I have a feature request about this but I'm not sure how we'd go about it...

When selecting the `Preferences -> Edit configuration` menu from Oni, it'd be nice if we seeded the file with:
```
module.exports = {
  //add custom config here, such as
  //"oni.useDefaultConfig": false
}
```

I don't know if it'd be better to ship with a default `~/.oni/config.js` with these contents or only add those contents when the user selects the option and the file doesn't exist.  This is similar to how Atom and VSCode behave and it helps the user understand the required format, especially since we now require the boilerplate string `module.exports`.

extr0py:
Thanks for the suggestion! I created this issue to track:
https://github.com/extr0py/oni/issues/252

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "oni-vim",
-  "version": "0.1.8",
+  "version": "0.1.9",
```
keforbes: Just a reminder, you might want to create a new tag for this release so GitHub sees it.

PR #248, opened 2017-02-22
move Developer Tools under Help menu

keforbes:
Most users won't care about displaying the DevTools so it shouldn't be given its own Menu.  Move DevTools into the Help menu instead.


extr0py:
Thanks @keforbes , good point. I have this also accessible via the command palette `Ctrl+Shift+P`, which is how I'm using it anyway.

PR #254, opened 2017-02-24
242 - fix and features

bert88sta:
Added editor.cursorLineOpacity and editor.cursorColumnOpacity to Config.ts, and toned down cursor line and column


extr0py:
Thanks @bert88sta ! Looks great

PR #255, opened 2017-02-24
Add config for debugging Oni in VSCode

keforbes:
I created a configuration for debugging Oni from within VSCode.  It will allow you to set breakpoints in VSCode and use the debugger.  Personally, I like having a separate application for my code/debugger rather than trying to use the built-in Developer Tools from Chrome, and VSCode has good support for typescript.

With that said, if you're personally offended that I would suggest people use VSCode to develop Oni then I'll close this Pull Request. :smile:  I was just happy to finally get this to work so I figured I'd share it; it's no problem if you don't want this committed.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,35 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            // Debug files that configure Electron (main.js, Menu.js, etc.)
+            "name": "Oni Main Process",
+            "type": "node",
+            "request": "launch",
+            "cwd": "${workspaceRoot}",
+            "program": "${workspaceRoot}/main.js",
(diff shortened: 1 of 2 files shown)


keforbes:
GitHub's diff viewer is highlighting my comments in red because the json specification technically doesn't allow for comments.  VSCode handles it just fine though.

extr0py:
Interesting :) I'm not offended at all, I'm actually a big fan of VSCode... I tried really hard to use it before starting Oni, but if their extensibility model was a bit more open, and it was easier to integrate Neovim, this project might not even exist.  Maybe if there was a debugger in Oni I'd lean towards leveraging that instead... but until that point this is great.

I'm glad that you have this set up to debug the Main process too - that's not easy to do at the moment.

PR #256, opened 2017-02-25
Move color schemes to core, and force completeopt even if settings ar…

extr0py:
- Clean up some of the overridable settings to make it easier to get the expected behavior with init.vim


TalAmuyal:
Works for me :)

extr0py:
Great! Thanks for the confirmation, @TalAmuyal ! I'll bring this in.

PR #257, opened 2017-02-25
WIP: Performance - Incremental Buffer Updates in insert mode

extr0py:


extr0py:
This results in **much** improved / faster auto completion for the typescript language service. Re-parsing the entire file for each keystroke is non-trivial...

PR #258, opened 2017-02-26
Added check and method for neovim api info

bert88sta:
If this needs any changes, let me know

### DIFF
--- a/browser/src/NeovimInstance.ts
+++ b/browser/src/NeovimInstance.ts
@@ -128,10 +128,8 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                             this.emit("window-display-update", args[0][0], args[0][1])
                         } else if (pluginMethod === "api_info") {
                             const apiVersion = args[0][0]
-                            if (apiVersion.minor < 1) {
-                                if (apiVersion.patch < 6) {
-                                    alert("Please upgrade to at least Neovim 0.1.6")
-                                }
+                            if (apiVersion.api_level < 1) {
+                                alert("Please upgrade to at least Neovim 0.1.6")
                             }
 
                         } else {
@@ -170,7 +168,7 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
 
                     // set title after attaching listeners so we can get the initial title
                     this.command("set title")
-                    this.callFunction("OniApiInfo",[])
+                    this.callFunction("OniApiInfo", [])
                 })
             }, (err) => {
                 this.emit("error", err)
--- a/browser/src/NeovimInstance.ts
+++ b/browser/src/NeovimInstance.ts
@@ -126,6 +126,14 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                             this.emit("event", eventName, eventContext)
                         } else if (pluginMethod === "window_display_update") {
                             this.emit("window-display-update", args[0][0], args[0][1])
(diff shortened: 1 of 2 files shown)


browser/src/NeovimInstance.ts
```diff
@@ -126,6 +126,14 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                             this.emit("event", eventName, eventContext)
                         } else if (pluginMethod === "window_display_update") {
                             this.emit("window-display-update", args[0][0], args[0][1])
+                        } else if (pluginMethod === "api_info") {
+                            const apiVersion = args[0][0]
+                            if (apiVersion.minor < 1) {
```
kopischke: If you target API version 0.1.6 or above, shouldn't this be `< 2`?

kopischke: … also, thinking about it, shouldn't this be able to handle the case the Neovim API hits a major version > 0? I.e.
```javascript
if (apiVersion.major < 1 && apiVersion.minor < 2 && apiVersion.patch < 6) {
  // handle incompatible version
}

Comment (justinmk) [inline browser/src/NeovimInstance.ts]: Only the API level should be checked (that's the entire purpose of having an API level). The nvim major/minor versions are irrelevant regarding API compatibility.

Comment (justinmk) [inline browser/src/NeovimInstance.ts]: Also these fields didn't exist in api_info result before 0.1.6.

Comment (kopischke) [inline browser/src/NeovimInstance.ts]: @justinmk sorry, I didn't follow the whole API level versioning discussion closely : the version numbers are not semantic? Then my comments are academic, if course: true as to pure code logic but irrelevant to the issue at hand ;).

Comment (bert88sta) [inline browser/src/NeovimInstance.ts]: I'll go in and fix this later, it was pretty late when I wrote this

**Comment by justinmk** [comment]
@kopischke the version numbers are semantic. The API level is more correct for comparing API capabilities. API level might not change between Nvim releases.

File: browser/src/NeovimInstance.ts [hunk 2]
```diff
@@ -126,6 +126,12 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                             this.emit("event", eventName, eventContext)
                         } else if (pluginMethod === "window_display_update") {
                             this.emit("window-display-update", args[0][0], args[0][1])
+                        } else if (pluginMethod === "api_info") {
+                            const apiVersion = args[0][0]
```
justinmk: The fields won't exist at all in older versions

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -113,3 +113,13 @@ function OniUpdateWindowDisplayMap()
...
     call OniNotify(["window_display_update", context, mapping])
 endfunction
+
+function OniApiInfo()
+    if (has_key(api_info(),'version'))
+        call OniNotify(["api_info",api_info()['version']])
+    else
+        call OniNotify(["api_info",{api_level:0}])
+    endif
+
```
extr0py: Could you please remove these extra newlines? No linting set up for us in VimL :)

extr0py:
Looks good except for some minor whitespace issues in `init.vim`. I'll pull this in once that's addressed. Thanks for the PR, @bert88sta !

extr0py:
Thanks @bert88sta for the fix, and @kopischke & @justinmk for the feedback!

PR #260, opened 2017-02-28
WIP: 182 - Implement `Find all References` for typescript language service

extr0py:


browser/src/index.tsx
```diff
@@ -251,6 +251,10 @@ const start = (args: string[]) => {
             return
         }
 
+        if (key === "|") {
+            pluginManager.findAllReferences()
+        }
```
keforbes: Can I ask why you picked `|` as the key to invoke this feature?  Is this a common key to invoke find-all-references?  I'm surprised you'd map this key directly given that `|` is already mapped by default in NeoVim.

extr0py: Good question, I actually just picked something quick and easy for testing. 

Do you have any ideas for a better default key for this? Another option would be to have it exposed in the app menu.

keforbes: Does it have to be a single character?  I'd probably go for something like `g|` since `g` is usually used for the weird extra commands.  That probably wouldn't help you for this method though.

Do we have to map it to a key now?  Could we add it to the command palette instead and then provide a config for mapping keys to Oni commands?  That'd be the more extensible solution so users could map any other commands in the palette to arbitrary keys.

browser/src/index.tsx
```diff
@@ -128,6 +128,10 @@ const start = (args: string[]) => {
         liveEvaluationOverlay.setLiveEvaluationResult(file, blocks)
     })
 
+    pluginManager.on("find-all-references", (references: any) => {
+        alert(references)
```
extr0py: Still need to finish wiring this up

vim/core/oni-plugin-typescript/src/index.ts
```diff
@@ -63,6 +45,13 @@ export const activate = (Oni) => {
             })
     }
 
+    const findAllReferences = (textDocumentPosition: Oni.EventContext) => {
+        return host.findAllReferences(textDocumentPosition.bufferFullPath, textDocumentPosition.line, textDocumentPosition.column)
+            .then((val: any) => {
+                return null
```
extr0py: ...and this

extr0py:
Thanks @bert88sta and @keforbes for the suggestion. I've removed the command for now and made this only accessible via the command palette (`Ctrl+Shift+P`).

PR #262, opened 2017-02-28
Fix #250, set menu per window on Windows and Linux

### LINKED ISSUE #250: Cross-communication between multiple instances of Oni
I don't know if this is related to #247 or not so I'm filing it as a new issue.  I've noticed that if I have multiple instances of Oni running, only the most recently launched Oni will display DevTools.  It looks like there's some cross-communication going on between instances and only the last DevTools listener is active.  It's possible there are other listeners that are mixed up as well but the DevTools one is the most noticeable to me.

multiple_instances

keforbes:
`Menu.setApplicationMenu(menu)` is for use on macOS where all windows of an application share the same menu.
https://electron.atom.io/docs/api/menu/#menusetapplicationmenumenu

`BrowserWindow.setMenu(menu)` is for use on Windows and Linux where each window is given its own menu.
https://electron.atom.io/docs/api/browser-window/#winsetmenumenu-linux-windows

This fixes #250 for me on Linux but I don't have a Mac to test it on so I'll just assume the previous line still works as desired.  I don't know if #250 was ever an issue on Mac.

### DIFF
--- a/main.js
+++ b/main.js
@@ -64,7 +64,13 @@ function createWindow(commandLineArguments, workingDirectory) {
     let mainWindow = new BrowserWindow({ width: 800, height: 600, icon: path.join(__dirname, "images", "Oni_128.png") })
 
     const menu = buildMenu(mainWindow)
-    Menu.setApplicationMenu(menu);
+    if (process.platform === 'darwin') {
+        //all osx windows share the same menu
+        Menu.setApplicationMenu(menu)
+    } else {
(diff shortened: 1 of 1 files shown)


extr0py:
Nice find! Setting it per `BrowserWindow` makes sense. I'll test #250 on a Mac when I get a chance - I imagine it may have a similiar issue (last menu wins). But this seems like it would improve things for Windows and Linux. Thanks @keforbes !

PR #263, opened 2017-03-01
Fix #261: Remove writing to log file from process

### LINKED ISSUE #261: Linux, can't access out.log after 'npm install -g'
This is specific to Linux but I don't think it's specific to my environment.  `npm install -g` requires superuser permission to install into `/usr/lib/` so I have to run `sudo npm install -g oni-vim`.  This is fine, it's expected of any `npm -g` command.

The problem is that after installing oni, it tries writing to a log file which still requires superuser permission.
```
Error: EACCES: permission denied, open '/usr/lib/node_modules/oni-vim/cli/out.log'
```
If I run `sudo oni` then it'll have permission to access the `out.log` file and launches fine.

Or, I can fix this by running
```
$ sudo touch /usr/lib/node_modules/oni-vim/cli/out.log
$ sudo chmod 777 /usr/lib/node_modules/oni-vim/cli/out.log
```

There are a couple things I think we could do here:
1. Don't write to out.log by default.  On a "good" execution, nothing is written to this log so it isn't needed in the majority use case.  We could maybe add a command-line flag to write to a log file when desired.  If we took this route, the command-line argument could just take the absolute path of where to write the log file and bypass this whole issue.
2. At "install-time" create this log file and set its permissions.  I'm not sure how to hook into the install process for Oni but I assume this would be the easier approach.

extr0py:
Remove creating `out.log` when starting Oni - there isn't much interesting that goes in there, anyway.

### DIFF
--- a/cli/oni
+++ b/cli/oni
@@ -6,11 +6,8 @@ var os = require("os")
 var fs = require("fs")
 
 var processOptions = {}
-var log = require("path").join(__dirname, "out.log")
-var out = fs.openSync(log, "a")
-var err = fs.openSync(log, "a")
 processOptions.detached = true
-processOptions.stdio = ["ignore", out, err]
+processOptions.stdio = ["ignore", null, null]
 
 var isWindows = os.platform() === "win32";
 


extr0py:
FYI - @keforbes

keforbes:
Removing the log entirely is definitely the easiest solution. 😄  If the log file never really provides any benefit then this is probably the right course of action.

PR #266, opened 2017-03-02
Fix #252, create default config.js if one doesn't already exist

### LINKED ISSUE #252: Add default `config.js` when going to Preferences -> Edit configuration
@keforbes had a good idea with this PR:
https://github.com/extr0py/oni/pull/243

When we first edit the configuration, and the configuration doesn't exist, we should add a template / placeholder that shows the format, so that it is easy to modify.

Later enhancements could be to add the default settings in there, and add custom completion for that file.

keforbes:
As I requested in #243, if the user selects `Edit Oni Config` from the Menu or Command Palette and the file doesn't already exist, create some default content showing the format.

### DIFF
--- a/browser/src/Services/Commands.ts
+++ b/browser/src/Services/Commands.ts
@@ -7,6 +7,7 @@
 import { remote } from "electron"
 
 import * as Config from "./../Config"
+import { IBuffer } from "./../neovim/Buffer"
 import { INeovimInstance } from "./../NeovimInstance"
 import { PluginManager } from "./../Plugins/PluginManager"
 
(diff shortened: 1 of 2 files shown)


browser/src/Services/Commands.ts
```diff
@@ -22,9 +23,27 @@ export const registerBuiltInCommands = (commandManager: CommandManager, pluginMa
         new CallbackCommand("oni.editor.gotoDefinition", "Goto Definition", "Goto definition using a language service", () => pluginManager.gotoDefinition()),
 
         // Menu commands
-        // TODO: Generate config.js if not already built
-        new CallbackCommand("oni.config.openConfigJs", "Edit Configuration", "Edit configuration file ('config.js') for ONI", () => neovimInstance.open(Config.userJsConfig)),
-        new CallbackCommand("oni.config.openInitVim", "Edit Neovim Configuration", "Edit configuration file ('init.vim') for Neovim", () => neovimInstance.open("$MYVIMRC")),
+        new CallbackCommand("oni.config.openConfigJs", "Edit Oni Config", "Edit configuration file ('config.js') for ONI", () => {
+            let buffer: null | IBuffer = null
+            neovimInstance.open(Config.userJsConfig)
```
extr0py: Cool! Thanks for adding this!

extr0py:
Thanks @keforbes ! This is very helpful in terms of lowering the bar for customization and modifying the config. Looks great!

Issue #177, opened 2017-01-20
Why is oni scanning my ~/.Trash ?


jordwalke:
When I started oni and used the command pallet (command shift p), it gave me a message:

```
Error: Error: ENOENT: no such file or directory, stat '/Users/me/.Trash/SomeFile 5.09.00 PM'
```

Why is it scanning the trash in the first place when opening the command pallet? I think it should have a blacklist of directories perhaps. Or maybe this should be included in wildignore and oni should ignore it?

extr0py:
Seems like it might be related to #125 (at least, the error message).

Right now,  in a non-git folder, we recursively read all directories / files for the 'quick open' service. Having a blacklist sounds good - for example, VSCode has the `files.exclude` setting in its config, would make sense to have a similiar option. 

I like the idea of linking it to the whitelist property, too - seems like #20 might apply in terms of how we synchronize the VIM/JS configuration & keybindings.

Regarding the configuration, a reasonable set of defaults could be:
```
["**/.git", "**/.svn", "**/.hg", "**/.DS_Store", "**/.Trash"]
```

keforbes:
I believe this defect can be closed now that #268 has been merged in.

extr0py:
Thanks @keforbes ! I'll close this since the fix is in now. @jordwalke , feel free to reopen if you are still seeing issues.

jordwalke:
Awesome!

Issue #125, opened 2016-12-25
QuickOpen crash


bert88sta:
I get this error:

Error: Error: ENOENT: no such file or directory, stat '/Users/Bret/.wine/dosdevices/d:'

running from a non-git (my home) directory, I'd assume there's some sort of issue with how oni handles not being able to open certain files.

extr0py:
I'm able to reproduce too with my non-git home directory (just a different file shows up EBUSY: hiberfile.sys). 

Seems like the `recursive-readdir` library we're using at the moment will fail as soon as it hits an issue, and not return a list of files. Couple of options for fixing this:
1) Use an alternative library that handles these cases more robustly (`shelljs` might be a good choice - I'd like to bring this in for some other scenarios as well - it has an ls command that might work better)
2) Use a default command for UNIX / Windows that doesn't require an additional node library

Also, we might want to update the Menu component to handle asynchronicity better - right now, we synchronously populate the menu which blocks. It'd be nice if we showed a spinner or something and items as they populate. I'll log a separate task for this shortly.

keforbes:
> Seems like the recursive-readdir library we're using at the moment will fail as soon as it hits an issue, and not return a list of files.

Someone created a [fix](https://github.com/jergason/recursive-readdir/issues/42#issuecomment-269523863) for this but the author of `recursive-readdir` [refuses](https://github.com/jergason/recursive-readdir/pull/43#issuecomment-264189392) to merge it in.  I tried following the directions for manually installing that person's `#patch-2` but my `npm` just hangs when I try installing it.

I tried getting `shelljs` to work but it wouldn't connect to neovim.  I didn't have any compilation errors and everything looked like it should've worked but I would just get an empty electron window when I tried launching oni.  Looks like I was hitting an error in `bundle.js` for:
```
require('./commands.json').forEach(function (command) {
```
saying "`__webpack_require__(...).forEach is not a function`" though I have no idea how replacing `recursive-readdir` with `shelljs` could lead to that error.

So yeah, I guess I just wanted to complain.  I'm still playing with this.

keforbes:
It looks like the issue I hit with `shelljs` is a known issue in the current version.  It'll be fixed in the next version in case you're still interested in using `shelljs` in oni.
https://github.com/shelljs/shelljs/issues/667

keforbes:
I believe this defect can be closed now that #268 has been merged in.

PR #268, opened 2017-03-03
use glob rather than recursive-readdir to fetch files in Fuzzy Finder

### LINKED ISSUE #267: Ctrl-P error ENOENT
When I try to push ctrl-p I get a pop-up with this error:
`Error: Error: ENOENT: no such file or directory, stat '/home/nikolay/.config/chromium/SingletonCookie`
and nothing else happens.

keforbes:
This should fix #267, #177, and #125.  `recursive-readdir` has a known issue with broken symlinks but it doesn't look like that defect will ever be fixed.  Use the [node-glob](https://github.com/isaacs/node-glob) library instead.  This library ignores all hidden (`.*`) directories by default so I just added a default ignore of `node_modules` since that directory always has thousands of files in it.  We can tell it to include hidden directories and then explicitly exclude things like `.git`, `.svn`, `.Trash` if we want but I'm not sure if anyone would ever want to look inside hidden directories so this default seems sensible to me.

Also, I added an animated "loading files" indicator in case this takes a while.  The operation is pretty quick after I excluded `node_modules` but I'm sure there will still be still be users with large directories.

### DIFF
--- a/browser/src/Services/QuickOpen.ts
+++ b/browser/src/Services/QuickOpen.ts
@@ -7,17 +7,15 @@
 import { execSync } from "child_process"
 import * as path from "path"
 
+import * as glob from "glob"
 import * as _ from "lodash"
 import * as Q from "q"
-import * as recursive from "recursive-readdir"
 
 import * as Config from "./../Config"
 import { INeovimInstance } from "./../NeovimInstance"
 import * as PromiseHelper from "./../PromiseHelper"
 import * as UI from "./../UI/index"
 import * as Git from "./Git"
-
-const recursiveQ = Q.denodeify<string[]>(recursive)
 
 export class QuickOpen {
     private _seenItems: string[] = []
@@ -39,6 +37,13 @@ export class QuickOpen {
 
     public show(): void {
         const overrriddenCommand = Config.getValue<string>("editor.quickOpen.execCommand")
+
+        UI.showPopupMenu("quickOpen", [{
+            icon: "refresh fa-spin fa-fw",
+            label: "Loading Files...",
+            detail: "",
+            pinned: false,
+        }])
 
         // Overridden strategy
         if (overrriddenCommand) {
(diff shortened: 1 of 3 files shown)


browser/src/Services/QuickOpen.ts
```diff
@@ -40,6 +38,13 @@ export class QuickOpen {
     public show(): void {
         const overrriddenCommand = Config.getValue<string>("editor.quickOpen.execCommand")
 
+        UI.showPopupMenu("quickOpen", [{
```
extr0py: Cool idea here to add a loading indicator.

extr0py: I logged #269 to build on this and better support async in general with Ctrl-P - it'd be nice to have a gradual loading experience, where you can search results as they are being populated.

browser/src/Services/QuickOpen.ts
```diff
@@ -62,10 +67,12 @@ export class QuickOpen {
                         })
                 } else {
                     // TODO: This async call is being dropped, if we happen to use the promise
-                    return recursiveQ(process.cwd())
-                        .then((files: string[]) => {
-                            this._showMenuFromFiles(files)
-                        })
+                    return glob("**/*", {
```
extr0py: Later, for #269, we could use the `match` event to incrementally populate the list.

browser/src/Services/QuickOpen.ts
```diff
@@ -7,18 +7,16 @@
 import { execSync } from "child_process"
 import * as path from "path"
 
+import * as glob from "glob"
```
extr0py: Thanks for finding a replacement for `recursive-readdir`. It's a bummer they never wanted to take that PR / fix the issue around handling errors...

browser/src/Services/QuickOpen.ts
```diff
@@ -62,10 +67,12 @@ export class QuickOpen {
...
                     // TODO: This async call is being dropped, if we happen to use the promise
-                    return recursiveQ(process.cwd())
-                        .then((files: string[]) => {
-                            this._showMenuFromFiles(files)
-                        })
+                    return glob("**/*", {
+                        nodir: true,
+                        ignore: ["**/node_modules/**"], // all hidden dirs (start with '.') are skipped by default
+                    }, (_err: any, files: string[]) => {
+                        this._showMenuFromFiles(files)
```
extr0py: If there is an error, files might be empty, and then QuickOpen crashes for me. Perhaps we should validate we actually have an array for `files` somewhere

extr0py:
Thanks for this fix, @keforbes ! The loading indicator is a huge improvement.

PR #272, opened 2017-03-05
Extr0py/code climate

extr0py:
- Add code climate statistics


extr0py:
TypeScript support is just  a wrapper around tslint... so abandoing for now.

PR #275, opened 2017-03-05
Fix #265 - Enable clicking on the `QuickOpen` and `Command` palettes

### LINKED ISSUE #265: Enable clicking on items in Fuzzy Finder and Command Palette
If you try clicking on an entry in the Fuzzy Finder or Command Palette, the click event is only registered by NeoVim (moving the cursor).

click

extr0py:
- Refactor the `selectMenuItem` action to an `ActionCreator`, so that it can be hooked up to the UI via mapStateToDispatch
- Add `redux-thunk`, because the action creator is conditional based on some info in the state
- Wire up the click events through `Menu` and `MenuItem`
- Add CSS styles for hover and mouse over
- Temporarily disable pointer events on the `overlays` (errors, buffer scrollbar), since they don't handle mouse events today, they are rendered on top of all other UI. Long-term, we'll need to move this to a more reasonable location in the DOM (refactor to a component)

### DIFF
--- a/browser/src/UI/ActionCreators.ts
+++ b/browser/src/UI/ActionCreators.ts
@@ -7,8 +7,8 @@
  * http://redux.js.org/docs/basics/Actions.html
  */
 
-import { Rectangle } from "./Types"
 import * as Events from "./Events"
+import { Rectangle } from "./Types"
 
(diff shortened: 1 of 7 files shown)


extr0py:
Fix #265

PR #278, opened 2017-03-06
add editor.exclude config for Fuzzy Finder

keforbes:
Add support for `editor.exclude` in config.js to affect Fuzzy Finder.  Default is still `**/node_modules/**`.  This works great outside of git repositories and I'm *pretty* sure it works within git repositories.  The exclude list for `git-ls` only applies to untracked files and I'm still including `--exclude-standard` which includes the whole `.gitignore` file so I don't have too many files left to try excluding.


extr0py:
Awesome! Thanks @keforbes !

Issue #166, opened 2017-01-16
Add scrollbar to Menu & Code Completion


extr0py:
Currently, there is no scrollbar, and no limit to the number of items shown in completion (and you can't really navigate beyond the bounds with <C-n> & <C-p>).

To address this, need to do the following:
- Add a `view` concept to the reducers - this would include the view size and the start position of the view
- Update the `next` and `prev` actions to adjust the view (for example, if you are at the top of the current view, it would move the view upwards when dispatching `prev`).
- Add a scrollbar component to the UI to display this

Later:
- Add mouse events to the scrollbar component

bryphe:
Would be great to look at integrating a library like `react-virtualized` for this.

PR #280, opened 2017-03-07
Fix wrapping around first/last item in menu

keforbes:
This Pull Request fixes two specific issues in the Fuzzy Finder and Command Palette:

1. When on the first item, hitting `Ctrl+P` to go to the previous item would run `selectedIndex -1 % length` but for whatever reason, in javascript, `-1 % x` equals `-1` and not `x-1`.  So rather than wrapping around to the last item in the list it was setting `selectedIndex` to `-1` and the highlight would just disappear.  Every subsequent `Ctrl+P` would just keep decrementing to `-2`, `-3`, etc.
2. When on the last item, hitting `Ctrl+N` to go to the next item would run `selectedIndex +1 % length`.  Unfortunately, the `length` property was based on the full list of options, not the number of options currently displayed.  So the highlight would disappear until you hit `Ctrl+N` enough times to wrap around the *actual* list.

To fix these issues, I did the following:
1. If `selectedIndex = 0`, just set the `selectedIndex` to `length - 1`.
2. Set `length` to `Math.min(10, length)` so if there are more than 10 entries in the list we'll just loop over the 10 visible entries.

There's one thing I don't like about this fix.  `Menu.tsx` is hardcoded to just display the first 10 items but in `Reducer.ts`, where we're handling the next/previous operation, we don't know how many items `Menu.tsx` truncated the list to.  I was hoping to set a variable that was accessible by both classes so they could stay in sync but `Menu.tsx` has a `IMenuProps` and `Reducer.ts` has a `IMenu`.  I couldn't find a way to access a `maxDisplayItems` in both classes.  So I just hardcoded `10` in `Reducer.ts` with a big ugly `TODO`.  If you have a way to access a variable in both of those classes please let me know so I can clean it up.

### DIFF
--- a/browser/src/UI/Reducer.ts
+++ b/browser/src/UI/Reducer.ts
@@ -91,6 +91,9 @@ export const reducer = (s: State.IState, a: Actions.Action) => {
 
 export const popupMenuReducer = (s: State.IMenu | null, a: Actions.Action) => {
 
+    // TODO: sync max display items (10) with value in Menu.render() (Menu.tsx)
+    let size = s ? Math.min(10, s.filteredOptions.length) : 0
+
     switch (a.type) {
         case "SHOW_MENU":
             const sortedOptions = _.sortBy(a.payload.options, (f) => f.pinned ? 0 : 1).map((o) => ({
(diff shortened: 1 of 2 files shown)


keforbes:
I guess I should mention that this fix is only relevant until #166 is implemented.  If the menu was changed to display all items with a scrollbar then there wouldn't be a difference between the number of "displayed" items and the number of "actual" items.

extr0py:
Thanks for the fix, @keforbes . This looks like a better experience until #166 is implemented... it doesn't make sense for it to just disappear when going upwards. Glad you caught that bug.

Regarding the hardcoded `10`, I think the TODO is reasonable for now, since it would go away once #166 is fixed.

Issue #163, opened 2017-01-14
Unable to scroll auto complete list beyond limit on autoCompletion


Rafi993:
I opened simple code snippet when and tried auto complete. I am unable to scroll beyont the limit to see other functions in suggestion. 
![screenshot from 2017-01-14 12-18-37](https://cloud.githubusercontent.com/assets/8092120/21953072/f85cb9a8-da53-11e6-937b-9eb34dd10c80.png)

In the above screen shot I was not able to scroll beyond **and** snippet in the auto completion.

Rafi993:
I was able to solve this issue by removing FirstTen in Autocompletion.tsx that used _.take to get the first ten entries

extr0py:
Thanks @Rafi993 ! Yes, I think we need to add a scrollbar + a way to manage the active view of these components (the Menu also has this same problem).  Appreciate the PR!

extr0py:
I realized why the firstTen was used now... There is some expensive logic that iterates over all the entries at the moment. We'll need another way of filtering them and a different scrollbar mechanism.

Rafi993:
@extr0py I agree. The scroll in auto completion feels clunky re-writing that module would be better.

bryphe:
Seems like the same issue as #166 - I'll resolve this as a duplicate.

PR #285, opened 2017-03-09
Fix wrapping around first/last item in autocomplete

keforbes:
Same basic concept as #280.  I didn't realize we had the same logic in autocomplete or I would've fixed them both at the same time.


extr0py:
Awesome! I'm hoping as part of #163 we can address some of the duplication between the autocomplete / menu - it's essentially the same operation so we should be able to consolidate it a bit. Thanks for fixing this!

PR #286, opened 2017-03-09
Reduce duplication in .less files

keforbes:
I noticed there were hardcoded colors in some `.less` files that were already defined in `common.less` so this is just some cleanup to use the variable from `common.less` instead.  The only visual change I made was to the `detail` color in the menu and autocompletion.  I think the detail text was too dim so I got rid of the alpha value but kept the color the same.

### DIFF
--- a/browser/src/UI/components/AutoCompletion.less
+++ b/browser/src/UI/components/AutoCompletion.less
@@ -1,9 +1,10 @@
+@import (reference) "./common.less";
 
 .autocompletion {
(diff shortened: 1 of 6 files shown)


keforbes:
Ok, one more visual change.  In the menu, make the mouse hover color slightly dimmer than the selected item color so you can tell the difference.

extr0py:
Thank you for cleaning this up! The changes you made to the detail text and hover color look great - it's a subtle but significant improvement in the completion / popup menu.

PR #288, opened 2017-03-10
Fix for 287

bert88sta:
Sorry, just me being an idiot as usual. I forgot the quotes. -_-


extr0py:
Error message looks better now! On Windows, I'm still seeing the error message. The binary I'm included is at 0.1.6, but it doesn't seem to have version defined on api_info(). Need to investigate more this weekend...

Issue #287, opened 2017-03-10
Error: 0, Vim(call):E121: Undefined variable: api_level


extr0py:
Seeing an error on launch in Windows:
![image](https://cloud.githubusercontent.com/assets/13532591/23804355/6b7a8ce0-056e-11e7-96cc-7e504cefb12f.png)

I think this is related to the recent api_level change made by @bert88sta 

I wonder if the packaged neovim binary for Windows is older - the potential fix could be either to revise the logic, or upgrade our neovim binary.

bert88sta:
@extr0py : Nope, i'm just forgetting quotes when I send back {api_level:0}, check #288

extr0py:
With #288, the error message is improved, but it's confusing on Windows - neovim 0.1.6 is included in the package, but I still get the error. I need to check and see if I should upgrade the binary or if the version in api_info is coming across differently

justinmk:
It shouldn't have changed. But I definitely recommend using the latest binary, tons of fixes for Windows in 0.2

extr0py:
Thanks @justinmk, look forward to testing out these fixes!

extr0py:
Fixed by #289

extr0py:
Clipboard integration seems to work much better in 0.2.0 on Windows... awesome!

Issue #123, opened 2016-12-23
Build windows version of Neovim as part of Oni


bryphe:
Right now, we use a prebuilt Neovim binary, but have no pipeline to build Neovim from source for Windows. It'd be helpful to hook that up as part of the build chain, so that we could add the `gui_oni` feature.

CrossR:
Since #875 was merged in, this (and #326) can be closed I think, and we can make a separate more focused ticket on the final part of this which is to add a `gui_oni` feature. 

From the comments on #119 and browsing the VimR neovim fork, the only real thing to add is a single ifdef to `eval.c` and pass a flag for that to cmake, and then the feature should be baked in.

I've tested this on my fork of neovim and it does indeed bake in the `gui_oni` feature, allowing my `init.vim` to swap to using `if has ('gui_oni')` instead of `if exists('g:gui_oni')` as it is currently.

So we could get a fork of neovim into the org with the feature added, which is easy enough. Creating the binaries may be a little more tricky, but thats just because I've not had to make the binaries for Windows / Mac before, which I assume someone else has! Then as far as I understand it, if the binaries are in the same form, we can just point the `oni-neovim-binaries` script at the new release in our fork. 

From there we can start to look into the more interesting bits, I see that VimR also has a few tiny changes to neovim (https://github.com/neovim/neovim/compare/master...qvacua:master), or I know that a web-assembly build was mentioned too.

bryphe:
Thanks for the feedback & investigation here, @CrossR ! 

> Since #875 was merged in, this (and #326) can be closed I think, and we can make a separate more focused ticket on the final part of this which is to add a gui_oni feature.

Sounds good - perhaps it even makes sense to make a ticket on the `oni-neovim-binaries` repo? It sounds like you've figured out where we'd need to add it.

> So we could get a fork of neovim into the org with the feature added, which is easy enough. Creating the binaries may be a little more tricky, but thats just because I've not had to make the binaries for Windows / Mac before, which I assume someone else has! 

Same here, I haven't actually tried to build on Windows 😄 Much less set up appveyor for it.

I haven't heard too many people asking for this recently (perhaps the `g:gui_oni` is sufficient), so it might make sense to hold off for now. Especially since you added the capability to load a custom `init.vim` for Oni, which helps here too.

We can bring it back as an issue on the `oni-neovim-binaries` repo though if it is something that would be useful. I'll close this one for now, and we can open a new issue to track the `gui_oni` feature if necessary.

Issue #99, opened 2016-12-15
Include pre-built binary of Neovim for OSX


extr0py:
For Windows, in order to make installation more convenient, a pre-built binary of Neovim is included. We could do the same for OSX.

extr0py:
Need to hook up an OS X build of Neovim, and add the `gui_oni` feature (so `has("gui_oni")` returns `1`)

extr0py:
Using #326 to track real OSX build.

PR #289, opened 2017-03-12
Upgrade Neovim binaries to 0.2.0-828-g11b08bb0

extr0py:


extr0py:
Fix #287

justinmk:
I'm surprised the nvim tree is copied to oni's source tree. Any plans to do that at build-time instead? Or instead expect the user to have an existing nvim?

extr0py:
Yes, for sure, I'd prefer to make this build-time instead - I'd like to build and package both the Windows and OSX binaries so it is a one-step process for install (like VimR) on those platforms. 

I have #123 tracking this - unfortunately I just haven't had time to pursue it in-depth, so I took the shortcut of checking in the built binaries.

extr0py:
Also #99 for the OSX pre-built binary.

Issue #279, opened 2017-03-06
Get matching package name `oni`, or reserve + update name


extr0py:
There is lots of confusion around this repo being `oni` and the npm package being `oni-vim`.

We should look at ways to match these up. Some discussion was tracked in #277 

A couple ideas:
- See if the owner for `oni` on npm would be willing to give us the package name
- Update the name to something that isn't already reserved.

Let me know if you have ideas!

bert88sta:
@extr0py is your npm username also extr0py?

extr0py:
Yep, same name on npm.

I started looking at `oni` vs `onivim` and it seems like in general `onivim` is more available:
1) `onivim.io` is available 
2) Subreddit is available: http://www.reddit.com/r/onivim
3) npm package `onivim` is available

I snagged all of those above :)

For the next release (likely `0.1.11`), I'm planning on making the switch.  If we stick with just `oni`, it'd be harder to get the rest of those collateral. Let me know if you have any concerns @keforbes @bert88sta

bert88sta:
Oh, I talked to the guy who owns oni and he said he'd just give it to you

keforbes:
If the guy who owns oni on npm is willing to give it to us I'd rather keep that name.  It's good to grab onivim on things like reddit since oni has another meaning, but on software-specific sites I think we should use the name oni.

bert88sta:
This is his/her email to me

Hello

Sure, you can have the package name. My project isn't really serious.

To transfer the package to you I just run:

npm owner add extr0py oni

right?

Cheers,
Till

Is that right? I don't know about npm enough to verify

extr0py:
Sweet! Thanks @bert88sta for reaching out - you rock.

Yes, `npm owner add extr0py oni` is all that would be needed. We should also check with Till and see if he has a twitter handle or a website he'd like us to plug - we should put a special thanks message in our README.md like "Special thanks to Till for handing over the `oni` NPM package name", with a link to his website/twitter. Not that we can get a lot of traffic now, but maybe one day ;)

extr0py:
And ya, since you did all this legwork, I agree - having the `oni` name would be awesome, we'll stick with that as our name. I do prefer having a short, easily identifiable name.

I have the onivim subreddit and onivim.io domains - it's a bummer we can't get just `oni` for those, but not the end of the world.

bert88sta:
Hello

I just transferred the package.

You don't have to thank me in the README, but if you really want to you can
[...]

PR #291, opened 2017-03-12
Update package.json to use `oni` package name, and add reference to Till Arnold in 'thanks' section

extr0py:


extr0py:
Thanks @tillarnold for letting us have the `oni` package name, and thanks @bert88sta for reaching out!

extr0py:
Fix #279

PR #294, opened 2017-03-14
Fix #236, refresh Config object anytime config.js changes, notify any listeners

### LINKED ISSUE #236: Configuration: When config.js changes, trigger a reload of settings
When a user is editing `config.js`, it'd be awesome to see the updates once the user saves. There are a couple of items we'd need for this:

- An event to listen for when configuration changes - like `Config.on("changed", (newConfig) => { ... })`
- A config file listener in Config that watches for file changes (or the save event on config.js) and reloads the new settings
- Updated the places that care about settings / setting changes to opt-in to listening for these. An example might be for the cursorLine/cursorColumn functionality

keforbes:
The main difficulty here was that `Config.ts` is just a collection of methods, not a class or even a module.  If it was a class I could've had it `extend EventEmitter` but that would require changing how the class is accessed.  Also, we would've had to create an instance of the class, and singletons are discouraged in typescript.  Modules can't extend classes either so I decided to just leave it as a collection of methods and implement a simple [Observer pattern](https://en.wikipedia.org/wiki/Observer_pattern).

At first I thought I'd have to change every class that uses `Config` to register a listener for changes.  However, most classes that use `Config` just fetch the value as needed, rather than fetching the value and referring to it later.  So most classes will pull the latest value when they want and they don't care if (or when) it changed.  The only exception I found was in `index.tsx`, where most of the initialization takes place.  So that's the only class that currently registers listeners.

The only place I see where properties won't automatically change is in `NeovimInstance.ts`, with `oni.loadInitVim` and `oni.useDefaultConfig`.  This is because we only launch neovim once.  We *could* kill and re-launch neovim if these properties change but that seems quite drastic so I'll let @extr0py do it if he wants.

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -86,14 +86,34 @@ Performance.mark("Config.load.start")
 
 export const userJsConfig = path.join(getUserFolder(), "config.js")
 
-let userConfig = {}
-let userRuntimeConfig = {}
+let Config = {}
 
-if (fs.existsSync(userJsConfig)) {
-    userRuntimeConfig = global["require"](userJsConfig) // tslint:disable-line no-string-literal
+function applyConfig(): void {
+    let userRuntimeConfig = {}
+    if (fs.existsSync(userJsConfig)) {
+        userRuntimeConfig = global["require"](userJsConfig) // tslint:disable-line no-string-literal
+    }
+    Config = { ...DefaultConfig, ...DefaultPlatformConfig, ...userRuntimeConfig }
 }
 
-const Config = { ...DefaultConfig, ...DefaultPlatformConfig, ...userConfig, ...userRuntimeConfig }
+
+applyConfig()
+// use watch() on the directory rather than on config.js because it watches
+// file references and changing a file in Vim typically saves a tmp file
+// then moves it over to the original filename, causing watch() to lose its
+// reference. Instead, watch() can watch the folder for the file changes
+// and continue to fire when file references are swapped out.
+// Unfortunately, this also means the 'change' event fires twice.
+// I could use watchFile() but that polls every 5 seconds.  Not ideal.
+fs.watch(getUserFolder(), (event, filename) => {
+    if (event === "change" && filename === "config.js") {
+        // invalidate the Config currently stored in cache
+        delete global["require"].cache[global["require"].resolve(userJsConfig)] // tslint:disable-line no-string-literal
+        applyConfig()
+        notifyListeners()
+    }
+})
+
 Performance.mark("Config.load.end")
 
 export function hasValue(configValue: string): boolean {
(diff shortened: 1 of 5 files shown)


extr0py:
We could consider refactoring this to a class, like:

```
class Config {
...
}
```

And then export it like:
```
const _config = new Config(...);
export const getConfig = () => this._config;
```

Which is basically creating a singleton and exposing the `getConfig` function. The nice thing about that is it becomes easier to unit test. But we can look at that refactoring later.

Thanks for tackling this!

browser/src/Config.ts
```diff
@@ -86,14 +86,33 @@ Performance.mark("Config.load.start")
...
+    let userRuntimeConfig = {}
+    if (fs.existsSync(userJsConfig)) {
+        userRuntimeConfig = global["require"](userJsConfig) // tslint:disable-line no-string-literal
+    }
+    Config = { ...DefaultConfig, ...DefaultPlatformConfig, ...userRuntimeConfig }
 }
 
-const Config = { ...DefaultConfig, ...DefaultPlatformConfig, ...userConfig, ...userRuntimeConfig }
+applyConfig()
+// use watch() on the directory rather than on config.js because it watches
```
extr0py: Thanks for the detailed comment here! If we are consistently getting double events, we could look at `debounce`'ing the event: https://lodash.com/docs/4.17.4#debounce

browser/src/Config.ts
```diff
@@ -107,3 +126,22 @@ export function getValue<T>(configValue: string): T {
 export function getUserFolder(): string {
     return path.join(Platform.getUserHome(), ".oni")
 }
+
+let listeners: Function[] = []
```
extr0py: For this, we don't actually need to create our own registration mechanism - we could just hijack the event emitter:

```
import { EventEmitter } from "events";
...
const configEvent = new EventEmitter()

export function registerListener(callback: Function): void {
configEvent.on("config-update", callback);
}

function notifyListeners(): void {
configEvent.emit("config-update")
}
```

browser/src/index.tsx
```diff
@@ -256,7 +245,28 @@ const start = (args: string[]) => {
         pendingTimeout = null
     }
 
-    instance.setFont(Config.getValue<string>("editor.fontFamily"), Config.getValue<string>("editor.fontSize"))
+    const configChange = () => {
```
extr0py: Cool, thanks for hooking this up! Excited to try it out.

browser/src/index.tsx
```diff
@@ -256,7 +245,28 @@ const start = (args: string[]) => {
...
+            UI.showCursorLine()
+        }
+
+        if (cursorColumn) {
+            UI.showCursorColumn()
+        }
+
+        remote.getCurrentWindow().setFullScreen(Config.getValue<boolean>("editor.fullScreenOnStart"))
+        instance.setFont(Config.getValue<string>("editor.fontFamily"), Config.getValue<string>("editor.fontSize"))
+        renderFunction()
```
extr0py: I don't think we need to call renderFunction explicitly (as it is already called by `requestAnimationFrame`, but I don't think this will hurt either).

extr0py:
Overall looks great! Would you mind checking out if we can replace the `Observer` implementation in `Config.ts` with the `EventEmitter`? Otherwise I'm good to bring this in.

extr0py:
Oh and by the way... I wanted to try out BountySource. I put a bounty on this bug:

https://www.bountysource.com/issues/42243372-configuration-when-config-js-changes-trigger-a-reload-of-settings

Wanted to try out that process and see how it works - not much, but at least it would be some beer money as a thank you for contributing!

keforbes:
Thanks for the code review!  I'll try to tackle all your suggested changes when I get a chance.

keforbes:
I tried synchronizing this branch and evidently messed things up.  I'll have to cleanup this branch now...

PR #297, opened 2017-03-17
Fix #236, refresh Config object anytime config.js changes, notify any listeners

### LINKED ISSUE #236: Configuration: When config.js changes, trigger a reload of settings

keforbes:
Refactored #294 following @extr0py's suggestions.  `Config` is now a class accessed via singleton and uses `EventEmitter`.  This required changing every class that uses `Config` to fetch the singleton first.

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -137,7 +137,12 @@ class Config extends EventEmitter {
     private applyConfig(): void {
         let userRuntimeConfig = {}
         if (fs.existsSync(this.userJsConfig)) {
-            userRuntimeConfig = global["require"](this.userJsConfig) // tslint:disable-line no-string-literal
+            try {
+                userRuntimeConfig = global["require"](this.userJsConfig) // tslint:disable-line no-string-literal
+            } catch (e) {
+                // TODO: display this error to the user somehow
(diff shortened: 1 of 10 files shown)


extr0py:
Nice! I tried this out and it works well. Cool to see the results instantly when changing the fonts or the cursorLine settings. Thanks for investigating this and doing the refactor!

extr0py:
I'll bring in this PR. Check out the bounty and let me know how easy it is to claim it:
https://www.bountysource.com/issues/42243372-configuration-when-config-js-changes-trigger-a-reload-of-settings

Thanks @keforbes !

keforbes:
Ok, I tried claiming the bounty.  I had to create an account with bountysource to claim it and now I guess you have to approve it.

extr0py:
I approved it! Hopefully it went through

Issue #247, opened 2017-02-22
Add server option to connect to specific nvim session


xelra:
It would be great if ONI could add a `--server` option so that I can connect to a specific, already running session of Neovim.

That would also open up the possibility of using it in combination with SSH forwarding, to connect to remote Neovim sessions.

bert88sta:
Doesn't Neovim already come with a client mode? I'm not familiar with it, but it might be something to look into. If it's lackluster, this might be something to look into, but we don't want to rewrite a feature that might be improved later in neovim core

extr0py:
Ya, I was wondering the same thing as @bert88sta - doesn't Vim have a clientserver mode already? Agree if Neovim is planning to support this, it'd make sense to have this in Neovim core. If not, we could create like a `neovim-node-server` that exposes the `msgpack-rpc` layer through a websocket, and Oni could connect to that. @justinmk might have context whether this is coming into neovim core, though.

xelra:
You can tell Neovim to create a specific socket with `NVIM_LISTEN_ADRESS`, but the only client right now that can connect to that is Neovim-Qt with its `--server` option. It's a planned feature for Neovim too though.

How can this be moved to core functionality? Oni would always need the `--server` option or how else are you going to tell it to which socket to connect?

hassec:
Hi @bryphe, has there been any more thought about adding this feature?  
I saw #299 attempted this but was closed.

hassec:
Hi guys,

I would actually like expand on this remote nvim  idea. I think that this is a really great opportunity to enable true remote development.  

Imagine the scenario which many of us probably share, where you mostly work/compile code on a powerful remote server/cluster. 

To really have nice c++ autocompletion you editor will have to know about all the dependencies and the whole stack. Having all of this locally on my computer just isn't efficient and a huge burden.

As far as I know, there is not a single truly effective solution for an effortless workflow. 
I currently just run neovim with plugins serverside, which is great, but I would love to have a more modern UI (e.g tooltips, hover, peak to definition....).

Allowing oni to connect to a remote neovim would be a first step in the right direction of allowing you to edit your code from any computer no matter how powerful. 
Once that works, one would need to think about plugins...  Like I said before the autocompletion should probably run on the host since that's where the code and CPU performance lies. But there might be extensions that make more sense locally. 

So allowing a plugin to run either on the client or server would probably be the best.

Sorry for the long rambling, just figured I would leave this here.
[...]

PR #299, opened 2017-03-18
WIP: Fix #153 & #247 - Add `--server` support to ONI

### LINKED ISSUE #153: Remote capabilities
It seems that oni uses a packaged neovim binary instead and calls that. I was trying to get oni to connect to an already running neovim instance and wasn't able to do so.

You can do this with neovim-qt, e.g.

```
NVIM_LISTEN_ADDRESS=/tmp/nvim.sock nvim --headless
nvim-qt --server /tmp/nvim.sock
```

extr0py:
Fix #153 & #247 - add `--server` support

- Add capability to connect to remove neovim instance

### DIFF
--- a/browser/src/NeovimInstance.ts
+++ b/browser/src/NeovimInstance.ts
@@ -1,8 +1,11 @@
 import * as cp from "child_process"
-import { remote } from "electron"
 import { EventEmitter } from "events"
 import * as path from "path"
+
+import { remote } from "electron"
+import * as minimist from "minimist"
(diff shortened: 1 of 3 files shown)


extr0py:
This needs to be revisited since a lot has changed since the original PR.

Some of the work done in #622 would allow for some interesting ways to address this - including potentially having neovim connected to a remote server live side-by-side with a local neovim instance, for example.

I'll close this for now as the changes aren't relevant anymore, but will revisit shortly.

Issue #81, opened 2016-12-10
Override Electron menu with new menu


bryphe:
Would be nice to have a custom menu, with some options like:
- Edit config.json
- Edit keybindings.json

As well as expose some of the IDE-level functionalities:
- Go-to definition
- etc..

In developer mode, we'd of course want to preserve the developer toolbar

extr0py:
PR #147 by @Rafi993  gives us a framework to make this much easier!

CameronDiver:
In addition is there currently a way to hide the menubar? I didn't see any configuration for it, and I think it'd be a good options.

keforbes:
There currently isn't a way to hide the menubar but that should be a pretty straightforward config option.  I'll see what I can do.

bryphe:
Doesn't look like there is any action need here - will use new issues to track changes in the menubar.

PR #300, opened 2017-03-19
#81, hide menubar if oni.hideMenu is set

keforbes:
As requested in #81, add option to hide menu bar.  The menu can be temporarily displayed with `Alt` when this option is set.

I believe this code should work but I wrote it on Ubuntu (with Unity) so I can't really test it out since there isn't a "menu bar" to hide.  Hopefully someone else can give it a try.  If it doesn't work, I can play with it in my Gnome 3 environment tomorrow.


extr0py:
Works great on Windows. Easy to test with the auto-reload of the config as well. Thanks for the PR!

PR #301, opened 2017-03-19
Reload init.vim when it changes

keforbes:
Similar to reloading `config.js` when it changes, reload `init.vim` when it changes (and `oni.loadInitVim` is set).

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -126,6 +126,14 @@ class Config extends EventEmitter {
         return path.join(Platform.getUserHome(), ".oni")
     }
 
+    public getInitVimFolder(): string {
+        if (Platform.isWindows()) {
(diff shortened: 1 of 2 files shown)


justinmk:
Many users' vimrcs are not prepared for re-sourcing. E.g., if they defined autocmds without an augroup, re-sourcing the file will create duplicate autocmds each time, which will slow down the session more and more each time the vimrc is sourced.

extr0py:
Given @justinmk 's concern here, I would prefer to hold off on this PR. It might mean that someone's `init.vim` is basically not editable in oni (or at least, saveable), and it would be hard to understand why. If someone wanted this behavior, they could bring in a plugin like [vim-reload](https://github.com/xolox/vim-reload)

keforbes:
Makes sense to me.  Closing PR.

Issue #201, opened 2017-02-07
Beginning external status line implementation


bert88sta:
I think at this point, getting an external status-line skeleton going would make us prepared for RFC 5686 in Neovim when it inevitably gets merged in. I literally wouldn't know where to even start on something like this, so any help or suggestions would be appreciated.

Once we have something like this up and running, I would hope to see something capable of being customized to no end. I personally love my status line in tmux, using plugins that enable notifications for weather, email, etc. I think having (optionally, but as a feature) this going would be attractive to new users.

extr0py:
I'll see if I can get a chance to put together the skeleton shortly, and then we can work together on finalizing it and prototyping some of the extensions, if you're up for it!

extr0py:
I'll see if I can get a chance to put together the skeleton shortly, and then we can work together on finalizing it and prototyping some of the extensions, if you're up for it!

bert88sta:
@extr0py thanks, and if there's some way you could keep me up to date, I think seeing how this project grows will help me understand how this project actually works. Right now there's like 3 files that handle logic that I understand. Anything GUI related as of right now is still magic to me :)

extr0py:
Sure, I'll loop you in on the PR requests as I go. Yes, there's definitely some complexity - I'm trying to keep the UI components (autocomplete, quickinfo, pop up menu, etc) as decoupled as I can from Neovim - so they are almost entirely driven by the store state. There will be opportunites to clean things up / simplify too as we start adding some of these new UI extensions.

I'm thinking about some of these as initial first steps:
- Rename overlay-ui -> ui
- Create <Window> container
- Create container like:
	<Oni>
	<Editor>
        <Window>
          <Neovim>
          <...>
        </Window>
        </Editor>
	<Status>
        </Oni>

- Move background into that concept

Then, create some hooks for the status bar - I'd like to have some JS extensibility via the Oni plugin interface, and we'll need to make sure it will support the messages coming from Neovim's msgpack-rpc API as well

Govinda-Fichtner:
@bryphe Is there a way to disable the statusbar of neovim? Checking the documentation oni there is `statusbar.enabled`, but setting this to true still show the statusbar of oni and neovim. Is there a documentation about this that I missed?

CrossR:
@Govinda-Fichtner, setting `    "statusbar.enabled": false,` will only disable the Oni status bar, ie: 
[...]

PR #303, opened 2017-03-19
WIP: #201 - External Status Line

extr0py:
This is work for the following:
- Create an API plugin for StatusBar
- Create state + React component for StatusBar (which will involve some DOM changes / changes to how we render the Neovim piece of the UI)

In addition, I'd like to create a couple of sample status bar implementations:
- A `git` status bar component to show the current branch
- A `weather` example to show the current weather

Once the external UI is available, we can also plug that into the core work above.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,6 +1,10 @@
 - Create API for StatusBar
 - Create state for status bar in redux store
 - Create StatusBar react component as part of external UI
+    - Refactor some of the pieces in `index.tsx` to a <NeovimEditor /> react component
+        - Pass in `Pluginmanager` as a dependency
+        - Change `resize` behavior - listen to DOM element mutation directly, instead of window resize
+        - Make sure on `unmount` neovim gets disposed + render loop is removed
 - Create way for plugins to get settings (and be notified of settings changes)
 - Create `git` status bar component to show current branch
 - Create `core` integration to show 
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,8 @@
+- Create API for StatusBar
+- Create state for status bar in redux store
+- Create StatusBar react component as part of external UI
+- Create way for plugins to get settings (and be notified of settings changes)
+- Create `git` status bar component to show current branch
+- Create `core` integration to show 
+- Add sample `weather` statusbar plugin
+- Show filename
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -20,8 +20,8 @@
         - [ ] Tab UI
         - [ ] Status bar
     - [ ] Website at onivim.io
+        - [ ] Youtube Demo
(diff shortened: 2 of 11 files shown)


bert88sta:
Anything i could do to speed this up for you? I'm stoked that this has been started, but I'm not too familiar with how the plugin-host business works... Always a first time to learn I suppose

extr0py:
The initial work will be a bit hard to parallelize - but there is some help I'll need shortly for this:

1) Some example plugins leveraging the status bar - I want to include a git integration out of the box, but also an example of something more interesting like a github notifications or weather statusbar plugin

2) Improving our documentation - our README.md is a bit hard to digest at the moment. I want to have a more concise demo gif / youtube video, and then distill some of the information in a better format (like a [gitbook](https://www.gitbook.com/@extr0py/dashboard))

bert88sta:
Just testing this as you go, I definitely get some weirdness with text selection

extr0py:
Oh ya, it's in a bad state right now, I gutted the whole UI temporarily to create an <EditorHost /> component. That will open the door for some really interesting UI integrations, after this work is done-

bert88sta:
I don't actually know if the work you're doing on changing the UI into a component is going to help with the gui-tabline, but if you think it will let me know. Most of the actual vim boilerplate isn't too bad to get the information we need for an external tabling

extr0py:
Thanks @bert88sta ! Yes, unfortunately I haven't been able to spend much time on this PR - but the changes will definitely help us. The UI will be in a more sane format so that we can add something like the tabline above, and be able to have some first-class styling for it.

With the component-ized UI:
```
   return <div className="container vertical full">
             <div className="container full">
                <EditorHost editor={this.props.editor} />
                <Cursor />
                <CursorLine lineType={"line"} />
                <CursorLine lineType={"column"} />
                <SignatureHelpContainer />
                <QuickInfoContainer />
                <SignatureHelpContainer />
                <MenuContainer />
                <AutoCompletionContainer />
            </div>
            <div className="container fixed">
                <StatusBar />
            </div>
          </div>
```

We'd be able to easily add a `<TabBar />` at the top of the UI. The other pieces that would be helpful would be thinking about what sort of state we need in the redux store (certainly an array of `TabInfo` - which would have things like `fileName`, `isModified`, maybe `filetype` for an icon), and then how to populate that state from Neovim

bert88sta:
Neovim API has several functions for getting info about the tab bar and tab pages. I *think* the only "global" tab-info function is the one that lists the IDs of the tabs from left to right, the rest are info about the current tab. I'll try to put something together soon

extr0py:
This one got crazy to merge, so I ported over the work into a new PR: #444. Tracking the ideas around tab lines in #361.

PR #306, opened 2017-03-20
Fix #304, refactor fuzzy finder logic to use fuse.js

### LINKED ISSUE #304: Highlight the wrong parts while searching in the command palette?
image

_"Open"_ keyword should be highlighted.

keforbes:
As suggested by @extr0py, replaced fuzzy finder logic with `fuse.js` library.  Most of it was pretty straightforward, except the old logic returned an array of indices to highlight while `fuse.js` returns arrays of `[start, end]` indices.  So I had to change some logic to expect `number[][]` rather than just `number[]`.


keforbes:
Personally, I think `fuse.js` behaves better than the old logic but there are still a couple scenarios where it's lacking.  Searching for `NeovimInstance.ts` works well but searching for `browser/src/NeovimInstance.ts` requires more extra characters than should be needed.  So I think this is an improvement but not quite perfect.  It's possible some more tweaking of fuse's options would give us better results though.  I haven't played with any of the options like `distance`, `threshold`, and `location`.

extr0py:
Ya, I just tried it out, and the `fuse` algorithm seems more forgiving - especially if I type like `NeovimInstance` and then add `browser/src` on it.

The only concern I had when I was testing it out is that sometimes it seemed slow - for example, if I typed `NeovimInstancebrows` and then started pressing backspace, it felt much slower than the previous implementation. 

When I profiled, I saw that `fuse` was doing a _lot_ of work:
image
(The `BitapSearcher` looks like the core of Fuse)

I don't want to regress performance here though - I think there are strategies we could use to make Fuse work - like potentially some memoization for the backspace case. I also wonder if there is a lighterweight algorithm we could use though - I'm curious what algorithm CtrlP or Sublime use for their fuzzy-finders.

extr0py:
Interesting discussion here: https://www.quora.com/How-is-the-fuzzy-search-algorithm-in-Sublime-Text-designed

extr0py:
One more optimization I'm thinking about - maybe we could do a "pre-pass" of the strings, and do a quick filter where we filter for only items that contain all the characters. I think this would improve performance greatly, because most of the issues I saw only occurred once all the characters were typed in.

keforbes:
Ok, I pushed a change that caches searches longer than 5 characters since the search takes more time as the search string gets longer.

I'm basing the cache on `process.cwd()` so we'll reset the cache any time the current working directory changes.  Unfortunately, I tried testing this with `:cd <dir>` in neovim and it didn't update the `process.cwd()`.  I thought your `DirChanged` event would update `process.cwd()` but I didn't see it work.  Am I missing something?

keforbes:
It looks like I'm not even seeing the `DirChanged` event firing, so that'd explain why `process.cwd()` isn't updating...

Also, I'm seeing some weird behavior where oni launches but doesn't connect to neovim until I hit `<Enter>` so it's possible my dev environment is all screwy again and that explains the lack of `DirChanged` event.
connect

I guess I'm just saying my code my work correctly even though I'm not seeing the correct behavior. :smile:

extr0py:
For caching, I was playing with this now and I realized it would cause problems when we start hooking up this async behavior:
https://github.com/extr0py/oni/issues/269 (supporting incremental results in QuickOpen). It's also fragile in that, if a file gets added outside Oni, it might not show up in the list - this would be confusing / frustrating.

I tried this out really quickly for filtering:
```
    const filteredOptions = options.filter((o) => {
        const combined = o.label + o.detail

        for (var i = 0; i < searchString.length; i++) {
            if (combined.indexOf(searchString[i]) === -1)
                return false
        }

        return true
    })
...
    const fuse = new Fuse(filteredOptions, fuseOptions)
```

And it seemed like it addressed the performance concern I had, and doesn't have the drawbacks above - it would work fine for the incremental / async population as well.

It could be cleaned up a bit, and it could be optiimzed further to take into account the number of characters, but perhaps we could try that instead? Thanks for looking at this!

extr0py:
Oh, and did you figure out what was happening with the DirChanged event? it sounds like there is maybe some `init.vim` error or something popping up. Unfortunately we don't have a good way to bubble that up at the moment... but running neovim separately or in the terminal can help isolate this

keforbes:
Ok, I changed it to use your suggestion.  I tweaked the logic a bit to remove all duplicate characters from the `searchString` so we're not checking them multiple times but otherwise it's the same.

No, I never figured out what's going on with the `DirChanged` event.  It's just an annoyance right now, it doesn't prevent me from doing anything.

extr0py:
Cool, thanks! I tested this out and the behavior seems greatly improved now, and the performance is smooth.

Bummer about the `DirChanged` event - as you mentioned above, if Oni isn't connecting until you hit enter, that means usually that there is some VimL error on launch. You could try clearing out your `init.vim`, and if that doesn't work, it means probably one of the core changes caused a regression. If you set `oni.loadInitVim` and `oni.useDefaultConfig` to false, that should disable both of these too. Keep me posted if you're not able to get that figured out.

This PR looks great though. I'll bring it in. Thanks for the contribution!

PR #307, opened 2017-03-21
remove unused dependency

jasonszhao:


extr0py:
Thanks for cleaning this up, @jasonszhao !

PR #308, opened 2017-03-21
Fix #305: Reduce boilerplate around actions in UI

### LINKED ISSUE #305: Refactoring: Use `bindActionCreators` in `browser\src\UI\index.tsx`
All the ActionCreators essentially have to be duplicated + re-exported in the UI namespace. Redux comes with a utility that can do this for us with `bindActionCreators`. We should export a `UI.Actions` instead, and update the callers.

In some cases, if there is logic in these wrapped items, we'll need to refactor it into an action creator that is thunk'd.

extr0py:
- Use the `bindActionCreators` utility from Redux to avoid having to redefine the same function multiple times

### DIFF
--- a/browser/src/Plugins/PluginManager.ts
+++ b/browser/src/Plugins/PluginManager.ts
@@ -157,15 +157,15 @@ export class PluginManager extends EventEmitter {
             }
 
             if (!pluginResponse.error) {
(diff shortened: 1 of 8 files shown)


extr0py:
Fix #305

Issue #312, opened 2017-03-21
Integrate electron-packager to generate OS X and Windows packages


extr0py:
- Bring in `electron-packager` to create OS X bundle (.app) and Windows bundle (.exe).
- This should add some new npm commands - `npm run package:windows`, `npm run package:osx`

extr0py:
Electron packager: https://github.com/electron-userland/electron-packager

extr0py:
The end goal is to have download links to lower the barrier to entry to trying ONI

extr0py:
http://electron.rocks/electron-angular-packing-and-distribution/ looks helpful

extr0py:
We now generate linux (deb, rpm, tar.gz), osx (dmg) and windows (exe, zip) packages for tags. Will create a new release on Sunday!

PR #316, opened 2017-03-21
WIP: Fix for #312 - Add electron-packager to create release binaries

extr0py:


extr0py:
Need to switch to `electron-builder`

keforbes:
I think you should do a new release as soon as you get this working.  Release binaries will be really cool!

extr0py:
Sounds like a plan - I'm hoping to have this all set by the weekend - perfect time for a new release!

extr0py:
Will need a couple of follow up issues for this:
- Move some `dependencies` to `devDependencies` if they are being bundled in the webpack `bundle.js`, so they are not included twice
- Create `x64` build for Windows

extr0py:
There are still some improvements we can make here, but we can track them as separate issues. I'm searching the first binary release will be a bit rocky, but we can continue to improve.

PR #318, opened 2017-03-22
Fix some npm warnings

keforbes:
Running `npm install` would get the following warnings:

```
npm WARN deprecated @types/redux-thunk@2.1.0: This is a stub types definition for Redux Thunk (https://github.com/gaearon/redux-thunk). Redux Thunk provides its own type definitions, so you don't need @types/redux-thunk installed!
npm WARN deprecated @types/redux@3.6.0: This is a stub types definition for Redux (https://github.com/reactjs/redux). Redux provides its own type definitions, so you don't need @types/redux installed!
...
npm WARN The package mocha is included as both a dev and production dependency.
```

This Pull Request resolves those warnings.


extr0py:
Awesome! Thanks for fixing these.

PR #319, opened 2017-03-22
BREAKING CHANGE, Use consistent naming convention for config options

keforbes:
As discussed in Gitter, I felt the config options should be more consistent in the namespaces used.  For now, the current convention is:

`debug` - internal properties no user should care about.  Currently just `incrementalRenderRegions` and `maxCellsToRender`, though I couldn't find any evidence `maxCellsToRender` is still being used.

`oni` - properties that affect **Oni** features, mostly unrelated to Neovim.  Things like menus, fuzzy finder, etc.

`editor` - properties that affect the viewable "neovim area" even though they don't affect Neovim itself.  Things like highlights, colors, etc.  This includes the background settings, which had been under a `prototype` namespace.

I've updated the README.md accordingly.  Let me know if you disagree with any of the changes I made though.


extr0py:
Thanks for bringing some sanity to our config settings! I like that you made sure to explicitly calling this out in the title.. will make it easy for me to see when I put together the release notes. Looks great to me. I'm glad we'll have some standards in place here moving forward.

Issue #281, opened 2017-03-08
Line Numbers stuck behind


v3ss0n:
![screenshot_20170308_133853](https://cloud.githubusercontent.com/assets/419606/23693817/c9dace3c-0404-11e7-8b66-8e09cba86c29.png)

keforbes:
How are the line numbers stuck behind?  Are you saying that isn't line `53` you're on?  I'm not seeing an issue in that screenshot.

v3ss0n:
5,4,3,2,1,53,1,2,3,4,5 . Is that working as intended? how it works

extr0py:
Ya, by default we set `:set relativenumber`. You can turn this off by running `:set norelativenumber`

This is a somewhat opinionated default, since it isn't on by default in Neovim. I'd be okay changing this back, if it is confusing.

In the meantime, the workaround would be to add `:set norelativenumber` to your `init.vim`.

extr0py:
Thinking about this, I believe it is better to have line numbers (not relative) as the default - that's a bit more friendly to people trying out ONI for the first time. @bert88sta - would you be up for changing our default here? I wanted to give you a bounty as a thank-you for your contributions

bert88sta:
@extr0py: I'll get it taken care of here soon, doing a lot of financial aid business for college at the moment

extr0py:
Cool, no rush, getting financial aid set up is much more important! 💵 💵 Good luck!

extr0py:
This default has been removed, so this is now addressed. Thanks @bert88sta !

PR #323, opened 2017-03-26
BREAKING CHANGE: Fix #322 & #281 (mouse fixes, and remove default relativeline setting)

### LINKED ISSUE #322: Inaccurate Clicking
This is on OSX, don't know if that has anything to do with the problem but it's become quite frustrating

https://i.gyazo.com/75e6af8103323a0ba854222008a8fdf7.gif

The gif may not be entirely clear, but I'm clicking continuously and it doesn't move up to the next line until maybe 1/3 or 1/2 of the way up the next line

bert88sta:
This addresses #322 and #281 

I think what I did in Mouse.ts is a hacked together version of floor, but I could be entirely wrong. It works

### DIFF
--- a/browser/src/Input/Mouse.ts
+++ b/browser/src/Input/Mouse.ts
@@ -69,8 +69,8 @@ export class Mouse extends EventEmitter {
         const mouseY = evt.clientY
 
         return {
-            line: Math.round(mouseX / this._screen.fontWidthInPixels - 0.5),
-            column: Math.round(mouseY / this._screen.fontHeightInPixels - 0.5),
+            line: Math.floor(mouseX / this._screen.fontWidthInPixels),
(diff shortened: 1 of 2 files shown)


keforbes:
Why didn't you just use `Math.floor()`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/floor

bert88sta:
Because I'm young and inexperienced :P 
I'll change it.

extr0py:
Looks good to me! I updated the title so I remember to call this out as a breaking change in the release notes. I also associated a bounty with this - so please submit a claim on bountysource (https://www.bountysource.com/issues/43365739-inaccurate-clicking). Thanks for the fix!

PR #325, opened 2017-03-28
Fix #99: Add prebuilt OSX binary to github

### LINKED ISSUE #99: Include pre-built binary of Neovim for OSX

extr0py:
This is to streamline the experience with the release binaries for #312 - ensuring that MacOS will work out-of-the box with the app package.


extr0py:
Fix #99

keforbes:
We need to update the README.md now

> For Windows, a pre-built x86 binary of NeoVim is included.
> 
> For OSX and Linux, there is no included pre-built binary. Please Install Neovim and ensure the 'nvim' binary is available. Version 0.2.0 is required.

extr0py:
Good point - updated the README in a separate commit.

Issue #327, opened 2017-03-29
Windows Paths Issue?


am2605:
Hi Folks,

Forgive me if this has already been addressed (I did search the issues but couldn't find anything).

I've just installed this on Windows 10, and it looks really quite an exciting project.  However I'm having problems opening files in Windows.

If I select File -> Open and browse to say c:\Users\AMyers\_viminfo, it appears that the backlashes get dropped and it opens a new file called c:UsersAmyers_viminfo - screenshot attached.

Is this a known issue?  If not, is there anything I can do to assist with debugging it?

Thanks!

Andrew.

![capture](https://cloud.githubusercontent.com/assets/119977/24432859/962e3962-1470-11e7-9e4a-f4b0e89615f5.PNG)

keforbes:
You're right, I don't think this has been addressed.  I wrote that feature on Linux and I don't have a Windows box to test with so I'm sure it's my fault.  I'll have to see how Neovim expects this to work.  I'm afraid to just blindly replace `\` with `/` since paths with spaces might be sent with `\ ` (`<backslash><space>`).  I might be able to quote the path, though we're already quoting the whole command with `execute ":e <file>"` under the hood and if I try quoting the path with `'` that would probably prevent variables from being expanded (if that matters).  And yes, I'm just rambling at this point.

As a potential workaround, try using the Fuzzy Finder with `<Ctrl-P>` and see if that opens files correctly.

If you're willing to look at the code to debug it, `Menu.js` is sending `":e " + name` to `index.ts` which executes `"exec \"" + command + "\""`.  There's probably some combination of quotes or string replacements we can do on that string.  If Fuzzy Finder works, you an compare that with `QuickOpen.ts` which is running `"e! " + fullPath`.

am2605:
Thank you for the reply.  Fuzzy Finder does work so that's a good 
workaround.

I have made a change to QuickOpen.tsx which has made it work, but I"m 
not sure if there's any other conseqences of this.  The change is:

Line 358: change from

instance.command("exec \"" + message + "\"")

to

instance.command("exec " + JSON.stringify(message) + "")

justinmk:
> instance.command("exec " + JSON.stringify(message) + "")

In this case, that's an improvement because the `\` slashes look like escape characters to the vim command processor, since the message is dumped as a raw string to `:exec`...

But the filename arg should also be wrapped in vim's `fnameescape()`...

This seems harder than it should be :<

extr0py:
[...]

PR #328, opened 2017-03-29
Fix for #327: Open command doesn't work on Windows

extr0py:
- Create a `normalizePath` helper method to coerce the `\` backslashes to `/` forward slashes, will are supported on all platforms
- Create helper method to handle running commands on an array of files
- Wire up the existing file menu commands (`:e`, `:sp`, and `:tabnew`) to leverage these helpers

### DIFF
--- a/Menu.js
+++ b/Menu.js
@@ -1,5 +1,6 @@
 const electron = require('electron')
 const os = require('os')
+const path = require("path")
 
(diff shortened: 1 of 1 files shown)


extr0py:
- Validated menu still works correctly on OSX

extr0py:
Fix #327

PR #330, opened 2017-03-29
Fix #329 - Issues with spaces when setting runtime paths

### LINKED ISSUE #329: Root path with spaces breaks `Oni`
Discovered this while working on #312 - if Oni lives in a path that has spaces, like:
- `C:/program files (x86)/Oni/oni.exe`

the app will fail to load. The issue that the runtimepaths set via `set rtp+=` aren't being set properly.

This is easy to repro by cloning Oni into a directory of the form like: `\test folder with spaces\oni`.

It's blocking the installation work though because on Windows most installations will go in `Program Files (x86)`

extr0py:
Fix #329 - handle issues with space characters in paths when setting runtime paths.

It looks like on all platforms, the space key in the path needs to be escaped in this scenario.

### DIFF
--- a/browser/src/NeovimInstance.ts
+++ b/browser/src/NeovimInstance.ts
@@ -417,9 +417,6 @@ function startNeovim(runtimePaths: string[], args: any): Q.IPromise<any> {
     const nvimProcessPath = Platform.isWindows() ? nvimWindowsProcessPath : Platform.isMac() ? nvimMacProcessPath : nvimLinuxPath
 
     const joinedRuntimePaths = runtimePaths.join(",")
-                                // Escape spaces in runtime path
-                                .map(p => p.split(" ").join("\\ "))
-                                .join(",")
 
     const shouldLoadInitVim = Config.instance().getValue<boolean>("oni.loadInitVim")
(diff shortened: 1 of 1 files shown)


justinmk:
I would instead suggest using `"let &rtp.='...'"` as [neovim-qt does](https://github.com/equalsraf/neovim-qt/issues/237#issuecomment-279879407). The single-quotes make the string literal.

keforbes:
@justinmk, I'm not sure if this is a corner case or a non-existent case, but is it possible for runtime paths to include variables like `$PATH` or `$HOME`?  If so, wouldn't single quotes prevent the variable expansion?

justinmk:
But this only appending stuff to the runtime path, right. Notice the `.=.

extr0py:
Ah, great - thanks for sharing that. I'll make the change.

For the environment variables, I don't think this code should be impacted, as we're adding all the runtime paths for the built-in plugins (and other plugins the plugin manager cares about). In other words, they should all be fully resolved.

extr0py:
Looks like it works great! Thanks for the suggestion, @justinmk

PR #333, opened 2017-03-30
Tweak error marker css

keforbes:
The error marker looks great for small messages but didn't look right when the message got long.

This Pull Request changes this:
error
to this:
error_new

I also deleted a couple CSS rules that didn't seem to be having any effect.  I deleted them and didn't notice any change in behavior but if they were there for a purpose let me know and I'll put them back.

### DIFF
--- a/browser/src/UI/components/Error.less
+++ b/browser/src/UI/components/Error.less
@@ -27,11 +27,6 @@
         position: absolute;
         left: 200px;
         background-color: rgb(60, 60, 60);
-
-        display: flex;
-
-        align-items: center;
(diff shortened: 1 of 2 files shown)


extr0py:
Nice, thanks for fixing this! I really just hacked it together quick, so I'm glad you went through and cleaned it up and fixed those problems. It looks much nicer now after your changes.

This reminded me of a grievance I have with the current design - that if an error is near the end of the line, the error marker overlaps it and it becomes hard to fix. I created issue #335 to track a new design (Atom has a good model for this, I think)

PR #336, opened 2017-03-31
Fix #332, hide 'Open Neovim Config' menu unless oni.loadInitVim is true

### LINKED ISSUE #332: MYVIMRC not set?
Not sure if this is another Windows issue or not.  

If I select File->Preferences -> Edit Neovim Config, it loads an empty file.  If I enter some text there and save it, it creates a file named $MYVIMRC in the directory I launched the app from.

Am I supposed to set this myself, or should oni be giving it a value?

keforbes:
As discussed in #332, hide the `Open Neovim Config` menu item unless `oni.loadInitVim` is true.  This is because we were trying to open `$MYVIMRC` and that environment variable is not set unless `init.vim` was loaded.

Note that because of https://github.com/electron/electron/issues/528 we can't just add/remove that menu item; we have to rebuild the entire `Menu` object to modify it.  I added some logic to only rebuild the menu if `oni.loadInitVim` changes, rather than rebuilding every time the user saves `config.js`.  However, this does mean we build the menu once as Electron starts and then (if `oni.loadInitVim` is true) build the menu again when the config is done loading.

Also, I had mentioned I wanted to add some comment to the default `config.js` telling users this menu item would exist.  I ultimately decided against it.  The existence of this menu item wouldn't affect their decision to enable `oni.loadInitVim` or not, it's just a nice-to-have feature.

### DIFF
--- a/Menu.js
+++ b/Menu.js
@@ -5,7 +5,7 @@ const os = require('os')
 const { Menu, app, shell, dialog } = electron;
 
 
-const buildMenu = (mainWindow) => {
+const buildMenu = (mainWindow, loadInit) => {
     let menu = []
 
     // On Windows, both the forward slash `/` and the backward slash `\` are accepted as path delimiters.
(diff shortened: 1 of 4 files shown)


extr0py:
The change looks great to me! Thanks for the quick fix. I think it will help alleviate some confusion - I'm hoping to move some additional configuration over like line numbering, colorscheme, etc to the `config.js` concept soon, so there are less reasons to need to turn on `oni.loadInitVim`.

Good point about the comment. I think it does make sense to potentially add some comments in the default `config.js` to explain some of the features / defaults, but you're right in this case it may have been superfluous.

I'm thinking that maybe later it would make sense to move the menu-building logic over to the _renderer_, and then pass the menu template to the _main_ process by passing it via ipc. That might make it easier to enable other dynamic menu options later (like turning on/off language service-specific functionality depending on if the current filetype supports it).

PR #347, opened 2017-04-03
Fix #342, issue with first launch when ~/.oni doesn't exist

### LINKED ISSUE #342: Issue starting for first time
Hey, love the project and support excited about its future!

I had a problem first starting it though. Electron started, and I just got a blank white screen. After some debugging (I only managed it with npm run start), I found that I was missing a .oni folder. I had to make it manually. 

I thought I'd let you know since it can be a real downer for first time starters to just get a blank screen.

keforbes:
The issue was with the `fs.watch` command when setting a listener for `config.js` changes.
```
"Uncaught Error: watch /home/keforbes/.oni ENOENT", source: fs.js (1318)
```
Check to see if the file exists before setting a listener on it.  This means the first launch of Oni won't automatically reload changes to `config.js`; they'll take effect on the next launch.

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -110,7 +110,7 @@ class Config extends EventEmitter {
         // and continue to fire when file references are swapped out.
         // Unfortunately, this also means the 'change' event fires twice.
         // I could use watchFile() but that polls every 5 seconds.  Not ideal.
-        if (fs.existsSync(this.userJsConfig)) {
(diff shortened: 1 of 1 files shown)


keforbes:
Added a small tweak so we watch `~/.oni` if it exists, even if `config.js` doesn't yet exist.  This means the watcher won't be set for the first launch of Oni but on subsequent launches of Oni changes to `config.js` will automatically take effect, even when the file is first created.

Issue #20, opened 2016-12-04
Synchronizing VimScript keybindings.


jordwalke:
For UI interactions such as the popup (command-p) window (and canceling source editor popup menus), how should people configure keyboard mappings? How do they configure these, while also configuring Vim keyboard mapping (for insert/command mode). For example, it might make sense to apply all the `cmap` mappings while the "command P" window is open.
If `cmap` doesn't make sense to integrate with, then perhaps a new "mode" for mappings makes sense. But in that case, would users configure 90% of their mappings in Vim, and then 10% of them in `~/.oni/config.json`? It would be nice if keyboard mappings could be specified entirely in one place.

Edit: Summarizing thoughts, tradeoffs, observations:

- No one wants to write VimScript (how can I blame them).
- Everyone knows JS/JSON so it is preferable.
- But we don't want a ton of startup time overhead to wait for a JS engine to warm up or wait for latency to cross the language bridge several times (if we can avoid it).
- We also don't want half our editor config in VimScript and half in JS/JSON long term so if there are two ways to do something it would be nice if there was a migration path to have everything in one language, without paying any performance costs (a suggested approach is ahead of time compiling JS/JSON to VimScript - which also lets you use that same config in any build of Vim).

extr0py:
Good question, it's just hard coded in a poor way right now. My initial thinking was to have a ~/.oni/keybindings.json, and create commands for these, like 'oni.menu.nextItem' with JSON bindings, and get to the point where 99% of the mappings could be done in the json file. 

It does have the downside of having two places for bindings, and being foreign to power users. To bridge the gap, there are a few options I was thinking about:
- Add a way to execute commands from VimScript, like: `OniCommand("oni.menu.nextItem")` that could be part of key mappings
- Add some information about the UI state in VimScript, sort of like how pumvisible() works

Long-term, my vision for this project is to make modal editing through neovim more friendly and attractive to users of VSCode/Atom/etc, without having to dive into creating a _vimrc/init.vim... but still retaining much of the flexibility.

Thanks for calling out some ideas here and trying out the app, great to have your feedback!

jordwalke:
> Long-term, my vision for this project is to make modal editing through neovim more friendly and attractive to users of VSCode/Atom/etc, 

Where do I sign?

jordwalke:
[...]

PR #351, opened 2017-04-04
[WIP] #6 - Language Server Integration

extr0py:
This is a WIP for #6 - some experimentation around working against a canonical language service protocol provider. 

The goal for this is a proof-of-concept working against the `omnisharp-client` to provide C# language services via the LSP. VSCode has provided several helpful libraries like `vscode-jsonrpc` to make it pretty easy to talk to a LSP from JavaScript.

This is a hack right now but will be cleaned up once it is working.


jordwalke:
Cool!
I'm curious, to what extent does this use the VIM language service shims?

https://github.com/tjdevries/nvim-langserver-shim
https://github.com/autozimu/LanguageClient-neovim

extr0py:
Good question! This is something I deliberated about at length. 

For Oni, I decided _against_ using the existing VimL language service shims. I also discussed this a bit in #6 but the main reason is that one of the goals of this project is to __move away from VimL and leverage JavaScript__ (or BuckleScript/ClojureScript/JavaScript) for the plugin infrastructure. This is part of my broader vision to lower the barrier to entry to Neovim and modal editing in general. Taking a large VimL dependency to implement this would be counter to that vision. We also have some open issues around key-bindings and configuration (#20 - thanks for your input on those!), I hope I can revisit those shortly for 0.4 - but I'd like to see those moved as well to JavaScript.

Aside from just the fuzzy this-is-my-vision reasoning, there were also more practical technical considerations to pursue a non-VimL-driven approach:
- There are some aspects of the protocol __that will not be handled by the VimL approach__. One example is the `completion details` part of the protocol - there isn't a great way to handle this with Vim/Neovim's _omnifunc_ capabilities. There are other cases like CodeLens where it is more straightforward to take advantage of this rich pieces of functionality in the `Oni` layer than the `Neovim` layer.
- __Multiplexed Neovim Instances__ - I have some longer term plans to update the UX to rival or at least have parity with Atom, VSCode, and Brackets - I want Oni to be a beautiful editor, from the install to the opening experience. We have a long way to go in that aspect, but one piece I'm investigating is multiplexing the Neovim processes in order to support some advanced UI scenarios, different tab/window experiences, etc - @justinmk sums up some of the considerations here nicely: [Future of UI](https://www.reddit.com/r/neovim/comments/5ym9p3/future_of_ui/). The difficulty with a VimL LSP solution in this world is that the Neovim instances would not have much knowledge of each other (worst case - spawning new instances of LSP for every window/split/tab), whereas a LSP implemented in the Oni layer could be shared.
- __Excellent JS Tooling__ - VSCode actually has some great node modules for integrating the LSP into a JS app, like https://github.com/Microsoft/vscode-languageserver-node. Leveraging this as a dependency will make it easier to keep the protocol up-to-date.
- __Windows Support__ - Oni should work out of the box on Windows. All of the language server clients in VimL have a dependency on Python, which is not available out of the box on Windows. 

I also called out a more selfish reason in #6 - I personally am not very productive in VimL. I spent some time going down that rabbit hole when I was learning Vim, but I also wasted a lot of time. I've come to my terms with the fact that I'll never be another tpope :smile: Given that I'm building this in my spare time, I'd prefer not to use VimL.

When I look at things like the [Stack Overflow Developer Survey](https://stackoverflow.com/insights/survey/2017) and see JavaScript at the top of the programming languages engineers use, it affirms to me that this is the proper direction for Oni to pursue.

Let me know if you have feedback or ideas though! As developers, it is easy to fall into the trap of [Not Invented Here](https://en.wikipedia.org/wiki/Not_invented_here) and continuously reimplement the wheel, so its important to critically evaluate when we choose not to leverage an existing solution. 

The existing LSP VimL clients are awesome and both of the authors have done incredible work, but the end goals / vision are simply different.

jordwalke:
> move away from VimL and leverage JavaScript (or BuckleScript/ClojureScript/JavaScript) for the plugin infrastructure

If only we had a Buckle(VimL)Script backend!

>  I hope I can revisit those shortly for 0.4 - but I'd like to see those moved as well to JavaScript.

To summarize my original position is that JS would be a great target for scripting this configuration as long as it can eventually be taken out of the startup path, generating a raw VimL config script, and that can be cached so that startup is very fast. Seems that this should be doable without anyone touching a line of vim script. In fact, this could even produce a highly condensed VimL cached boot script that's faster than any VimL you would be likely to write by hand.

> The difficulty with a VimL LSP solution in this world is that the Neovim instances would not have much knowledge of each other (worst case - spawning new instances of LSP for every window/split/tab), whereas a LSP implemented in the Oni layer could be shared.

I'll have to take your word for that.  Seems like an oversight somewhere in either LSP or in NeoVim if it's truly not possible.

> I also called out a more selfish reason in #6 - I personally am not very productive in VimL. 

You have no idea how much I can relate.

> The existing LSP VimL clients are awesome and both of the authors have done incredible work, but the end goals / vision are simply different.
 
My thoughts were that they just might create 90% of the needed functionality and could possible even be extended to include the remaining 10%.

Ultimately, I don't predict the LSP layer is going to be too large anyways. That's kind of the point - isn't it? The bulk of language plugins happens behind the service - it's a good time to be alive.

My only suggestion is that for anything not dynamic (configuration, theming, mappings, settings, workspaces),  that it "eject" to VimL which can be cached and loaded without ever needing to loads tons of JS. LSP integration is beyond that use case. I'm just thinking it's a way that Oni can set itself apart, by maintaining super fast startup time.

extr0py:
This is blocked on OSX by omnisharp/omnisharp-node-client#411

extr0py:
Alright, seems like this is good to go... I logged several follow-up issues though:
#382 #381 #380 #379 #378 #377 #376 #375

keforbes:
Don't forget to claim your bounty!
https://www.bountysource.com/issues/39723750-microsoft-language-server-protocol

extr0py:
Cheers, thank you very much @keforbes ! 😄  Just submitted the claim. I'll use it to post up more bounties.

Issue #314, opened 2017-03-21
Move background to React component


extr0py:
In `browser\src\UI\index.tsx`, we have this chunk of code:

```
export function setBackgroundColor(backgroundColor: string): void {
    const config = Config.instance()
    const backgroundImageElement: HTMLElement = document.getElementsByClassName("background-image")[0] as HTMLElement
    const backgroundColorElement: HTMLElement = document.getElementsByClassName("background-cover")[0] as HTMLElement
    const backgroundImageUrl = config.getValue<string>("prototype.editor.backgroundImageUrl")
    const backgroundImageSize = config.getValue<string>("prototype.editor.backgroundImageSize") || "cover"

    backgroundImageElement.style.backgroundImage = "url(" + backgroundImageUrl + ")"
    backgroundImageElement.style.backgroundSize = backgroundImageSize
    backgroundColorElement.style.backgroundColor = backgroundColor
    backgroundColorElement.style.opacity = config.getValue<string>("prototype.editor.backgroundOpacity")
}
```

This should be factored out to a React component that is rendered in `RootComponent`, and again driven by state in the redux store.

FrigoEU:
Hey, I'd like to do this, but was wondering what should be in the redux store and what shouldn't. I guess the backgroundColor should be in the store, and the use of the config singleton will stay like that? Or should/will the config object also migrate to the store at some point?

extr0py:
Hi @FrigoEU , sorry for the slow response on my end! Good question about the Redux store. I could see us migrating these config options at some point to the redux store, but I looked at your PR and you did the right thing here - added `backgroundColor` to the redux store and leveraged the rest from config values. I opened #353 to track discussion around this but it isn't pressing and certainly shouldn't block this PR. Overall your PR looks great - thanks for taking this on!

extr0py:
Fixed by #352

PR #352, opened 2017-04-05
Move background to React component (#314)

FrigoEU:
Hey,

I really like this project and felt like contributing something to get a feel for the codebase. I tried to follow your existing codestyle, but feel free to ask me to change anything/everything. I tried to add the Background component to the RootComponent, but that didn't really work out too well since the text editor needs to be "in between" the background and the rest of the React UI in terms of DOM nodes. I couldn't really figure out how to do it differently, but I'm not really a React expert yet.

Simon

### DIFF
--- a/browser/src/UI/ActionCreators.ts
+++ b/browser/src/UI/ActionCreators.ts
@@ -55,12 +55,12 @@ export const setCursorPosition = (screen: IScreen) => (dispatch: Function) => {
     dispatch(_setCursorPosition(screen.cursorColumn * screen.fontWidthInPixels, screen.cursorRow * screen.fontHeightInPixels, screen.fontWidthInPixels, screen.fontHeightInPixels, cell.character, cell.characterWidth * screen.fontWidthInPixels))
 }
 
-export const setColors = (foregroundColor: string) => (dispatch: Function, getState: Function) => {
-    if (foregroundColor === getState().foregroundColor) {
+export const setColors = (foregroundColor: string, backgroundColor: string) => (dispatch: Function, getState: Function) => {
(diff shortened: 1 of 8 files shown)


keforbes:
We have a linter that checks for style errors as part of the build process.  Your commit failed for these (minor) issues:
```
oni/browser/src/UI/Reducer.ts[30, 32]: trailing whitespace
oni/browser/src/UI/Reducer.ts[32, 58]: Missing trailing comma
```

FrigoEU:
Oops, I missed that. Fixed.

extr0py:
Thanks for the kind words on the project! The change looks great to me. 

> I tried to add the Background component to the RootComponent, but that didn't really work out too well since the text editor needs to be "in between" the background and the rest of the React UI in terms of DOM nodes

Yes, this is definitely painful. I like the solution you chose for now - just having ReactDOM.render directly to another element. I'm working on another PR - #303 - which refactors the UI structure a bit, so we end up with something like this:
```
export class RootComponent extends React.Component<IRootComponentProps, void> {

        return <div className="container vertical full">
            <div className="container full">
                <EditorHost editor={this.props.editor} />
                <Cursor />
                <CursorLine lineType={"line"} />
                <CursorLine lineType={"column"} />
 ...
            </div>
            <div className="container fixed">
                <StatusBar />
            </div>
          </div>
```

Your solution here will fit nicely with that PR because we can just pick up your <Background /> component and place it in the right level there.

extr0py:
Looks great, nice work & thanks for the contribution @FrigoEU ! I'll merge this in.

Issue #353, opened 2017-04-06
Move config options to Redux store


extr0py:
One question that has come up a few times as we're working on some new UI work is whether config options belong in the Redux store. 

If the config options are _not_ in the Redux store, that means the React components are not _pure functions_ and that has some implications on hot-reloading, unit testing, etc. 

If the config options _are_ in the Redux store, it means there might be a bit of extra book-keeping / maintenance when adding new 

However, given the tradeoffs, I believe the right approach is to move these to the Redux store, and when we get the config changed event, we could populate a `configuration` branch of our state tree with these values.  We could have a generic `setConfigurationValue` action that would take a key and a value, to make these easily accessible in our components.

keforbes:
Fixed by #356.

PR #356, opened 2017-04-07
Move config options to Redux store (#353)

FrigoEU:
This bothered me when I was making the Background component, so I thought I'd take a stab at this.

Some remarks: I might have went overboard with the type safety. I write a lot of PureScript so I like type safety a lot, but adding type safety to the configuration object was a bit annoying, so I can totally understand if you want me to dial it down. Secondly, I had to make this State.readConf function to make reading the IConfigValues interface safe, since just indexing into it (eg. state.configuration["editor.backgroundImageUrl"]) is unsafe since we us "suppressImplicitAnyIndexErrors": true.

Just let me know how far you want the type safety to go, I'll adjust to what you feel is right.

Secondly I only updated the components to read from the redux store. Services that want to read from the config keep their own reference to the instance, which is OK for now I think. The Error.tsx component isn't hooked up to redux, so I left that for now.

Again, this is just my interpretation so I'll change anything/everything you want me to. 

I tried to keep the separate commits small and consistent to aid with code review.

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -1,11 +1,13 @@
 import { EventEmitter } from "events"
+
 import * as fs from "fs"
+import * as _ from "lodash"
 import * as path from "path"
 
 import * as Performance from "./Performance"
 import * as Platform from "./Platform"
 
-interface IConfigValues {
+export interface IConfigValues {
     // Debug settings
     "debug.incrementalRenderRegions": boolean
     "debug.maxCellsToRender": number
@@ -177,6 +179,10 @@ class Config extends EventEmitter {
 
     public getValue<K extends keyof IConfigValues>(configValue: K) {
         return this.Config[configValue]
+    }
+
+    public getValues(): IConfigValues {
+        return _.cloneDeep(this.Config)
     }
 
     public getUserFolder(): string {
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -5,70 +5,117 @@ import * as path from "path"
 import * as Performance from "./Performance"
 import * as Platform from "./Platform"
 
+interface IConfigValues {
+    // Debug settings
+    "debug.incrementalRenderRegions": boolean
+    "debug.maxCellsToRender": number
+    "debug.fixedSize": {
+        rows: number,
+        columns: number,
+    } | null
+
+    // Production settings
+
+    // Bell sound effect to use
+    // See `:help bell` for instances where the bell sound would be used
+    "oni.audio.bellUrl": string
+
+    // The default config is an opinionated, prescribed set of plugins. This is on by default to provide
+    // a good out-of-box experience, but will likely conflict with a Vim/Neovim veteran's finely honed config.
+    //
+    // Set this to 'false' to avoid loading the default config, and load settings from init.vim instead.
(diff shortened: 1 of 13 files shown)


browser/src/Config.ts
```diff
@@ -79,23 +128,23 @@ class Config extends EventEmitter {
...
         "editor.fontFamily": "Monaco",
     }
 
-    private WindowsConfig: any = {
+    private WindowsConfig: Partial<IConfigValues> = {
         "editor.fontFamily": "Consolas",
     }
 
-    private LinuxConfig: any = {
+    private LinuxConfig: Partial<IConfigValues> = {
```
extr0py: Cool, I didn't know about `Partial` - that's really helpful for this scenario. I think it would be useful for some of the React typings, especially how `connect` works with `mapStateToProps` and `mapDispatchToProps`. Nice!

browser/src/UI/ActionCreators.ts
```diff
@@ -174,6 +176,16 @@ export const setCursorColumnOpacity = (opacity: number) => ({
     },
 })
 
+export function setConfigurationValue<K extends keyof Config.IConfigValues>(k: K, v: Config.IConfigValues[K]): Actions.ISetConfigurationValue<K> {
```
extr0py: I like the way the typing here too was set up, that's really cool

extr0py:
This is excellent! I didn't know some of those typing constraints were possible - I learned something new with `Partial` and with the construct used like:
```
export function setConfigurationValue<K extends keyof Config.IConfigValues>(k: K, v: Config.IConfigValues[K]): Actions.ISetConfigurationValue<K>
```
I didn't realize we'd be able to get that level of typing granularity.

> Secondly I only updated the components to read from the redux store. Services that want to read from the config keep their own reference to the instance, which is OK for now I think. The Error.tsx component isn't hooked up to redux, so I left that for now.

This sounds totally reasonable to me. I'd like to incrementally refactor these to be bound to the redux store. 

The only concern I have is that it looks like all of the reducers in `browser/src/UI/Reducer.ts` have to add the `<K extends keyof Config.IConfigValues>` to accept the new Action union type. I wonder if th ere is a way to limit that - reducers like the `autoCompletionReducer` and `popupMenuReducer` probably shouldn't need to know the details of the Config typing. Could you please take a look at that and see if there is a way we can make most of the reducers agnostic of the type?

One way to address that might be to have a separate union type that doesn't contain the `ISetConfigurationValue` action, and use those for the downstream reducers. 

Otherwise, the PR looks great, and is a big improvement both in our typing robustness and improving the architecture of our React components. Let me know what you think about the reducer typing and I'll merge it in afterwards. Thanks for taking this on!

FrigoEU:
Ok, I split them up as you asked. I'm not 100% satisfied but couldn't find any other solution. If I ever come up with anything else I'll make another PR.

extr0py:
Thanks for making the change! Looks good to me. I'm not sure if there is a better solution either, but we can always make a PR if we find another way to handle the typing. Overall this PR is a major improvement in how we handle typing of Config options. Later, I'm also thinking it would be nice to have completion and quickinfo specific for the `config.js` file - having this typing info would make that goal more achievable. 

Nice work, appreciate the contribution! I'll merge this in shortly.

Issue #335, opened 2017-03-31
UI - Error Marker: Error Marker overlaps error at end of line


extr0py:
The error marker UI is very helpful for real-time feedback of issues and problems in the code, however, our current design is impractical if the error is at the end of the line.

![image](https://cloud.githubusercontent.com/assets/13532591/24554927/2704b618-15e4-11e7-9f1a-c003b9ec176a.png)

In this case, the error overlaps the problem, so it is very difficult to fix.

A good inspiration for fixing this would be Atom's error message:
![image](https://cloud.githubusercontent.com/assets/13532591/24554958/3f0f2d1a-15e4-11e7-8a5e-aeafb8386e1a.png)

It shows up below the line, which makes it easier to work with. This adds a bit of complexity, because we have to decide whether to show it above or below the line (we'd want to show it above if the line is near the bottom of the screen). But the user experience would be much better.

keforbes:
Fixed by #357.

Issue #33, opened 2016-12-05
Synchronizing colors:


jordwalke:
I [commented about synchronizing keyboard shortcuts](https://github.com/extr0py/oni/issues/20) and the various options for doing so.

TL;DR:
> - No one wants to write VimScript (how can I blame them).
> - Everyone knows JS/JSON so it is preferable.
> - But we don't want a ton of startup time overhead to wait for a JS engine to warm up or wait for latency to cross the language bridge several times (if we can avoid it).
> - We also don't want half our editor config in VimScript and half in JS/JSON long term so if there are two ways to do something it would be nice if there was a migration path to have everything in one language, without paying any performance costs (a suggested approach is ahead of time compiling JS/JSON to VimScript - which also lets you use that same config in any build of Vim).

Now, I'd like to make an Issue to track the same kind of configuration experience for color schemes and fonts including colors of objects that don't have a corresponding element inside of stock Vim. (popup menus etc). It would be nice to configure everything in one place, just like keyboard mappings. It would also be great if every color scheme could work out of the box with UI extensions with a nice default color that synced to the color scheme bgcolor etc - but then *also* allow colorschemes to add customizations for these UI extensions. It would even be preferable if color schemes could be *defined* in JS, and then executed ahead of time to spit out regular vimscript config. Imagine if all the color blending logic was done in JS, and the output is raw vim highlight commands (this is much like our shift to css in JS).

extr0py:
Great feedback. I've been wondering how to handle color schemes. One option was just to stick with the default Vim colorschemes. 

I was taking a look at some other editors to see how they define colorschemes:
1) VSCode - uses TextMate themes
2) Atom - uses less files (makes sense based on the way it renders)

I thought at first about just using the vim color themes, but we run into the problem where the UI extension don't respect the themes at all. One option I looked briefly at was reading the set of highlights and applying them to the UI - trying to pick good defaults for the built-in UI elements like the menu, etc. 

I like the idea of having the colorscheme configuration be in JS too though, to be consistent with the other configuration. The hardest part with that is to bridge the gap so that users can bring in the themes they are comfortable with. 

So from that that, a `colors.js` file might look like this:
```
const config = {
"base": "onedark.vim",
"ui.bgColor": "black",
"ui.autoComplete.border": "gray",
"ui.autoComplete.background": "rgba(200, 200, 200, 0.5)",
..etc...
} 

export default config
```
[...]

PR #357, opened 2017-04-08
Redesigns the error text component (#335)

FrigoEU:
This is my initial attempt at (#335). It's really rough, especially the CSS is a bunch of magic numbers everywhere. Also the "algorithm" to decide whether to show the tooltip at the top or bottom is just a very simple heuristic, but I think it might work pretty well with very little performance impact. I haven't figured out how to do accurate text measuring so far.

I thought I'd show you and see what you think. Any comments very welcome :).

### DIFF
--- a/browser/src/UI/components/Error.tsx
+++ b/browser/src/UI/components/Error.tsx
@@ -27,12 +27,16 @@ export class Errors extends React.Component<IErrorsProps, void> {
             if (this.props.windowContext.isLineInView(e.lineNumber)) {
                 const screenLine = this.props.windowContext.getWindowLine(e.lineNumber)
 
+                const xPos = this.props.windowContext.getWindowPosition(e.lineNumber, e.startColumn).x
                 const yPos = this.props.windowContext.getWindowRegionForLine(e.lineNumber).y - (padding / 2)
                 const isActive = screenLine === this.props.windowContext.getCurrentWindowLine()
 
+                const showTooltipTop = this.props.windowContext.dimensions.height - this.props.windowContext.getWindowLine(e.lineNumber) <= 2
+
                 return <ErrorMarker isActive={isActive}
+                    x={xPos}
                     y={yPos}
-                    height={this.props.windowContext.fontHeightInPixels}
+                    showTooltipTop={showTooltipTop}
                     text={e.text}
                     color={e.color}/>
             } else {
@@ -65,8 +69,9 @@ export class Errors extends React.Component<IErrorsProps, void> {
 }
 
 export interface IErrorMarkerProps {
+    x: number
     y: number
-    height: number
+    showTooltipTop: boolean
     text: string
     isActive: boolean
     color: string
@@ -78,23 +83,40 @@ export class ErrorMarker extends React.Component<IErrorMarkerProps, void> {
 
     public render(): JSX.Element {
 
-        const positionDivStyles = {
+        const iconPositionStyles = {
             top: this.props.y.toString() + "px",
         }
+        const textPositionStyles = {
+            left: this.props.x.toString() + "px",
+            // Tooltip below line: use top so text grows downward when text gets longer
+            // Tooltip above line: use bottom so text grows upward
+            top: this.props.showTooltipTop ? "initial" : this.props.y.toString() + "px",
+            bottom: this.props.showTooltipTop ? "calc(100% - " + this.props.y.toString() + "px" : "initial",
+            borderColor: this.props.color,
+        }
 
-        let className = this.props.isActive ? "error-marker active" : "error-marker"
+        const className = [
+            "error",
+            this.props.isActive ? "active" : "",
+            this.props.showTooltipTop ? "top" : "",
+        ].join(" ")
 
-        const errorDescription = this.config.getValue("editor.errors.slideOnFocus") ? (<div className="error">
-            <div className="text">
-                {this.props.text}
-            </div>
-        </div>) : null
-
-        return <div style={positionDivStyles} className={className}>
-            {errorDescription}
+        // TODO change editor.errors.slideOnFocus name
+        const errorDescription = this.config.getValue("editor.errors.slideOnFocus") ? (
+            <div className={className} style={textPositionStyles}>
+                <div className="text">
+                    {this.props.text}
+                </div>
+            </div>) : null
+        const errorIcon = <div style={iconPositionStyles} className="error-marker">
             <div className="icon-container" style={{color: this.props.color}}>
                 <Icon name="exclamation-circle" />
             </div>
+        </div>
+
+        return <div>
+         {errorDescription}
+         {errorIcon}
         </div>
     }
 }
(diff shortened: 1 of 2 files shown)


keforbes:
Adding pretty pictures of this change.

Single-line error:
tooltip

Multi-line error:
large

Above line when close to bottom of window:
above

Functionally, this works beautifully, nice job!  I have some opinions about the colors though.  Is there a reason why you hardcoded that purple?  Did you take it from somewhere else in the current Oni styling?  I wonder if we could grab one of the colors from the current neovim theme to dynamically set the background color of that tooltip.  Or maybe just leave it gray as it was before.  Also, I liked how the previous implementation had a `0.8` opacity so it didn't *completely* cover the adjacent lines while still being readable.

I'm perfectly happy with the functionality of this error tooltip so we could merge it in, but I'd like to have a bit of a discussion about what color we want to give the background.  That's the only thing that stands out for me.  I'm afraid purple won't look good against all possible color schemes.

FrigoEU:
Oh yeah I'm definitely open to changing the background. I think I got the color from somewhere in the UI, I think from the line numbers, but I suppose it's something that should be themed. I tried to look around a bit, but I couldn't immediately figure out how the css/color structure and theming support works or would work, so I just tried to make the first iteration of something and build on that.

I'm personally not a fan of opacity, since it makes the message harder to read. I like it on the small error markers on the right since they're always visible, but this one is only visible when you're on that line. It's your call though, I'll change it if you want to.

bert88sta:
Getting this quasi-synced with the color scheme in vim is a must for the
future  I don't know if this is doable now

keforbes:
> Getting this quasi-synced with the color scheme in vim is a must for the future.  I don't know if this is doable now.

Ok, I thought it'd be relatively easy since we're already pulling out `backgroundColor` and `foregroundColor` from Neovim.  I'm not sure which theme property we'd use for the error background but I thought we had a set of colors to pull from.


> I'm personally not a fan of opacity, since it makes the message harder to read. I like it on the small error markers on the right since they're always visible, but this one is only visible when you're on that line.

That's fair, you can just move the cursor to a different line if you care that the error popup is covering up an adjacent line.

browser/src/UI/components/Error.tsx
```diff
@@ -78,24 +83,41 @@ export class ErrorMarker extends React.Component<IErrorMarkerProps, void> {
...
-
-        const errorDescription = this.config.getValue("editor.errors.slideOnFocus") ? (<div className="error">
-            <div className="text">
-                {this.props.text}
-            </div>
-        </div>) : null
-
-        return <div style={positionDivStyles} className={className}>
-            {errorDescription}
+        const className = [
```
extr0py: Minor suggestion here - you can use the `classnames` (https://github.com/JedWatson/classnames) module to make this a bit easier, without needing the `join(" ")`. We already have it in the project, the 'oni\browser\src\UI\components\LiveEvalMarker.tsx` uses it

extr0py:
I'm very sorry for the slow response on my end! Had some family obligations this weekend so wasn't on the computer much. This is an awesome PR @FrigoEU - great to see the design & usability improvements! Excellent work. 👍 👍 👍 

I really like the restyling and the thought you put in as to whether it should open above or below.

I agree with the sentiment around the background color - it would make sense for this to be consistent with the quick info, completion, and popup menu UI. We have the `@background-color` constant in .less we could leverage. Regarding the colorscheme - makes sense to sync once we are able to do that - that work is tracked by #33. But it won't be straightforward until we address that issue, so just making it match with the other elements would be a reasonable goal. 

I experimented a bit and using the `@background-color` constant along with a drop shadow (`box-shadow: 3px 3px rgba(0, 0, 0, 0.2)` doesn't look too bad and is relatively consistent:
image

The only other thing I was thinking about was - does it make sense for the errors to be centered under the item? Something like this:
image.

This is the sort of experience and change that helps move us towards the __Modern UX__. Looking at the current experience and this PR, it's night and day, and it is leaps and bounds ahead of the UI that is possible with the terminal limitations. Awesome job! 👍 👍 

The only item I'd like to see addressed is making the background color consistent with the other popup elements - otherwise I'm ready to merge it in. Can't wait to start using it! 😄

FrigoEU:
I made the changes to use the @background-color variable and the dropshadow as mentioned above. I think it's not a bit idea to try it out for a bit and then see what everybody thinks about it. There are still issue with the errors, like when you have two on one line they go over/under eachother, so this is definitely not the end. 

Just wanted to say thanks for the enthousiasm and encouragement I've been getting on all my PR's, it makes all the difference.

FrigoEU:
One thing that I noticed and that is really annoying is that autocompletions are positioned under the error messages. I can fix this with some z-index magic, but I think that's a pretty bad idea. As I can see, the completions get rendered inside the "overlay-ui" container which is present in the index.html so I think that's quite good, easy to find. The errors however get positioned inside the DOM by the OverlayManager, which just appends itself to the document on startup and then renders all its overlays inside, so it's always later in the DOM than the containers that are initially present in the index.html.

It hurts my head a little bit trying to figure out the proper structure to handle the correct level positioning of all these overlay elements. As I can see we have these "kinds" of overlays:

* QuickOpen / Ctrl-P
* QuickInfo
* Autocompletions
* Error markers / Errors
* Live Evaluation
* Scrollbar

Honestly, I don't think we'll have that many more overlay kinds popping up over time so maybe it's not that big a deal, but it is a bit annoying that some are managed by the OverlayManager (the bottom 3), and others are just rendered directly in the RootComponent (top 3). I'm more of a fan of keeping everything in the RootComponent hierarchy, but that's just me. I'd love some guidance on the direction we want to take here, if any.

extr0py:
> I made the changes to use the @background-color variable and the dropshadow as mentioned above. I think it's not a bad idea to try it out for a bit and then see what everybody thinks about it. There are still issue with the errors, like when you have two on one line they go over/under eachother, so this is definitely not the end.

Sounds great! I checked out the changes and they fit in well with the UI. Thanks for making the updates. As you mentioned we can always tweak it later, but this moves much closer to the experience you get with something like Atom!

> Just wanted to say thanks for the enthousiasm and encouragement I've been getting on all my PR's, it makes all the difference.

Definitely, thank you for taking the time to learn about and contribute to the project, and for being so open to feedback! Really appreciate your contributions.

> It hurts my head a little bit trying to figure out the proper structure to handle the correct level positioning of all these overlay elements. As I can see we have these "kinds" of overlays:

Unfortunately, the overlay management system is a mess at the moment. IMO it is the most convoluted and confusing part of the codebase at the moment. It hurts my head too and I wrote it 😦 The main problem is there is a bunch of complexity in terms of managing the 'overlays' today - because the Neovim instance is one giant `<div>`, and when there is a split or changes in the split size, there are a bunch of hacky hooks to get the errors and scrollbar to render 'on top' of the appropriate area. 

I'd like to look at _handling window management explicitly_ and _multiplexing Neovim instances_ to solve this problem.  This has been alluded to in a couple PRs, but I logged this issue - #362 - to explicitly track this. 

The goal with that work would be to keep everything in our React component hierarchy, so we don't hit those annoyances you mentioned. Let me know if you want to chat more. PR #303 is the first step I had in refactoring some of the React components and creating a React component to host the Neovim process - #362 might be more doable after that PR.

extr0py:
I'll bring this change in! Thanks for the contribution, @FrigoEU , excellent work!

Issue #98, opened 2016-12-15
Implement notification UX


bryphe:
Part of this wouldn't be fully baked until we get the Neovim PR to decouple the UI, however, we could still use the UI paradigm for alerts / existing errors in code

jordwalke:
This other issue might be related:

https://github.com/extr0py/oni/issues/39

It seems much better than the Atom UI notifications feature.

extr0py:
#365 makes progress towards this - I think it makes sense to keep open to track the alert / pop-up notification scenario, though.

PR #358, opened 2017-04-10
Fix config live reload, pop-up config.js parse errors

keforbes:
I was already logging to the console when `config.js` has parsing errors but I wanted to put the error somewhere the user would actually see it.  I changed the `console.log` to an `alert` so it now pops up in the user's face.

While testing this out, I was getting an error in `index.tsx` that `UI.Actions.setConfigValue` was not a method.  It turns out the method was actually named `setConfigurationValue`.  This was introduced in #356.  I changed the method name in `ActionCreators.ts` to match the expected name used in `index.tsx` and changes to `config.js` automatically take effect once again.

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -203,8 +203,7 @@ class Config extends EventEmitter {
             try {
                 userRuntimeConfig = global["require"](this.userJsConfig) // tslint:disable-line no-string-literal
             } catch (e) {
-                // TODO: display this error to the user somehow
-                console.log("Failed to parse " + this.userJsConfig + ": " + (<Error>e).message) // tslint:disable-line no-console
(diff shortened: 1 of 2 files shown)


extr0py:
Thanks @keforbes  for catching this regression and fixing it! Hopefully with #98 we can have a nicer way to show the errors eventually (this is a general problem not specific to the PR, lots of places where `alert` is being used), but just tried it out and the functionality is great.

FrigoEU:
Ajajaj that's my bad. I've been thinking this cast is really annoying: 

```
export const Actions = bindActionCreators(ActionCreators as any, store.dispatch)
```
And my first commit PR I already get bitten by it... I'll think about how we can keep this typesafe.

PR #359, opened 2017-04-11
Improve type safety of ActionCreators

FrigoEU:
As evidenced by (#358), the cast to any UI/index.tsx of the ActionCreators takes away a lot of the type safety that is present in Actions.tsx and ActionCreators.tsx. This PR makes sure we keep that type safety.

It even uncovered a bug immediately!


extr0py:
Nice, thanks for making this proactive fix to address the typing - this will definitely improve our "safety net"

Issue #313, opened 2017-03-21
Move `showNeovimInstallHelp` to redux+react


extr0py:
Right now, UI/index.tsx has a method that looks like this:

```
export function showNeovimInstallHelp(): void {
    const element = document.getElementById("overlay-ui")
    ReactDOM.render(<InstallHelp />, element)
}
```

This should be refactored to be driven by Redux + React

extr0py:
It could be potentially moved to `RootComponent`, we just need to have some redux state + actions to control rendering it.

keforbes:
Fixed by #360.

PR #360, opened 2017-04-11
Refactor InstallHelp into redux + RootComponent (#313)

FrigoEU:
This should take care of #313. I added the if-statement to the RootComponent. It was the only way to keep the current behaviour, namely showing none of the other UI elements when we're showing the install help. I'm ok with this placement, but if you want it somewhere else, ie. more localized in the InstallHelp component alone, let me know.

### DIFF
--- a/browser/src/UI/ActionCreators.ts
+++ b/browser/src/UI/ActionCreators.ts
@@ -186,6 +186,10 @@ export function setConfigValue<K extends keyof Config.IConfigValues>(k: K, v: Co
     }
 }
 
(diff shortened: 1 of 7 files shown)


extr0py:
Awesome, this is great! Having the `if` in RootComponent for this scenario makes sense to me. Maybe later we can have some sort of splash screen or something, but this works fine for now.

extr0py:
Meant to merge this in yesterday - bringing it in now. Thanks for the contribution, @FrigoEU !

PR #363, opened 2017-04-11
Fix #343, install to Oni/ rather than oni/ on Windows

### LINKED ISSUE #343: Windows Installer: Fix Casing of `Oni`
Currently in 0.2.1, the Windows installer installs Oni to `C:\Program Files (x86)\oni`. It should be `C:\Program Files (x86)\Oni` instead.

keforbes:
I have no idea if this works (no Windows box to test on) but it seems reasonable to me.  I'm really just committing this Pull Request to see how Appveyor treats it. :smile:

### DIFF
--- a/build/BuildSetupTemplate.js
+++ b/build/BuildSetupTemplate.js
@@ -18,13 +18,15 @@ shelljs.cp(sourceFile, destFile)
 const packageJsonContents = fs.readFileSync(path.join(__dirname, "..", "package.json"))
 const packageMeta = JSON.parse(packageJsonContents)
 const { version, name } = packageMeta
+const prodName = packageMeta.build.productName
 
 // Replace template variables
 
 const valuesToReplace = {
     "AppName": name,
+    "ProdName": prodName,
     "AppExecutableName": `${name}.exe`,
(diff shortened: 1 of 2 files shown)


keforbes:
Build succeeded.  Would anyone be willing to test the output from this build?

extr0py:
Awesome! I'll test it out tomorrow 😄

extr0py:
I tested it out in Windows, and it looks much improved! Good idea to use that build metadata, so we don't have to duplicate the name (and removing the existing duplication).

I still see a few of places where `oni` shows up instead of `Oni`, which looks out of place in Windows.
image

image
image

I'm thinking it just makes sense to set `"AppName": prodName` so it updates all the entries needed.

extr0py:
Thanks for your help with this! It will definitely improve the Windows experience.

keforbes:
I changed the case of the Start Menu entry.  I'm afraid to change `"AppName"` to `prodName` globally though.  I know Windows isn't case-sensitive but I'm afraid it may have unintended consequences.

extr0py:
Sure, I can try making that later and testing it out.

extr0py:
I'll merge this in now - thanks for your help with this!

Issue #362, opened 2017-04-11
UI: Externalize Window Management


extr0py:
There are several shortcomings to how Oni manages windows today.
- __Convoluted code path__ - the overlay / overlaymanager concepts are unnecessarily complex and confusing
- __Difficult to extend__ - it's non-trivial to add new functionality, like a git blame marker plugin, or something like markdown preview

For the overlay scenarios, it'd be really nice if each split had its own 'container' div. This really isn't possible today, as Neovim as a whole manages rendering to various splits.

I'd like to look at the possibility of 'multiplexing' Neovim instances to address these shortcomings. In effect, this would allow us to move window management to Oni, and allow us to really build on that experience - we could have first-class animations and much simpler overlay code if we manage the windows explicitly. Each 'split' would be its own Neovim process.

My original concern with this was that it would be difficult to persist settings between the splits - for example, marks, history, etc. @justinmk had a great write up on how this could work:
https://www.reddit.com/r/neovim/comments/5ym9p3/future_of_ui/

FrigoEU:
For me, this issue is super important, but also still very vague. I love Vim's editing experience but hate its extensibility story. Now most extensions that I'm interested in (and spacemacs is my biggest influencer here) are a git interface, keybinds explorer, compilation feedback, search, etc. These are all panels that are being drawn over/around the editor. 

I don't know how you see this, and I don't know how I think it should be, but I think if we could present a nice API to plugins (and I think React can help here) next to the "internal" panels, it would mean the world.

extr0py:
>  I love Vim's editing experience but hate its extensibility story.

I'm with you 100% here. This is exactly what motivated the creation of this project.

> For me, this issue is super important, but also still very vague

Definitely! I have some ideas in my head, but haven't had much chance to experiment yet... I think it's worth calling out some of the primary goals:
- Git interface
- Git blame annotations
- Keybindings explorer
- Compilation feedback
- Search
- Live Preview (React components, HTML, Markdown)
- File Explorer
- Plugin Manager

I see three broad classifications of these:
- _Floating Panels_ - these are independent pieces of UI that exist independently or on top of the editors. The log panel you implemented is a good example. The file explorer and git interface also potentially fit this, but would be on the side.
- _Overlays_ - these are UI entry points that place UI on top of an editor window. We have lots of examples today - the completion, the error bubble, the quick info bubble. Other items I could see extending this would be things like `git blame`
- _Additional Editor / Window Types_ - preview modes would fall into these. 

[...]

Issue #39, opened 2016-12-07
Better :messages log


jordwalke:
Here's a proof of concept for Atom's notifications that get rid of the spammy user experience of Atom's notifications. It would be cool to render Vim's `:messages` log in this manner:

https://github.com/atom/notifications/issues/29

extr0py:
Cool design, I hadn't seen it before, I like the idea a lot! It's nice to be able to opt-in to getting barraged with all the details. Would make a lot of sense to integrate in terms of messages, errors, etc.

I believe this RFC will make it easier to get at the messages:
https://github.com/neovim/neovim/pull/5686

bryphe:
This PR https://github.com/neovim/neovim/pull/7466 looks like it could help us here too-

PR #365, opened 2017-04-13
Notifications

FrigoEU:
Hey, 

This is a pretty big PR that corresponds to #98 most closely, in the future maybe to #39 and the design is stolen for the most part from https://github.com/atom/notifications/issues/29. I started from the atom css and markup but ending up rendering it in a normal table.

I split up the commits in 3: 
* The first one is the biggest one, adding the actual component, hooking it up to Redux, adding layout, adding some examples etc.
* The second adds some extra actions for showing/hiding, and adds a command to show the panel
* The third implements the notifications in the NeovimInstance and the Config

This is not really finished yet, but far enough along to show I think. Here's some screenshots, mixing the example data and some actual notifications:

screen shot 2017-04-13 at 11 18 25
screen shot 2017-04-13 at 11 18 38

Use the new command "Show Notifications" to see it pop open. Note that the first few notifications are examples.

What still needs to happen:
* Remove examples. I thought I'd leave them in if anyone wants to check
* Refactor some more console statements. I figure this can be done in a later PR
* The biggest: position the panel correctly. Right now I render it straight over the editor, hiding a good chunk of it. This is obviously not OK, but I'm not sure yet how to handle this in a robust way. I'd basically need to shrink and grow the .oni-text-editor when the changeNotificationVisibility actions come in, but this also comes into contact with #362 so I thought I'd hold off on that just for now until somebody else can take a look at it.
* Right now the only way you can view this new notifications panel is via a command. I think it would be nice to integrate it with #201 once it's ready.

Some other remarks on the code:
* I think emitting some new events from the NeovimInstance was the best idea, to avoid tight coupling. I just followed along with how it already works.
* Config was already an EventEmitter, but it wasn't emitting anything yet. I have it emit an event now + added a method to check for parsing errors. I wish I didn't have to do both, but because of how the Config initialises, you can never listen to the first parsing error being emitted so I "solved" it like this. Any input more than welcome.
* I made the CSS and markup as simple and straightforward as I could, but I again introduced a bunch of variables etc that need to be part of some design system/theming solution/etc.


extr0py:
Wow, this change looks awesome! Thanks a lot for including the screenshot. The UI looks really great and since it was inspired by Atom it has that modern UX feel.

This seems like it would be a convenient place for hosting the compile errors & warnings as well, at some point. The redux state changes look solid to me.

I added a few comments inline, and then just had these questions:
- Is it possible to style the scrollbar that comes up? At least on Windows, it looks pretty out-of-place:
- I'm wondering if the current component should be named `Logs`/`LogRenderer`? When I think of Notifications, I think of the pop-up notifications like:
image
I imagine we'll have the need for both styles of Notifications, so naming the component in the PR `Logs` will mean we can use `Notifications` for that pop-up experience.

Another thing I'm thinking about in the longer-term is how to keep Oni keyboard-friendly with these new components. Something like the vimium behavior might be really helpful - I created #366 to discuss this. Not something we need to tackle now but longer-term it'd be useful, especially as we add more of these rich UI experiences to Oni.

Excellent work, this looks great!

browser/src/NeovimInstance.ts
```diff
@@ -97,7 +97,7 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
 
         this._initPromise = startNeovim(this._pluginManager.getAllRuntimePaths(), filesToOpen)
             .then((nv) => {
-                console.log("NevoimInstance: Neovim started") // tslint:disable-line no-console
+                this.emit("logInfo", "NevoimInstance: Neovim started")
```
extr0py: Would you mind fixing my typo while you're here? 😄 I used `NevoimInstance` instead of `NeovimInstance`

browser/src/NeovimInstance.ts
```diff
@@ -326,7 +326,7 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
         this._initPromise.then(() => {
             this._neovim.uiTryResize(columns, rows, (err?: Error) => {
                 if (err) {
-                    console.error(err)
+                    this.emit("logError", err)
```
extr0py: Thanks for hooking these up to the new logging infrastructure

browser/src/UI/components/Notifications.less
```diff
@@ -0,0 +1,91 @@
+@import (reference) "./common.less";
+
+@icon-size: 30px;
+@font-family-monospace: Consolas, "Liberation Mono", Menlo, Courier, monospace;
+
+@component-padding: 10px;
+@border-color: rgb(25, 25, 25);
+// @font-family-monospace:
```
extr0py: Looks like there are a few unused lines here

browser/src/UI/components/Notifications.less
```diff
@@ -0,0 +1,91 @@
+@import (reference) "./common.less";
+
+@icon-size: 30px;
+@font-family-monospace: Consolas, "Liberation Mono", Menlo, Courier, monospace;
```
extr0py: This might be a good candidate to have in `common.less` - same for a few other constants.

browser/src/UI/components/Notifications.tsx
```diff
@@ -0,0 +1,147 @@
...
+                case "info": return "notification-info"
+                case "warning": return "notification-warning"
+                case "error": return "notification-error"
+                case "fatal": return "notification-fatal"
+                default: return ""
+            }
+        }
+    }
+}
+function mapStateToProps(s: State.IState): Partial<INotificationsProps> {
```
extr0py: Nice use of `Partial` here! I need to remember to leverage that..

FrigoEU:
I renamed it all to Log/Logs, it fits better indeed. I cleaned up some of the CSS + corrected display of long stack traces. 

I don't have a windows PC at my disposable right now, so I didn't do the scrollbars yet. I'll see what I can do later on.

extr0py:
Awesome, thanks @FrigoEU ! This looks excellent. Appreciate you taking the time to do the rename... we need to get the `rename` IDE functionality in here to make that easier :smile:

We can track the scrollbar issue separately, I'm happy to take a look at that since my main dev machine is Windows.

The only thing left is that we should remove the placeholder logs, could you please take those out? Otherwise I'm set to merge this in!

FrigoEU:
Oki, I removed the examples. 

Yes, it would be nice to track the scrollbars in a different issue. Thanks for any help you can give, otherwise I'll find some windows machine to run Oni on later;

extr0py:
Awesome, thanks for addressing those! I'll open an issue for the windows scrollbar. I'm merging in now - thanks for the contribution, @FrigoEU !

extr0py:
FYI - logged #370 to track the windows styling issues.

PR #368, opened 2017-04-14
Further improvements to error tooltip

FrigoEU:
I added some more improvements to the Errors component. The commits explain pretty well what's going on.

Commit 1 groups errors on the same line together. It just takes the most left message as the place where to put the arrow. The color gets decided kinda at random. I think the IErrorWithColor interface should rather have something like a severity (info, warning, error) so we can pick the color more intelligently. Screenshot: 

screen shot 2017-04-14 at 11 41 47

Commit 2 is the gnarliest one. There were a few problems with errors not being shown, or shown at the wrong place, when lines get wrapped. I think I fixed them all. The hardest part is keeping the 3 kinds of "lines" and "columns" apart. I tried to make it as consistent as possible, using the following names: 

* "Buffer" lines & columns are just the lines in the file, irrespective of what we're rendering, wrapping, whatever. Compilers, linters etc emit their messages with these kind of lines & columns.
* "Window" lines & columns. These are the same as Buffer lines, but 1-indexed from the top of the screen.
* "Screen" lines & columns. These are also 1-indexed from the top, but stricltly look at how the text is layed out. So if one buffer line is wrapped over 3 lines in the editor, it means it takes 3 screen lines.

Also getting a hold of the "offset" on the left side of the screen caused by eg. ```set numbers``` is awful, so I had to deduced it from the info I did have. If we could just ask that to neovim that would be much easier!

Commit 3: Squiggles were not shown correctly on wrapped lines + when startColumn and endColumn were the same, nothing was shown. You pretty much always want something to be shown imo, so I added a "minimum" squiggle of width = 1 cell.

Commit 4: Using the above functions from Commit 3, I added support for showing errors to the left if there is not enough real estate on the right side.

screen shot 2017-04-14 at 11 51 04

Phew!

### DIFF
--- a/browser/src/UI/Overlay/WindowContext.ts
+++ b/browser/src/UI/Overlay/WindowContext.ts
@@ -113,6 +113,9 @@ export class WindowContext {
         // It would be much easier if we could just ask neovim to give this to say how much the offset is, but I haven't found how so far
     }
 
+    /**
+     * Returns position of window line + column in pixels on the screen
+     */
     public getWindowPosition(windowline: number, column: number): Rectangle {
         const linePosition = this.getWindowRegionForLine(windowline)
 
--- a/browser/src/UI/Overlay/WindowContext.ts
+++ b/browser/src/UI/Overlay/WindowContext.ts
@@ -58,11 +58,11 @@ export class WindowContext {
         return typeof this._lineMapping[line] === "number"
     }
 
-    public getCurrentWindowLine(): number {
+    public getCurrentScreenLine(): number {
         return this._eventContext.winline
     }
 
-    public getWindowLine(bufferLine: number): number {
+    public getStartScreenLineFromBufferLine(bufferLine: number): number {
         return this._lineMapping[bufferLine]
     }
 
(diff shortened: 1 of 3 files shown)


extr0py:
This is really cool! It addresses one of the issues I started having with the new error marker system - sometimes they would overlap. Having this coalescing behavior is huge improvement in usability for that scenario!

And the `WindowContext` code to try and reason about the position of the window/screen/buffer/lines is one of the most complicated pieces of the code. I really like your explicit comments around some of the operations to make sense of that - I wish I would've done more of that earlier. The details around `getColumnOffset` are particularly helpful 👍 👍 The way you differentiated between `Buffer`, `Window`, and `Screen` lines makes a lot of sense. Thanks for renaming some of those methods to make it clear!

> Also getting a hold of the "offset" on the left side of the screen caused by eg. set numbers is awful, so I had to deduced it from the info I did have. If we could just ask that to neovim that would be much easier!

Definitely... I like the way you derived both of those from the `wincol`/`winline` - at least you solved that problem in a way that others won't have to worry about it! Even once we do the window management work in #362, we'll still have this issue for line numbers, unless we externalized those too - so it's great that this is there.

The code overall looks excellent to me!

The only thing I'm wondering about is if we can include the source of the error. For example, in this case,  I have two errors - one from `tslint` and one from `typescript`:
image

We set a key here in `index.tsx`:
```
	    pluginManager.on("set-errors", (key, fileName, errors, color) => {
	        errorService.setErrors(fileName, errors);
	        color = color || "red";
	        errorOverlay.setErrors(key, fileName, errors, color);
	        const errorMarkers = errors.map((e) => ({
	            line: e.lineNumber,
	            height: 1,
	            color,
	        }));
	        scrollbarOverlay.setMarkers(path.resolve(fileName), key, errorMarkers);
	    });
```
I think we could potentially use that. Right now they are a bit verbose, like `typescript-compiler` and `ts-tslint`. But I'm thinking we could have something like this:
image

Sort of inspired by this:
image

It'd be awesome to have the source next to the error (typescript/tslint in our case!). And you're right, it would be better to have some standard error types like warning/error/etc, instead of an arbitrary color sent from the error provider. 

This change would be important for my workflow, where I tend to ignore the yellow (tslint) errors in lieu of the red (typescript) errors. The colors don't matter as much to me as the source - I like to have a way to quickly differentiate between the compiler and lint issues. 

> Commit 4: Using the above functions from Commit 3, I added support for showing errors to the left if there is not enough real estate on the right side.

Nice - this a huge improvement!

Thanks for continuing to improve this, it's awesome to see the progress.

FrigoEU:
Just a heads up, I'm trying out some new design ideas for the errors since I'm not satisified with how it is currently. I'll make sure I bring in your request for mentioning from which plugin the errors are coming. Do you want this mostly when there are errors from multiple sources on the same line? Or just always?

Secondly this weekend I've been working on a PureScript plugin for Oni. Some stuff works really well, other things I don't know how to do. I'll let you work on the omnisharp plugin first before I bombard you with questions though :)

extr0py:
> Just a heads up, I'm trying out some new design ideas for the errors since I'm not satisified with how it is currently. I'll make sure I bring in your request for mentioning from which plugin the errors are coming. Do you want this mostly when there are errors from multiple sources on the same line? Or just always?

Sweet! I'm excited to check out the new design ideas. :smile: Having the source always present would be helpful - I like to be able to quickly differentiate if it is a typescript or tslint issue, for example, even if it is the only error on the line. The way Atom had it in that screenshot seemed pretty nice. Thanks for looking at this!

> Secondly this weekend I've been working on a PureScript plugin for Oni. Some stuff works really well, other things I don't know how to do. I'll let you work on the omnisharp plugin first before I bombard you with questions though :)

Awesome! I think that will make you the first person to write a plugin for Oni.  I'm trying to wrap up the initial language client work, and then let me know what questions you have. I'm going on vacation + camping the next week, so I may be a bit slow to reply. One recommendation I would have is to create a generic language service, if there isn't one already. I look through [langserver.org](http://langserver.org/) and unfortunately didn't see one for PureScript. The advantage of creating a language server is that other editors can benefit too. For TypeScript, for example, I'd like to either move the implementation we have today to a language server, or use an existing one like https://github.com/sourcegraph/javascript-typescript-langserver

Keep me posted though if you have any questions!

FrigoEU:
I agree a language server for purescript would be really nice, but that would take me much longer than a weekend to get running :). I talked about it with the maintainer of the ide engine of purescript and he's all for it, so maybe one day we can pool our efforts and get it running. In the meantime, maybe an in-between layer that translates psc-ide's protocol to LSP might be possible... We'll see! :)

extr0py:
> I agree a language server for purescript would be really nice, but that would take me much longer than a weekend to get running :).

Ah ya, makes sense 😄 In that case, `oni/vim/core/oni-plugin-typescript` has examples of some of the extensibility points. The `package.json` is particularly important because Oni decides which events to send to the plugin side based on that. Good luck!

bert88sta:
I'm not sure if you'll have fixed this or not, but opening a vertical or horizontal split whilst an error is present (under cursor) causes a bit of weirdness

bryphe:
With #804 , looks like this is set now, so I'll close this out. Thank you for the ideas, discussion, and proposal around this!

PR #386, opened 2017-04-19
Fix #374: Add 'editor.fontLigatures' boolean setting, and pick up when config c…

### LINKED ISSUE #374: Disable ligatures
Hi,

Although I like ligatures for coding with fonts like Fira Code, some other fonts (Space Mono) have ligatures that I would not like to see in the code (e.g. fi).
Is it possible to disable ligatures as a setting somewhere?

Many thanks,

Nikos

extr0py:
Adds `editor.fontLigatures` setting.

If _true_ (default), ligatures are enabled.
If _false_, ligatures are disabled.


extr0py:
Fix #374

PR #390, opened 2017-04-19
[Cleanup] change lists of else-if statements to switch statements

keforbes:
I had intended to only add a clause for the `set_icon` event so we wouldn't get the `Unhandled command: set_icon` warning anymore.  The `set_title` event tells us the window title to set, the `set_icon` event just tells us the title to use when the window is minimized.  Even the Neovim docs say:
```
In windowing systems not distinguishing between the two, "set_icon" can be ignored.
```

After making that change though, I felt compelled to change the list of `else if` statements into a `switch` statement.  I then looked for more occurrences of these lists of `else if` statements and only found one other offender.

So yeah, functionally, this Pull Request does nothing.  It just cleans up some annoyances in the code I've been wanting to address.

### DIFF
--- a/browser/src/NeovimInstance.ts
+++ b/browser/src/NeovimInstance.ts
@@ -340,26 +340,24 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
 
     private _handleNotification(_method: any, args: any): void {
         args.forEach((a: any[]) => {
-            const command = a[0]
-            a.shift()
-
-            if (command === "cursor_goto") {
+            const command = a.shift()
+            switch (command) {
+            case "cursor_goto":
                 this.emit("action", Actions.createCursorGotoAction(a[0][0], a[0][1]))
-            } else if (command === "put") {
-
+                break
+            case "put":
                 const charactersToPut = a.map((v) => v[0])
                 this.emit("action", Actions.put(charactersToPut))
-            } else if (command === "set_scroll_region") {
+                break
+            case "set_scroll_region":
                 const param = a[0]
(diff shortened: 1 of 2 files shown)


bert88sta:
Ive heard that hash tables aren't more efficient than switch cases in JS,
is this true because hash tables are implemented differently or because
switch statements are optimized into a hash table

bert88sta:
Just asking because in python this would all go into a dictionary lookup.

keforbes:
It's my understanding that switch statements are still preferred in javascript when there are more than 3 conditions.
http://stackoverflow.com/questions/2922948/javascript-switch-vs-if-else-if-else
http://archive.oreilly.com/pub/a/server-administration/excerpts/even-faster-websites/writing-efficient-javascript.html
https://www.safaribooksonline.com/library/view/high-performance-javascript/9781449382308/ch04s02.html

Of course, with all that said, I really did it because I find switch statements easier to read. :laughing:  I doubt these methods were performance bottlenecks.

bert88sta:
Interesting, I've seen some tests where larger lists  (100) in switch
statements are slower than object lookup

keforbes:
Oh definitely, object lookup would be faster than `else-if` or `switch` statements.  I don't think it's worthwhile to create an object of callback functions for this scenario though.  Like I said, I doubt this is a performance bottleneck, I just find it easier to read.

extr0py:
> Of course, with all that said, I really did it because I find switch statements easier to read. 😆

Thanks for doing this refactoring! Agreed 👍 The big if..else blocks are pretty annoying. Sometimes they start out as just a condition or two and then grow organically out of control.

PR #394, opened 2017-04-20
Fix #392, 'setMenu of null' with multiple instances

### LINKED ISSUE #392: Cannot read property `setMenu` of null
Occasionally, this error pops up:
image

It never occurs on first launch. Only when launching a second instance, and not all the time.

The following repro seems consistent:
1) Open `oni` at command prompt
2) Open a second instance of `oni`
3) Close second instance
4) Open another instance

At this point, I see the `Cannot read property 'setMenu' of null` error message. Might be a bug with how we are handling the lifecycle of the browserwindow or menu.

keforbes:
This is more of a workaround than a true fix but it works fine.  Basically, we're setting the `rebuild-menu` event listener on `ipcMain` but there is only a single instance of `ipcMain` across all Oni windows.  When one of the instances receives a `rebuild-menu` event from one of the other instances it would try calling `mainWindow.setMenu()` and get this error.

My fix is to just ignore the event if `mainWindow` is null.  So it avoids the error but I feel like in the future we should find a way to isolate the Oni instances a little better.

### DIFF
--- a/main.js
+++ b/main.js
@@ -70,7 +70,9 @@ function createWindow(commandLineArguments, workingDirectory) {
     })
 
     ipcMain.on('rebuild-menu', function(_evt, loadInit) {
-        updateMenu(mainWindow, loadInit)
+        if (mainWindow) {
+            updateMenu(mainWindow, loadInit)
+        }
     })
 
     // and load the index.html of the app.
--- a/main.js
+++ b/main.js
@@ -70,6 +70,8 @@ function createWindow(commandLineArguments, workingDirectory) {
     })
 
     ipcMain.on('rebuild-menu', function(_evt, loadInit) {
+        // ipcMain is a singleton so if there are multiple Oni instances
+        // we may receive an event from a different instance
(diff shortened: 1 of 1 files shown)


extr0py:
> My fix is to just ignore the event if mainWindow is null. So it avoids the error but I feel like in the future we should find a way to isolate the Oni instances a little better.

Sounds good for now. At least this will suppress the error. I think that the event object returned from `ipcMain` has a `WebContents` instance through the `sender` event property of the object it returns. It looks like we need the `BrowserWindow` instance though - but maybe we can derive that from the `WebContents`. That way, we could pass in the `BrowserWindow` that is relevant for the event to `rebuildMenu`.

I'm also wondering if later there is a way to handle more of the menu logic in `browser/src` instead of building it from `main.js` - I think we might be able to call those methods directly from the browser process without needing to ipc out, and that might help us out here.

In the meantime this will definitely help alleviate that pain point - thanks for the fix, @keforbes !

Issue #404, opened 2017-04-26
Ctrl+PageUp/Ctrl+PageDown bindings don't work


unclechu:
I have bindings like this;
```viml
" moving tabs
nnoremap <C-S-PageUp>   :tabm-1<CR>
nnoremap <C-S-PageDown> :tabm+1<CR>

" jump by half of screen by pageup/pagedown
nmap <PageUp>     <C-u>
nmap <PageDown>   <C-d>
xmap <PageUp>     <C-u>
xmap <PageDown>   <C-d>
" default jump by pageup/pagedown with shift prefix
nmap <S-PageUp>   <C-b>
nmap <S-PageDown> <C-f>
xmap <S-PageUp>   <C-b>
xmap <S-PageDown> <C-f>
```
But only these works in the Oni:
```viml
nmap <PageUp>     <C-u>
nmap <PageDown>   <C-d>
xmap <PageUp>     <C-u>
xmap <PageDown>   <C-d>
```

extr0py:
Thanks for the issue! There are a few places where we are hard-coding keybindings in the code (while working on a full solution for #20). Out of the box, `<C-PageUp>` and `<C-PageDown>` switch between active Oni processes, but I think I may be the only one using that functionality and it isn't well documented.

They are hardcoded here in `browser/src/index.tsx`:
```
        if (key === "<f12>") {
            commandManager.executeCommand("oni.editor.gotoDefinition", null)
        } else if (key === "<C-p>" && screen.mode === "normal") {
            quickOpen.show()
        } else if (key === "<C-P>" && screen.mode === "normal") {
            tasks.show()
        } else if (key === "<C-pageup>") {
            multiProcess.focusPreviousInstance()
        } else if (key === "<C-pagedown>") {
            multiProcess.focusNextInstance()
        } else {
            instance.input(key)
        }
```

For the `<C-pageup>` and `<C-pagedown>` bindings, we could do an interim solution while #20 is in progress:
- Create __commands__ to represent these, like `oni.nextActiveProcess` and `oni.previousActiveProcess`
- Create a VimL bridge for executing commands (like `OniExecuteCommand('oni.nextActiveProcess')`), and then in my `init.vim` I can include these bindings. Once the work for #20 is complete, I could include these in `config.js` instead.

bert88sta:
[...]

Issue #366, opened 2017-04-13
Keyboard Accessibilty for non-Vim/Neovim features


bryphe:
One major selling point for me in using Vim/Neovim is keyboard accessibility. In fact, the reason I started using Vim in the first place was because I had trouble coding on an airplane and hated touchpads ✈️ 😄 

At the same time, Oni is exploring bringing in some modern, non-terminal UX features that are present in more modern IDEs. The log viewer in #365 is an example - some other examples might be a file browser, a git repo history browser, markdown preview, etc. I believe it is still important to preserve the keyboard accessibility of these features in a way that makes sense to Vim users.

I'd like to propose adding capabilities similiar to `Vimium` (https://github.com/philc/vimium)  or `cVim` in the browser - these let you specify a specific hotkey, like `f` and jump to elements on the page.

The behavior could be as follows:
- Hotkey to trigger behavior: `<Alt+s>` for navigation with the current buffer/window/split, `<Alt+Shift+s>` for navigation across Oni
- Once the hotkey is triggered, we would show labels against all interactive elements, something like this:
![image](https://cloud.githubusercontent.com/assets/13532591/25020034/c72b9c9e-2041-11e7-8548-bb489c0a864f.png)

This would make it easy to quickly navigate across various pieces of UI.

Some example use cases:
- Trigger a button in a sidebar menu
- Trigger a status bar item
- Open a markdown preview window, use this behavior to click a link
- Open a separate browser window, use this behavior to navigate

justinmk:
I wonder what mechanism could be added in core nvim to help with this. E.g. for the examples of an embedded web browser, or a graphical file tree, one can ask: what do users want carried over to those components?

The most basic thing that stands out is key mappings. As I user I would want to be able to ctrl-w to the graphical window like any other text window, as well as control its size, and have buffer-local mappings there.

These graphical windows could be backed by a special buffer type (or even just a blank, normal buffer with a name like `[External]` and `filetype=extfoo`).  Then users can set up mappings using the typical approach, on a FileType event: `au FileType extfoo nnoremap <buffer> a b`. The external UI would probably need to ignore the cursor position reported by nvim in these buffers.

What's the minimum needed from core nvim to support this somewhat elegantly for external UIs?

jordwalke:
[...]

PR #408, opened 2017-04-27
WIP: #404 - Remove hardcoded PageUp key bindings

extr0py:
As part of this, this would be a good time to sketch out how configurable keybindings could look like on the Oni side (and then later, build optimizations on top of it, along with heavier synching of VimScript keybindings like in #20)

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,21 @@
+Start using JS for keybindings:
+- `keybindings.js`
+- Format like:
+```
+{
+    "<C-PageUp>": "editor.quickOpen.show",
+    "<C-PageDown>": {
+        "normal": "editor.quickOpen.show",
+        "insert": "..."
+    },
+    "<Enter>": (context) => { ... }
+}
+```
+
+- Add configuration: default keybindings
+- Load keybindings asynchronously
+
+- Add performance profiling for loading keybindings
+- Refactor the hardcoded bindings to use the new format
+
+- Update documentation
--- a/browser/src/Services/Commands.ts
+++ b/browser/src/Services/Commands.ts
@@ -11,6 +11,8 @@ import { IBuffer } from "./../neovim/Buffer"
 import { INeovimInstance } from "./../NeovimInstance"
 import { PluginManager } from "./../Plugins/PluginManager"
 
+import * as MultiProcess from "./MultiProcess"
+
 import * as UI from "./../UI/index"
 
 import { CallbackCommand, CommandManager } from "./CommandManager"
@@ -58,6 +60,10 @@ export const registerBuiltInCommands = (commandManager: CommandManager, pluginMa
                             "Show Logs",
                             "Show all logs in the bottom panel",
                             () => UI.Actions.changeLogsVisibility(true)),
+
+        // Inter-process management
+        new CallbackCommand("oni.focusNextInstance", "Focus Next Instance", "Focus the next running instance of Oni", () => MultiProcess.focusNextInstance()),
+        new CallbackCommand("oni.focusPreviousInstance", "Focus Previous Instance", "Focus the previous running instance of Oni", () => MultiProcess.focusPreviousInstance()),
 
         // Add additional commands here
         // ...
--- a/browser/src/Services/MultiProcess.ts
+++ b/browser/src/Services/MultiProcess.ts
@@ -6,12 +6,10 @@
 
 import { ipcRenderer } from "electron"
 
-export class MultiProcess {
-    public focusPreviousInstance(): void {
-        ipcRenderer.send("focus-previous-instance")
-    }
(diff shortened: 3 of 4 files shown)


extr0py:
WIP to Fix #404

bert88sta:
@extr0py 

I think the first thing we need to decide about is whether or not we're going to send keys on a "round trip" or whether oni should intercept bindings it recognizes. On one hand, I like the idea of an init.vim config because our convert to vimkey method doesn't handle complicated bindings well (leader key, for example, which I know for sure I want to be able to use in an oni config for control of UI). On the other hand, there are definitely benefits to be had handling certain actions in-house (not having to come up with a  complicated scheme for interpreting keymappings). I think this is going to clear up well as more UI business becomes externalized

Regardless, the problem above doesn't change my overall thoughts too much. Here's my idea

1) Define a set of actions that a keybinding could take, i.e scrollDown, sendKeyToVim, etc.

2) People can execute a set of actions linearly that they assign to a key. If we're clever and let them send keys to neovim, this allows us to make macros that work across oni and neovim both (!) I might want to do a UI selection and then completion, and then tell neovim to do some manipulation afterwards all in one keybinding.

extr0py:
@bert88sta Yes, good point, I don't think this has been called out explicitly.

> I think the first thing we need to decide about is whether or not we're going to send keys on a "round trip" or whether oni should intercept bindings it recognizes.

Definitely. My plan is to handle the keys directly, for the keys that Oni knows how to handle. This is for a couple of reasons:
- The round-trip through (Neo)Vim incurs a non-negligible delay (10-20ms), so if we can handle it directly, that helps performance.
- There are some features I'd like to implement that would be outside of Neovim. Things like #366.

But Neovim has rich input handling, especially for the text-edit experiences, so for editing, I'd prefer to delegate to Neovim.

I know this is heresy, but I've been looking at the Emacs model for keybindings. Emacs' functional approach combined with key binding is pretty powerful. The JS runtime has the potential to have similiar flexibility. Here's sort of an API I was thinking about:
```
// My plugin.js

const activate = (Oni) => {

    const showRefactorMenu = () => {
        Oni.showPopupMenu(["Extract Method", "Sort Imports", ...], doRefactor)
    }

    const doRefactor = (selectedOption) => {
        // do some refactoring stuff..
    }

    Oni.keys.bind("<C-A-r>", Mode.All, refactor)
}
```

As you mentioned, being able to also send keys to Neovim, or execute a function, would be pretty powerful too. Would just need to expose that in the `Oni` plugin object, which would be straightforward. That hybrid behavior - of doing stuff in Oni and having Neovim manipulate, etc is basically how most of the current UI enhancements are implemented anyway - it makes a lot of sense to expose those to keybindings and plugins, as I'm sure people can think of really creative things to do with that extensibility.

I'd also like to have a way for people more comfortable in the VimScript world to be able to customize the Oni-facing functionality as well - this is sort of along the lines of your #1 suggestion:

```
// Some plugin exposes a 'command' - a named string mapped to the function

const activate = (Oni) => {

    const showRefactorMenu = () => {
        Oni.showPopupMenu(["Extract Method", "Sort Imports", ...], doRefactor)
    }

    const doRefactor = (selectedOption) => {
        // do some refactoring stuff..
    }

    Oni.commands.register("command.refactor")
}

// And then later, in `init.vim`

nnoremap <C-A-r> :call OniCommand("command.refactor", ..someOptionalArgs)
```

I'm holding off on this PR for now, to think through that model a bit more. Let me know if you have any ideas - thanks for your thoughts!

bert88sta:
I think we're definitely in agreement about the functionality of Oni-side
bindings.

This is one of those situations where I'm again going to tout object lookup
as being superior to if or switch statements, especially if we're checking
 hundreds of key chords.

I know that vim round trip is slow, but if I wanted to use my leader key,
(space bar) in a keybind, Say <Space>gb for git blame overlay.  How can we
avoid sending that to neovim?

extr0py:
> I think we're definitely in agreement about the functionality of Oni-side
bindings.

Awesome!

> This is one of those situations where I'm again going to tout object lookup
as being superior to if or switch statements, especially if we're checking
 hundreds of key chords.

Yep, sounds like we're on the same page. Having dynamic / configurable bindings will necessitate object lookup anyway.

> I know that vim round trip is slow, but if I wanted to use my leader key,
(space bar) in a keybind, Say <Space>gb for git blame overlay.  How can we
avoid sending that to neovim?

For this scenario, we might need to have an Oni-side setting for a leader key (and potentially have it sync / integrate with the Vim-side configuration).

bert88sta:
Alright, I'm not sure exactly how neovim does it, but getting a leader key chord is usually based on a timeout IIRC, I'm going to dig through vimdoc and see if I can find more implementation details

extr0py:
No longer needed with #633

PR #409, opened 2017-04-27
Fix #377 - Refactor the Oni language service types to use `CompletionItemKind` from the vscode language server

### LINKED ISSUE #377: Refactor code to use `CompletionItemKind` constants instead of string constants
In the auto-completion dialog, we are using string constants instead of `CompletionItemKind` members - we should switch to using the language protocol constants instead of Oni-only constants.

extr0py:
This is an incremental step towards moving to a LSP-backed implementation of the TypeScript language service. There are several enums and entries, like `CompletionItemKind`, that were created one-off based on the typescript standalone server implementation, but this refactors to use the language server types provided in `vscode-languageserver-types`, so that we don't need a duplicate or parallel implementation.


extr0py:
WIP to Fix #377

PR #410, opened 2017-04-27
[WIP] Fix #380 - Use sourcegraph's typescript language server

### LINKED ISSUE #380: Implement TypeScript services as language server
The code we are using currently for the TypeScript language functionality should either be refactored to a language server, or preferably, we should leverage a common language server functionality like here:
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_signatureHelp

extr0py:
WIP for #380 

It looks like there are several issues with using the typescript language server:
- Completions are slow (could be due to the document sync strategy, or completion resolution not being implemented, or a combination of the two).
- QuickInfo data is not in the expected format.

- Live evaluation will need to be brought out of the LSP.

Should log follow up issues in https://github.com/sourcegraph/javascript-typescript-langserver/ to track these. Once those are addressed, it'll be viable to replace the existing typescript implementation with their LSP.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
+    "javascript-typescript-langserver": "2.0.0-pre.25",
     "minimist": "1.2.0",
     "mkdirp": "0.5.1",
     "neovim-client": "2.1.0",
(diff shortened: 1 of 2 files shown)


extr0py:
Will close this for now, but keep the branch around - will revisit as that project matures (or I get more time to contribute back to it)

PR #411, opened 2017-04-28
[WIP] Fix #384 - Python Language Server support

### LINKED ISSUE #384: Integrate Python language server
Palantir has a python language server implementation: https://github.com/palantir/python-language-server

extr0py:
This change leverages the `pyls` language server from https://github.com/palantir/python-language-server.

There are a few tweaks and additional settings to support, but that will make Oni more robust at supporting an array of language servers.


extr0py:
WIP for #384

extr0py:
Remaining work:
[ ] Validate on OSX (and ensure there are no regressions to C# / go language)
[ ] Update documentation to describe installation steps

extr0py:
There actually is a newline issue as well... It looks like the `\r\n` returned by the python language server isn't probably understood by `messageReader` from `vscode-jsonrpc`. Need to figure out if this is a bug in `pyls` or `vscode-jsonrpc`. I worked around it temporarily by changing `pyls` in the way it sends down the header info - but not sure if it is the proper fix.

extr0py:
Looks like the `\r\n` issue is due to python's newline conversion on Windows. The `\n` gets expanded to `\r\n` on Windows (see: https://stackoverflow.com/questions/2536545/how-to-write-unix-end-of-line-characters-in-windows-using-python), so this ends up with `\r\r\n`, which `vscode-jsonrpc` is not equipped to parse.

extr0py:
Tracking newline issue: https://github.com/palantir/python-language-server/issues/53

extr0py:
Validated the server works well on OSX... only thing left is documentation 😄

extr0py:
Need to double-check Go & C# on OSX

extr0py:
C# works fine, but looks like there is a regression on Go - potentially related to the `file:///` prefix. Need to investigate that. It would also be helpful to have a bit more logging in the console when things go wrong - looks like the `text/didOpen` call is hanging.

extr0py:
Validated Go/C#/Python on Windows & Mac with latest changes.

PR #415, opened 2017-04-29
Fix # 391: Errors overlay the completion logic.

extr0py:
This change disables the error details while in insert mode, and shows them again in other modes. 

This is really a hack, the better long-term solution is to move the error info to the redux store, and use the information there to gate the rendering. This change unfortunately just adds complexity to the `ErrorOverlay` concept, which is already overly complex, due to how the overlay/window management is implemented. Looking forward to implementing #362 to improve this story.


extr0py:
This was a blocker for the C# autocompletion, which is the main feature implemented so far, so I wanted to get it in prior to pushing the release.

PR #417, opened 2017-04-29
Add zip package to OSX

extr0py:


extr0py:
Looks like the build issues may be related to: https://github.com/electron/electron/issues/9323

extr0py:
Seems like packing a zip would need extra work, because by default macOS zip publishes to squirrel auto-updater.

PR #418, opened 2017-04-30
Move some dependencies to be bundled in the webpack bundle

extr0py:
- Reduce bundled dependency footprint


extr0py:
Fix #419 by including the needed dependencies

PR #428, opened 2017-04-30
Fix #426: Use common utility for spawning script

### LINKED ISSUE #426: v0.2.3-rc2 - OSX: C# plugin not loading
Getting an error on initialize of the omnisharp-server

extr0py:
- Use the common `spawnNodeScript` utility for spawning the language service
- Pass along environment variables to spawned processes

### DIFF
--- a/browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts
+++ b/browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts
@@ -25,9 +25,9 @@ export const ProtocolConstants = {
     },
 }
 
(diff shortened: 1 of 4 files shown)


browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts
```diff
@@ -25,9 +25,9 @@ export const ProtocolConstants = {
     },
 }
 
-export const wrapPathInFileUri = (path: string) => "file:///" + path
```
extr0py: Should try reverting this change, I don't believe it is necessary and may break Windows

PR #439, opened 2017-05-02
Fix #298 - Initial spectron validation (integration tests)

### LINKED ISSUE #298: Build: Integrate `spectron` tests for simple validation
Currently, our CI integration only runs the unit tests. It's easy to break "integration scenarios", though - and having the app actually boot and do some simple edits would go a long way in terms of keeping the app stable, especially across platforms.

Spectron (https://github.com/electron/spectron) seems like a really convenient way to do this. 

Our first test should simply:
1) Start `oni`
2) Validate the neovim process gets started
3) Validate a command can get processed (for example, an `echo` command) without error

In addition, this should be hooked up to run as part of `CI` builds, and easily run from the command-line via `npm run integration-tests`

extr0py:


extr0py:
Fix #298 - Add basic integration tests

extr0py:
Seems like the error now is a permission issue with the shellscript - need to make sure git tracks the executable permission: http://stackoverflow.com/questions/33820638/travis-yml-gradlew-permission-denied

extr0py:
Fix #298

extr0py:
Disabled Linux tests for now, it seems there are issues with the virtual display driver - and it will be more work to set up an install of Neovim on travisCI. Having validation across OSX and Windows covers a large surface area.

Issue #383, opened 2017-04-19
Language Client: Show language server status in status bar


extr0py:
Dependent on #201 - but it's helpful to get an update of what the language server is doing. In some cases, like Omnisharp, the initial load can be slow since it downloads a binary - it'd be helpful to have an indicator / spinner / error icon to show the state of the language server.

Galbar:
Is this a duplicate of #485 ?

extr0py:
Derp... this sure is a duplicate! Thanks for catching it, @Galbar ! I'll close this one out since #485 has a bit more info.

PR #444, opened 2017-05-04
[WIP] Fix #201 - External status line

### LINKED ISSUE #201: Beginning external status line implementation

extr0py:
This is implementing the status bar changes on top of the DOM refactoring / `IEditor` refactor. This enables the creation of UI that lives more cleanly outside of the DOM rendering.

Here's a screenshot of it with some hardcoded values:
image

Need to make it look a bit nicer and implement the actual plugins. It'd be nice to implement some default color-coding and vary up the font colors a bit for a cleaner look.

The lightning bolt by `typescript` is a placeholder for #383 - I'd like to use that or some icon to show the status of the language service. The lightning bolt would be active when there is a language server available for a particular language.


keforbes:
I don't have strong feelings about this, I'm just curious.  Would this feature prevent users from using plugins like airline and powerline?

extr0py:
Good question @keforbes - it's worth calling out and should definitely be documented.

This wouldn't preclude using airline and powerline - the default config wouldn't use airline/powerline (or vim's default statusline), so by default they would not be used. However, if the user did override the default (`oni.useDefaultConfig` with `false`) and used their own `init.vim` (`oni.loadInitVim` with `true`), then they would get both statuslines.

My plan is to add an extra configuration value, like `oni.statusbar.enabled`, that can be set, to support that scenario. If someone is using airline or powerline, they'd want to set this to false, unless they want two statusbars 😄 

I'll add this configuration value, and also add information to the documentation. Thanks for calling it out.

bert88sta:
Hey, I've been keeping tabs on this and I want to say: Good work! It's getting to look like a really nice, sleek status line. The only thing that's going to suck (and this is something simple enough that i could probably help with it ) is color scheming. As a chore, once this is in, we're going to have to add probably 10-20 of the most popular color schemes and have the status line conform to them. Also we should make the insert|visual|normal|replace block overridable in terms of color, that way people can fix a color theme we might not support. I'm hoping this really becomes a selling point for oni in terms of configurability of the status line. 

In addition, we might want to throw in some of the typical stuff from powerline/airline, such as percentage scrolled, dirty marker, whether or not it's a new file etc.

Keep up the great work!!!

I'm hoping that over the summer I'll have a lot more time to contribute since I'm graduating high school and won't have any more dumb summer assignments

extr0py:
Thanks @bert88sta ! Ya, I like the way it is looking with the default theme too. I hope that I can get some more time to polish the rest of the UI - I want Oni to be sleek and fast 😄

> The only thing that's going to suck (and this is something simple enough that i could probably help with it ) is color scheming. As a chore, once this is in, we're going to have to add probably 10-20 of the most popular color schemes and have the status line conform to them.

Agree with the colorschemes - we still need to address #33, which will help this problem in general. I'm hoping that work can help us automatically leverage some of the colors, but you're right, it might help to make some of the other status elements more configurable.

> In addition, we might want to throw in some of the typical stuff from powerline/airline, such as percentage scrolled, dirty marker, whether or not it's a new file etc.

Ya, for sure. For the dirty marker / new marker, I'd like to have a 'buffer list' above - a lot like the tabs in Atom. But we could potentially add some of these as optional items and flag them with configuration - and it's pretty easy to add new ones with the current API. Some other things I'd like to add (but might track in a separate issue) are the current git branch, the filetype + language server status, and error / warnings count. Lots of possibilities - I hope it's easy to extend and customize.

And congrats on graduation, that's really exciting! 👨‍🎓 🎉 😄

Issue #379, opened 2017-04-19
Language Client: Implement syntax highlighting


bryphe:
Language protocol-supported syntax highlighting can be built using the documentSymbol request:
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_documentSymbol

jordwalke:
Is there a really good example of an editor using this without a Flash of Unstlyed Content or other lags? Does VSCode currently use this? Does going out of process make performance more difficult? I'm not skeptical, just curious.

extr0py:
VSCode actually does this - but for large files there is a large flash of unstyled content. My plan right now is to leverage the Vim syntax highlighting for fixed items - like keywords, block styling, strings, comments, etc - and use the `documentSymbol` request for more dynamic items, like variable names.

But ya, as soon as you go out of process, you always incur some latency - so you get that flash of unstyled content. The variable name highlighting is pretty useful IMO - but there could be a setting to disable it if it is too distracting. There are also ways we can look at optimizing the performance, too, across the pipeline in various ways (for using IPC vs stdio, to only requesting symbols for the active 'window', pre-emptively loading symbols for a doc, etc). It seems doable to minimize this, since it's not really noticeable in VSCode after the initial "flash".

bryphe:
This is partially implemented via TextMate grammars, behind the `experimental.editor.textMateHighlighting.enabled` flag here: https://github.com/onivim/oni/wiki/Configuration#textmate-highlighting

bryphe:
Textmate highlighting is now enabled by default with #1639

PR #445, opened 2017-05-06
Fix #443: Implement fixes to support Go langserver (generalize language server support)

### LINKED ISSUE #443: Go-langserver
Hi
Could you implement the golang server at https://github.com/sourcegraph/go-langserver/ ?
I for some reason cannot make it run. I copied the Omnisharp extension replaced the references to cs/omnisharp-server to go/go-langserver but it does not work. I open a go file it states Successfully activated but literally nothing happens. No autocomplete, no defs on hover, no linting, nothing.

extr0py:
This implements several improvements to the Language Service in order to support other language services.

- Generalize the language client initialize to support either running a shell command (which is needed for running the python, go, rust, etc language services), or a script module (omnisharp, typescript, etc).
- Update QuickInfo to further support `MarkedString`.
- Fix bug in go-to definition on OSX / Linux to properly support file paths.
- Remove the `getHighlights` call which isn't supported by any language service for now - it seems like this is causing a crash in the go language service

There are a couple of caveats with this service:
- The Go language service doesn't seem to support Windows yet (sourcegraph/go-langserver#113).
- Completion isn't currently not supported.
- Looks like highlighting via keywords could be supported,  but still need #379 implemented on the Oni side to support this.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,9 +1,0 @@
-- Set up initialization path to differentiate between a node script starting and a shell executable
-- Need configuration variables to be passed to plugin?
-
-
-interface ServerOptions {
-    shellCmd?: string
-    module?: string
-    args?: string[]
-}
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,2 @@
+- Set up initialization path to differentiate between a node script starting and a shell executable
+- Need configuration variables to be passed to plugin?
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,2 +1,9 @@
 - Set up initialization path to differentiate between a node script starting and a shell executable
 - Need configuration variables to be passed to plugin?
+
+
+interface ServerOptions {
+    shellCmd?: string
+    module?: string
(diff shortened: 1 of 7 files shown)


extr0py:
WIP to Fix #443

fasa123:
For autocompletion https://github.com/nsf/gocode is used It is pretty much the default autocomplete provider (honestly I haven't seen any other). Due to the nature of Golang you will find multiple tools for those "IDE-Tasks" (refactor, reference counting, autocomplete etc.). 

The keyword highlighting is pretty irrelevant since Golang only has like 16 keywords. Obviously it's a nice thing to have but manually specifying the keywords shouldn't be the problem with only 16.

extr0py:
Cool, thanks for the info! It looks like that `gocode` provider supports Vim/Neovim out of the box - but let me know if there are issues with Oni. 

There's an open issue here as well for sourcegraph's language server: sourcegraph/go-langserver#77 but doesn't look like much action there.

Issue #229, opened 2017-02-20
Performance: Improve performance of FastBuffer/walker


extr0py:
Looking at the latest profile of typing in ONI after the plugin fix to remove IPC (which removes that as the hottest codepath), I see a couple of new hot points:
![image](https://cloud.githubusercontent.com/assets/13532591/23133193/26742354-f745-11e6-8a24-b3cebf501d71.png)

These are part of https://github.com/neovim/node-client - called when it is decoded the msgpack messages. 

There are a couple of avenues to optimize this path:
- Reduce the data we send over msgpack-rpc. One place that impacts this is sending over the entire buffer on updates (`OniNotifyBufferUpdate` in `vim\core\oni-core-interop\plugin\init.vim`). Reducing this potentially decreases the amount of time we spend copying in FastBuffer, but walker is still pretty hot.
- Investigate alternatives to those libraries - for example, walker does a really  general object traversal - maybe there are some shortcuts we can take to improve the performance? If so, we can push this back to the neovim/node-client project and any other JS consumer would benefit too.

bert88sta:
@extr0py: 

I've done a bit of preliminary work on this. I've added this to the init.vim

```
function OniNotifyBufferUpdate()

    if !exists("b:last_change_tick")
        let b:last_change_tick = -1
    endif

    if !exists("b:last_cursor_line")
        let b:last_cursor_line = 1
    endif

    if b:changedtick > b:last_change_tick
        let b:last_change_tick = b:changedtick
        let buffer_lines = getline(1,"$")
        if b:last_cursor_line < line(".")
            let buffer_lines = getline(b:last_cursor_line, line("."))
        elseif b:last_cursor_line > line(".")
           let buffer_lines = getline(line("."),b:last_cursor_line)
        "elseif b:last_cursor_line == line(".")
        "    let buffer_lines = getline(".")
        endif

        let context = OniGetContext()
        call OniNotify(["buffer_update", context, buffer_lines])
        let b:last_cursor_line = line(".")
    endif
endfunction```

Now I'm assuming that whatever handles the update logic doesn't account for it just being a string.

The commented out lines will stop oni from starting. I dont' know why or where. However, if i source the file while oni is open, I can't produce a vim error. I think the reason is because getline(".") returns a string instead of the list you expect. Unfortunately, this is the most common case for editing, where it's being typed on one line. If you could point me in the direction to fix this I think we'd see a drastic improvement

extr0py:
[...]

PR #456, opened 2017-05-19
Perf: Optimize Layers / Rendering Strategy

extr0py:
__Issue:__ I was benchmarking a simple scenario in the performance timeline - just pressing `j` and `k` repetively back-and-forth. This gesture has been laggy for a while, so I wanted to get a sense of where the bottlenecks are.

__Defect:__ Significant amount of time was being spent in the `Update Layer Tree` and `Composite Layer` events. It looks like some of the DOM rendering is being done in a non-optimal way, especially in terms of finding re-paint ares.

__Fix:__ Help the electron renderer out by creating layers using `will-change: transform`. This change splits the background, the editor, and the cursor into separate layers. I also noticed that in some degenerate cases, the layer for the editor could get very, very large - consuming extra memory and time in the composite step. This was due to the offscreen elements. Switching from using `style.left = -10000px` to `display: none` addresses this issue.

The next blip that shows up on the timeline is the way we pack / unpack `msgpack` protocol messages. I'll use #229 to look at using the `msgpack-lite` library which seems to be quite a bit faster for some scenarios. In addition, the window management work in #362 will open up some new opportunities to speed up rendering further.


extr0py:
Also bump up React version to 16.5 and bring in the `react-addons-perf` module for debugging extraneous re-renders.

Issue #460, opened 2017-05-22
Closing One Instances Closes all


mrWinston:
First of all, Thanks for all the effort. I really like Oni so far! I've just encountered one little Problem:

When i close one instance of Oni with ":q!", all other instances also close. This does not happen, when i terminate an instance with the window managers "x" button. Below is a gif showing the problem: 
![](https://media.giphy.com/media/3o7btYCdCzE83gLNcI/giphy.gif)

I'm running ONI 0.2.3-1235 on KDE Neon (Ubuntu 16.04, 64bit)

extr0py:
Cheers @mrWinston , thanks for the kind words! 

I've run into this issue every so often too, but forgot to log an issue for it - so thanks for opening it up and including the gif. Definitely a bug

extr0py:
Resolved by #461

PR #461, opened 2017-05-23
Fix for #460: Close only single instance

extr0py:
__Issue:__ When using `:q` to close the app, all open Oni windows also close. This is very frustrating and annoying.

__Defect:__ The app uses electron's `makeSingleInstance` functionality, which means there is only one 'main' process and multiple 'browser' processes. When using `remote.app.quit()` this would close _all_ instances.

__Fix:__ When an individual instance quits, just close the browser window instead of killing the whole app and all browser windows.


extr0py:
Fix #460

PR #466, opened 2017-05-29
Fix #465: Update QuickInfo to open underneath, when it is close to the top

### LINKED ISSUE #465: QuickInfo: Move below line if too close to top
This has always been an issue, but I'm noticing it more and more as I start adding additional language support.

image

The quick-info UI gets cut-off at the top. When the line is close to the top, we should show the quick-info below the line. Both quick-info and the signature help use some common elements, but for signature help, we don't want the same behavior, because it would conflict with the completion dialog.

extr0py:
- Update QuickInfo positioning so that it renders underneath the line when it is close to the top.


extr0py:
Fix #465

PR #468, opened 2017-05-30
Fix #376, off-by-one in autocomplete substring

### LINKED ISSUE #376: Autocompletion: Inconsistent behavior if there is no substring match
This occurred frequently in the C# completion strategy as I was testing, but it is not actually specific to that strategy, as this reproduces in TypeScript as well.

### Case 1:
`window.|` -> choose `setTimeout`

__Expected__: `window.setTimeout|`
__Actual__: `windowsetTimeout|`

### Case 2:
`private _onBufferUpdate(args: Oni.|)` -> choose `EventContext`

__Expected__: `private _onBufferUpdate(args: Oni.EventContext)`
__Actual__:  `private _onBufferUpdate(args: OniEventContext|`
(note missing parentheses)

It seems that cases where there are characters following the completions are problematic, so need to look at both those cases and evaluate a holistic fix.

keforbes:
Oni's autocomplete feature works by taking the selected item (e.g., `addEventListener`) and removing the subset of characters already typed (e.g., `add`).  Unfortunately, if no subset was provided it would delete the last typed character (#376).  This is because there was an off-by-one error where we were always starting at the character *before* the cursor.  So when given `add` we were looking for `ad`.  This works fine when there are extra characters to lose but when the substring is `""` we were stepping the cursor back one character and *then* searching for the substring, which would lose that previous character.

The fix for this defect is to step one character forward to include the character under the cursor rather than moving one character backward in the initial case.

### DIFF
--- a/browser/src/Services/AutoCompletionUtility.ts
+++ b/browser/src/Services/AutoCompletionUtility.ts
@@ -8,10 +8,10 @@ export function getCompletionStart(bufferLine: string, cursorColumn: number, com
 
     cursorColumn = Math.min(cursorColumn, bufferLine.length)
 
-    let x = cursorColumn - 1
+    let x = cursorColumn
     while (x >= 0) {
 
-        const subWord = bufferLine.substring(x, cursorColumn)
+        const subWord = bufferLine.substring(x, cursorColumn + 1)
 
         if (completion.indexOf(subWord) === -1) {
             break


extr0py:
Nice catch, thanks @keforbes ! Embarrassing mistake by me... That 'off-by-one' case was a major pain point for completion. 

Awesome, looks like this addresses both cases in #376

PR #469, opened 2017-06-01
Fix #453: Prevent text from wrapping in editor elements

### LINKED ISSUE #453: Text improperly wrapped when opening README.md
The `README.md` in the repo is showing some wrapped text:
image

extr0py:
**Issue**: In some cases, the spans for text content are sized just a bit too small, causing the text to wrap

**Defect**: It seems that in some cases, with fractional fonts, there are fractional errors in the calculation that cause the element to be sized just a bit too small. This does not reproduce for all fonts / environments.

**Fix**: Experimented with a few different ways to fix this:
- Add a buffer to the element (using `Math.ceil` for the calculation). This didn't work in all cases
- Add `whitespace: nowrap` to the elements to prevent wrapping

The latter fix seemed to work reliably, and there was not a performance regression based on my testing, so I went with that scoped fix.

### DIFF
--- a/browser/src/Renderer/ElementFactory.ts
+++ b/browser/src/Renderer/ElementFactory.ts
@@ -32,6 +32,7 @@ export class RecycleElementFactory {
         }
 
         const elem = document.createElement("span")
+        elem.style.whiteSpace = "nowrap"
         this._rootElement.appendChild(elem)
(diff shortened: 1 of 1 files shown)


extr0py:
Fix #453

PR #472, opened 2017-06-02
Fix #471, enable opening a file by drag-dropping it onto Oni

### LINKED ISSUE #471: Feature request: Open a file by dragging the file to the panel
It would be nice that ONI support opening a file by dragging the file over to the panel

keforbes:
I'll let @extr0py judge whether this is the correct file to stick the event listener in, but this code works for me.  Fixes #471.

### DIFF
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -296,13 +296,11 @@ export class NeovimEditor implements IEditor {
             ev.preventDefault()
 
             let files = ev.dataTransfer.files
-            if (files.length === 1) {
-                this._neovimInstance.open(files[0].path.split("\\").join("/"))
-            } else {
-                // apparently FileList doesn't support iterators so I have to manually iterate
-                for (let i = 0; i < files.length; i++) {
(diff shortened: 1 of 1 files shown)


keforbes:
Ok, assuming @extr0py agrees with the location of this code, I'm pretty happy with this functionality.  If a single file is drag/dropped, it replaces the current editor; if multiple files are drag/dropped, the first file replaces the current editor and the remaining files are opened in new tabs.

extr0py:
Looks great to me! Thanks for the quick fix on this, @keforbes 👍 👍 

In terms of location of the code, I think we might eventually have to shuffle it a bit once #362 is implemented, because in that case there may be multiple `<EditorHost />` components, so we'd want to put the drag/drop handler on those individually. But until that is implemented this is great, and it will be easy to move eventually.

I tried it on Windows and it works well!

mjedmonds:
Hi all, this functionality isn't working for me on macOS - any ideas why?

CrossR:
Hi @mjedmonds , it isn't working for me now on Windows... At a guess with the recent changes to the architecture to support  multiple instances of neovim and what not (ie #1682), this has broke now there is multiple instances around.

If you create an issue we can track adding it back again, and possibly some tests to prevent this happening again, though I expect tests for this would be very awkward.

Issue #122, opened 2016-12-23
QuickInfo doesn't appear until file is modified


keforbes:
There's probably a better way to describe this but I've noticed when I first open Oni, the QuickInfo doesn't appear at all.  Yet if I make some modification to the file, QuickInfo suddenly starts working.  Just entering insert mode isn't enough, I actually have to make a change.

In my example below, QuickInfo doesn't work.  I then hit `o` to open a new line, exit insert mode, and suddenly QuickInfo is working.
![quickinfo_init](https://cloud.githubusercontent.com/assets/817509/21460383/bfb7bc3e-c904-11e6-9e35-cccbc4655f98.gif)

extr0py:
Interesting... It looks like somehow the typescript plugin isn't being fully initialized until a change is being made. It also seems like the syntax highlighting from the plugin isn't working.

Do you know what version of typescript is installed? You can check in `node_modules/typescript/package.json`. It might be worth bumping that to the latest, just to rule out a typescript version issue.

You can also debug the typescript plugin by changing the `debugging: false` line (line 13) in `vim\core\oni-plugin-typescript\package.json` to `debugging: true`. With that change, when the plugin initializes, it'll pop open a debugger window. I wonder if there is a script error or something there.

Unfortunately, I'm not able to reproduce this on my build (on either Win10 or Linux) - which OS are you on?

keforbes:
This is on Linux.  I'm using typescript 2.1.4.  Setting `debugging: true` didn't help.  I've been looking at the console output and I don't see anything special.  No change in console output before and after modifying the file either.

extr0py:
I think I am seeing something similiar now - I don't get any highlighting, etc, until I move around the file. Need to investigate the plugin lifecycle and see if it was waiting for a specific event from Vim before it kicks off the integration.

keforbes:
Update for this defect, even though no one else seems to be seeing it...

In `PluginManager.ts`, line 172 says:
```
if (!pluginResponse.error) {
```
If there is no error, display QuickInfo.  If there is an error, hide the QuickInfo (no-op).  It seems prior to editing the file, `pluginResponse.error` exists.  Unfortunately, it's an empty object `{}` so it doesn't really help me with troubleshooting.  Also, `pluginResponse` doesn't have a `payload` object so there'd be nothing to display even if I wanted to.  After editing the file, `pluginResponse` does not have an `error` object at all and it has a `payload` object with the contents of the QuickInfo.

[...]

PR #476, opened 2017-06-05
Fix #372, use readFile() to load plugins rather than readFileSync()

### LINKED ISSUE #372: Performance: Make plugin initialization path async
In `Plugin.ts`, the plugin initialization path is synchronous:

```
        if (fs.existsSync(packageJsonPath)) {
            this._oniPluginMetadata = PackageMetadataParser.parseFromString(fs.readFileSync(packageJsonPath, "utf8"))
```

Making this async (by factoring out to an initialize method that returns a promise, instead of doing two synchronous File I/O calls) will help improve startup performance and responsiveness.

keforbes:
In #372, you recommended using a Promise rather than two synchronous calls to `fs` (`existsSync` and `readFileSync`).  However, the docs for [File System](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback) say:
> Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

So instead of refactoring to use a Promise I removed the call to `exists` and just call `readFile` (not `readFileSync`) to make it asynchronous.  I didn't profile it to test whether this solution is faster and I also don't know if there are any benefits to a Promise which an asynchronous call does not provide.  So let me know if this solution isn't acceptable.  I did test the change though and it seems to work as far as I can tell.

### DIFF
--- a/browser/src/Plugins/Plugin.ts
+++ b/browser/src/Plugins/Plugin.ts
@@ -30,21 +30,20 @@ export class Plugin {
         const packageJsonPath = path.join(this._pluginRootDirectory, "package.json")
         this._channel = channel
 
-        if (fs.existsSync(packageJsonPath)) {
-            this._oniPluginMetadata = PackageMetadataParser.parseFromString(fs.readFileSync(packageJsonPath, "utf8"))
+        fs.readFile(packageJsonPath, "utf8", (err, data) => {
+            if ( ! err) {
+                this._oniPluginMetadata = PackageMetadataParser.parseFromString(data)
 
-            if (!this._oniPluginMetadata) {
-                Log.error(`[PLUGIN] Aborting plugin load, invalid package.json: ${packageJsonPath}`)
-            } else {
-                if (this._oniPluginMetadata.main) {
-
+                if (!this._oniPluginMetadata) {
+                    Log.error(`[PLUGIN] Aborting plugin load, invalid package.json: ${packageJsonPath}`)
+                } else if (this._oniPluginMetadata.main) {
                     this._oni = new Oni(this._channel.createPluginChannel(this._oniPluginMetadata,
                         () => this._onActivate()))
 
                     this._commands = PackageMetadataParser.getAllCommandsFromMetadata(this._oniPluginMetadata)
                 }
             }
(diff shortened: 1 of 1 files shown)


keforbes:
The appveyor build failed because it couldn't download a dependency:
> Error: GET https://github.com/electron/electron/releases/download/v1.6.0/chromedriver-v2.21-win32-ia32.zip returned 500

If you poke the "re-build" button it should work but I don't have permission to click that button myself. :stuck_out_tongue_winking_eye:

extr0py:
Poked it and it passed now 👍 👍  Those intermittent download failures are annoying, though! 

I took a look and I'm a bit nervous about this change - just because I believe that there may be points downstream that could fail. For example, places that assume the plugin loading is synchronous. There is an example of that here in `PluginManager.ts`:

```
    private _createPlugin(pluginRootDirectory: string): Plugin {
        const plugin = new Plugin(pluginRootDirectory, this._channel)

        if (plugin.commands) {
            plugin.commands.forEach((commandInfo) => {
                this._commandManager.registerCommand(new CallbackCommand(commandInfo.command, commandInfo.name, commandInfo.details, (args?: any) => {
                    this._sendCommand(commandInfo.command, args)
                }))
            })
        }

        return plugin
    }
```

Basically, this assumes that the plugin metadata gets parsed synchronously to get the commands from it, and then registers those. So now that the plugin initialization is asynchronous, we'd need to wait to look at the commands until the plugin is fully "activated" - I think with your change this might not work as is expected (there's now a race condition). I think there could potentially be other repercussions here too, in terms of timing of events from Vim to having the plugin process it.

The timing is a bit sensitive already - I'm looking at cases where plugins get activated but don't get the buffer-enter event (maybe related to #122?). I think this might require a more holistic look across the plugin-loading pipeline, to make sure the new async behavior doesn't have any ripple effects

keforbes:
Closing pull request due to Fear, Uncertainty, and Doubt.  😄

PR #477, opened 2017-06-05
Fix #369, replace findParentDir.sync with Promise

### LINKED ISSUE #369: Performance: TSLint - Replace `findParentDir.sync` in `getLintConfig` with async call / promise
While looking at some performance issues, I saw that `findParentDir` is called synchronously in tslint - need to refactor that to be asynchronous and return a promise instead.

The file is `vim\core\oni-plugin-tslint\lib\index.js`. This is called whenever we decide to do linting, which right now is on buffer enter and buffer save.

keforbes:
`find-parent-dir` has a dependency on a deprecated call (https://github.com/thlorenz/find-parent-dir/issues/4) so I decided to replace it with [find-up](https://github.com/sindresorhus/find-up), which does basically the same thing but has support for Promises.  Also, `find-up` returns the full path of the filename rather than the parent dir.  This reduced some code since every usage of `find-parent-dir` was `join`ing the directory name and filename.

Now, with that said, I don't know anything about Promises.  I tried to emulate the Promise code found elsewhere in Oni but it's very likely I made a rookie mistake.  Please take a look and let me know if you have any suggestions.

### DIFF
--- a/browser/src/ProjectConfig.ts
+++ b/browser/src/ProjectConfig.ts
@@ -39,27 +39,27 @@ const DefaultConfiguration: IProjectConfiguration = {
  * Search upward for the relevant .oni folder
  */
 export function getProjectConfiguration(filePath: string): Q.Promise<IProjectConfiguration> {
-    const oniDir = findUp.sync(".oni", { cwd: filePath })
-
-    if (!oniDir) {
-        return Q(DefaultConfiguration)
-    }
-
-    return loadConfigurationFromFolder(oniDir)
+    return findUp(".oni", { cwd: filePath })
+    .then((oniDir: string) => {
+        if (!oniDir) {
+            return DefaultConfiguration
+        }
+        return loadConfigurationFromFolder(oniDir)
+    })
 }
 
-function loadConfigurationFromFolder(folder: string): Q.Promise<IProjectConfiguration> {
+function loadConfigurationFromFolder(folder: string): IProjectConfiguration {
 
     const launchPath = path.join(folder, "launch.json")
 
     if (!fs.existsSync(launchPath)) {
-        return Q(DefaultConfiguration)
+        return DefaultConfiguration
     } else {
        const launchInfo: ILaunchConfiguration = JSON.parse(fs.readFileSync(launchPath, "utf8"))
        const config = {...DefaultConfiguration, ...{
            launchConfigurations: [launchInfo],
        }}
 
-       return Q(config)
+       return config
     }
 }
--- a/browser/src/ProjectConfig.ts
+++ b/browser/src/ProjectConfig.ts
@@ -9,7 +9,7 @@ import * as path from "path"
 
 import * as Q from "q"
 
-const findParentDir = require("find-parent-dir") // tslint:disable-line no-var-requires
+const findUp = require("find-up") // tslint:disable-line no-var-requires
 
 export type LaunchType = "execute"
 export type Request = "launch" // attach later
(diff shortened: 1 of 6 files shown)


extr0py:
Cool, it does look like that library is a bit more convenient in terms of syntax for Oni.

> This reduced some code since every usage of find-parent-dir was joining the directory name and filename.

Excellent!

> Now, with that said, I don't know anything about Promises. I tried to emulate the Promise code found elsewhere in Oni but it's very likely I made a rookie mistake. 

Yes, Promises are a little tough to understand at first...  These are a couple decent references:
https://developers.google.com/web/fundamentals/getting-started/primers/promises
https://davidwalsh.name/promises
http://www.mattgreer.org/articles/promises-in-wicked-detail/

The cool thing is that they can be chained, which avoids having tons of nested callbacks. It seems like handling asychronous actions in JavaScript started with callbacks, then moved towards Promises, and is now moving towards async/await: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

browser/src/ProjectConfig.ts
```diff
@@ -39,13 +39,13 @@ const DefaultConfiguration: IProjectConfiguration = {
  * Search upward for the relevant .oni folder
  */
 export function getProjectConfiguration(filePath: string): Q.Promise<IProjectConfiguration> {
-    const oniDir = findParentDir.sync(filePath, ".oni")
+    const oniDir = findUp.sync(".oni", { cwd: filePath })
```
extr0py: Seems like we could actually make this `async` now (which is the non-blocking code-path). Something like:
```
return findUp(".oni", {  cwd: filePath })
.then((oniDir) => {
      if (!oniDir) {
          return DefaultConfiguration
      }
      return loadConfigurationFromFolder(oniDir)
})
```

extr0py:
Added a couple comments about the Promises - we can actually simplify and remove the `Q.defer` and just return the values directly. Let me know if you have any questions!

keforbes:
Ok, I made that change to `ProjectConfig.ts`, let me know if there are any other changes needed.

extr0py:
Thanks for making all the changes, @keforbes ! I just had one other request to see if we can use the async version of `findUp` in tslint - since the promises are all set there, it should be an easy switch, and helpful for performance.

keforbes:
> see if we can use the async version of findUp in tslint

I might not be understanding this.  Are you suggesting I do:
```
 .then(() => {
    findUp("tsconfig.json", { cwd: currentWorkingDirectory })
 })
.then((project) => {
```

Given this call of `findUp` returns another promise, I think I'd have to nest the remaining promises like this:
```
 .then(() => {
   findUp("tsconfig.json", { cwd: currentWorkingDirectory })
    .then((project) => {
      //all other code
    }
 })
```

Let me know if I'm missing something.

extr0py:
Close - you can actually chain promises. Your code would work, it's just missing the `return` statement, so if it looks like this:

```
 .then(() => {
    return findUp("tsconfig.json", { cwd: currentWorkingDirectory })
 })
.then((project) => {
```

This is nice because you can eliminate that nesting (which happens a lot with callbacks). 

Then, there is an even shorter version, because if you use a lambda and omit the `{` and `}`, the `return` is implied. So this form actually means the same as the one above... just a bit more concise:
``` 
.then(() => findUp("tsconfig.json", { cwd: currentWorkingDirectory }))
.then((project) => {
```

Thanks!

keforbes:
Cool, thanks for explaining it @extr0py!

extr0py:
Sure thing! Thanks for taking this on, @keforbes , and addressing all th e feedback. I'll bring this in - it helps move us to leverage Promises (and hopefully, in the longer run, `async` and `await`).

PR #480, opened 2017-06-07
Fix #446 - Make configuration values available to plugins

### LINKED ISSUE #446: Plugins: Expose configuration values to plugins
For the language services, and plugins in general, it's important to expose the configuration values.

There should be an `configuration` member on the `Oni` object exposed to plugins, with the following:

`onChanged`: event, fired whenever there is a new configuration
`getValue(settingName: string)`: gets the value of a particular configuration setting

This is especially important for the language service providers, to set paths for language services.

extr0py:
- Add `Oni.configuration.getValue` API available to plugins
- Simplify plugin infrastructure by removing over-complicated sandboxing


extr0py:
TODO: Add issue for supporting configuration values for pyls/go-langserver paths.

PR #488, opened 2017-06-10
Fix #389 - Integrate with ocaml-language-server

### LINKED ISSUE #389: Integrate OCaml/ReasonML language server
Language server here: https://github.com/freebroccolo/ocaml-language-server

Wanted an excuse to learn more about these languages 👍 👍

extr0py:
Initial support for OCaml / ReasonML, based on https://github.com/freebroccolo/ocaml-language-server (WIP)]

__Remaining items:__
- [x] Syntax Highlighting
- [x] Full document sync on ocaml-language-server


freebroccolo:
Glad to see someone trying this. Please let me know if you run into any problems.

jordwalke:
FYI: You can set the environment variable `export MERLIN_LOG=~/Desktop/log` which will have merlin log its internal steps/logic to that file (assuming your Oni picks up the environment vars).

extr0py:
Nice! Thanks for that tip, @jordwalke ! I'm seeing an issue where the `Js` namespace isn't picked up (in both VSCode and Oni), so I'm wondering if it is a merlin configuration issue. Will see if that logging can help me get more info.

jordwalke:
Yeah, if you're building with bucklescript, the project must first be built in order for the Js namespace to be picked up. Does that help?

extr0py:
Yep, building using BuckleScript - I followed the getting started tutorial here:  https://facebook.github.io/reason/gettingStarted.html

So I have a project with a `bsconfig.json` and using `bsb -w` to compile. I replaced the `print_endline` with this: `let promise = Js.Promise.resolve(1);`, and it compiles OK, giving me:
```
var promise = Promise.resolve("a");

exports.promise = promise;
```

But in VSCode, I see this:
screen shot 2017-06-21 at 3 32 13 pm

With the message `Unbound module 'Js'`.

I'll try out the merlin logging - maybe that will give a clue. I wonder if I need to have a `.merlin` file for it to pick up the BuckleScript modules?

extr0py:
Think I need to a `.merlin` file - maybe https://github.com/BuckleScript/bucklescript/issues/939 is relevant. Adding `"generate-merlin": true` to my `bsconfig.json` doesn't seem to be quite enough.

extr0py:
Actually, the `.merlin` file is there now - will try and get the logs. I wonder if I need to call out the JS library. It looks like it is only including the `ocaml` library but not any of the `js` or `es6` ones. Probably something missing in my `bsconfig.json`.

extr0py:
Working now - I restarted the language service and it is picked up (in both Oni and VSCode). Looks like the `"generate-merlin": true` was the only change that was needed in my `bsconfig.json`. I wonder if it is worth calling out in the "Getting Started" section of the Reason documentation?

extr0py:
BTW, @freebroccolo , nice work on the language server! It's great. All the problems so far have been on the Oni side 😉

PLAN.md
```diff
@@ -1,8 +1,8 @@
-- Syntax highlighting
-    - Fix regression / change in highlighting
-    - Support document symbol
-    - Only update symbol if the LS supports it
+- Switching files between workspace
+    - reason-of-life - good example
 
+- Semicolon issue? Why is there always a semicolon error, even though examples compile fine without it?
```
jordwalke: This sounds like something isn't picking up `refmt` in your PATH. Could it be the case? How did you install merlin/reason/ocaml?

extr0py: Thanks for looking at this, @jordwalke ! I've been going through the docs and I believe this was actually a misunderstanding on my part. I've been spending time jumping back and forth between OCaml/Reason syntax (learning OCaml primarily, and then playing with Reason) - I think this was a case where I didn't fully context switch between the two. I was caught up in some examples from BuckleScript w/o semicolons and was porting them over to Reason and missed some semicolons. Sorry for the noise, I don't believe there's a real issue here.

extr0py:
Fixes #379 as well with the latest changes.

extr0py:
Logged this issue to track the Python issue with document symbols: https://github.com/palantir/python-language-server/issues/73

For now, I'll just disable `textDocument/documentSymbol` for python.

PR #489, opened 2017-06-12
Fix #407, use editor.quickOpen.execCommand to filter files in quickOpen

### LINKED ISSUE #407: How to integrate fzf to oni Fuzzy Finder?
I try to integrate [fzf](https://github.com/junegunn/fzf) to oni Fuzzy Finder by set : 
```json
editor.quickOpen.execCommand = "fzf"
```
But there is no change, oni still use git ls-files

keforbes:
This pull requests changes the format of `editor.quickOpen.execCommand` to require a `${search}` variable to be defined.  That variable is used when filtering the files in Quick-Open.  The code seems right to me but my use of `execSync` always throws an error and I don't know why.  It's acting like `fzf` is exiting with a code of `1` but I don't see how.

So this code isn't functioning yet, I just wanted to create the pull request so I had a place to complain because I don't understand why it's failing.

### DIFF
--- a/browser/src/Services/QuickOpen.ts
+++ b/browser/src/Services/QuickOpen.ts
@@ -52,7 +52,7 @@ export class QuickOpen {
         if (overriddenCommand) {
             try {
                 // replace placeholder ${search} with "" for initial case
-                const files = execSync(overriddenCommand.replace("${search}", '""'), { cwd: process.cwd() })
+                const files = execSync(overriddenCommand.replace("${search}", ""), { cwd: process.cwd() })
                     .toString("utf8")
                     .split("\n")
                 this._showMenuFromFiles(files)
--- a/browser/src/Services/QuickOpen.ts
+++ b/browser/src/Services/QuickOpen.ts
@@ -6,6 +6,7 @@
 
 import { execSync } from "child_process"
 import * as path from "path"
+import * as Log from "./../Log"
 
 import * as glob from "glob"
 import * as _ from "lodash"
@@ -37,7 +38,7 @@ export class QuickOpen {
 
     public show(): void {
         const config = Config.instance()
-        const overrriddenCommand = config.getValue("editor.quickOpen.execCommand")
+        const overriddenCommand = config.getValue("editor.quickOpen.execCommand")
         const exclude = config.getValue("oni.exclude")
 
         UI.Actions.showPopupMenu("quickOpen", [{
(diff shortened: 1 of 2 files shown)


keforbes:
Looks like the logic is sound.  I can set `execCommand` to `ls` or `find` and it works correctly.  It must be something specific to `fzf` that I just can't get to work.

extr0py:
Ya, seems like the `execSync` will always throw if the exit-code is non-zero, according to the docs here: https://nodejs.org/api/child_process.html

And looks like the exit code for fzf is non-zero even in success cases.

A couple options:
- We could wrap the execSync call in a try/catch, and ignore the error code.
- We could use an alternative to execSync (like `spawn` - sort of like how `tslint` spawns the command and gets a promise of the output). This means we'd have to restructure the code a bit, because the reducers aren't designed to work with an asychronous flow - we might want to move the asynchronous processing to QuickOpen, for example. This would be helpful for performance though, since `execSync` blocks while the shell command is executed, but using the asynchronous versions would be much faster.

One other issue I saw is that in the initial case - with `fzf -f` - there is an error of "query string required". We might need to make the command something like `fzf -f "{searchString}"`.

That's great that it is working for non-fzf scenarios, though!

keforbes:
> One other issue I saw is that in the initial case - with fzf -f - there is an error of "query string required". 

Yeah, you can see the second commit in the pull request where I removed that behavior.  I was originally sending `""` as the query in the initial case so `fzf` had an argument but then I tested with `ls -a ${search}` and `find . -name \*${search}\*` and for those scenarios, `""` would fail.  You can always set the `execCommand` to `fzf -f "${search}"` in the config rather than putting those quotes in the code.

keforbes:
> We could wrap the execSync call in a try/catch, and ignore the error code.

I've committed a change where we'll fallback to our normal filtering mechanism if the `execSync` call fails.  Unfortunately, when the error fires it doesn't provide any stdout so I can't just ignore the error and use the output.

> We could use an alternative to execSync. This means we'd have to restructure the code a bit, because the reducers aren't designed to work with an asychronous flow - we might want to move the asynchronous processing to QuickOpen

That's above my pay grade.  I don't know how to move the code outside of the action/reducer code since no one is calling that method directly.  I can't just update references.

I'm tempted to merge in this code since it does what it sets out to do.  I think the errors with `fzf` are specific to `fzf` and we can always make the code asynchronous later.  I think it's a step in the right direction at least.  Although, if you can explain how you envision moving this code into `QuickOpen` I can certainly give it a try.

extr0py:
> I'm tempted to merge in this code since it does what it sets out to do. I think the errors with fzf are specific to fzf and we can always make the code asynchronous later. I think it's a step in the right direction at least. Although, if you can explain how you envision moving this code into QuickOpen I can certainly give it a try.

Sounds good. This is an improvement over the existing implementation since it seems like the setting wasn't working at all previously, so it's great that it's working now for `ls` and `find`.

And apologies for the slow response time on this! Thanks for making the change & the investigation. I'll bring this in now.

PR #493, opened 2017-06-21
Fix #492: Add the `editor.scrollBar.visible` setting to control visibility of buffer scrollbar

### LINKED ISSUE #492: Hide scrollbar option
Is there an option in the config to hide the scrollbar? If not, is it possible to add it?

extr0py:
__Issue:__ There is no way to disable the buffer scroll bar, which some users may find distracting or unnecessary.

__Fix:__ Add a setting for `editor.scrollBar.visible` that can be modified in the `config.js`:

scrollbar setting


extr0py:
Fixes #492

PR #498, opened 2017-06-23
Fix #495: Clean up list jumpiness in autocompletion menu

### LINKED ISSUE #495: Jumpy Lists:
I had some feedback about the autocomplete documentation experience:

onicomplete

One thing I noticed, is that it has this particular characteristic that might be suboptimal. That is, when changing the highlighted row in the completion popup, the physical location of *other* items (below it) change. I believe that UIs that don't move elements in lists unless due to scrolling are easier to navigate (unlike accordion style menus), because the user's mental model of where everything is on screen needs to be updated less frequently, or can be updated more quickly.

Scrolling in general doesn't have the same downsides, because all items move in unison - so it only requires that the user understand "every item is where it used to be but moved 10px upwards", which is much more compressed than having to know which items moved upwards, and which moved downwards.

extr0py:
__Issue:__ The autocompletion list is jumpy, so it is a jarring experience for users when the list item changes.

__Fix:__ Move the details to the bottom of the completion menu. I also address another jumping issue - sometimes the autocompletion menu would expand/collapse, which was also a jarring experience. In addition, I made the highlighting a bit more subtle, and added an overflow experience for the details:

completion-v2

### DIFF
--- a/browser/src/Input/Mouse.ts
+++ b/browser/src/Input/Mouse.ts
@@ -15,9 +15,7 @@ export class Mouse extends EventEmitter {
         this._editorElement = editorElement
         this._screen = screen
 
-
         // MUSTFIX: Temporary hack to remove mouse handling to prototype completion handler.
-
         document.body.addEventListener("mousedown1", (evt: MouseEvent) => {
             const { line, column } = this._convertEventToPosition(evt)
 
--- a/browser/src/Input/Mouse.ts
+++ b/browser/src/Input/Mouse.ts
@@ -15,14 +15,15 @@ export class Mouse extends EventEmitter {
         this._editorElement = editorElement
         this._screen = screen
 
-        document.body.addEventListener("mousedown", (evt: MouseEvent) => {
+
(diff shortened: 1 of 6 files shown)


extr0py:
I'll need to fix some issues with the Mouse handling (the `Mouse.ts` strategy is catching all mouse events right now). In addition, I'll need to find a way to expand the completion details, so that users can open the details fully without needing the mouse.

extr0py:
Also need to look at integrating the scrollbar with the quick-info UI.

jordwalke:
Looks pro. I imagine one day being able to map keyboard shortcuts just to scroll the doc browser (which could apply to other scrollable dialogs too).

extr0py:
> I imagine one day being able to map keyboard shortcuts just to scroll the doc browser (which could apply to other scrollable dialogs too).

Definitely! This will be a good case to think about along with #20 and #408.

extr0py:
[ ] Need to fix the mouse interaction and remove the temporary hack

Issue #485, opened 2017-06-10
Status Bar: Add language server info to status bar


extr0py:
Splitting this out from #201 .

One challenge with the language server is that it's opaque to the user - often times there are interesting things happening behind the scenes (like downloading a package, or an error, etc), and these don't get bubbled up. This is especially problematic if it fails - there is no indication to the user, things just don't work.

Having the current language server status in the bar would be really helpful:
![image](https://user-images.githubusercontent.com/13532591/27004754-fa10549c-4dc3-11e7-9f93-0ed799ac97a6.png)

The icon could change in the following scenarios:
- __Spinner__ - the spinner would show while the language server is loading.
- __Lightning__ - the icon would show a lightning bolt when it is connected. It might be nice to have a small animation when a language server request is processed.
- __Error__ - an icon shown if the language server encounters an error. 

It would also be helpful to leverage tooltip text so that users can drill into the current status or error status of the language server.

extr0py:
This was added by #524

PR #507, opened 2017-06-30
[WIP] Git StatusBar Item

bert88sta:
This doesn't work, probably for several reasons. The component itself shows up fine, but a call to 
Oni.git.getBranch() doesn't work.

### DIFF
--- a/browser/src/Plugins/Api/Git.ts
+++ b/browser/src/Plugins/Api/Git.ts
@@ -1,0 +1,14 @@
+import { execSync } from "child_process"
+import * as Q from "q"
+
+/**
+ * API instance for interacting with Oni (and vim)
+ */
+export class Git implements Oni.Git {
+    public getBranch(): Q.Promise<String> {
+        let cmd = "git rev-parse --abrev-ref HEAD"
+        const branchName = execSync(cmd).toString("utf8")
+        return Q.resolve(branchName)
+    }
+
+}
--- a/browser/src/Plugins/Api/Git.ts
+++ b/browser/src/Plugins/Api/Git.ts
@@ -6,7 +6,7 @@ import * as Q from "q"
  */
 export class Git implements Oni.Git {
     public getBranch(): Q.Promise<String> {
-        let cmd = "git rev-parse --abrev-ref HEAD"
+        let cmd = 'git rev-parse --abbrev-ref HEAD'
         const branchName = execSync(cmd).toString("utf8")
         return Q.resolve(branchName)
     }
--- a/browser/src/Plugins/Api/Oni.ts
+++ b/browser/src/Plugins/Api/Oni.ts
@@ -2,11 +2,11 @@ import * as ChildProcess from "child_process"
 import { EventEmitter } from "events"
 
 import { IPluginChannel } from "./Channel"
-
 import { Commands } from "./Commands"
 import { Configuration } from "./Configuration"
 import { Diagnostics } from "./Diagnostics"
 import { Editor } from "./Editor"
+import { Git } from "./Git"
 import { StatusBar } from "./StatusBar"
 
 import { DebouncedLanguageService } from "./DebouncedLanguageService"
@@ -25,6 +25,7 @@ export class Dependencies {
  */
 export class Oni extends EventEmitter implements Oni.Plugin.Api {
 
+    private _git: Oni.Git
     private _configuration: Oni.Configuration
     private _dependencies: Dependencies
     private _editor: Oni.Editor
@@ -56,6 +57,9 @@ export class Oni extends EventEmitter implements Oni.Plugin.Api {
     public get statusBar(): StatusBar {
         return this._statusBar
     }
+    public get git(): Oni.Git {
+        return this._git
+    }
 
     constructor(private _channel: IPluginChannel) {
         super()
@@ -66,6 +70,7 @@ export class Oni extends EventEmitter implements Oni.Plugin.Api {
         this._editor = new Editor(this._channel)
         this._commands = new Commands()
         this._statusBar = new StatusBar(this._channel)
+        this._git = new Git()
 
         this._channel.onRequest((arg: any) => {
             this._handleNotification(arg)
--- a/definitions/Oni.d.ts
+++ b/definitions/Oni.d.ts
@@ -3,6 +3,9 @@ import { EventEmitter } from "events"
 import * as types from "vscode-languageserver-types"
 
 declare namespace Oni {
+    export interface Git {
+        getBranch(): Q.Promise<String>
+    }
 
     export interface EventCallback<T> {
         (val: T): void
--- a/vim/core/oni-core-statusbar/index.js
+++ b/vim/core/oni-core-statusbar/index.js
@@ -6,6 +6,7 @@ const rgb = (r, g, b) => `rgb(${r}, ${g}, ${b})`
 const activate = (Oni) => {
     const React = Oni.dependencies.React
 
+    const gitBranchItem = Oni.statusBar.createItem(0, -2)
     const filePathItem = Oni.statusBar.createItem(0, -1)
     const lineNumberItem = Oni.statusBar.createItem(1, -1)
     const modeItem = Oni.statusBar.createItem(1, -2)
@@ -65,6 +66,11 @@ const activate = (Oni) => {
         filePathItem.setContents(element)
     }
 
+    const setBranch = (branchName) => {
+        branchElement = React.createElement("div", null , branchName)
+        gitBranchItem.setContents(branchElement)
+    }
+
     Oni.on("mode-change", (evt) => {
         setMode(evt)
     })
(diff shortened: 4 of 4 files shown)


bert88sta:
I suspect that I've done something incorrect with the promise in the getBranch() function

extr0py:
This looks like a great start, @bert88sta ! Nice job implementing the API. I just checked out your branch and I can see the `Oni.git` object that is exposed to the plugin now.

One convenient thing about the plugin API is you can actually test it in the console. Here's a gif of how I tested your API:

I figured I'd detail it out just in case it helps you - these are the steps I took:
debugging new api

1. Opened the developer console (Control+Shift+P -> Open DevTools)
2. The Oni plugin object is available in the console, which is really convenient - so I started by running `Oni.git` just to make sure that object was set up correctly.
3. I then made sure `getBranch()` was actuallyi returning a promise
4. I then ran `Oni.git.getBranch().then((branchName) => console.log(branchName))`, which logs out the result of the promise. As you can see, the branch name isn't quite what I expected (my branch is `pull-507`).

I don't have a lot of "debugging guides" right now but there are lots of tips and tricks that can make life easier, I need to collect them somewhere!

Looking at the code of getBranch() though, there is one problem with it - we're making a _synchronous_ call, which will block the UI and any keyboard interactions. This kills interactive performance, and it can be nefarious if you have a bunch of these synchronous actions happening sort of intermittently (like we might need to detect if a branch changes).

I was researching this a bit, and another way we could get the branch name - that might be easier to make async - is to read from the `./git/HEAD` file in the repo (using the `fs.readFile` API - https://nodejs.org/api/fs.html) . This is actually what vim-fugitive does to get the branch (well, read the file at least). The other nice thing about this, is that we can watch the file for changes using `fs.watch`: https://nodejs.org/docs/latest/api/fs.html , and figure out when the branch has changed.

Just some thoughts. I hope that helps. Great work so far! Can't wait to have my branch show up in the statusbar.

bert88sta:
Whoa thanks for the help!

The abbrev-ref is weird, the same command run in the terminal gives me the branch name. We'll probably move over to the first line of HEAD

bert88sta:
AH! Notice I spelled "abbrev" with one "b". It works now!

My only concern with filesystem watching is if people migrate to a different project we're going to have to use a similar git command anyway to get the root git folder. There is an async version of exec from child_process, but I keep getting errors when I try to use it... Not completely sure about that.

I'm also wondering (weloming suggestions) for what colors we should use, and how we should do status. I personally love the agnoster zsh theme for git things, but we can work on this. Hopefully I'll have some work done later today

extr0py:
> AH! Notice I spelled "abbrev" with one "b". It works now!

Nice find! 👍 

> My only concern with filesystem watching is if people migrate to a different project we're going to have to use a similar git command anyway to get the root git folder. There is an async version of exec from child_process, but I keep getting errors when I try to use it... Not completely sure about that.

I think we could also look upward, recursively, to see if there is a `.git` folder - we wouldn't actually need to run a command. All git repos have a `.git` folder - so if we don't eventually find one, we're not in a git repo. In any case - whether use the file system or spawn a process - we'd want to use the async version of the APIs, so that they don't block the UI.

> I'm also wondering (weloming suggestions) for what colors we should use, and how we should do status. I personally love the agnoster zsh theme for git things, but we can work on this. Hopefully I'll have some work done later today

Ya, this is a tough one - here's a quick mock-up I did:

image
(with the git branch slightly darker) - that's using the [fa-code-fork](http://fontawesome.io/icon/code-fork/) icon, a background-color of `rgb(20, 20, 20)`, and a foreground color of `white`. I think as long as it looks reasonable with the default theme it'll be cool.  When #33 comes in we can do some more interesting things 😄 

Hope that helps!

bert88sta:
I meant colors (or potentially an asterisk) for dirty and staged changes,
or anything else git related that we want oni to track. Also, the upward
search is alright, we just have to know when we're in root.

extr0py:
Oh I see - sure, modelling that after the shell integration seems reasonable. We could start with just the branch and then add some things like files added/modified, files changed, etc. Modelling the coloring after some existing shell integration sounds good.

bert88sta:
@extr0py : I can't get the exec() to work async because I'm not familiar with how callbacks (?) should work. Can you point me in what might be the right direction so I can learn?

I'm probably going to check out the file I/O option next.

extr0py:
For sure - just as an FYI, there is wrapper I've been using in `src\Plugins\Api\Oni.ts` for plugins called `execNodeScript` (this is specific to running node modules / javascript files though, but it uses `ChildProcess.exec` as well).

The tslint plugin, in `vim\core\oni-plugin-tslint\lib\index.js`, actually wraps the call in a promise:
```
        return Q.nfcall(Oni.execNodeScript, tslintPath, processArgs, { cwd: workingDirectory })
            .then((stdout, stderr) => {

                const errorOutput = stdout.join(os.EOL).trim()

                const lintErrors = JSON.parse(errorOutput)
```

The `Q.nfcall` is from the [Q](https://github.com/kriskowal/q) promise library, and converts a function that uses a callback to one that uses a promise. The (stdout, stderr) => { ... } is called when the promise is completed. This is convenient for other parts of the code because the promise gets chained.

You could also use a callback directly, like:
```
Oni.execNodeScript(tslintPath, processArgs, { cwd: working Directory }, (stdout, stderr) => {
// stuff that happens eventually, once the process completes - stdout would contain the output of the process
})
```

(Oh, and `Oni.execNodeScript` is just a small wrapper around `exec`).

Hope that helps some. The async stuff can get complicated fast, let me know if you have any more questions.

bert88sta:
That should help a ton, the issue I was having was that I couldn't return a promise as per the interface definition

Galbar:
Hi! a week ago or so I started working on this as a way to get to know the code base. This is what I got:
```js
const childProcess = require("child_process");

const activate = (Oni) => {

    const gitBranchIndicator = Oni.statusBar.createItem(1, -3);

    const getBaseDir = (filePath) => {
        return filePath.substr(0, filePath.lastIndexOf('/') + 1)
    }

    const displayBranch = (error, stdout, stderr) => {
        if (error && error.code) {
            // Branch not available
            gitBranchIndicator.hide();
        }
        else {
            const React = Oni.dependencies.React;
            const gitBranch  = React.createElement("div", null, stdout);
            gitBranchIndicator.setContents(gitBranch);
            gitBranchIndicator.show();
        }
    };

    const updateGitBranch = (filePath) => {
        childProcess.exec("cd " + getBaseDir(filePath) + " && git rev-parse --abbrev-ref HEAD", displayBranch);
    }

    Oni.on("buffer-enter", (evt) => {
        updateGitBranch(evt.bufferFullPath)
    })

};

module.exports = {
    activate,
};
```
It does branch per buffer, instead of per folder where oni was executed and it allows the branch to change on the fly.

The issue I found is that `gitBranchIndicator.hide()` does not hide the item.

PS: Note that `getBaseDir()` approach is naive and probably is not cross-platform.

extr0py:
Thanks for sharing, @Galbar ! That looks really solid and clean, I like the approach you took.

> The issue I found is that gitBranchIndicator.hide() does not hide the item.

I just noticed this too while implementing #485, sorry about that. I have a fix in progress with PR #524 

> PS: Note that cd and getBaseDir() are probably not cross-platform.

Which platform did you test on? I can check see if it works on others. I'm glad to see you are using the async version of `exec` 👍

Galbar:
Glad you liked my approach, @extr0py!
I tested it on Linux.

Galbar:
Looking at this PR I see that the integration of git inside the Oni object is done in the new file ` browser/src/Plugins/Api/Git.ts`. Looking throughout the project I found the file ` browser/src/Plugins/Api/Git.ts` which also contains git functionality. To avoid duplication I think they should be bundled together for both internal and external Oni usage. What do you think, @bert88sta and @extr0py ?

extr0py:
> I tested it on Linux.

Cool, you're right, I think this may not work in Windows - but there should be an equivalent way. Instead of using `cd <dir> && ` in the exec command, you can actually pass the `cwd` parameter to exec to specify the current working directory - which should accomplish the same thing: https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

> Looking at this PR I see that the integration of git inside the Oni object is done in the new file browser/src/Plugins/Api/Git.ts. Looking throughout the project I found the file browser/src/Plugins/Api/Git.ts which also contains git functionality. To avoid duplication I think they should be bundled together for both internal and external Oni usage.

Definitely, it would be great to move this in a common place. Long-term, I want to take a more modular approach with Oni - splitting it up into smaller, reusable chunks. Moving Git to a common place that is usable both internally and externally is in line with that.  Thanks for thinking about this!

One possibility for this reuse is putting the git functionality on the `Oni` plugin interface, and making the `Oni` instance available internally too - it's basically 90% there anyway. There's a 'global' instance of `Oni` defined in `browser\src\Plugins\AnonymousPlugin` - we could just have a cleaner way of accessing it and reuse it globally. That way, anything defined on that API is reusable both internally and externally for plugins.

bert88sta:
I'm going to close this out and let @Galbar submit his.

Galbar:
I opened a merge request with the git status bar indicator #540

PR #517, opened 2017-07-03
Fix #511: Only execute completion requests if LS supports it

### LINKED ISSUE #511: LSP: Check if completion is enabled before submitting request
- For some LSPs (like the golang LSP), completion is not supported. We should be respecting the initialize argument and not sending completion requests.

extr0py:
__Issue:__ We always ask for completions.. even if the language server doesn't support it. This causes unnecessary errors.

__Fix:__ Use the completion provider capability to check before executing the request.


extr0py:
Fixes #511

PR #518, opened 2017-07-03
WIP: #484 - Fix OSX behavior after quit

### LINKED ISSUE #484: can't open a new window after quit (macos)
:q will close a buffer or window as expected, but if the window is closed and the app still running, I would expect to be able to open a new window with the 'New' or 'Open' command in the menu.  Instead I get this error:

screen shot 2017-06-08 at 10 52 53 am

I'd also expect to be able to open a second window with a shortcut (cmd+n)/file menu option similar to MacVim instead of only being able to do so through the command line `oni`, though that might be a feature request more than a bug.

extr0py:
Fix #484 

__Issue:__ In OSX, it is common for the application to be left running even though there is no window.

__Fix:__ Two parts:
- `createWindow` isn't being called correctly on the activate event, which is executed when the user clicks on the icon in the dock after closing
- the menu items don't work correctly when there is no browser window. For this, we can invert the control and pass in a function that returns a browser window


extr0py:
Splitting this out into separate fixes, and need to address #679 prior to taking this-

PR #536, opened 2017-07-13
Fix #229: Improve msgpack-layer perf by using `msgpack-lite`

### LINKED ISSUE #229: Performance: Improve performance of FastBuffer/walker

extr0py:
Fix #229 - improve performance of msgpack layer


extr0py:
This is one of the largest responsiveness bottlenecks at the moment.

PR #539, opened 2017-07-14
[WIP] preliminary incremental buffer updates

bert88sta:
This is a re-start of an old bit of vimscript I wrote to avoid sending the entire buffer between oni and nvim.

I don't think I have all of the code I need to get behavior 100% normal, but this is a start


browser/src/Services/BufferUpdates.ts
```diff
@@ -52,5 +52,15 @@ export class BufferUpdates {
...
             }
         })
+        
+        this._neovimInstance.on("buffer-update-incremental", (args: Oni.EventContext, bufferLine: string, lineNumber: number) => {
+            // const lastLine = args.line
+            // this._lastArgs = args
+            // this._lastBufferLines = bufferLine
+            // this._lastBufferVersion = args.version
+
+            // If we can send incremental updates, and the line hasn't changed, just send the incremental change
```
bert88sta: I'm not sure what of this I should keep in... it only receives one line, whereas the whole buffer update takes them all

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -12,6 +12,44 @@ function OniNotify(args)
...
+            let context = OniGetContext()
+            call OniNotify(["incremental_buffer_update", context, buffer_lines, line(".")])
+        else
+            let buffer_lines = getline(1,"$")
+            let context = OniGetContext()
+            call OniNotify(["buffer_update", context, buffer_lines])
+        endif
+        let b:last_cursor_line = line(".")
+    endif
+endfunction
```
bert88sta: For some reason, this blocks load. It works after keypress, so I'm not sure

bert88sta:
@extr0py: thoughts?

browser/src/Services/BufferUpdates.ts
extr0py: I think here, we'd want to update `_lastBufferLines`, for sure - since that is sent over sometimes to the plugin. Calling `notifyBufferUpdateIncremental` is important because that's how we update the language services with the current line, so they can give us up-to-date completion. We'd also want to update the `_lastBufferVersion` since there will be a new change tick now.

vim/core/oni-core-interop/plugin/init.vim
extr0py: If it's blocking load, the means Neovim hit an error while executing it on startup. It might be worth trying to call it once Neovim is loaded to see if you can capture the error. 

Regarding `last_cursor_line` and `last_change_tick` - if we remove those, then in insert mode, we'd just send over the single line? And in normal mode, we'd send over the whole buffer? It might be good to remove those, so I can see what it looks like simplified without those. That does seem like it would give a nice perf boost when typing in insert mode!

extr0py:
The overall strategy looks good to me! Long-term, we might be able to replace it with the `LiveUpdate` PR in progress in Neovim. In the short-term, though, only sending over the current line when in insert mode would be a big perf boost, especially for large files. We only send the current line over to the plugins, anyway, so in that insert mode case, there isn't a need to send over the entire file. I think this could help a lot with typing responsiveness in insert mode!

A couple things we should look at before merging:
- Fix the error causing Neovim to block load
- Remove extra stuff around `last_cursor_line` and `last_change_tick`

Thanks @bert88sta for looking into this!

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -12,6 +12,44 @@ function OniNotify(args)
...
+        let context = OniGetContext()
+        call OniNotify(["buffer_update", context, buffer_lines])
+    endif
+    if b:changedtick > b:last_change_tick
+        let b:last_change_tick = b:changedtick
+        if mode() == 'i'
+            if b:last_cursor_line < line(".")
+                let buffer_lines = getline(b:last_cursor_line, line("."))
+            elseif b:last_cursor_line > line(".")
+            let buffer_lines = getline(line("."),b:last_cursor_line)
```
extr0py: Minor thing here, looks like indentation is off

browser/src/Services/BufferUpdates.ts
bert88sta: @extr0py:  So it's ok to just update lastBufferLines with just the single line that gets sent over? It seemed like that was always updated with the whole buffer, not just a single line.

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -19,27 +19,34 @@ function OniNotifyBufferUpdate()
...
+        if mode() == 'i'
+            let buffer_line = [getline(".")]
+            let context = OniGetContext()
+            call OniNotify(["incremental_buffer_update", context, buffer_line, line(".")])
+        else
+            let buffer_lines = getline(1,"$")
+            let context = OniGetContext()
+            call OniNotify(["buffer_update", context, buffer_lines])
+        endif
+        let b:last_cursor_line = line(".")
```
extr0py: Can we remove this reference of `b:last_cursor_line` as well?

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -19,27 +19,34 @@ function OniNotifyBufferUpdate()
 
     if b:changedtick > b:last_change_tick
         let b:last_change_tick = b:changedtick
-        let buffer_lines = getline(1, '$')
-        let context = OniGetContext()
-        call OniNotify(["buffer_update", context, buffer_lines])
+        if mode() == 'i'
```
extr0py: Looks much simpler now! Thanks 👍

bert88sta: Yeah, the last_cursor_line business was an attempt to handle more complex things like multiline deletes and such, but It's a bit hard to make all the scripts

browser/src/Services/BufferUpdates.ts
```diff
@@ -52,5 +52,14 @@ export class BufferUpdates {
                 this._pluginManager.notifyBufferUpdate(args, bufferLines)
             }
         })
+
+        this._neovimInstance.on("buffer-update-incremental", (args: Oni.EventContext, bufferLine: string, lineNumber: number) => {
+            this._lastArgs = args
+            // this._lastBufferLines = bufferLine
```
extr0py: I believe this needs to be set, similiar to how incremental update is handled above. I noticed the completion doesn't seem to be working correctly when typing in insert mode in TypeScript files - I think this may be the root cause. (For example, typing `window` on a new line behaves differently on this branch vs master).

browser/src/NeovimInstance.ts
```diff
@@ -128,6 +128,12 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                             const bufferLines = args[0][1]
 
                             this.emit("buffer-update", eventContext, bufferLines)
+                        } else if (pluginMethod === "incremental_buffer_update") {
```
extr0py: Sorry you have a merge conflict here - I moved `NeovimInstance.ts` to the `neovim` folder. Most of the code is the same though, so hopefully this can be brought over to that new file without too much pain. Let me know if you want help here.

extr0py:
I just tried again - I'm still seeing the completion issues.

Here's a gif of the __expected__ behavior on `master`:
inc-completion-current

Note that as the value is typed from `w` -> `wi` -> `win`, the set of completions is changed.

Here's the behavior I'm seeing on this PR:
inc-completion-539

This appears to be a regression from this change, so we'll need to fix this. 

I see that a failure in the `getCompletions`  code in the typescript language service, in this block:
```
        while (col >= 0) {
            var currentCharacter = currentLine[col];
            if (!currentCharacter.match(/[_a-z]/i)) {
                break;
            }
            currentPrefix = currentCharacter + currentPrefix;
            col--;
        }
```

It seems like the values in the current line are out of sync with the column value (so somehow the current line isn't making it all the way here, or something), and the `col` ends up indexing a value outside of the current line. Likely this is occurring further up, or their is some timing change that is impacting this code path.

bert88sta:
I'm honestly not sure why this is happening. the *only* difference between
the prior implementation and this implementation is lastBufferLines. Other
than that, this uses the same methods. I'll give it a look later tonight or
tomorrow since I have off work

bert88sta:
Actually, this bit from the index.js:

```const activate = (Oni) => {
    Oni.on("buffer-update", (args) => {
        const fullText = args.bufferLines.join(os.EOL)
        const words = fullText.split(/\W+/)

        lastBuffer = args.bufferLines

        currentWords = Object.keys(words.reduce((prev, cur) => {
            prev[cur] = cur
            return prev
        }, {}))
            .filter(w => w.length >= 3)
    })

    const getCompletions = (textDocumentPosition) => {
        if (textDocumentPosition.column <= 1)
            return Promise.resolve({
                completions: []
            })

        let currentLine = lastBuffer[textDocumentPosition.line - 1];
        let col = textDocumentPosition.column - 2
        let currentPrefix = "";

        while (col >= 0) {
            const currentCharacter = currentLine[col]

            if (!currentCharacter.match(/[_a-z]/i))
                break

            currentPrefix = currentCharacter + currentPrefix
            col--
        }
```

I'm going to take a wild guess and say that lastBuffer is supposed to be the whole thing, but is getting assigned to the single line that gets sent over from neovim. I don't know if this would throw an error or not

extr0py:
> I'm going to take a wild guess and say that lastBuffer is supposed to be the whole thing, but is getting assigned to the single line that gets sent over from neovim. I don't know if this would throw an error or not

Ya, good catch! I think this logic isn't quite right in `BufferUpdate.ts`:
```
        this._neovimInstance.on("buffer-update-incremental", (args: Oni.EventContext, bufferLine: string, lineNumber: number) => {
            this._lastArgs = args
            this._lastBufferLines = [bufferLine]
            this._lastBufferVersion = args.version

            this._pluginManager.notifyBufferUpdateIncremental(args, lineNumber, bufferLine)
        })
```

Seems like `this._lastBufferLines` should be updated as:
```
this._lastBufferLines[lineNumber - 1] = bufferLine
```

Kind of like how it is updated in the `buffer-update` case above.

bert88sta:
I changed that, but I still can't seem to get completions for webkit*

extr0py:
Changes look good! Thanks for fixing it up. Once the conflict is resolved, I'll bring it in.

bert88sta:
Aha! fixed it. Now we'll just see if I linted my code well enough :/

extr0py:
Thanks for making the fixes, @bert88sta ! Looks like it's working for me, and it certainly improves perf in insert mode for large files. Nice work, appreciate the contribution!

Issue #486, opened 2017-06-10
Status Bar: Add git branch status


bryphe:
Splitting out from #201 - it'd be great to see the current git branch as an optional setting in the status bar. This is one item I miss with the new status bar from #201.

bert88sta:
I've got some code for this, I made a component and everything, I'm just not sure where it should go. I added a method to the Git.ts that this needs to run, so it seems like oni-core-statusbar/index.js is not the right place. Any thoughts?

extr0py:
Awesome! That's really great. I've been missing have my git branch in the status bar, I'd be stoked to have this back.

Just to give you an idea of my long-term vision - I'd like to split up pieces of Oni into subcomponents (kind of like how Atom does it). For example, move `oni-core-statusbar` to an npm package, and same with other plugins. And move some existing functionality into plugins / packages (like quickopen / fuzzy finder).

For git integration in general, I was thinking of creating an `oni-core-git` plugin under `vim/core`. To start with, this would be focused mainly on the statusbar. But then later I was planning on extending it a bit, especially to bring in things like diffing and git blame. 

So if you're up for that - adding a new plugin (which you can start by copying `oni-core-statusbar` to a new folder `oni-core-git`) - that would be awesome. Otherwise, I think `oni-core-statusbar` is reasonable, because we can always split that out later.

Thanks for looking at it, excited to check it out. Keep me posted if you need any help.

bert88sta:
I meant more along the lines that I didn't know if I could put it in it's own file in there or not. I'm not sure about importing Git into index.js, doesn't seem very "decoupled" and also I don't know if javascript is just going to let me import a Typescript file...

Sorry these are just basic questions haha. I like code, not architecture

extr0py:
> Sorry these are just basic questions haha. I like code, not architecture

Haha not at all, the module / import system is pretty confusing in JavaScript as it is. Combine that with different strategies we have in TypeScript+Webpack vs JavaScript, and it is certainly a recipe for confusion.

It sounds like you're adding the functionality in `Git.ts` and then using it in the statusbar. You're right, you can't import that, but we can give access to it in other ways. In that case, we might want to extend the `Oni` object we expose to plugins to have those methods available. 

Having the API available through `Oni.git`:
- `Oni.git.isGitRepository(path)`
- `Oni.git.getTrackedFiles()`
- `Oni.git.getBranch()`
- `Oni.git.onBranchChanged(callback)`

[...]

PR #540, opened 2017-07-15
#486 - Add git branch to statusbar

Galbar:
* Add git branch indicator to status bar

* Add `getBranch()` to `Git` in `Services` and expose it to plugins through `Oni.services.git`
* Expose ui icon creation to plugins through `Oni.ui.createIcon()`

screenshot from 2017-07-15 13-47-00

### DIFF
--- a/browser/src/Plugins/Api/Ui.ts
+++ b/browser/src/Plugins/Api/Ui.ts
@@ -1,0 +1,14 @@
+import { Icon, IconProps, IconSize } from "../../UI/Icon"
+
+export class Ui {
+    constructor(private _react: any) {
+    }
+
+    public createIcon(props: IconProps): any {
+        return this._react.createElement(Icon, props)
+    }
+
+    public get iconSize(): any {
+        return IconSize
+    }
+}
--- a/browser/src/Services/Git.ts
+++ b/browser/src/Services/Git.ts
@@ -7,6 +7,10 @@
 import { exec, execSync } from "child_process"
 import * as Q from "q"
 
+interface IExecOptions {
+    cwd?: string
+}
+
 export function isGitRepository(): Q.Promise<any> {
     const deferred = Q.defer<boolean>()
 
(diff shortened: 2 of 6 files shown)


bert88sta:
@Galbar: Thanks for picking up after me! Do you think you could potentially model this after the zsh git integration?

Galbar:
@bert88sta: How do you mean?

bert88sta:
@Galbar: I'm not sure of the syntax exactly, I believe it's  * to indicate dirty, + to indicate commits or something along those lines. I'll get a link for you.

bert88sta:
@Galbar: 
image

browser/src/Plugins/Api/Ui.ts
```diff
@@ -0,0 +1,14 @@
+import { Icon, IconProps, IconSize } from "../../UI/Icon"
+
+export class Ui {
+    constructor(private _react: any) {
+    }
+
+    public createIcon(props: IconProps): any {
+        return this._react.createElement(Icon, props)
```
extr0py: Cool, thanks for including both this and `Oni.services.git` in our API surface. Great idea!

browser/src/Services/Git.ts
```diff
@@ -31,3 +35,20 @@ export function getUntrackedFiles(exclude: string[]): Q.Promise<string[]> {
     const untrackedFiles = execSync(cmd).toString("utf8").split("\n")
     return Q.resolve(untrackedFiles)
 }
+
+export function getBranch(path?: string): Q.Promise<string> {
+        const options: IExecOptions = {}
```
extr0py: Glad we can share this functionality between the core code and plugins.

extr0py:
Change looks great to me! Thanks @Galbar!

There are a few things we could look at as next steps:
- @bert88sta 's feedback regarding additional status info (ie, files modified/uncomitted/etc) . We might be able to get some of this info from the output of the git 
- Detecting when the branch has changed, since currently we only update on entering a buffer (for example, a user changed branch in the terminal). We could potentially add an event on the Git shared object like `branch-changed` that the plugin could subscribe to.
- A configuration option to enable / disable the statusbar (ie, `git.statusBar.enabled`)

But IMO we can tackle those points as separate issues and bring in the PR as-is, and continue to build on it. This is something I've missed in my workflow so I'm excited to have it back! Nice work! 👍 👍

PR #548, opened 2017-07-22
Fix #547 - positioning issues with error markers from the LSP

### LINKED ISSUE #547: Error markers are off by one line
- Error squigglies & markers seem to be off by a line for all language-service implementations.
- For python language service, all errors are showing up as informational.

extr0py:
There was an off-by-one issue - the error markers assumed a 'one-based' protocol for the line numbers, as this is what the typescript language server uses. However, the language server protocol specifices a 'zero-based' protocol - since much of the code is common between the strategies, one of them wasn't going to work correctly.

The fix is to move to a consistent protocol, with the LSP taking precedence, so the code now handles a 'zero-based' protocol.


extr0py:
Fixes #547

Issue #562, opened 2017-08-02
Quickinfo panel too small


josemarluedke:
The quick info panel usually is too small to have a good, fluid read of the description. Bellow is an example of what I mean. 

<img width="609" alt="screen shot 2017-08-02 at 4 05 05 pm" src="https://user-images.githubusercontent.com/230476/28898793-a27f1544-779c-11e7-8898-3035b6422f19.png">

Happy to fix this if provided guidance. Thanks.

extr0py:
Thanks for logging the issue, @josemarluedke (and for the great screenshot).  Glad this is tracked! It's a very visible bug and certainly hurts the polish and usability of the app.

It seems like there are a couple issues here:

- The `title` element is wrapping - we should add the `white-space`: `nowrap` style to the title to accomodate this. I think this would be enough to solve the issue in the screenshot.
- If there is a description, we should set a `min-width` to some value - like perhaps `300px`?

Both of these could be updated in either `/oni/browser/src/UI/components/QuickInfo.tsx` and `/oni/browser/src/UI/components/QuickInfo.less`

One other issue I notice is that if the quickinfo is near the edge of the screen, it gets truncated:
![image](https://user-images.githubusercontent.com/13532591/28931550-5a017994-782b-11e7-8f18-f319ef8550a1.png)

This is due to the sizing logic we have in the `QuickInfo` component:
```
        const innerCommonStyle = {
            "position": "absolute",
            "opacity": this.props.visible ? 1 : 0,
            "max-width": (document.body.offsetWidth - this.props.x - 40) + "px",
        }
```

When the cursor is close to the edge, the `max-width` becomes small! That's something we could take a look at separately, though. I think what would be ideal here is to have some fixed width, and use a triangle 'beak' to point at the relevant location... something like this:
![image](https://user-images.githubusercontent.com/13532591/28932010-e9b9e548-782c-11e7-9b2c-fd406ae59f62.png)

If you're up for taking this on, or any pieces of it, that would be great! Let me know if there is anything I can do to help, or if you hit any road blocks on the way.

extr0py:
Using @josemarluedke 's fix and it's a much nicer experience now! 😄 

I was going to close this out, but there's still an issue when it gets clipped near the edge of the screen - I'll leave this issue open to track that.

bryphe:
The remaining layout issue on the edge of the screen should now be addressed by #766

PR #563, opened 2017-08-03
Use pre for white-space in quickinfo

josemarluedke:
Partially solves #562.

I went with `pre` instead of `nowrap` because of this:

## Using `nowrap`
screen shot 2017-08-03 at 9 52 33 am

## Using `pre`
screen shot 2017-08-03 at 9 52 16 am

### DIFF
--- a/browser/src/UI/components/QuickInfo.less
+++ b/browser/src/UI/components/QuickInfo.less
@@ -18,6 +18,7 @@
         .title {
             width: 100%;
             margin: 8px;
+            white-space: pre;
         }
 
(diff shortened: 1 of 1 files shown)


extr0py:
Thanks for making this change @josemarluedke . Looks great to me!

And good call on using `pre` instead of `nowrap` - the experience is much better in that case where there are line breaks in the title string.

I'll bring this in once the build has completed.

extr0py:
Merged - thanks for your contribution, @josemarluedke !

Issue #566, opened 2017-08-04
Add 'Open Folder' option to Oni


extr0py:
Splitting this out from #506 - there should be a way in Oni to easily open a folder. This is pretty common functionality in other editors, like VSCode. It'd be helpful to have an `Open Folder` menu option in the `File` menu for Oni.

As far as I know, Vim/Neovim don't have a concept of opening a folder, so we'd have to define what that means for Oni.

Right now, this would do the following:
- Set the working directory to the folder

In addition, this could hook up with some other integrations:
- Set the root of a file explorer window, when this is implemented
- Load per-folder specific configuration (like .editorconfig, `.oni` configuration, etc).

cyansprite:
Perhaps when this happens we could set off an async task to gather up the files (git ignore in mind) and cache these files, I'm working on a neovim plugin for searching, replacing and destroying, I can look into how to integrate that with Oni (is there a way to integrate python and typescript? I can try converting it), and the maybe add some kind of caching system after I figure out the open folder thing. 
I was planning on taking some common code and make an indexing system which is closely related to "opening a folder" and caching.

Thanks for listening I will try to get some coding in tonight!
One step at a time just some ideas...

extr0py:
> Perhaps when this happens we could set off an async task to gather up the files (git ignore in mind) and cache these files, 

Sounds reasonable! As a side benefit, this would also help significantly with `QuickOpen` performance - if we could cache those results instead of having to ask git or the tool of choice everytime.

> I'm working on a neovim plugin for searching, replacing and destroying, I can look into how to integrate that with Oni (is there a way to integrate python and typescript? I can try converting it), and the maybe add some kind of caching system after I figure out the open folder thing.

VSCode actually has something similiar... Their search functionality is powered by a tool called RipGrep, which they package via npm as: https://www.npmjs.com/package/vscode-ripgrep . It's actually built in Rust but you could use this strategy for any language that can compile cross-platform.

The problem with Python is it's not always guaranteed there's a python runtime (at least on Windows machines), so I've been avoiding writing core functionality in Python.

It might make sense to create a standalone binary (at least if it fits the scenario you're looking at) - that way it can be leveraged between Oni, Neovim, or even other editors like Atom/VSCode. Otherwise of course any web language would work!

I'll try and send out a PR of the 'Open Folder' option to change the working directory, so that we have an entry point in which we can build out some of this additional functionality.

[...]

Issue #506, opened 2017-06-29
Fuzzy Files: Not opening files from current directory


extr0py:
Splitting this out from #503 (opening multiple files) - from @CrossR:

> Normally I'd use some form of fuzzy file opener to just open up more tabs/splits, but I'm also having an issue where that isn't opening the current directory...I can open a separate issue for that though.

> This occurs with 0.25 and 0.26rc in Windows 10.

And it looks like the behavior is as follows: 

> It's not empty, it's just regardless of where my file is open, it only shows files relative to oni in some fashion, I think the electron folder. If I swap the search command to the suggested windows one in Config.ts, It swaps to the program files folder that oni is installed in.

extr0py:
Just a couple of follow-up question on my end (and thanks @bert88sta for drilling into the behavior)
- How are you opening up Oni? Is it via the start menu or some other way?
- What does ":pwd" return? This is the current working directory Neovim knows about.

Right now the Fuzzy Files / QuickOpen experience uses the current working directory from Neovim, and it defaults to wherever Oni was opened. 

It sounds like the expectation here is to set the working directory along with opening the file, which makes sense to me. It doesn't really make sense for Oni to have the working directory stick around as the install directory when you explicitly open a file.

Maybe a behavior we can start with here is that, if the working directory hasn't been changed (is the install folder), we should just change it with the first file that is opened - that would smooth out this experience.

This also aligns a bit with the `autochdir` setting in Vim/Neovim.

A couple of other things I'm thinking about that don't have answers right now:
- For projects like Node or C#, where we can determine a root folder, like a `package.json` or `*.csproj` respectively - would it make sense to be smart and look upwards to grab that folder? This is something a plugin could also handle, or the language client could handle, since it does this traversal anyway.
- For multi-selection - default to the last open file?
- Maybe it makes sense to have an `Open Folder` option that just changes the working directory?

CrossR:
So, normally I'm opening Oni via the start menu. Using :pwd like that returns F:\Program Files\Oni, which is the install directory I have.
Opening it via cmd/powershell, :pwd instead returns the folder I was in when I opened Oni.

So yeah, it is set to wherever Oni is started from. From the shell that is the current folder, but from the Start menu this is just the Program files folder.

[...]

Issue #473, opened 2017-06-02
Oni open binaries will cause crash


gynet:
If a user uses Oni to open a binary, e.g Oni.exe, node.dll in the package, it will crash within seconds.

extr0py:
Thanks for logging the issue! I'm assuming you're on Windows?

I'm able to reproduce as well on Windows 10. It looks like Oni gets the `disconnect` even from Neovim (which I believe means the neovim process died).

This doesn't reproduce in `nvim-qt.exe` - so I'm wondering if it is related to the default / core set of plugins we use for Oni integration. In particular, the buffer update might be causing problems for these large binary files.

gynet:
Yes, I am using Windows 10

keforbes:
Interesting, this isn't a problem on Linux; it loads binary files just fine.  I tried with nvim.exe, which is 3.8MB so I don't think large file sizes are a problem.  I wonder if this issue only occurs on Windows.

cyansprite:
I will look into this too since I am making a plugin and it sometimes opens very, very large files :).

PR #568, opened 2017-08-05
Add basic bookmark support

cyansprite:
# 'Fixes' #506 - Fuzzy files not changing to current directory
Whenever you open Oni from the menu or clicking on the icon, it can
cause the fuzzy finder to try to find through-out the install directory,
which made Oni unusable and have to crash it or kill it via task manager
or kill -9 on linux, etc.
This checks if you are in the install directory, if so then open
bookmarks instead of searching through files.

- Implement oni.bookmarks so user can set these.
- Comment some sections of the code I have concerns about.
- Add different icon type for bookmarks.
- If bookmark change directory of vim to the bookmark
  (which triggers DirChanged so also Oni.)

##### TODO: Before adding more bookmark support it really needs to wait until we have an "Open Folder" #566 solution,so that we know how to better handle bookmarks and how to combine that with opening folders, but maybe having a key combo always to be able to open bookmarks (this would effect opening folders too?).

### DIFF
--- a/browser/src/Services/QuickOpen.ts
+++ b/browser/src/Services/QuickOpen.ts
@@ -33,6 +33,13 @@ export class QuickOpen {
             } else {
                 neovimInstance.command("vsp! " + fullPath)
             }
+
+            // TODO If we export the icon, change from arg.icon to whatever
+            // Or have modes
+            if (arg.icon === "chain") {
+                neovimInstance.command("cd! " + arg.detail)
+            }
+
         })
     }
 
@@ -55,7 +62,7 @@ export class QuickOpen {
                 const files = execSync(overriddenCommand.replace("${search}", ""), { cwd: process.cwd() })
                     .toString("utf8")
                     .split("\n")
-                this._showMenuFromFiles(files)
+                this._showMenuFromFiles(files, "file-text-o")
                 return
             } catch (e) {
                 Log.warn(`'${overriddenCommand}' returned an error: ${e.message}\nUsing default file list`)
@@ -63,15 +70,21 @@ export class QuickOpen {
         }
 
         // Default strategy
-        //  If git repo, use git ls-files
+        //  If in exec directory, show bookmarks to change cwd to
+        //  Otherwise, If git repo, use git ls-files
         //  Otherwise, find all files recursively
         const openPromise = Git.isGitRepository()
             .then((isGit) => {
+                if (process.execPath === process.cwd()) {
+                    const bookmarks = config.getValue("oni.bookmarks")
+                    this._showMenuFromFiles(bookmarks, "chain")
+                    return
+                }
                 if (isGit) {
                     return Q.all([Git.getTrackedFiles(), Git.getUntrackedFiles(exclude)])
                         .then((values: [string[], string[]]) => {
                             const allFiles = _.flatten(values)
-                            this._showMenuFromFiles(allFiles)
+                            this._showMenuFromFiles(allFiles, "file-text-o")
                         })
                 } else {
                     // TODO: This async call is being dropped, if we happen to use the promise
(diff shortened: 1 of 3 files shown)


browser/src/Services/QuickOpen.ts
```diff
@@ -33,6 +33,13 @@ export class QuickOpen {
             } else {
                 neovimInstance.command("vsp! " + fullPath)
             }
+
+            // TODO If we export the icon, change from arg.icon to whatever
+            // Or have modes
+            if (arg.icon === "chain") {
+                neovimInstance.command("cd! " + arg.detail)
```
extr0py: Oh cool! I just added a `neovimInstance.chdir` method in my PR (#577), but I missed adding the `!` - that's a good call.

browser/src/Services/QuickOpen.ts
```diff
@@ -55,46 +62,60 @@ export class QuickOpen {
...
         }
 
         // Default strategy
-        //  If git repo, use git ls-files
+        //  If in exec directory, show bookmarks to change cwd to
+        //  Otherwise, If git repo, use git ls-files
         //  Otherwise, find all files recursively
         const openPromise = Git.isGitRepository()
             .then((isGit) => {
+                if (process.execPath === process.cwd()) {
```
extr0py: We might want to make this a helper method, like `isInstallDirectory(process.cwd())` - we might need to differentiate the logic a bit between platform. I believe for Windows this should work, but for OSX and Linux this logic might not be quite correct. Refactoring it to a method will help encapsulate this without the if getting too crazy.

extr0py: Also, we may want to move this check up, as it looks like it is only in the `Git` strategy. This could be problematic because, if the install directory is not controlled by git, we wouldn't hit this bookmark logic.

extr0py:
Cool, thanks for taking initiative to start this, @cyansprite ! Appreciate it! 😄 

One question I had - I think the default cause for this will be that the user _does not_ have any bookmarks set. If we bring up the open-bookmark experience, would it just show an empty list? I'm wondering if we can smooth out that initial experience.

Some options:
- We could prompt the user to open a folder
- We could show some experience that describes how bookmarks work

I think this will be a really common scenario when a user downloads Oni and opens it for the first time, without any bookmarks set - I want to make sure that is a smooth and intuitive experience.

cyansprite:
I added some comments on the issue let me go get those and yeah I will alter the files when I get home, also will pull your pull request so i can use your nvim.chdir instead if the command.

'
Also i noticed that if you select loading files it tried to open loading files..
Going to fix that and also going to detect if bookmarks is empty then offer to direct to oni config.
Also with my plugin im making I've researched file types that are never to be opened so i compiled a list and I think i can write some kind of binary checker to ensure we don't open binary files or show them such as .exe
'
The second should help #473

cyansprite:
Sorry for choppy messages I'm on my phone and lazy:p

cyansprite:
I updated it, it works on windows I'm not done testing yet or fixing everything... and don't ask why it's another commit on top of it... I'm in a rush >.>

extr0py:
Cool, looks like good progress!

> and don't ask why it's another commit on top of it... I'm in a rush >.>

No worries, I squash the commits anyway when I merge, so don't feel too much pressure there 😄

extr0py:
Sorry, didn't mean to close, hit the wrong button! Look forward to seeing it in action.

cyansprite:
Oh hey cool, I was going to fix it but if you want to be my guest :).

I will test it on linux today and also make a .gif to show it off.

extr0py:
> I will test it on linux today and also make a .gif to show it off.

Sweet, a .gif sounds great! I just merged in my change to add `chdir` (#570), I hope the merge isn't too bad for you

cyansprite:
Nah I just copied your file over until you merged, so now I just need to rebase. I'm going to clean up the commit probably manually.

cyansprite:
Making a new repo and PR... I made the other on windows and me and windows terminal just... we don't like each other :P.

PR #573, opened 2017-08-06
Fix #561 - Don't call requestAnimationFrame unless something has changed

### LINKED ISSUE #561: CPU usage even when idle
Oni is using some CPU even when idle. This is a common problem among electron apps, but usually can be minimized.

One key issue is that the render animation loop is triggered, even when idle or nothing is changed. This should be updated so that we only `requestAnimationFrame` when there is actually something to draw.

extr0py:
__Issue:__ Oni is always using the CPU, even when idle. This can be observed via task manager, taking up a small percentage of CPU.

__Defect:__ Oni is always calling `requestAnimationFrame`, even if there are no changes from Neovim to render.

__Fix:__ Only call `requestAnimationFrame` if there is an action or change from Neovim.

### DIFF
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -54,6 +54,7 @@ export class NeovimEditor implements IEditor {
     private _screen: NeovimScreen
 
     private _pendingTimeout: any = null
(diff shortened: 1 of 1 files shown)


extr0py:
Fixes #561

Issue #554, opened 2017-07-25
Opening fuzzy finder in large directories.


Bretley:
Occasionally I accidentally open Oni from my home dir, and it tries to recursively list everything in there. Once it's started, Oni becomes pretty much unusable. I don't know that there's a great fix for this, but it's annoying when I screw up.

extr0py:
Yep, unfortunately the current strategy is synchronous:
```
                // replace placeholder ${search} with "" for initial case
                const files = execSync(overriddenCommand.replace("${search}", ""), { cwd: process.cwd() })
                    .toString("utf8")
                    .split("\n")
                this._showMenuFromFiles(files)
                return
```

The fix here would be to switch to executing the search command in an async manner, and streaming the results to the QuickOpen UI. That way the UI doesn't 'freeze up', and results can be accessed as we pick them up from the search command. In addition, it's common to cache locally recent results.

Thanks for opening the issue! This is definitely something that needs to be addressed.

cyansprite:
I will try to look into this.

cyansprite:
Looking into what strategy to use

extr0py:
> Looking into what strategy to use

Cool! I think the most important thing here is to be able to replace the `execSync` in `getTrackedFiles` and `getUntrackedFiles` - that'd be the first step.

The other part is, the way the Menu is currently structured, is it sort of expects to have all the items in one-go. I was thinking about changing the API a bit:
- Have an option to show some sort of 'loading' experience, so that the user knows that it is loading. Right now, we do that with an extra menu option, but it'd be nice to have this baked into the UI as something separate.
- Have the ability to add items as we discover them - this will make the experience much quicker and less painless, and is what most other fuzzy-finders do.

In addition, another goal would be to add caching, so that we can show items we've seen before right away.

Thanks for looking at this! Hope that helps - keep me posted if you have questions

cyansprite:
I'm thinking about how we use quickopen and about files, bookmarks, and was playing around with an idea similiar to fzf.vim where quickopen will become an interface or abstraction and we have a bookmark , file , folder,
Buffer(s) lines, etc quick open. Trying to figure out if that would be a preferred strategy also adding files as they load instead of batch loading.

cyansprite:
And yes first and foremost ditch the batch load :)

cyansprite:
And yes the loading indicator shouldn't be a selectable item, put it on the left or right side of the text field or something and then go away or show a different icon after.

[...]

Issue #434, opened 2017-05-01
Command + v (paste) not working or not implemented?


josemarluedke:
Hey there! Fist, great project, looking forward to it evolve nicely.

I was wondering if command (or ctrl in windows) + v (or c to copy) is implemented or it is just not working for me? I tried to use the edit menu but nothing happens as well. 

Using the last build (v0.2.3-rc2) on MacOs Sierra.

Thanks.

bert88sta:
Check keyboard.tsx, as of right now any command-prefixed keys are just
discarded

extr0py:
Hi @josemarluedke , cheers, thanks for reporting the issue!

Out of the box, Vim / Neovim does not support the standard platform hotkeys for cut/copy/paste (Control/Command + c and Control/Command + v mean different things in Vim).

I've been thinking about how to handle this as part of the goal of "lowering the barrier to entry" to modal editing. I believe that it makes sense to add support for the platform specific hot keys - I'm thinking of the following behavior:

__Visual Mode__
<Control/Command + c> - copy the selected text to the clipboard

__Insert Mode__
<Control/Command + v> - paste the clipboard text

Neovim has clipboard management out of the box, so this potentially be implemented as just `inoremap` and `vnoremap`s in the default config.

extr0py:
The Edit Menu should be using the vim metaphors (yank and paste), but it might not be persisting them on the clipboard.

keforbes:
I have mixed feelings about this.  I can understand wanting to lower the barrier to entry by having the expected key combos for copy/paste, but at the same time, this would get in the way of a Vim-savvy user who wants to use (e.g.) visual block mode.  At the very least, I think this should be configurable.  We should probably define a "beginner mode" and a "true Vim experience" which can be toggled.

extr0py:
Thanks for the feedback, @keforbes, these are great points. I agree, there isn't really a clear answer here.

I guess one option would be to set those mappings when we have the default plugin / Oni behavior active (controlled by the `oni.useDefaultConfig` flag). There is the unfortunate conflict with visual block mode and the usage of <Ctrl+c> in visual mode, but perhaps it is reasonable in terms of lowering the bar if it is documented. I'm leaning towards adding these mappings in the default config but disabled when the `oni.useDefaultConfig` flag is false... I'll keep this around a bit though to stew on it and see if there is any other feedback.

cyansprite:
> Visual Mode
<Control/Command + c> - copy the selected text to the clipboard

[...]

Issue #430, opened 2017-05-01
Lowering the Bar: Gamify & Improve Vimtutor


bryphe:
This came about from discussion with @JosephFerano in the gitter channel. The idea is to have an improved vimtutor experience - a way to teach users modal editing in a fun, paced environment.

It is generally accepted that the way to start learning vim is to open up vimtutor and spend 45 minutes or so getting a feel for the keys, but having an actual game and measured objectives / achievements around it would make it much more fun and enjoyable for new users to pick up Oni. The existence of other experiences like https://vim-adventures.com/ show that there is a demand for this.

This would fit in well with Oni's goal of lowering the bar to modal editing, and help funnel new users that are curious about modal editing to Oni.

CrossR:
@bryphe with the recent tutor bits we've added I think this issue can be closed?
We've still got other tutorials to make, but it looks like specific issues are raised for them each time.

bryphe:
@CrossR , sounds good to me! Agreed, any other issues we can track separately 👍

PR #574, opened 2017-08-06
Basic Bookmark Support/Quick Open Enhancements

### LINKED ISSUE #506: Fuzzy Files: Not opening files from current directory

cyansprite:
# Basic bookmark support
##### Fixes #506 
This checks if you are in the install or home directory, if so then open
bookmarks instead of searching through files which can cause freezing
of Oni.

- Implement oni.bookmarks so user can set these.
- Make sure we don't click on spinning icon or bookmark info message as
  a file
- Comment some sections of the code I have concerns about.
- Add different icon type for bookmarks.
- If bookmark change directory of vim to the bookmark
  (which triggers DirChanged so also Oni.)
- If user doesn't yet have bookmarks supplied open config through
  command so that is the default open oni when one doesn't exist.


cyansprite:
will submit gif later

cyansprite:
### Screen gif :)
<a href="http://imgur.com/O2rilC6"></a>

##### This would be the syntax for the Readme.md
```md
![basic-bookmark-support](http://i.imgur.com/O2rilC6.gifv)
```

extr0py:
Trying this out right now - thanks for the gif! 😄

extr0py:
Tried it out by running the packaged executable on Windows. Cool, looks like the functionality is working great! I was able to add a bookmark and open it.

A couple of suggestions / ideas:

### First round experience
- Can we change the text for the  no-bookmarks menu option to read "Add Bookmark" and add detail text that reads "Select this to open configuration and add a bookmark."
- Could we add a menu option, in the initial case, to also "Open Folder" (to hook in the functionality")?

### General
That's cool that the bookmark defaults to the 'explore' experience. I wonder if that makes sense for 'Open Folder' too, at least until Oni gets a first-class file explorer experience?

Could you please also update the README.md to highlight the functionality? It'd be great to have documentation on:
- the new setting (`oni.bookmarks`) 
- the new entry point (`<C-0>`)

Thanks for putting this together! Look forward to having this in the next release 👍

cyansprite:
>> Can we change the text for the no-bookmarks menu option to read "Add Bookmark" and add detail text that reads "Select this to open configuration and add a bookmark."

Yeah the help message string was just something random to get the point across, I was going ask for your preference. I'm thinking about how we use quickopen and about files, bookmarks, and was playing around with an idea similiar to fzf.vim where quickopen will become an interface or abstraction and we have a bookmark , file , folder, 
Buffer(s) lines, etc quick open. Trying to figure out if that would be a preferred strategy also adding files as they load instead of batch loading.

>> Could we add a menu option, in the initial case, to also "Open Folder" (to hook in the functionality")?

Initial case? Do you mean open folder then select a folder and add that to bookmarks?

Yeah I can document it in a bit out right now.

keforbes:
I've only seen the gif (haven't played with it myself) so correct me if I'm wrong, but are we really only showing bookmarks when you're in your home directory?  I understand most users won't be interested in opening files in Oni's install dir but I don't think it's a safe assumption that no one will ever want to open files in their home directory.

If this is a way to mitigate the effect of scanning large directories, I'm fine with showing bookmarks *first* in the list and show them *immediately* as we fetch the directory contents asynchronously, but I don't think you should explicitly prevent getting the home directory's contents.  And again, if I'm misunderstanding something and you *are* still fetching the home directory's contents then you can ignore this rant. 😄

extr0py:
Yep, that's good feedback! Thanks @keforbes - I think it lines up with what @cyansprite mentioned in terms of next spite in the comments for #554:

> I'm thinking about how we use quickopen and about files, bookmarks, and was playing around with an idea similiar to fzf.vim where quickopen will become an interface or abstraction and we have a bookmark , file , folder,
Buffer(s) lines, etc quick open. Trying to figure out if that would be a preferred strategy also adding files as they load instead of batch loading.

In other words, it sounds like we could not 'special-case' the bookmarks but show them alongside files (as well as buffers). Now that we have 'Open Folder' functionality, we could potentially only show/search files if the user has _explicitly_ opened a folder. There's a precedence for this - as its how VSCode behaves. I think this would help the experience because you'd have to opt-in to the degenerate cases (specifically opening the Oni install directory folder, or the home folder), as opposed to just being surprised that Oni is scanning everything.

cyansprite:
Basically yes I understand going to fix it next PR but currently this is the best I can do without breaking Oni :)

cyansprite:
I think I agree with your open folder idea that we don't search files unless we have a folder open.

We would kill the need for a special case scenario, since the user wouldn't mean to open those folders, and in the case someone does want to search through their home directory we could have options to ignore .files/.dir and ignore binary files, and since I'm converting it to async this should be feasible.

extr0py:
> We would kill the need for a special case scenario, since the user wouldn't mean to open those folders, and in the case someone does want to search through their home directory we could have options to ignore .files/.dir and ignore binary files, and since I'm converting it to async this should be feasible.

That sounds great! The special case scenario doesn't sit well with me, in retrospect, because there is a lot of explaining you have to do to a user to tell them how the feature works and what the expectation around it is. Being able to say "You need to Open Folder before your files show up in QuickOpen" is pretty easy to explain.

Your plan sounds great, I really like the idea of saving the folder and tracking so that it can be easy to get back to it.

In terms of the Open Bookmarks - I wonder if, to keep it simple, we could just always show the bookmarks too in QuickOpen? Otherwise, I'm okay with the default key sequence.

Thanks for following up on this, @cyansprite ! Looking forward to all the QuickOpen improvements! 👍

cyansprite:
> In terms of the Open Bookmarks - I wonder if, to keep it simple, we could just always show the bookmarks too in QuickOpen? Otherwise, I'm okay with the default key sequence.

Hmm... That sounds like a neat idea.

cyansprite:
Haha, I love it.
@extr0py this what you have in mind :)
<a href="http://imgur.com/lYr126v"></a>

extr0py:
Wow, thanks for the gif and putting it together so quickly 😄 

> If you like it, I think we should just always show files regardless if the folder is opened...

I do like it, it shows the user that the 'bookmark' feature is available. The only concern I have is that, we might continue adding items here - it won't be clear to the user what things they need here. I think it also might be a bit confusing how 'Bookmark' differs from 'Open Folder'. But I'm okay trying it out and then refining it!

> If you like it, I think we should just always show files regardless if the folder is opened...
Async of course, tracking the folder is kind of weird... on startup we chdir, so we would have to add
the special case for on start up, and then what if the user opens from command line? Do we respect that as a folder or not?

I'm thinking these might not need to be mutually exclusive. We could have the behavior of - if a folder is opened, show the files, otherwise, show an 'Open Folder' option. It seems like to me we'd want to handle opening from a folder in the command line the same as if they had selected 'Open Folder'.

So I think in that case we'd still need the special case of the install directory - if the working directory is the install directory, treat it like there was no 'open folder' action.

My goal is to keep things as simple as possible for the user (reduce the _cognitive load_ of the experience). 

Another thing I'm thinking about later is having a special start screen for Oni, if there is no file opened - that would show things like opening QuickOpen, adding a bookmark, maybe starting an improved vimtutor (#430), etc.

cyansprite:
> I do like it, it shows the user that the 'bookmark' feature is available. The only concern I have is that, we might continue adding items here - it won't be clear to the user what things they need here.

What other items? The intent would be show bookmarks if available or show this. I can't think of anything else to add unless we list the directories recursively ( Not files , ) so they can choose to open what they are looking for. I don't think we should add anything else to c-p. I am working on doing buffers, buffer lines and other things, but it will be a different mapping because it's a different function.

> I think it also might be a bit confusing how 'Bookmark' differs from 'Open Folder'. But I'm okay trying it out and then refining it!

The only difference between the two are that opening a Bookmark would be opening a folder they saved so they can use the fuzzy finder rather than having to find the directory they want with open folder. So essentially they are the same.

> So I think in that case we'd still need the special case of the install directory - if the working directory is the install directory, treat it like there was no 'open folder' action.

On Linux the install directory, or when I open it from the menu by click on the icon, it opens from the home directory, are we going to treat this as the install directory so if we are in the home folder as well?

> Another thing I'm thinking about later is having a special start screen for Oni, if there is no file opened - that would show things like opening QuickOpen, adding a bookmark, maybe starting an improved vimtutor (#430), etc.

Sounds wonderful, also should be an option whether to show start/welcome screen or to open a blank file.

extr0py:
> What other items? The intent would be show bookmarks if available or show this. I can't think of anything else to add unless we list the directories recursively ( Not files , ) so they can choose to open what they are looking for. 

Oh ya, there is nothing else today - I was thinking about the future, if we decided to add more functionality, it might blur the line between what goes in QuickOpen vs the command palette. It's hypothetical now, so we don't need to do anything until it becomes an actual problem. Was just thinking out loud.

> The only difference between the two are that opening a Bookmark would be opening a folder they saved so they can use the fuzzy finder rather than having to find the directory they want with open folder. So essentially they are the same.

Yep, sounds like a bookmark is really just a shortcut to open a folder. Sounds reasonable.

> On Linux the install directory, or when I open it from the menu by click on the icon, it opens from the home directory, are we going to treat this as the install directory so if we are in the home folder as well?

Good point - I guess if this is the case, we really can't get out of special casing both those directories. If the user clicks "Open Folder" and picks the home directory or install directory, we should let them do that. But we shouldn't open it by default, like we would for other directories. Does that sound reasonable?

> Sounds wonderful, also should be an option whether to show start/welcome screen or to open a blank file

Definitely! 😄

cyansprite:
> Oh ya, there is nothing else today - I was thinking about the future, if we decided to add more functionality, it might blur the line between what goes in QuickOpen vs the command palette. It's hypothetical now, so we don't need to do anything until it becomes an actual problem. Was just thinking out loud.

So how do you feel about loading all the folders when no folder is opened? Considering it will be async?

> Good point - I guess if this is the case, we really can't get out of special casing both those directories. If the user clicks "Open Folder" and picks the home directory or install directory, we should let them do that. But we shouldn't open it by default, like we would for other directories. Does that sound reasonable?

Reasonable yes, just need to move some code around to get it to play nice :).

extr0py:
> So how do you feel about loading all the folders when no folder is opened? Considering it will be async?

My preference would be not loading the folders. I tested the same scenario out in VSCode and Atom and they don't load the folders, either. It also is a safety net from us / the user hitting issues like #177

> Reasonable yes, just need to move some code around to get it to play nice :).

Haha, sounds good

cyansprite:
> My preference would be not loading the folders. I tested the same scenario out in VSCode and Atom and they don't load the folders, either. It also is a safety net from us / the user hitting issues like #177

Then let's try to be a step above them?
I wrote a binary that lists all the dirs recursively under my home dir, ignores all dirs starting with . (hidden directories) in .8 seconds.

Would you like to try it out? (It's for linux so if you are on mac or linux you should be fine.)

cyansprite:
Here's a example of how it would work.
This is from my home directory, async loading folders (You can't tell because it gets done so fast...)
These are binaries I'm making for files and directory loading...
I can add args so the user can pass in ignores they would like and we will have defaults so they can pick from them.

<a href="http://imgur.com/byhc495"></a>

extr0py:
That's a nice find that you don't need two calls to `git ls-files` - that's great!

browser/src/Services/QuickOpen.ts
```diff
@@ -4,102 +4,144 @@
...
-            }
+            // replace placeholder ${search} with "" for initial case
+            this.loadMenu(overriddenCommand.replace("${search}", ""), icon)
+            return
         }
 
         // Default strategy
-        //  If git repo, use git ls-files
-        //  Otherwise, find all files recursively
-        const openPromise = Git.isGitRepository()
```
extr0py: I like that you consolidated this strategy, to reuse the existing code path (and just have `git` be another search command). Thanks!

cyansprite: sure thing

browser/src/Services/QuickOpen.ts
```diff
@@ -4,102 +4,144 @@
...
         const config = Config.instance()
         const overriddenCommand = config.getValue("editor.quickOpen.execCommand")
-        const exclude = config.getValue("oni.exclude")
+        // const exclude = config.getValue("oni.exclude")
+
+        //  If in exec directory or home, show bookmarks to change cwd to
+        if (this._isInstallDirectoryOrHome()) {
+            const bookmarks = config.getValue("oni.bookmarks")
+            let icon = "chain"
+            const helpMessage = "No bookmarks yet! Select this to open config! (shows example)"
```
cyansprite: oh I forgot to add that, I did on my other commit but I abandoned it because I started going in a different direction. 

And I will add an open folder option as well.

Also, do you want to test out my binary? I don't know how to add it to the install or where, I' would assume in bin...

extr0py: > oh I forgot to add that, I did on my other commit but I abandoned it because I started going in a different direction.

> And I will add an open folder option as well.

Cool, thanks!

> Also, do you want to test out my binary? I don't know how to add it to the install or where, I' would assume in bin...

Hmm, do you have a repo for it somewhere? And does it work on Windows? I would be interested in a fast search program. I'd be interested in how it compares to [RipGrep](https://github.com/BurntSushi/ripgrep) - I was thinking about replacing the default strategy with that (and bundling it with Oni) at one point.

cyansprite: https://github.com/cyansprite/ListFiles

cyansprite:
Open Folder support
<a href="http://imgur.com/Alxfh0I"></a>

cyansprite:
C:/Users/brand/Desktop/oni/browser/src/Services/QuickOpen.ts[216, 1]: File should end in newline
>.> 
C:/Users/brand/Desktop/oni/browser/src/Services/QuickOpen.ts[218, 1]: Consecutive blank lines are forbidden
>.> 
File is 217 lines long, I give up on the formatting :P.

cyansprite:
Next thing to do...
@extr0py can you test out my binary, see if it's suitable.

Add buffer lines support, I'm thinking <c-/> for like I'm doing a search, except do it with the fuzzy finder in QuickOpen, then this will jump to the line.

cyansprite:
Buffer lines fuzzy finding, complete :)

<a href="http://imgur.com/1U99reW"></a>

extr0py:
Oh wow, the buffer search is a cool idea - neat!

I'm thinking about some other sorts of searching too (searching _in files_). I like Sublime / VSCode / Atom much better than the vim experience for doing things like find & replace.

extr0py:
Looks like one of my recent changes caused your AppVeyor build failure - I removed the 'Q' library and it looks like there is a conflict in Git.ts. So you might need to port over the changes had to convert that to use ES6 Promises instead of Q.

extr0py:
Oh and I'll try out the binary in a bit. I'm interested as to how it compares to something like [ag](https://github.com/ggreer/the_silver_searcher) or [RipGrep](https://github.com/BurntSushi/ripgrep). I've been toying with pulling ripgrep in as the default non-git search strategy for fuzzy finding.

cyansprite:
> Oh and I'll try out the binary in a bit. I'm interested as to how it compares to something like ag or RipGrep. I've been toying with pulling ripgrep in as the default non-git search strategy for fuzzy finding.

Wellll the binary I'm talking about it specifically for listing files and directories.
I don't know much about ripgrep but I didn't know ag could list files without searching through them.

> Looks like one of my recent changes caused your AppVeyor build failure - I removed the 'Q' library and it looks like there is a conflict in Git.ts. So you might need to port over the changes had to convert that to use ES6 Promises instead of Q.

Okay I will merge the two.

> Oh wow, the buffer search is a cool idea - neat!

Thanks I really liked the idea so I just went for it :).

> I'm thinking about some other sorts of searching too (searching in files). I like Sublime / VSCode / Atom much better than the vim experience for doing things like find & replace.

Definitely. the search and replace and destroy plugin (sarada for short) is a vim remote plugin and WIP, I can make a stand alone binary that would do what we want, give me some time and I will put it together :P, the algo is sound.

With excluding files it's faster than ag in bigger directories on linux, but it's still a WIP... I think I can make a binary hell of a lot faster because the plugin actually lets you select lines you can ignore to replace and destroy the entire line if you want.

extr0py:
> Wellll the binary I'm talking about it specifically for listing files and directories.
> I don't know much about ripgrep but I didn't know ag could list files without searching through them.

Ya, actually `ag` is recommended frequently to go with the `Ctrl-P` plugin - https://stackoverflow.com/questions/28260090/what-is-the-proper-vim-ag-gctrlp-user-command-value-for-windows. And I think `ripgrep` is even faster. There's even an NPM package with a cross-platform binary, too: https://www.npmjs.com/package/vscode-ripgrep , so it'd be straightforward to add as a dependency and package with Oni.

I tried out your binary on a VM, seems like it works well - it'd be interesting to compare the performance though to `ag`/`ripgrep`. There are some other features they have, like respecting the .gitignore functionality, search-in-files, etc.

cyansprite:
> I tried out your binary on a VM, seems like it works well - it'd be interesting to compare the performance though to ag/ripgrep. There are some other features they have, like respecting the .gitignore functionality, search-in-files, etc.

Listing files git ls-files will always be faster than rg or ag because it already has the list with .gitignore in mind.

Searching through files I think it would be easiest to just use ripgrep, it's trusted by a lot of people and I think my plugin is going in a different direction. Mine starts up with vim, when started it creates 16 p-threads, then it ignores .gitignore (i use git ls-files too :P) and then it caches those, as long as you don't change your dir it keeps them, then it pushes to the process daemon and then writes to a file and then loads it (vim doesn't have a good way to handle live streams... trying to find a work-around.)

cyansprite:
Notice something while testing buffer lines, the text is too small >.>

keforbes:
This feature looks like it'll be amazing when it's all done!  Way to go!  @extr0py, are you sure you don't want to hold off on v0.2.8 until this feature is in? :laughing:

cyansprite:
@keforbes thanks :)

do we have a show error popup?

cyansprite:
<a href="http://imgur.com/QXrWKzI"></a>

extr0py:
> This feature looks like it'll be amazing when it's all done! Way to go! @extr0py, are you sure you don't want to hold off on v0.2.8 until this feature is in? 😆

Definitely, it's really cool to see a new 'navigation paradigm'. I like how easy it is to open and put in a split! Nice work @cyansprite 👍 

I'm planning on doing the 'final' release for 0.2.8 next weekend - I did the pre-release just so I could start testing the binaries, see if the space was resized, and make sure there isn't any regression in the functionality. So if it's in before then I'll definitely pick it up!

> @extr0py do we have a show error popup?

Sort of... I kept starting one, and not needing it,  what I started is in `browser\src\UI\components\Message.tsx`. I was intended to be an alert message, though.

@FrigoEU also created an error log that looks like this:
image

Which you can get to via 'Show Logs'.

What's the scenario / case you want to show it?

extr0py:
And I realized I missed this earlier:

cyansprite:
@extr0py 
> What's the scenario / case you want to show it?

I was thinking, when we close a tab if it's modified we don't close it this is because bw is not bw! but that's preferable in some cases because the user might not realize its modified which is hard to do but anyways... When we try to quickopen we do e!/vsp!/sp!, if it's modified and we use e, then that buffers changed will be lost, so I was thinking showing some kind of dialog saying are you sure modified blah blah.

>One feature I do want - though - and this is common for pretty much all the new pieces of UI (like file explorer, git integration, find files, errors, etc) - is that I can navigate using the same paradigms and functionality I use in VIM - in other words, I should be able to use h/j/k/l and all the other tools I'm comfortable with. This is a part of the reason I wanted to build a new editor - most editors that expose extensive functionality like that, even if they have some vim bindings, don't have it set up in a holistic way.

HELL yes, reason why I quit vscode, because some places I couldn't JUST use the keyboard and I couldn't use hjkl.

> I haven't 100% figured this out yet, so if you have any thoughts, that would be great! I know that I don't like the implementation in VSCode though - because when you get your search results, they are always clipped by the left pane. For my workflows, I'd prefer the flexibility to either have them on the bottom, or on the side. But not sure exactly how it should look like yet - might be worth creating an issue to start thinking about ideas!

I see the issue I will look at it there!
thanks!

cyansprite:
Next up is quick open buffers so you can easily cycle through opened buffers with ease :).

If you think if any ideas feel free to let me know :)

cyansprite:
Oh. Color schemes would be cool!!

extr0py:
> I was thinking, when we close a tab if it's modified we don't close it this is because bw is not bw! but that's preferable in some cases because the user might not realize its modified which is hard to do but anyways... When we try to quickopen we do e!/vsp!/sp!, if it's modified and we use e, then that buffers changed will be lost, so I was thinking showing some kind of dialog saying are you sure modified blah blah.

This would be excellent to have! I think this will be a major pain point for users - being able to warn when we navigate away from an unsaved buffer would be very helpful. In addition, it would be nice to have this check on quitting too - it's too easy to quit Oni when you have unsaved changes. It might be worth tracking this in another issue / PR

> Next up is quick open buffers so you can easily cycle through opened buffers with ease :).

Sounds awesome!

> If you think if any ideas feel free to let me know :)

One thing I was thinking about, and this is something similiar to VSCode and other editors, is having symbol prefixes mean special things in QuickOpen. For example, if you type the following prefixes:
- `:` - vim command mode for current window
- `/` - buffer search (integrate with the buffer search you added)
- `!` - oni command mode (same commands shown in Ctrl+Shift+P)
- '`' - show available vim marks
- `&` - show available buffers

I think Nuclide also had some cool ideas with their enhanced quick open experience. They basically showed a few tabs with these sorts of pivots, underneath the text - that might be a good way to enhance discoverability and let users know the sort of things they can do with QuickOpen.

Might make sense again to have this a a separate issue / PR - I don't want you to feel like you have to save the world with this PR! Let me know when you're ready to merge it and I'll test it out.

extr0py:
Oh, and regarding colorschemes, a good starting point might be to use `state.backgroundColor` and `state.foregroundColor` (and darken the background a bit). I tried it on `onedark` and `vim8solarized_light` and it looked pretty good... much improved from what we had before. On top of that, we could have customized colors like `oni.colors.menu.foreground`/`background`/`accent`, or something.

cyansprite:
> This would be excellent to have! I think this will be a major pain point for users - being able to warn when we navigate away from an unsaved buffer would be very helpful. In addition, it would be nice to have this check on quitting too - it's too easy to quit Oni when you have unsaved changes. It might be worth tracking this in another issue / PR

I agree, and I don't know if it's still an issue but whenever I would switch a buffer it would wipe by undo history. I will check again when I get time.

> '`' - show available vim marks

I've never used this are marks useful?

>: - vim command mode for current window
/ - buffer search (integrate with the buffer search you added)
! - oni command mode (same commands shown in Ctrl+Shift+P)
& - show available buffers

I haven't looked how are are matching typing yet but I can imagine it shouldn't be too hard.

> I think Nuclide also had some cool ideas with their enhanced quick open experience. They basically showed a few tabs with these sorts of pivots, underneath the text - that might be a good way to enhance discoverability and let users know the sort of things they can do with QuickOpen.

You will have to explain a bit I'm slightly confused :P.

> Might make sense again to have this a a separate issue / PR - I don't want you to feel like you have to save the world with this PR! Let me know when you're ready to merge it and I'll test it out.

I think it would be beneficial for you and me both to go ahead and merge this for testing and merge conflicts in the future. The only reason I would hold of is for the listing of files NOT in a git repo but...
A PR can fix that later :P.

> Oh, and regarding colorschemes, a good starting point might be to use state.backgroundColor and state.foregroundColor (and darken the background a bit). I tried it on onedark and vim8solarized_light and it looked pretty good... much improved from what we had before. On top of that, we could have customized colors like oni.colors.menu.foreground/background/accent, or something.

Going to start with vim colorschemes and buffers, then add in the oni colors, since vim colorschemes will be done just by modifying QuickOpen.

cyansprite:
Last gif showing what was accomplished.
You can tell I don't use my mouse much when it comes time to open a folder lol...

<a href="http://imgur.com/lsnUVrs"></a>

extr0py:
@cyansprite - nice! I like the icons you added for the bookmarks & open folder 👍

extr0py:
> Lol am I doing too much?

Haha, no way, I just want to make sure we can get pieces of this in the 0.2.8 release (if you want!). 

> I think it would be beneficial for you and me both to go ahead and merge this for testing and merge conflicts in the future. The only reason I would hold of is for the listing of files NOT in a git repo but...
A PR can fix that later :P.

Cool, I just took a pass over it. The buffer find is really cool!

There are only two issues I see:
- The finding for _non-git_ folders seems to have regressed. On master, if I use QuickOpen on a folder that has a bunch of stuff (like 'Documents' or 'Downloads'), I immediately see a loading spinner:
image
But with these changes, I don't see anything. I don't think we need to implement the full-external-binary solution, but we shouldn't introduce a regression for this experience (in other words, it should at least behave the same as it did before).
- Reading the whole set of lines the way we're doing right now is expensive - there's actually already a `BufferUpdates` object that keeps track of changes, and caches the last set of buffer lines. We should just use that object, instead of querying Neovim for the current buffer. For large files, this will be slow / expensive, especially if we're doing it every time we leave insert mode.

Once those issues are addressed, I'll bring it in!

cyansprite:
> The finding for non-git folders seems to have regressed. On master, if I use QuickOpen on a folder that has a bunch of stuff (like 'Documents' or 'Downloads'), I immediately see a loading spinner:

Right I will add that back in, we didn't finish on figuring out if we were using ripgrep so I was waiting until then but for now I will go ahead and add it back in.

> Reading the whole set of lines the way we're doing right now is expensive - there's actually already a BufferUpdates object that keeps track of changes, and caches the last set of buffer lines. We should just use that object, instead of querying Neovim for the current buffer. For large files, this will be slow / expensive, especially if we're doing it every time we leave insert mode.

will do, I need a better overview of the code -.-, or at least a way to ask, is this implemented yet?

cyansprite:
I decided to leave blank lines in buffer search for quickopen, I think it works better because sometimes i think it's odd if some numbers are missing :P.

Anyways, almost done :)

cyansprite:
I think the next PR to do with QuickOpen will be colors, buffers, and caching each until an autocmd happens...

So if 
- dirchanged recache loaded files
- Sourced file recache colors
- Buffers are monitored already just need a way to pull them into quickinfo, suggestion easiest way to do this? I can figure it out but it you have a suggestion to make my life easier that'd be great so I don't reinvent the wheel like with buffer lines :P.

My concern with loaded files is that what happens if a person makes a new file but doesn't change directories outside of vim?
We could use fs to watch them but that might be computation intensive... I've haven't looked much into it honestly.

And then the next will do with the 
One thing I was thinking about, and this is something similiar to VSCode and other editors, is having symbol prefixes mean special things in QuickOpen. For example, if you type the following prefixes:

: - vim command mode for current window
/ - buffer search (integrate with the buffer search you added)
! - oni command mode (same commands shown in Ctrl+Shift+P)
'`' - show available vim marks
& - show available buffers

You mentioned stories in keybindings, do you have a program you use to track it? because this be wonderful to have a layout of tasks, I mean I know what needs to be done... but getting confirmation from you and others would be nice as well as making sure I implement it correctly with functionality already existing.

cyansprite:
issue with the bufferupdates, it doesn't load on bufenter, it loads on buffer-update, or am I doing something wrong?

also I noticed
```ts
            const lastLine = args.line
            if (this._canSendIncrementalUpdates && lastLine === args.line) {
```

won't ```lastLine === args.line``` always be true?

cyansprite:
I think it's only when you are on the same line... sof it the lastline === args.line && buffernum? I'll try it out.

cyansprite:
So i think adding checking for buf num as well as line and send the event onbufenter might be an easy fix.

extr0py:
> So i think adding checking for buf num as well as line and send the event onbufenter might be an easy fix.
I will test it when I get time.

Thanks for catching that issue with BufEnter. Looks great :) This might explain why, in some cases, the language services wouldn't load until the cursor moved... nice. Sorry I wasn't able to give more help, was stuck at work - but looks like you were able to solve it without me anyway. Seems like you have the whole flow down between Neovim <-> Oni, that's great!

> You mentioned stories in keybindings, do you have a program you use to track it? because this be wonderful to have a layout of tasks, I mean I know what needs to be done... but getting confirmation from you and others would be nice as well as making sure I implement it correctly with functionality already existing.

I've actually just been keeping a giant 'TODO.md' file right now, nothing formal at all. I've heard good things about 'org-mode' for Emacs, so I'd like to learn more about that, and see if any of the concepts make sense in Oni to bring over. But ya it'd be nice to have something more granular than github issues. I've tried waffle.io for a small project - do you have any suggestions?

> I think the next PR to do with QuickOpen will be colors, buffers, and caching each until an autocmd happens...

Excellent, sounds like a lot to look forward to! I like where you're going with colorschemes, and the live preview is a really neat idea.

Just tried out the PR and it's working great. The regression is fixed in non-git directories, and the buffer behavior is working well. I'm ready to bring it in! 👍 👍 

I'll plan on merging it tomorrow morning, unless I hear otherwise 😄

cyansprite:
:) I try.

No reason from me to hold off. Fire at wil:).
Will try to get some more enhancements done today working on a yankring plugin right now as well because I think vim needs a better tracking system clipboard history and I don't like other plugins

extr0py:
Excellent! I just merged it in. Congrats @cyansprite on getting your first PR in - great work and thank you for the contribution!

extr0py:
> Will try to get some more enhancements done today working on a yankring plugin right now as well because I think vim needs a better tracking system clipboard history and I don't like other plugins

Sounds cool, look forward to it. I was experimenting with some clipboard integration work (because #434 is the main reason I switch out of Oni to other editors, I realized...) - that's in #620 . Might give you some ideas into how we could build a better system on top of the clipboard primitives electron provides.

cyansprite:
> Excellent! I just merged it in. Congrats @cyansprite on getting your first PR in - great work and thank you for the contribution!

Thanks.

PR #576, opened 2017-08-07
Fix #575 (FocusGained / FocusLost functionality)

### LINKED ISSUE #575: FocusGained vim event doesn't work
I use [vim-gitgutter]( https://github.com/airblade/vim-gitgutter) and also have this code to switch to normal line numbers when window looses focus:
```VimL
" Use relativenumber only in the focused window
if exists('&relativenumber')
  set relativenumber
  augroup WindowRNU
    auto!
    auto BufWinEnter,WinEnter,FocusGained * setlocal relativenumber
    auto WinLeave,FocusLost               * setlocal nornu
  augroup END
endif
```

Both don't work with Oni, because `FocusGained` and `FocusLost` doesn't fire.

extr0py:
- Listen to `BrowserWindow`'s `blur` and `focus` events
- Add an `executeAutoCommand` method to `NeovimInstance`
- Call `executeAutoCommand` on `blur`/`focus`

Fixes #575

### DIFF
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -9,7 +9,7 @@ import * as ReactDOM from "react-dom"
 
 import * as types from "vscode-languageserver-types"
 
-import { ipcRenderer } from "electron"
+import { ipcRenderer, remote } from "electron"
 
 import { IncrementalDeltaRegionTracker } from "./../DeltaRegionTracker"
 import { NeovimInstance } from "./../neovim"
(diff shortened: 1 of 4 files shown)


cyansprite:
Do we redraw on focus gain?

Whenever I resize my window it shows a blank whatever background color is (test on windows of latest build)

extr0py:
@cyansprite - sorry, just saw your comment here!

We do not explicitly redraw on `FocusGained` - we might though if there is something in an autocmd that causes Neovim to ask for a redraw (like printing a message, changing line number format, etc).

Resize would likely be a separate issue - this PR had some changes related to resize: https://github.com/extr0py/oni/pull/557

When you say latest build, is it the latest __master__ or the latest downloadable build in the `Releases` tab? If its on __master__, feel free to open up an issue for it! Otherwise, there have been quite a few changes since the last binary release, so it might be fixed in master now.

cyansprite:
@extr0py 
Sorry, I meant building the latest source.
I will test it in a bit and get back to you, whether to open an issue or not.

PR #594, opened 2017-08-10
Fix #586 - Consolidate 'Tab Open...' and 'Open...' options into 'Open Files...'

### LINKED ISSUE #586: Open vs Tab Open
If I launch Oni and go to Oni(File) -> Tab Open, all files will be opened in new tabs and `gt` will move between tabs.  If I launch Oni and go to Oni(File) -> Open, all files will be opened in new tabs but `gt` will *not* move between tabs.

I'm not sure what benefit there is to having both Open and Tab Open given that they seem to do the same thing.  The only difference between the two that I see is any files opened by Open are ignored when using `gt`.  If I open some files with Tab Open and other files with Open, `gt` will only iterate through the tabs opened by Tab Open.

I guess my recommendation is to just remove Open and keep the Tab Open functionality.  Though if we do that we should probably rename Tab Open to just Open.

extr0py:
Fixes #586

### DIFF
--- a/Menu.js
+++ b/Menu.js
@@ -50,8 +50,8 @@ const buildMenu = (mainWindow, loadInit) => {
             {
                 label: 'Open...',
                 click: (item, focusedWindow) => {
-                    dialog.showOpenDialog(mainWindow, ['openFile'], (files) => {
-                        executeVimCommandForFiles(":e", files)
+                    dialog.showOpenDialog(mainWindow, { properties: ['openFile', 'multiSelections'] }, (files) => {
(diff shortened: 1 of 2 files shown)


CrossR:
Was it intended as part of this to revert back to always opening a file in a new tab?
Ie, if I have an empty tab and unnamed tab, I'd expect using open file to open in there, rather than opening up a new tab and leaving me with an empty one.

extr0py:
@CrossR , Ah, no - thanks for catching that! I missed leveraging that code path you added for opening multiple files, but it certainly should be using that. I logged #600 so I don't forget about it.

PR #597, opened 2017-08-11
Upgrade to WebPack 3

extr0py:
- Upgrade to webpack 3 from webpack 1, for some of the improvements in tree-shaking and bundle sizing.
- Incorporate uglification to make our bundle size smaller.... down to 500 kb (which will improve load time).


extr0py:
When debugging locally, you'll want to make sure to use `npm run build:debug` or `npm run start` (live-reload Oni), as `npm run build` will do the uglification / minification, and be impossible to debug. @keforbes @bert88sta @cyansprite - just a heads up 😄  You'll also want to close all Oni windos, and do an `npm install` to pick up the new dependencies.

PR #599, opened 2017-08-11
Fix #596: Scrollbar not showing up due to state regression

### LINKED ISSUE #596: Scrollbar is gone
Is it just me?  Is it just Linux?  I don't have a scrollbar in Oni anymore.

no_scrollbar

extr0py:
The schema for active buffer info in the Redux store was changed, but this slipped by the TypeScript compiler for the `mapStateToProps` function for the scroll bar. This change fixes the logic so that the buffer correctly picks up the active buffer information.


extr0py:
Fixes #596

Issue #560, opened 2017-08-02
Reduce installation size and footprint


bryphe:
- Includes work for #350, which will decrease size / footprint on disk. Will be easier to fix that after the full set of fixes.

There are several other places we need to optimize the package size, though:
- Are all the node_modules needed? Are they being pruned correctly? Should they be flattened? This is a pretty huge culprit, at ~150MB.
- Remove unbuilt sources from package
- Does all of font awesome need to be included?
- Some of the images can be excluded

Long term, some packages should not be included by default - like `oni-plugin-csharp`. This is a contributing factor to the size of the `node_modules`. This will be easier once there is a plugin story for Oni.

I would like the install size to be on par with some of the other Electron editors (~50 MB), and the footprint ideally to be under 100MB.

extr0py:
`lodash` seems pretty huge too - we should only include the subset needed, and remove it from `node_modules` entirely - it's getting double-packed, both in `node_modules` and in the webpack bundle.

bryphe:
The initial set of pruning is done - we should track further items individually.

PR #606, opened 2017-08-12
Remove Omnisharp dependency, extract to separate plugin

extr0py:
- Extract C# functionality into Oni's first official plugin - [oni-language-csharp](https://github.com/extr0py/oni-language-csharp)
- The `omnisharp-client` dependency was very heavyweight - adding over 100MB to the uncompressed footprint of Oni. This will significantly reduce both the unpacked size, as well as the installer size (and installation speed). 
- This also features better handling in the case a plugin has javascript but doesn't provide an `activate` function.

Impacts #560


cyansprite:
Awesome :)

PR #612, opened 2017-08-12
Fix #600 - If [No Name] file isn't modified, 'Open File...' should replace it.

### LINKED ISSUE #600: 'Open File' should open in current file, if file is empty
This was implemented in , but regressed in PR #586, because it misses using that code path.

All the 'multi-open' files should go through that common code path that was added in #526 

It'd be good to factor this to use a `command` so that we have a common code path. There might be a case later too where we'd expose a setting on how the files open (for example, `tabe` vs `e`).

extr0py:
- This fixes a regression brought about the 'Open File...' consolidation - thanks for catching it @CrossR !
- Uses the `executeVimCommandForMultipleFiles`, which handles the empty-file case in a nicer way.


extr0py:
Fixes #600

Issue #598, opened 2017-08-11
"Open With" support


CrossR:
The last big thing that is preventing me from using Oni full time is proper integrations into my OS as an editor.

For me this means:
* Oni registered as a file editor
  * This seems to already be the case.
* "Open With" support.
  * Ie, I can double click a text file and it opens in Oni.
* "Recent File" support.
  * This looks to be implemented, since I've got it in my Windows 10 start menu, but clicking the files doesn't actually open them. I expect this is due to the above bullet point.

Really, its just the "Open With" support. With that, I can drop using Notepad++ on Windows 10 and start using Oni for my poking around files and such.

Is this a Windows specific issue, or just that its not been implemented? I can have a look into how you are supposed to do this with Electron, but I'm not the most knowledgeable with it.

extr0py:
Hi @CrossR ! Thanks for logging this issue. These sorts of blockers are really helpful to call out, as I'm sure this is a blocker for others as well.

On Windows, the 'Open With' functionality is controlled by a registry key. So we'd need to update our installer to build up those associations. There's a stackoverflow thread here:
https://stackoverflow.com/questions/26536030/inno-setup-file-association

And some more information about the tooling used to create our setup program (InnoSetup): http://www.jrsoftware.org/ishelp/

On OSX, it's controlled by a plist file - this has some more info:
https://stackoverflow.com/questions/28222787/mac-os-x-file-association-works-but-file-icon-not-changed

> With that, I can drop using Notepad++ on Windows 10 and start using Oni for my poking around files and such.

That would be awesome! 😄

CrossR:
Looks like getting the registry keys added isn't so bad, its just a few lines in the setup.template.iss file.
What is awkward is that you need to duplicate those lines for each and every extension you want to support...

If you look at the VSCode .iss file here, you can see its just massive : https://github.com/Microsoft/vscode/blob/master/build/win32/code.iss

Since its so easy to add yourself (Right Click -> Open With -> Look for another app on PC -> Point at oni.exe), I think what makes most sense is just adding a few for now, and then add to it as language servers are added?

That all said, even after Oni is registered as the editor, it doesn't open any files when I double click a file, even when I've said to open with Oni.

I'm less sure how I should go about doing this linking, it looks like on Windows at least the file path you want to load is stored in argv[1], so I need to grab that somewhere and then open the file it says. Just can't work out where the code actually enter when I use the open with dialog.

extr0py:
[...]

Issue #344, opened 2017-04-03
Windows Installer: Add option to add Oni to %PATH%


extr0py:
Oni should be available on the %PATH% in Windows, so that a user can type `oni` from a terminal and start up Oni.

The installer currently provides no option to do this - we should add a checkbox / option that states `Add Oni to %PATH%`

CrossR:
This was added with #613.

extr0py:
Verified it's working for me with the latest Windows installer - nice!

extr0py:
I guess you can't claim the bounty until the issue is closed - bountysource should give you a "Claim Bounty" option now.

Issue #350, opened 2017-04-03
Enable Asar packaging for installation


extr0py:
For v0.2, the `asar` option was set to `false` - meaning the app is not bundled into an asar archive. The electron-builder docs recommend keeping this as true, but excluding certain files. This can also help with the app launch speed. On all platforms, we get this warning due to disabling asar packing:

```
Packaging using asar archive is disabled — it is strongly not recommended.
```

To re-enable it, the files in the `bin` folder must be excluded.

extr0py:
The fix for this issue should:
- No longer have the warning that our app is not packaged via asar
- All app contents bundled in the `asar`, except for things that truly cannot like the Neovim binaries
- Work on all platforms (Windows, OSX, Linux)

extr0py:
Bringing this back, as the asar package unfortunately __regressed__ package size. The files / folders specified in `asarUnpack` are actually _still stored_ in the asar package - almost doubling the size!

Will need to follow up and see if there is a configuration change to avoid this, otherwise, we may need a custom packaging strategy or see if there is a possibility of pushing a fix to `electron-builder`.

When this is addressed, will need to update the app icon in the setup scripts (#613)

CrossR:
I've had a look at this now that my PR was merged.

Is my understanding correct that we'd need to unpack the entire extensions folder, certain node modules (ripgrep, neovim, LSP like CSS/Reason/etc, typescript, reason, oni-api), the vim and images folder, and then update all places where we try to do `node_modules/X/...` to `app.asar.unpacked/node_modules/X/...` and ensuring that whatever folder it was trying to access is unpacked?

I've got the very most basic version of it running again, with Oni launching and linked to nvim (though it fails with colourschemes since I don't have extensions unpacked).

bryphe:
> I've had a look at this now that my PR was merged.

Sweet! Thanks @CrossR !

> Is my understanding correct that we'd need to unpack the entire extensions folder, certain node modules (ripgrep, neovim, LSP like CSS/Reason/etc, typescript, reason, oni-api), the vim and images folder, and then update all places where we try to do node_modules/X/... to app.asar.unpacked/node_modules/X/... and ensuring that whatever folder it was trying to access is unpacked?

[...]

PR #613, opened 2017-08-13
[WIP] Windows Open With Support

CrossR:
Partial fix for #598.

This implements the Windows side of this, I'm not sure if other platforms have a similar issue.
Does this by adding the needed registry keys, and updating the opening of files to work for the new argument list.

Also added a fix for #344 whilst I was in this area.

I'm not especially a JS dev, so I need some advice on cleaning up the `BuildSetupTemplate.js`. Its functionally working, but its not the neatest code.
Similarly for the way I've changed main, it works but its not the nicest.

Oh, and what file types should we support right now? So far I've added text and JavaScript/TypeScript files, but there is obviously a tonne more we could add.

### DIFF
--- a/build/BuildSetupTemplate.js
+++ b/build/BuildSetupTemplate.js
@@ -44,11 +44,11 @@ Root: HKCU; Subkey: "Environment"; ValueType: expandsz; ValueName: "Path"; Value
 function replaceForRegKeys(ext, desc) {
 
     const addFileEditor = `
-Root: HKCR; Subkey: "${ext}\\OpenWithProgids"; ValueType: none; ValueName: "${prodName}.exe"; Flags: deletevalue uninsdeletevalue; Tasks: registerAsEditor;
-Root: HKCR; Subkey: "${ext}\\OpenWithProgids"; ValueType: string; ValueName: "${prodName}.exe"; ValueData: ""; Flags: uninsdeletevalue; Tasks: registerAsEditor;
-Root: HKCR; Subkey: "${prodName}.exe\\${ext}"; ValueType: string; ValueName: ""; ValueData: "${desc}"; Flags: uninsdeletekey; Tasks: registerAsEditor;
-Root: HKCR; Subkey: "${prodName}.exe\\${ext}\\DefaultIcon"; ValueType: string; ValueName: ""; ValueData: "{app}\\images\\oni.ico"; Tasks: registerAsEditor;
-Root: HKCR; Subkey: "${prodName}.exe\\${ext}\\shell\\open\\command"; ValueType: string; ValueName: ""; ValueData: """{app}\\${prodName}.exe"" ""%1"""; Tasks: registerAsEditor;
+Root: HKCR; Subkey: "${ext}\\OpenWithProgids"; ValueType: none; ValueName: "${prodName}"; Flags: deletevalue uninsdeletevalue; Tasks: registerAsEditor;
+Root: HKCR; Subkey: "${ext}\\OpenWithProgids"; ValueType: string; ValueName: "${prodName}${ext}"; ValueData: ""; Flags: uninsdeletevalue; Tasks: registerAsEditor;
+Root: HKCR; Subkey: "${prodName}${ext}"; ValueType: string; ValueName: ""; ValueData: "${desc}"; Flags: uninsdeletekey; Tasks: registerAsEditor;
+Root: HKCR; Subkey: "${prodName}${ext}\\DefaultIcon"; ValueType: string; ValueName: ""; ValueData: "{app}\\images\\oni.ico"; Tasks: registerAsEditor;
+Root: HKCR; Subkey: "${prodName}${ext}\\shell\\open\\command"; ValueType: string; ValueName: ""; ValueData: """{app}\\${prodName}.exe"" ""%1"""; Tasks: registerAsEditor;
 {{RegistryKey}}
 `
     shelljs.sed("-i", "{{RegistryKey}}", addFileEditor, destFile)
--- a/build/BuildSetupTemplate.js
+++ b/build/BuildSetupTemplate.js
@@ -39,7 +39,7 @@ const fileExtensions = {
 }
 
 const addToEnv = `
-Root: HKCU; Subkey: "Environment"; ValueType: expandsz; ValueName: "Path"; ValueData: "{olddata};{app}\\"; Tasks: addtopath; Check: NeedsAddPath(ExpandConstant('{app}\\bin'))`
+Root: HKCU; Subkey: "Environment"; ValueType: expandsz; ValueName: "Path"; ValueData: "{olddata};{app}"; Tasks: addtopath; Check: NeedsAddPath(ExpandConstant('{app}'))`
 
 function replaceForRegKeys(ext, desc) {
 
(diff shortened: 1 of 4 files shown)


CrossR:
Also I need to work out how to get the desktop icon working for this.
The icon is there since its working for .vim files and the app itself, it must just be the link to the icon that I've messed up.

extr0py:
Wow, this is awesome! Thanks for putting this together, @CrossR !

I was just trying this on Windows and it's really cool to have the shell integration hooked up. I could see it improving my workflow on Windows for sure.

> I'm not especially a JS dev, so I need some advice on cleaning up the BuildSetupTemplate.js. Its functionally working, but its not the neatest code.

I don't really see any problems with it. As long as it is still relatively simple and easy to understand what is going on, that should be fine. The only piece that confused me a bit was that we were re-injecting the {{RegistryKey}} placeholder each time we replaced it, and then file add the closing line. We could potentially build up that registry key section in its entirety, and then just replace {{RegistryKey}} placeholder with the whole section. Not a requirement though to change, just thinking out loud - I'm fine with it as it is.

I really like that you pulled out the file extensions into a separate object - making it easy to configure this later.

> Oh, and what file types should we support right now?

Just a few off additional ones off the top of my head:
- `.md`: `Markdown Files`
- `.json`: `JSON Files`
- `.jsx`: `JavaScript Files`
- `.tsx`: `TypeScript Files`

I'm sure we'll keeping adding more as we go :)

> Also I need to work out how to get the desktop icon working for this.

It looks like the problem here is this setting:
```
Root: HKCR; Subkey: "${prodName}${ext}\\DefaultIcon"; ValueType: string; ValueName: ""; ValueData: "{app}\\images\\oni.ico"; Tasks: registerAsEditor;
```

The image path is a bit more nested in our install directory - it looks like it should be this instead:
```
{app}\resources\\app\\images\\oni.ico
```

> Similarly for the way I've changed main, it works but its not the nicest.

Unfortunately this change wasn't working for me. It might not hit that `app.on('ready')` path in all cases.

There's another place we do the slice of the arguments:
```
    const shouldQuit = app.makeSingleInstance((commandLine, workingDirectory) => {
        createWindow(commandLine.slice(2), workingDirectory)
    })
```

So I think we should pull your code to a helper method - like `removeExecutableFromArguments`, and that can do the logic that you added - and then we'd use it in both places. The logic looks reasonable to me, the only thing we might want to change is the `process.argv.length === 2` condition - I wonder there are cases where we could get multiple files?

Hope that helps. Nice work - thanks again for putting this together! It really makes Oni feel more like it 'belongs' on Windows and isn't just an afterthought.

extr0py:
I merged the `package.json` for you - it was a conflict with my change to turn off the asar packing. The `oni.ico` path will still need to be updated in the script.

CrossR:
Got the image working now, it was just because I was using the path for the source code, rather than the installer location. It will probably need changing quite soon, as I assume the folder will move once the asar stuff is added back?

Got those other file types added, and tested that its easy enough to manually add more if a user wanted via the Windows interface.

Does look very nice just having the icons all sat there!
Directory Image

Just need to work out why I can't seem to open files once you use the Open With loading...
It gives an error of "No matching Autocommands", but works fine if I open Oni normally.

extr0py:
> It will probably need changing quite soon, as I assume the folder will move once the asar stuff is added back?

I think it will actually take some time to bring back the 'asar' packaging - unfortunately the behavior of the out-of-the-box asar packager actually increases the package size! So it will take some investigation and perhaps a custom packing strategy to make things work. I added a note in #350 to remind me to update the icon when we bring that back.

> Does look very nice just having the icons all sat there!

Ya, that's awesome! 👍 👍 

> Just need to work out why I can't seem to open files once you use the Open With loading...
It gives an error of "No matching Autocommands", but works fine if I open Oni normally.

Interesting - I'll see if I can repro this in a bit.

CrossR:
Looks to be something to do with the way the first file is opened, which is via the function I wrote...

If I try and open one file it fails. If I try and open 4 files, the first fails and the three after are fine.

extr0py:
> If I try and open one file it fails. If I try and open 4 files, the first fails and the three after are fine.

It seems like it's working for me now, after trying it (rebuilding the installer and everything):
- If I select _multiple files in explorer_ and Open, it actually opens 3 instances of Oni. It's kind of weird but at least all the files are opened (and it might be just how explorer behaviors)
- If I open _multiple files from the command line_, they all get opened, but they aren't visible in the "buffer line" until I start cycling through them (and the title says '1 of N').

Opening a single file seems to work fine for me - both from explorer and the command prompt.

Sounds like I'm seeing different behavior than you, though! Trying to figure out why..

A couple questions:
- By fails, do you mean the file doesn't load at all?
- Are you opening from explorer or the command line?

I'm excited to get this change in, it's so cool to see the icons hanging out in explorer now 😄

CrossR:
Hmm I'll have a look into those two cases, I'm not sure. I don't know if Explorer calls a single command for each file (which it looks like its doing if there is multiple instances of Oni) or if it passes an array of items.

Might be the sort of thing that gets sorted if more complex tab management is implemented (ie dragging a tab out of the Oni tab bar to make a new instance of Oni etc). The second case just sounds like something isn't fully getting updated.

> By fails, do you mean the file doesn't load at all?
Are you opening from explorer or the command line?

Yep, the file doesn't load at all.
I'm opening a file via the explorer, the issue comes if I then open a second file in Oni. Ie Double click file > Oni Opens with File > Menu > Open File > Select File > Doesn't Open. The first file loads fine, the second doesn't load at all.

extr0py:
> Yep, the file doesn't load at all.
I'm opening a file via the explorer, the issue comes if I then open a second file in Oni. Ie Double click file > Oni Opens with File > Menu > Open File > Select File > Doesn't Open. The first file loads fine, the second doesn't load at all.

Got it, thanks for the clarification, I can hit the same thing now. I'll try and look at it a bit tomorrow morning.

extr0py:
Ah, I just realized this is due to missing a space after `:tabnew` in the work I did to reconcile Open Files... So not related to the installer / arguments at all.

This line:
```
executeVimCommandForMultipleFiles(":tabnew", files)
```

Needs to be:
```
executeVimCommandForMultipleFiles(":tabnew ", files)
```

Sorry about that, not sure how I missed that. I'll send out a PR for the fix. I found I could reproduce this without installing, by opening a file, and then opening multiple files.

Besides that, I think this is good to go - was there anything else you wanted to look at, @CrossR ? I'm planning on pushing out a new release this weekend and it would be great to have this in there!

extr0py:
Opened #621 to track fixing that `:tabnew ` vs `:tabnew` regression.

CrossR:
There is two other parts to clean up, but I think they can be moved into a separate issue, as I think this would be useful to get into the next release.

* Opening multiple files into a single instance of Oni if multiple files are highlighted.
* Fixing the command line opening so it correctly opens multiple files into individual visualised tabs.

Do you want me to merge the `:tabnew` change into this branch or should I leave them separate?

extr0py:
> There is two other parts to clean up, but I think they can be moved into a separate issue, as I think this would be useful to get into the next release.

Sounds good to me! These both sound like incremental issues, given that neither scenario would've been enabled prior to this change.

> Do you want me to merge the :tabnew change into this branch or should I leave them separate?

Nah, that's okay - I merged it in with #621 . I'll go ahead and merge this in now (so we'll have both fixes in master).

extr0py:
@CrossR - merged! This is a huge improvement for the Oni experience on Windows. I'm really glad we have this functionality now - thanks for your contribution!

extr0py:
FYI, @CrossR , as a thank you, I put a small bounty on this issue:
https://www.bountysource.com/issues/43751451-windows-installer-add-option-to-add-oni-to-path

So make sure to go claim that 😄

CrossR:
Cheers! Thanks for letting me help out on a great project!

PR #615, opened 2017-08-13
Make left tab corner clickable as well

extr0py:
This makes the whole tab clickable, as opposed to just the name on the tab. Thanks for the idea, @cyansprite !


cyansprite:
👍

Issue #610, opened 2017-08-12
Scriptability of Oni


bert88sta:
How scriptable is Oni right now? I know that we have an existing plugin infrastructure, but if I miss the whole "Just add it to your .vimrc" mentality for small things

extr0py:
Great question. This is something I've been thinking about.

One thing I'd like to do is treat the `Config.js` as a mini-plugin, and have an `activate` method for it - so you can script against the `Oni` object.

You can also open the developer tools and make calls against the `Oni` object (which is the same thing the plugins get), but it's not super interesting - it'll be moreso once there are things like keybindings hooked up... 

Is there a particular sort of customization you were thinking about, that wasn't easy to do without rebuilding Oni?

bert88sta:
Maybe a dumb example, but hypothetically I might want to change the
background color of the whole status bar based on mode. I don't actually,
but it seems like something I *should* be able to do. I don't know that
it's going to be feasible in JS and electron, but basically any state that
isn't critical to the editor's correct functionality should be available to
the user

extr0py:
Not a dumb example at all, this is great, and is exactly what I'd like to support.

The way I'm thinking about this would be, in `Config.js`, we'd expose an activate method:
```
const activate = (Oni) => {
     
     Oni.on("mode-changed", (newMode) => {
          const newColor = newMode === "insert" ? "black" : "white"
           Oni.configuration.setValue("oni.colors", {
                 "statusbar.background": newColor
           })
     })
}

module.exports = {
   activate,

  //add custom config here, such as
  //"oni.useDefaultConfig": true,
  "editor.fontSize": "12px",
  "editor.fontFamily": "Fira Code",
  "oni.bookmarks": ["C:/Toronto2"],
  "oni.loadInitVim": true,
  "statusbar.enabled": true,
  "editor.scrollBar.visible": true,
  "editor.cursorLine": true,
  "editor.cursorLineOpacity": 0.1,
...
```

Something like that... It's a bit awkward having the function in the same space as the configuration, but I'd like to use 'activate' and 'deactivate' as reserved capabilities here. 

The setting of color properties ties into some of the thoughts @cyansprite had in #604 

Ideally, it should be really lightweight... in that I can save and see changes immediately, as well as be very straightforward - it should have premier autocompletion / quickinfo / etc.

extr0py:
[...]

Issue #604, opened 2017-08-12
Status bar colors not really an issue... more of enhancement/discussion


cyansprite:
I was playing around tonight... I'm pretty picky when it comes to colors.

<a href="http://imgur.com/MgS5vjb"><img src="http://i.imgur.com/MgS5vjb.gif" title="source: imgur.com" /></a>

achieved by (oni statusbar core), but my concerned is... should we have a color file, or should it just be part of the config file? It's what I ran into while playing around with the keybindings...

```js
        const getBackgroundColorForMode = (m) => {
            switch (m) {
                case "insert":
                    return Oni.configuration.getValue("statusbar.insertBackgroundColor", "rgb(0,200,100)")
                case "visual":
                    return Oni.configuration.getValue("statusbar.visualBackgroundColor", "rgb(0,100,255)")
                case "replace":
                    return Oni.configuration.getValue("statusbar.replaceBackgroundColor", "rgb(0,200,100)")
                case "operator":
                    return Oni.configuration.getValue("statusbar.operatorBackgroundColor", "rgb(255,100,0)")
                default:
                    return Oni.configuration.getValue("statusbar.normalBackgroundColor", "rgb(0,100,255)")
            }
        }
        const getForegroundColorForMode = (m) => {
            switch (m) {
                case "insert":
                    return Oni.configuration.getValue("statusbar.insertForegroundColor", "rgb(220, 220, 220)")
                case "visual":
                    return Oni.configuration.getValue("statusbar.visualForegroundColor", "rgb(220, 220, 220)")
                case "replace":
                    return Oni.configuration.getValue("statusbar.replaceForegroundColor", "rgb(220, 220, 220)")
                case "operator":
                    return Oni.configuration.getValue("statusbar.operatorForegroundColor", "rgb(220, 220, 220)")
                default:
                    return Oni.configuration.getValue("statusbar.normalForegroundColor", "rgb(220, 220, 220)")
            }
        }

```

CrossR:
I remember poking around the tickets and seeing this: https://github.com/extr0py/oni/issues/33

Obviously no idea if it changed since then, but it looks like goal at one point was to have a separate file for it.

I think that makes more sense, as potentially you'd want to share colourschemes etc, without having to remove your config from it too.

cyansprite:
Yeah, I agree it should be a separate file...
I'm not sure if it should be a .js or .tmTheme, or .json, or .less...
[...]

Issue #567, opened 2017-08-04
add c-j and c-k to quick info?


cyansprite:
### Can we/I add c-j and c-k so I can navigate using these?
<i>Might also be advantageous to add up and down for new users</i>
Hell... we could make this a setting too?
```ts 
else if (key === "<C-n>" || key === "<C-j>" || key === "<down>") {
                    UI.Actions.nextMenuItem()
                } else if (key === "<C-p>" || key === "<C-k>" || key === "<up>") {
                    UI.Actions.previousMenuItem()
```}

Also c-w to delete prev word, c-u to delete entire line etc...

bert88sta:
We're working on a way currently to make keybindings much easier, but it's going to take a while.

Bear with us :)

extr0py:
Definitely! There are a few open issues in terms of key bindings:
#20 , #366, #404 

But the case where the keybindings go through Oni (and don't round-trip through Neovim) is actually the simplest.

Exposing this as a setting would be a good solution in the interim while that's being worked out. Perhaps something like `keybindings.nextMenuItem` and `keybindings.previousMenuItem` would be reasonable for now? I definitely see these hardcoded keybindings as being an annoyance, so having a short-term solution that we can iterate on seems reasonable to me.

---

Just to give some ideas of what I'm thinking in terms of input, longer-term, is something a bit more general, like for this case:
`Oni.map("<C-n>", "menu.nextItem")`

But also be able to map to arbitrary functions:
`Oni.map("<C-Left>", () => Oni.windows.focusLeft())`

And access functions exposed by plugins:
`Oni.map("<C-Enter>", () => Oni.plugins.orgModeForOni.toggleTodo())`

And give some helpers to make doing transformations easier, something like:
(this is a pointless example,
`Oni.map("<C-J>", Oni.helpers.transformLine((line) => line.toLowerCase())`

A few other considerations I'm thinking about:
- Exposing Oni commands to Neovim (and vice-versa) - the interop between Oni/Neovim is the most interesting piece (#20 touches on this)
- Handling commands for non-Vim/Neovim features (#366)
- Handled chorded commands in Oni
- Handling different modes / filters for commands
- How these are specified in existing `config.js`

[...]

PR #617, opened 2017-08-14
Keybindings

cyansprite:
# First keybindings commit, will document later, opening PR for early reviews :)

starts fixing #567 , #20 , #366, #404

and abstraction for

#604 #33 #610


cyansprite:
I need to clean this up -.-

PR #618, opened 2017-08-15
Config Abstraction/Keybindings Birth WIP

cyansprite:
## Abstraction

We have the need for config other than just user config... such as keybindings, and colors. So I abstracted it, and now it should be easily manipulable for adding new kinds of configs.

- Allows for keybindings and config not to get muddied up.
- Will allow for oni.color not to be muddied with config/keybindings.

## Keybindings  #567 #20 #366 #404

Allow the user to select Onis keybindings

- You provide an array of strings, these strings will then test if you hit them with it's string alias/key, and if any of those are hit, fire the action.

### TODO : 
- I'm not sure if keybindings reload if you change the file...
- Ensure keybindings work with plugins.
- Make sure keybindings files work
- Add tasks action
- Document
- Add tab keybindings ([b ]b I think this is in vim somewhere I didn't look for it.)
- Check if keys are duplicate? I never had issues with this in vim... so it's up for debate, I always figured if I overrode a key and didn't notice I didn't need it anyway... and if I did I would realize I DID use that key... but some may see it differently. 

### Other issues this will affect:
- Make sure the abstraction will work well for oni specific colors. (status bar, foreground, background, blah blah)  #604 #33
- Make a nice activate/binding method for config #610 #619 

##### Example
```ts        
// Opened menu keys
        "keybindings.openedMenu.nextMenuItem"         : ["<C-n>"    , "<C-j>"  , "<down>"] ,
        "keybindings.openedMenu.previousMenuItem"     : ["<C-p>"    , "<C-k>"  , "<up>"]   ,
```


cyansprite:
Last gif showing what was accomplished.
You can tell I don't use my mouse much when it comes time to open a folder lol...

<a href="http://imgur.com/lsnUVrs"></a>

extr0py:
@cyansprite , thanks for putting together this PR and the thought you've put into the keybindings!

I like the abstraction you put together for the configuration - it's great to separate the loading / watching logic from all the values.

The model being proposed for the keybindings seems like it addresses the short-term problem of not being able to map keys - but what I want for the keybindings is to be able to bind to `commands` (specified by `Commands.ts` right now, _as well as functions_. 

Also, I'd like for our keybinding-story to be able to address all the switching / nesting / hardcoding in the `keydown` handler in `NeovimEditor.tsx`. This change gets rid of the key hardcoding, but we still need all the special cases for different pieces of UI being opened.

Ideally, this whole block could be replaced by something like:

```
const keyboard = new Keyboard()
keyboard.on("keydown", (key: string) => {
    const wasKeyHandled = inputManager.handleKey(key)

    // If the key couldn't be handled in the Oni layer, pass it to Neovim
    if (!wasKeyHandled) {
         this._neovimInstance.input(key);
    }
})
```

So there's really 4 things I'd like to see from the keymapping work:
- __Configurable__ - ability to specify via some configuration (either config.js or keybindings.js).
- __Multiple mappings per key__ - We should map _key_ -> _action_ (whether it's a command string, or a function). We might have multiple mappings for the same key, that would be dependent on the current mode, for example. This couldn't be expressed in the current model.
- __Handling commands AND functions__ - I want to handle both lambdas as well as existing commands defined in `Commands.ts`. It might mean that we have to port over some of the functionality, like `menu next`, `menu previous`, etc to commands, and have a way for the UI elements to listen for these commands.  I believe that we could reuse the commands registered as opposed to defining new settings.
- __Generalize all the special-casing in `NeovimEditor.tsx`__ - Some commands will have different behavior depending on the edit mode (`normal`, `insert`, `visual`, etc), or based on UI that is available (like, 'is a menu visible'). We should have a way to handle those cases. If we address that, than the key handling in `NeovimEditor` can just defer that to the input manager to resolve. This will guarantee that our key-handling is flexible enough to support new and novel elements.

The concern I have is that, if these 4 bullet points aren't meant, and we introduce a short-term paradigm, it will actually incur complexity / debt - because now we have to maintain backwards compatibility and also help our users to figure out two different paradigms. So my preference would be to have a holistic solution, as opposed to a short-term solution. The proposed solution solves the first bullet point, but not the subsequent 3. 

I'd like to see something that addresses all 4 of the points above.

I'll try and see if I can get time this weekend to put something more concrete together in terms of what I'm thinking, since some of the thoughts above might be too vague to be actionable.  I know the  lack-of-configurability with the input is a huge pain point right now, so I'm definitely on-board with solving it!

cyansprite:
> Ideally, this whole block could be replaced by something like:
```ts const keyboard = new Keyboard()
keyboard.on("keydown", (key: string) => {
    const wasKeyHandled = inputManager.handleKey(key)

    // If the key couldn't be handled in the Oni layer, pass it to Neovim
    if (!wasKeyHandled) {
         this._neovimInstance.input(key);
    }
})
```

I like the above example.

> Generalize all the special-casing in NeovimEditor.tsx - Some commands will have different behavior depending on the edit mode (normal, insert, visual, etc), or based on UI that is available (like, 'is a menu visible'). We should have a way to handle those cases. If we address that, than the key handling in NeovimEditor can just defer that to the input manager to resolve. This will guarantee that our key-handling is flexible enough to support new and novel elements.

```ts
// temp solution
"keybindings.openedMenu.nextMenuItem"
// more specific solution
{
   "keybindings.openedMenu": {
         "nextMenuItem": ["C-n", "c-j", "<down>"]
     } 
}
```

Also noticed i used c-j instead of C-n, we need to convert all c to C so we can still handle them because if it's c we ignore, and the user doesn't care about case or at least I don't want to.

> The concern I have is that, if these 4 bullet points aren't meant, and we introduce a short-term paradigm, it will actually incur complexity / debt - because now we have to maintain backwards compatibility and also help our users to figure out two different paradigms. So my preference would be to have a holistic solution, as opposed to a short-term solution. The proposed solution solves the first bullet point, but not the subsequent 3.

There is no way I would want to merge this right now, I just wanted to put a base review so we could discuss it much like you did :).

I'm kind of new to github and I'm not sure exactly how we should discuss things... so I open prs and issues... is there a better way of doing it?

cyansprite:
Or we could do plugin the actual command or lambda

extr0py:
> There is no way I would want to merge this right now, I just wanted to put a base review so we could discuss it much like you did :).

> I'm kind of new to github and I'm not sure exactly how we should discuss things... so I open prs and issues... is there a better way of doing it?

Oh yes, this is great! I don't think there is a better way to start a discussion than with some code 😄  It gets us out of talking in the abstract and into the real meat of the issue.

Sorry if it came across as negative / critical - didn't mean it like that at all. I appreciate the time, thought, and expertise you put in (after all, you're taking time out of your day to contribute!). So I try and get concerns out early, out of respect for your time! But ya I still feel relatively new to this too, so lots of learning for me - I'm still finding the right balance there.  😄 Thanks for being open to the considerations!

> If you could outline the specifics of what we need and build tasks off of a story that way you could define exactly what to complete would be great. I have a general idea and your bullets are a GREAT start.

Sounds good! I know it's vague right now but I wanted to at least set up what the "north star" looks like with those bullet points. I'll post up some more concrete ideas soon.

Thanks again for thinking about this, and getting the conversation started. It will be a huge step forward for this project to have this solved 👍

cyansprite:
> Sorry if it came across as negative / critical - didn't mean it like that at all. I appreciate the time, thought, and energy you put in (after all, you're taking time out of your day to contribute!). So I try and get concerns out early, out of respect for your time! But ya I still feel relatively new to this too, so lots of learning for me - I'm still finding the right balance there.  :smile: Thanks for being open to the considerations!

It wasn't negative at all, thanks for the feedback, I just want to make sure I'm doing the best I can.
The quick feedback is always appreciated, I would love to help Oni get to release 1.0 as quick, but as good/correct as possible. Not only for my own editing but I want to make an text editor/IDE I believe in.

> Sounds good! I know it's vague right now but I wanted to at least set up what the "north star" looks like with those bullet points. I'll post up some more concrete ideas soon.

Less that it's vague more of I want to manage it properly, maybe a check list?

extr0py:
> The quick feedback is always appreciated, I would love to help Oni get to release 1.0 as quick, but as good/correct as possible. Not only for my own editing but I want to make an text editor/IDE I believe in.

Cheers to that! Really appreciate your help.

> Less that it's vague more of I want to manage it properly, maybe a check list?

Ah ya, github has a check list feature, I just always forget to use it.. That's a good idea - I'll work on breaking down the thoughts I have into an actionable list.

cyansprite:
> Ah ya, github has a check list feature, I just always forget to use it.. That's a good idea - I'll work on breaking down the thoughts I have into an actionable list.

That would be nice but also thinking of something like trello.

cyansprite:
Putting this on hold until #633 is processed.

bryphe:
Looks like this is no longer applicable since #633 is in, and we now have colors customization with #950 -  so I'll close it out. Thanks for the thoughts around this!

PR #619, opened 2017-08-15
WIP for #610 - Provide Oni API object via 'activate' to the Config

extr0py:
This change adds an `activate` and `deactivate` method to the configuration, as follows:

- If `activate` is present on the Config, it will be called when the config is loaded, with the Oni API object passed in (the same object that goes to plugins). So the `activate` method of the config is essentially a mini-plugin - meaning you can do scripting / testing there, and port it over to a plugin eventually.
- If `deactivate` is present, it will be called when the script is unloaded. Right now, this would only be when the configuration is updated, but maybe that could be enhanced for per-project or per-folder configs.

This also updates the 'default' config with samples.

Here's an example:
config-activation

Some things we'll need later:
- Better handling of errors. If there is a script error, it will fail silently (although there will be a message in the dev tools).
- Better documentation of API surface - so people know what they can do with their config :)
- Premier intellisense - editing the config should have first-class intellisense!
- Unsubscription of event listeners... In the sample I showed, the deactivate doesn't automatically clear out the 'Oni.on("mode-change")...' event, which is problematic for reloading, because we keep piling listeners on top..


extr0py:
Related to #619 and #618

extr0py:
This isn't very useful at the moment, since there is not much relevant that can be done with the Config.js object, but I think it makes sense to go in now - so that as we expand the API, there actually _are_ interesting things that can be done with the config..

extr0py:
This will be a good incremental launching point for some of the input work / prototyping, so I'll go ahead and bring it in.

PR #620, opened 2017-08-16
WIP - #434 - Clipboard integration

extr0py:
This is a WIP addressing #434 - adding some system-level clipboard integration to Oni.

This does the following:
- Adds an `editor.clipboard.enabled` setting, which, when set to true, enables the system-integrated clipboard functionality
- If enabled, adds a handler for `onYank` that pushes the contents to the clipboard. In the future, we might want to specify or filter certain types of gestures (ie, 'deletes' vs 'yanks'), but that could come as a later setting.
- If enabled, in visual mode, <C-c> or <Cmd-c> (mac) will copy the selection to the system clipboard.
- If enabled, in insert mode, <C-v> or <Cmd-v> (mac) will paste the current clipboard items inline.

This is an initial proposal at 'bridging the gap' between what users expect from other apps in their environment, and from the vim-world of yank.


browser/src/Editor/NeovimEditor.tsx
```diff
@@ -256,6 +262,12 @@ export class NeovimEditor implements IEditor {
                 }
             }
 
+            if (key === "<C-v>" && this._screen.mode === "insert") {
```
cyansprite: should we make sure they have the config enabled first?
This would override my mapping I have in vim even if I say
```ts
"editor.clipboard.enabled": false,
```

cyansprite:
How about a fuzzy find Quickopen for clipboard history? 

As well as an autocompletion for clipboard, my plugin I was playing around with I was using s after pasting, it worked really nice, especially since I don't think I have EVER used s in vim...
However I know a lot of people use s for sneak and easymotion.

(Could choose which registers, would this be native or a plugin.)

browser/src/Editor/NeovimEditor.tsx
extr0py: Yes, good point, thanks! I just added a check for it around the keys that we are strong-arming into clipboard behavior when this setting is enabled (so disabling it should actually work now). This will be an interesting case for us to untangle with the input binding work, as:
- These should be applied as defaults (prior to loading the keybindings), if the `editor.clipboard.enabled` setting is true.
- There will be conditional applications based on the platform (Windows vs OSX vs Linux) - not too different from how we handle default configuration values today, though.
- They will need to be filtered / enabled based on the mode, and either execute an Oni-side command (`editor.clipboard.paste`), or a new set of keystrokes to Neovim (`y`).

extr0py:
> How about a fuzzy find Quickopen for clipboard history?

Sounds like a cool idea! It'd be interesting to see the set of registers too. I always forgot what are in my registers so I end up using just registers `a` and `b` 😄 

> As well as an autocompletion for clipboard, my plugin I was playing around with I was using s after pasting.

I'm thinking about a way to navigate some of the 'enhanced UI' elements without leaving the keyboard. That's kind of tracked in my mind here: #366. I was thinking of potentially using the `s` key too as a default (maybe `<Alt-s>`). But that's sort of a tangent, I'm still trying to figure out entry points for some of these new aspects...

cyansprite:
> Sounds like a cool idea! It'd be interesting to see the set of registers too. I always forgot what are in my registers so I end up using just registers a and b 😄

I will try to make some progress today on colors buffers and registers for QO.

Also do you care to test out my yank cycler plugin when its done? I can send a github link on here to the repo when I'm done.

And yes I know... It's a bit mind boggling sometimes:p.

extr0py:
> Also do you care to test out my yank cycler plugin when its done? I can send a github link on here to the repo when I'm done.

Definitely!

extr0py:
FYI @cyansprite - I just brought this in. Sounds like you'll want to update your `editor.clipboard.enabled` setting to `false` so it doesn't mess up your customizations.

PR #622, opened 2017-08-16
WIP / Prototype: File Explorer window

bryphe:
This is a first-round exploration of having Oni start 'multiplexing' Neovim instances. In this case, it would create a _separate editor window_ with _its own Neovim instance_. This gives us the opportunity to do interesting things, like have separate colorschemes or tabs in window.

In this case, the file explorer would have its own init.vim, and later, a custom renderer that will show icons / files / source control status with a higher fidelity. It will still load the user's init.vim if available, but this means we can start exploring some of the interesting graphical integrations (like layers) in #362 .

There is still _a lot_ of work to do, but this starts refactoring out some of the common hooks Oni puts into the NeovimInstance, and generalize/separates some of the global functionality from some of the editor-specific functionality. The code today makes several assumptions that there is only one NeovimInstance around, so this will start breaking those assumptions.

The goal with this is to bring in a first-class file explorer experience, with the keyboard interactions behaving as a user would expect from the neovim-world, and incrementally step into multiplexing processes.


PLAN.md
```diff
@@ -0,0 +1,52 @@
+- Create own directory plugin?
```
cyansprite: I would prefer our own...
netrw is also hard to use

PLAN.md
```diff
@@ -0,0 +1,52 @@
...
+            - WindowManager - multiple Neovim editor instances
+
+- Window focus management
+    - Create Window Manager
+    - Have focus managed
+    - Pull keyboard out externally, have focused window handle input
+        - Keyboard focus for active window
+    - Create `deactivate`, `activate` concepts for editor
+    - Push `onKeyDown` to editor
+    - How to keep from 'drilling in' to file?
```
cyansprite: ?

PLAN.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Rename scenario?
+    - Lightweight editor
+
+- Custom renderer
+    - Custom renderer
+
+- Share common code between SimpleEditor / OniEditor
+
+- Separate 'IDE' UI functionality from 'Core' functionality
```
cyansprite: 👍

browser/src/Editor/SimpleNeovimEditor.tsx
```diff
@@ -0,0 +1,120 @@
...
+    private _screen: NeovimScreen
+
+    private _pendingTimeout: any = null
+    private _pendingAnimationFrame: boolean = false
+
+    constructor(
+        private _config: Config.Config = Config.instance(),
+    ) {
+        // TODO: How to get rid of this?
+        this._neovimInstance = new NeovimInstance(new DummyPluginManager(), 100, 100)
```
cyansprite: which part?

PLAN.md
```diff
@@ -0,0 +1,52 @@
...
+- Window focus management
+    - Create Window Manager
+    - Have focus managed
+    - Pull keyboard out externally, have focused window handle input
+        - Keyboard focus for active window
+    - Create `deactivate`, `activate` concepts for editor
+    - Push `onKeyDown` to editor
+    - How to keep from 'drilling in' to file?
+
+- Rename scenario?
```
cyansprite: Will the lightweight/simple editor always be used for widgets?

browser/src/Editor/SimpleNeovimEditor.tsx
```diff
@@ -0,0 +1,120 @@
...
+            this._screen.dispatch(action)
+
+            this._scheduleRender()
+
+            if (!this._pendingTimeout) {
+                this._pendingTimeout = setTimeout(() => this._onUpdate(), 0)
+            }
+        })
+
+        const keyboard = new Keyboard()
```
cyansprite: I'm assuming when you say extract keyboard you mean have detached from an editor and whichever editor has focus will receive the events right?

Sounds like a nice polymorphic opportunity for editors.

The difference would be instead of an interface have an abstract, and then have a static instance/private instance for _currentEditor, and that way you don't have to change who keyboard is sending to because it can just be _currentEditor.sendKeys() and then _currentEditor can process it based on it's overrided method but can also pull common code from AbstractEditor, thoughts?

PLAN.md
extr0py: Agreed!

PLAN.md
extr0py: Sorry, some of these notes are really raw!

The 'drilling in' to the file was, when I was testing, I'd navigate a folder structure and open a file. The file would open _in the file explorer pane_ which isn't ideal. I think if we have our 'own' directory plugin, instead of netrw, we can handle this.

But basically there'd need to be a pipeline where the file-explorer-neovim handles the <enter> key, pushes it up to Oni (or maybe Oni is just using our Input work to watch the enter key), and then it would communicate to the main neovim instance to open the file. Just need to think about that flow a bit.

browser/src/Editor/SimpleNeovimEditor.tsx
extr0py: > I'm assuming when you say extract keyboard you mean have detached from an editor and whichever editor has focus will receive the events right?

That's exactly right!

> The difference would be instead of an interface have an abstract, and then have a static instance/private instance for _currentEditor, and that way you don't have to change who keyboard is sending to because it can just be _currentEditor.sendKeys() and then _currentEditor can process it based on it's overrided method but can also pull common code from AbstractEditor, thoughts?

Yes, I think that's a really good idea. It actually lines up a lot with what I've been thinking for an API. I'd like for plugins to know who the activeEditor is - something like:
`Oni.editors.activeEditor`

As well as knowing when the editor changes - something like:
`Oni.editors.onActiveEditorChanged((newEditor, oldEditor) => { ... })`

But I don't want the plugins to need to always be listening for active editor changed - I might just have a plugin that logs to the console when I change mode, like:
```
const activate = (Oni) => {
Oni.editors.activeEditor.onModeChanged((newMode) => console.log("New mode: " + newMode);
}
```

And I'd want to have that work across the currently active editor, regardless of which one it is. Sounds like a similiar case to input - I might want to send input to the active editor, and not care about the implementation.

So I think it's a great idea, thanks for sharing! I'd picture refactoring this a bit and have a common core editor class as you mention (with some of the shared functionality)

PLAN.md
cyansprite: I agree with you, if I open a folder inside of the folder pane, I don't want it to open in the folder pane, that would be a pain.

I would want it to open in the "active editor", or split if I use c-v/c-s starting from the most far right/far bottom (some users like left and top, in fact vim's standard is left and and top which I never understood but... we could check this value)

PLAN.md
cyansprite: This all sounds wonderful.
You say keep the keybindings from vim, which is a great idea.

And the mappings we could dynamically set the light weight windows to the mappings one would expect.
If I open a browser window, I want to use hjkl to navigate inside of them, if someone has mapped those to jkl;, then it should work for them as well, but do I expect other functionality like splits will open a new browser below?
To what extent will we handle vims mappings? Same with file explorers, it wouldn't be hard if you are saying use autocmds, just where do we draw the line?

browser/src/Editor/SimpleNeovimEditor.tsx
extr0py: Oops, I meant the `DummyPluginManager`. It'd be nice to not need to send in this object at all - rethinking the relationship between NeovimInstance <-> PluginManager and revisiting that dependency.

browser/src/Editor/SimpleNeovimEditor.tsx
cyansprite: having our own events that plugins can listen to sound great.
above with the other editors, we can have on load event with what kind of editor abstraction it is.

Also if we change the filetype of it to be something specific, we can have autocmds in vim for users to interact as well if they have something they would like to do in vim when a folder opens.

Also I know a lot of people have the need/want to open the folder pane on load to whichever side they want on open, this is something we need to think about as well.

Thanks for listening to my ideas, Oni just keeps more and more in line with what I was thinking for an editor.

extr0py: Yes, good points!

> above with the other editors, we can have on load event with what kind of editor abstraction it is.

This is interesting, and something I'm still trying to reason about - there are specific things we can do with a neovim-editor (like send it normal mode commands, execute VimL, etc), but how do plugins reason about this vs a non-neovim-backed editor? I like the idea of having a load event or some way to know the capabilities that the editor has.

> Thanks for listening to my ideas, Oni just keeps more and more in line with what I was thinking for an editor.

Great to hear :)

browser/src/Editor/SimpleNeovimEditor.tsx
cyansprite: Yeah I agree, there a few ways to deal with it... I wonder. Do we want separate neovim instances per multiplex? We could have just one and manage it that way, this might be cleaner and since there are a lot of functions already provided by neovim, we wouldn't have to make our own logic there.

Furthering this idea, we wouldn't need a ```FolderEditor``` to have a neoviminstance at all, since we technically wouldn't even <i>need</i> to ever contact neovim through the folder interface (assuming we implement our own.)

extr0py:
Another cool capability here would be to copy and paste files in the file explorer by using `yy` and `p` gestures.

bryphe:
Tracking the file explorer work now in #886 - great discussion here though!

PR #623, opened 2017-08-16
Fix 'No Matching Autocommand' message

extr0py:
Fix 'no matching autocommand' message by attaching no-op listeners to FocusLost/FocusGained


justinmk:
Instead of defining a dummy event(s), one can check `nvim_eval("exists('#FocusGained')")`. And this can be done in `nvim_call_atomic()` to avoid a (unlikely) race-condition.

extr0py:
Excellent, thanks @justinmk ! That sounds like a much cleaner approach.

Issue #624, opened 2017-08-16
Tabline says dirty after undolevels = 0


cyansprite:
Vim help:
```When all changes have been undone, the buffer is not considered to be changed.
It is then possible to exit Vim with ":q" instead of ":q!".
Note that this is relative to the last write of the file.  Typing "u" after
":w" actually changes the buffer, compared to what was written, so the buffer
is considered changed then.
```

So you can change the buffer, but the tab icon is still dirty.
I'm not sure what we need to detect in order to update the tab state, I can do some research on this.

extr0py:
Nice catch... I've seen a bit of flakiness around the dirty indicator so perhaps this is related.

I think right now we are using the 'b:changedtick' variable - perhaps we need to switch to use '&modified'?

cyansprite:
right because b:changedtick will change even if you type and then undo like wise undo redo (assuming you saved after a change)

so checking &modified <i>should</i> give us the correct answer each time, I just test and it and seems correct

cyansprite:
```
b:changedtick	The total number of changes to the current buffer.  It is
		incremented for each change.  An undo command is also a change
		in this case.  This can be used to perform an action only when
		the buffer has changed. 
```

extr0py:
> so checking &modified should give us the correct answer each time, I just tested it and it and seems correct

Cool, that does seem what we want! 

To implement this, right now, the `UI.Actions.bufferUpdate` takes a `version` parameter. Seems like we'd want to replace that with an `isDirty` flag. We'd have to modify it in a couple places (action creator, action, reducer, state).

In addition, we don't send up the `&modified` flag. We might want to add that to the context we send up from neovim -> Oni. If we have that, then we can pass that easily to the `bufferUpdate` redux action.

extr0py:
Just tried out the latest changes - works great! The behavior of the dirty flag is working as I'd expect it now 😄 Thanks!

extr0py:
Since this is closed now, @cyansprite , you should be able to submit the solution @ bountysource. https://www.bountysource.com/issues/48329508-tabline-says-dirty-after-undolevels-0

PR #625, opened 2017-08-17
Tabs modified

cyansprite:
Only concern I have is with TextDocumentItem, not sure what it's doing so... need to figure that out.

'Fix' for  #624


cyansprite:
-.- why does it take all the commits? I need to learn better git things...

cyansprite:
I figured it out -.-

Issue #591, opened 2017-08-10
Hidden buffers are shown in buffer tabline


Bretley:
I have a custom vimscript function that I use to effectively make a "popup" terminal show up.

```
let g:term_buf = 0
function! Term_toggle()
  if g:term_buf == bufnr("")
    setlocal bufhidden=hide
    hide
  else
    botright 15new
    try
      exec "buffer ".g:term_buf
    catch
      term
      let g:term_buf = bufnr("")
    endtry
    startinsert!
  endif
endfunction
```

This causes a new "no name" tab to appear, but it isn't accessible, I.e. clicking on it doesn't take me to an empty buffer, and I can't exit it by clicking on it.

extr0py:
Thanks for logging this @bert88sta ! Ya, sounds like we're not handling hidden buffers correctly - they shouldn't be shown in tabs.

extr0py:
@bert88sta - just curious - is there a reason why you are setting the buffer to hidden in this case?

cyansprite:
I do something similar it's so that if I use something that switches through buffer it hides the ones that are hidden, that way I don't switch through my buffers and accidentally open my terminal instead of a file I'm trying to get to.

cyansprite:
in ./browser/src/UI/components/Tabs.tsx
```ts
<div className="corner" onClick={props.onClickName}></div>
```

that way you don't have to click directly on the name but you can click the corner without an X.
This is an issue I ran into with things with short names.

in browser/src/neovim/Buffer.ts
```ts
public getOption(optionName: string): Promise<any> {
     return this._session.request<any>("nvim_buf_get_option", [this._bufferReference, optionName])
}
```

that way we can check buffer var of hidden/listed.
I think hidden might be fine to leave in the tabs since I jump from buffers a lot and don't use tabs, but don't show listed.

extr0py:
Good call! The tab issue is pretty straightforward, I'll open a PR for that.

For the buffers - we'll have to figure out the timing of when we make that request (seems like we'd have to check it for all buffers too). It would be nice to have this metadata for the buffers available in the Redux store, so the UI could treat it appropriately -

bryphe:
Unfortunately I realized to support some of the scenarios - like being able to navigate away from a file when it has edits, we'd need to use the `set hidden` option. This is important for gestures like go-to definition, applying refactorings across a workspace, etc. Without `set hidden`, we have to use `e!`, which clears pending edits (super annoying with go-to definition - #778).

[...]

Issue #564, opened 2017-08-04
Tabs: Implement scrollbar when there is overflow


bryphe:
The case when there is an 'overflow' (too many tabs) isn't handled today.

There should be the following:
- A minimal scrollbar to allow easy navigating via the mouse, if desired
- Focusing the proper tab - so even if you navigate via keyboard, that would get scrolled into view.

hoschi:
Instead of a scrollbar, is it possible to make the tabs container wrapping in to more lines than one? Like flexbox + flex-wrap. I know that behavior of IntelliJ and really liked it. With a scrollbar you not see all your opened tabs. Probably it is a problem that the editor space gets smaller so a render cycle in NVim must be triggered?

hoschi:
![multi line tabs](http://hadihariri.com/images/no-tabs-1.png)
from http://hadihariri.com/2014/06/24/no-tabs-in-intellij-idea/

extr0py:
@hoschi - interesting idea! I didn't think about this. Yes, it would definitely be doable, we'd just have it overflow to the next row. 

Perhaps we could make it an option - I think I would prefer the look of a single-line of tabs, but I could see where some would prefer this. Thanks for the suggestion!

hoschi:
@extr0py great :+1: 
Yeah, I think some people want to stick to the default vim behavior. An option would be good.

Shou:
You can also make the tabs shrink Chrome-style with Flexbox as a first measure before scrolling kicks in. I'd be glad to submit a PR if you're happy with the idea.

bryphe:
That sounds great @Shou, a PR would be awesome if you're up for it!

hoschi:
Can we make this configurable? My point of wrapping tabs was to show always the full name.

@Shou when you use flexbox for shrinking can you add wrapping with `flex-wrap` in one go?

bryphe:
Looks like the work this issue was tracking is complete - having both scrollbar + wrapping behavior. I'll cose this, but we can use a new issue to track the tab-shrinking option, if desired.

PR #626, opened 2017-08-17
Fix tabs not changedTick instead of modified/ Don't show hidden bufs/scrollbar on overflow

cyansprite:
If one would modify a buffer then undo it, instead of changing from
isDirty to not dirty (tabline with the little circle) it would remain
dirty because ticks update even when undoing...

Therefore check for modified instead.

This would have caused furthur issues whenever we implement
requesting not close on dirty.

Issue(s) : #624

### DIFF
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -386,7 +386,7 @@ export class NeovimEditor implements IEditor {
             UI.Actions.hideSignatureHelp()
             UI.Actions.hideQuickInfo()
 
-            UI.Actions.bufferEnter(evt.bufferNumber, evt.bufferFullPath, evt.bufferTotalLines)
+            UI.Actions.bufferEnter(evt.bufferNumber, evt.bufferFullPath, evt.bufferTotalLines, evt.hidden, evt.listed)
         } else if (eventName === "BufWritePost") {
             // After we save we aren't modified... but we can pass it in just to be safe
             UI.Actions.bufferSave(evt.bufferNumber, evt.modified, evt.version)
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -197,11 +197,11 @@ export class NeovimEditor implements IEditor {
         this._neovimInstance.on("mode-change", (newMode: string) => this._onModeChanged(newMode))
 
         this._neovimInstance.on("buffer-update", (args: Oni.EventContext) => {
-            UI.Actions.bufferUpdate(args.bufferNumber, args.version, args.bufferTotalLines)
+            UI.Actions.bufferUpdate(args.bufferNumber, args.modified, args.bufferTotalLines)
         })
 
         this._neovimInstance.on("buffer-update-incremental", (args: Oni.EventContext) => {
-            UI.Actions.bufferUpdate(args.bufferNumber, args.version, args.bufferTotalLines)
+            UI.Actions.bufferUpdate(args.bufferNumber, args.modified, args.bufferTotalLines)
         })
 
         this._render()
@@ -388,8 +388,8 @@ export class NeovimEditor implements IEditor {
 
             UI.Actions.bufferEnter(evt.bufferNumber, evt.bufferFullPath, evt.bufferTotalLines)
         } else if (eventName === "BufWritePost") {
-            // After save, there is always an additional change tick bump, so the `+1` is needed to account for that.
-            UI.Actions.bufferSave(evt.bufferNumber, evt.version + 1)
+            // After save we aren't modified anymore :) (0 means not modified in vim)
+            UI.Actions.bufferSave(evt.bufferNumber, evt.modified)
         } else if (eventName === "BufDelete") {
 
             this._neovimInstance.getBufferIds()
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -197,11 +197,11 @@ export class NeovimEditor implements IEditor {
         this._neovimInstance.on("mode-change", (newMode: string) => this._onModeChanged(newMode))
 
         this._neovimInstance.on("buffer-update", (args: Oni.EventContext) => {
-            UI.Actions.bufferUpdate(args.bufferNumber, args.modified, args.bufferTotalLines)
+            UI.Actions.bufferUpdate(args.bufferNumber, args.modified, args.version, args.bufferTotalLines)
         })
 
         this._neovimInstance.on("buffer-update-incremental", (args: Oni.EventContext) => {
-            UI.Actions.bufferUpdate(args.bufferNumber, args.modified, args.bufferTotalLines)
+            UI.Actions.bufferUpdate(args.bufferNumber, args.modified, args.version, args.bufferTotalLines)
         })
 
         this._render()
@@ -388,8 +388,8 @@ export class NeovimEditor implements IEditor {
 
             UI.Actions.bufferEnter(evt.bufferNumber, evt.bufferFullPath, evt.bufferTotalLines)
         } else if (eventName === "BufWritePost") {
-            // After save we aren't modified anymore :) (0 means not modified in vim)
-            UI.Actions.bufferSave(evt.bufferNumber, evt.modified)
+            // After we save we aren't modified... but we can pass it in just to be safe
+            UI.Actions.bufferSave(evt.bufferNumber, evt.modified, evt.version)
         } else if (eventName === "BufDelete") {
 
             this._neovimInstance.getBufferIds()
--- a/browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts
+++ b/browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts
@@ -68,14 +68,13 @@ export const getTextFromContents = (contents: types.MarkedString | types.MarkedS
 
 export const bufferUpdateToTextDocumentItem = (args: Oni.BufferUpdateContext): types.TextDocumentItem => {
     const lines = args.bufferLines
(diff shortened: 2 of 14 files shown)


cyansprite:
TextDocumentItem worries me because I'm not really sure where it comes from, I didn't get too look into it that much.

browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts
```diff
@@ -68,13 +68,14 @@ export const getTextFromContents = (contents: types.MarkedString | types.MarkedS
...
-    const { bufferFullPath, filetype, version } = args.eventContext
+    const { bufferFullPath, filetype/* , modified  */} = args.eventContext
     const text = lines.join(os.EOL)
 
     return {
         uri: wrapPathInFileUri(bufferFullPath),
         languageId: filetype,
-        version,
+        // modified,
+        version: 0,
```
extr0py: Unfortunately this may cause problems. This is used in our client implementation of the "Language Server Protocol" (https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md).

The `TextDocumentItem` is used to notify the language server that the file has changed, and it should update its in-memory representation. This is important for stuff like autocomplete, enhanced syntax highlighting, etc.

Some implementations rely on the `version` being incremented, otherwise they won't bother re-parsing or doing work to update that in-memory version... this would then come up as issues during completion.

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -110,10 +110,7 @@ let context.windowTopLine = line("w0")
 let context.windowBottomLine = line("w$")
 let context.byte = line2byte(line(".")) + col(".")
 let context.filetype = eval("&filetype")
-
-if exists("b:last_change_tick")
```
extr0py: Can we keep `version` here? The `version` seems like the right thing to send to the language servers, but `&modified` is definitely want we want for actually knowing if the buffer is dirty.

browser/src/UI/Actions.ts
```diff
@@ -34,7 +34,7 @@ export interface IBufferUpdateAction {
     type: "BUFFER_UPDATE",
     payload: {
         id: number,
-        version: number,
+        modified: number,
```
extr0py: We could change the `modified`s here to `boolean`

extr0py: (It might be more intuitive if it is a `boolean` - it would mean we'd need to convert it from a `number` to a `boolean` somewhere in the pipe - either before the action creator or in the action creator). If `modified` is a `number`, and I'm just glancing at it, it makes me think there  might be multiple modified states or something - so making it a `boolean` would make this more crisp.

extr0py:
Thanks for jumping in and making the change, @cyansprite ! This is awesome, having the `&modified` flag for dirty tracking is a much better experience.

> TextDocumentItem worries me because I'm not really sure where it comes from, I didn't get to look into it that much.

Yes, unfortunately I think if we remove the `version` there (or always send 0), it will cause problems, as the language servers use it and they might not update their version of the document. Which would then cause completion issues and stuff. 

Can we send both the `version` and `&modified` for now?

The only other item I was thinking about was the `modified` field on the buffer objects in `State.ts` and `Actions.ts` - can we change that to a `boolean`? I believe the value would be only ever 0/1 so that would make the 'intent' of the field more clear. 

Thanks again for putting this together! This is a huge improvement for the buffer line.

browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts
cyansprite: Oooooooh that's what that was when I was doing this I couldn't find where we made TextDocumentItem and it makes sense because it's language-server-protocol.

Okay yeah I have a plan to fix that.

vim/core/oni-core-interop/plugin/init.vim
cyansprite: yeah, I didn't realize the thing about the text document item, it seems weird to me that they check the version and not if it's modified, but either way, I will fix it.

browser/src/UI/Actions.ts
cyansprite: Yeah I was going to do that, but I left it as is because I know some vimmers are hard core numbers because c, but yeah I was thinking about changing it to bool too, so will do :).

cyansprite:
I agree with all your suggestions, I implemented this way for convenience and rapid prototyping, plus I was sleepy.

Fixing now

extr0py:
> I implemented this way for convenience and rapid prototyping, plus I was sleepy.

Haha, this is the same way I work too... Get it working in a fast / dirty way, then once it is working, I'll  go back and take the time to clean it up.

cyansprite:
So we use b:changeTicks, why don't we use 
```vim
changenr()						*changenr()*
		Return the number of the most recent change.  This is the same
		number as what is displayed with |:undolist| and can be used
		with the |:undo| command.
		When a change was made it is the number of that change.  After
		redo it is the number of the redone change.  After undo it is
		one less than the number of the undone change.

```

extr0py:
Interesting... I actually use `b:changedTick` because I saw that is what other completion providers were using. It seems like the difference is the 'undo' behavior - `b:changedTick` would always increase, whereas this would map to the change in the `undolist`. (I think YouCompleteMe is where I saw this)

I didn't know about `changenr` actually until just now - looking through the doc you posted, `changenr` would probably work OK too (and maybe even better, if the language service caches older versions). I haven't tested, but the only problem I could see is if there is a naive language service that only does the update if newVersion > oldVersion. But the most important scenario for this is the completion scenario - so as long as the version is incrementing as the user types - it seems like it would be fine.

cyansprite:
Also when I open oni I have to click any key for it to load while the editor has focus or it's just, black.

extr0py:
Thanks for the info / research!

> We actually don't send buffer updates until on save is this the design we want?
Well for the syntax, completions of course work.

We actually send buffer updates on any change (this happens via an `autocmd` that hooks `CursorMoved`/`CursorMovedI` events and calls `OniNotifyBufferUpdates`) - but in insert mode we only send a subset (the current line). It turns out there is no 'perfect' event for listening to buffer changes - Neovim has an issue tracking adding a better event, though.

> however, I don't think that's for everyone so changed tick would be better left untouched in this case. I think it's more hassle and just better to use changed tick, because then we would have to keep track or undo levels and changed tick, and then we would need to check what mode it's in to tell which we should check against.

Sounds reasonable to me. The latest changes look great!

cyansprite:
> Sounds reasonable to me. The latest changes look great!

Unless there is something else to add I think I'm done :).

cyansprite:
Fixing #591 while I'm here since it's right in line with modified

cyansprite:
Also small quick fix for #564

browser/src/Editor/NeovimEditor.tsx
```diff
@@ -386,10 +386,10 @@ export class NeovimEditor implements IEditor {
             UI.Actions.hideSignatureHelp()
             UI.Actions.hideQuickInfo()
 
-            UI.Actions.bufferEnter(evt.bufferNumber, evt.bufferFullPath, evt.bufferTotalLines)
+            UI.Actions.bufferEnter(evt.bufferNumber, evt.bufferFullPath, evt.bufferTotalLines, evt.hidden, evt.listed)
         } else if (eventName === "BufWritePost") {
-            // After save, there is always an additional change tick bump, so the `+1` is needed to account for that.
-            UI.Actions.bufferSave(evt.bufferNumber, evt.version + 1)
```
extr0py: Glad we can lose the `version + 1` hack here 👍

extr0py:
Just looked through the latest changes - looks great to me! I'm really happy the `version + 1` hack is gone, and that we now have a more robust way to know when the buffer is modified (and should be visible). I like the way you implemented this, too - it's a very natural extension of the state we have there today.

Looks like there are a couple of minor lint issues that the appveyor picked up (you can run `npm run lint` locally to reproduce them):
```
> oni@0.2.8 lint:browser C:\projects\oni
> tslint --project browser/tsconfig.json --config tslint.json && tslint vim/core/oni-plugin-typescript/src/**/*.ts
C:/projects/oni/browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts[110, 19]: Missing trailing comma
C:/projects/oni/browser/src/Plugins/Api/LanguageClient/LanguageClientHelpers.ts[126, 46]: Missing trailing comma
```

Once those are addressed, I'll merge this in.

Also, I put a small bounty on this issue as a thank you: https://www.bountysource.com/issues/48329508-tabline-says-dirty-after-undolevels-0

So make sure to claim that once this is in 😄

cyansprite:
Thanks:) I will fix the errors when I get home.

> Also, I put a small bounty on this issue as a thank you: https://www.bountysource.com/issues/48329508-tabline-says-dirty-after-undolevels-0

So make sure to claim that once this is in 😄

What's a bounty and it's not big deal I like to help and I love Oni:).

cyansprite:
All fixed :)

extr0py:
Just tested out the PR - the behavior feels so much better now, and the dirty mark actually behaves as I'd expect 😄 Great work, thank you for the contribution! I'll merge in now.

extr0py:
> What's a bounty and it's not big deal I like to help and I love Oni:).

I set up a bountysource page [here](https://www.bountysource.com/teams/oni) for Oni and there have been some really generous donations, even in the early stage Oni is in! The funds are for Oni development, and as part of that, I think its important to show appreciation to contributors to the project. And I really appreciate the improvements you've made so far!

So once this issue is closed, make sure to "claim" the bounty here: https://www.bountysource.com/issues/48329508-tabline-says-dirty-after-undolevels-0

Thanks again for your contribution! 😄

cyansprite:
> Thanks again for your contribution! 😄

No problem!
And wow I had no idea this was a thing, this is amazing thank you!

Issue #361, opened 2017-04-11
UI: Externalize tab lines


extr0py:
The goal for this WI is to externalize the tab lines, so that we can have a tab experience like Atoms:

![image](https://cloud.githubusercontent.com/assets/13532591/24913212/4cb3c4a4-1e85-11e7-944c-eb1f0404ab18.png)

There are a couple of parts to this:
- Create a `Tabs` and `Tab` component
    - `Tabs` should be a collection of `Tab` and handle things like overflow and the actively focused tab
    - `Tab` should have an optional icon, a filename, a dirty marker, and a close icon
- Add state in the Redux store to handle active tabs
- Connect the state in the redux store to the current vim tab state

extr0py:
The work in #201 is sort of a pre-req to this, as it contains some DOM refactoring that would make it easier. However, the `Tabs` and `Tab` component could be developed in isolation in the iterim

extr0py:
This was just merged in via @justinmk with this PR: https://github.com/neovim/neovim/pull/6583 . Look forward to integrating this.

extr0py:
Looks like this is available in the build of Neovim that's included - `:help ui-ext-tabline`.

extr0py:
Resolved by #557, which adds support for this, along with configuration values documented in the readme.

Issue #55, opened 2016-12-08
Feature - Snippets


bryphe:
Snippet support is huge in saving people typing. It's a prominent feature in every major text editor, for good reason.

UltiSnips is the de-facto standard in the VIM community, and ideally we would reuse that. The problem is that it requires Python support and at least Neovim binaries aren't built with Python on Windows.

_Python interpolation_ is a really cool feature UltiSnips has - in the spirit of this project, with a JavaScript-driven plugin API, it would make sense to introduce javascript interpolation. There are some potentially really neat integrations there - like showing the current date in specific formats, showing the lines in a buffer, expanding github issues by ID in a TODO comment, etc

So for those reasons, I believe it might make sense to create a UI-driven shell approach to snippets.

We would need to recognize multiple snippets formats:
- Existing Ultisnips formats
- TextMate snippet syntax (which would give us (VSCode Snippet support)[https://github.com/neovim/neovim/pull/5686]

bryphe:
This is in the latest build now! Will use new issues to track remaining work.

badosu:
@bryphe What are the snippets formats supported? Can it be disabled?

bryphe:
> What are the snippet formats supported?

Yes, I need to document this - it's the same snippet format as VSCode (which is basically the same as TextMate): https://code.visualstudio.com/docs/editor/userdefinedsnippets

> Can it be disabled?

There isn't a configuration setting at the moment - it might make sense to add one. The only entry point for it is Oni's completion menu:
![image](https://user-images.githubusercontent.com/13532591/37055475-ac3de7f6-2136-11e8-8791-9989ab21ae1e.png)

So it gets implicitly disabled if that is off. Let me know if it'll interfere with the native config we have: https://github.com/onivim/oni/wiki/How-To:-Make-Oni-closer-to-bare-Vim-experience

badosu:
I have the completion menu open, but I'd like to not have the snippets enabled (for stability and avoiding distractions).

WDYT of opening an issue for providing an option to disable snippets?

bryphe:
> WDYT of opening an issue for providing an option to disable snippets?

Sounds great! Makes sense to have it - easy fix too 👍

badosu:
@bryphe https://github.com/onivim/oni/issues/1719

bryphe:
Cheers, thanks @badosu ! I should get to it today

PR #635, opened 2017-08-20
WIP Quickopen Enhancements

cyansprite:
# Progress Made
### Can select more than one item now
   - c-i increments/decrements
   - On enter/c-v/c-s opens all the incremented in buffers and whichever you enter/c-v/c-s on it will give that one focus.
---
### Can quickopen for colorschemes
   - recaches each request due to not having a way to check if it updated outside of vim, it's fast so... I don't see this being a problem.
### Fixed quickopen overflow with small windows
   - Max width to cap at 600px (width before)
   - width defaults to 70%
   - details overflow-x: hidden so they aren't floating in random space...
### Force open bookmarks
   - Sometimes I want to open them even if I'm in a cwd other than home/install.
---
# TODO
* [x] Force open bookmark/select folder
* [x] Colorschemes activate on select instead of actually having to hit enter, this way you get a live preview of the colorscheme and you don't have to reopen each and everytime.
* [x] Ignore binary files for all file gathering.
* [x] Work on speed improvements
* [x] Async files for things other than git folders
* [x] State bg/fg colors for neovim to have for oni QO menus
* [x] Cache files on directory change/construction
* [x] Update ignore (we will use .gitignore kind of ignore)
* [x] Be able to type as things load...
* [x] Refine input field typing experience.
* [x] Make incrementing only available for certain ```QuickOpenTypes```, currently this would be files.

---
### Future TODO
- Cache files when we figure out projections/workspaces.
- Searching through search results
- Implement loc list with incrementing buffer lines and search results.
- Buffer lines for ALL active buffers.
- Preview and silent loading of buffers
- <s>* [ ] Make editing shortcuts for (nvim temporary buffer/instance)</s>
- * [ ] State popup bg/fg colors for neovim to have for oni AutoComplete menus
* [ ] Merge #633 changes once they are done.
## Gif Showing Basics Done so far.
<a href="http://imgur.com/WUyfzsp"></a>
* [ ] Scrollbar... this is going to be fairly complex due to us only using 10 items right now, but... it shouldn't be <i>that</i> bad.
* [ ] Figure out what color to make the highlighting. (Use a light bg you'll see what I mean.
* [ ] Make incrementing loading play with tabs better so it's not slow on many file selections.
* [ ] Quick Open support for buffers
* [ ] Have an option to ignore hidden files/folders
* [ ] Prefix support
  -  : - vim command mode for current window
  -  / - buffer search (integrate with the buffer search you added)
  - ! - oni command mode (same commands shown in Ctrl+Shift+P)
  - '`' - show available vim marks
  - & - show available buffers
  - Others....

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -47,8 +47,7 @@ export interface IConfigValues {
     // (can still be activated by pressing 'Alt')
     "oni.hideMenu": boolean
 
-    // glob pattern of files to exclude from fuzzy finder (Ctrl-P)
-    // "oni.exclude": string[]
+    // Ignore patterns following .gitignore rules for QuickOpen/Fuzzy Finding (Ctrl-P)
     "oni.ignore": string[]
 
     // bookmarks to open if opened in install dir
@@ -135,7 +134,11 @@ export class Config extends EventEmitter {
         "oni.hideMenu": false,
 
         // "oni.exclude": ["**/node_modules/**"],
-        "oni.ignore" : [".git", ".svn" ],
+        "oni.ignore" : [
+            ".git", ".svn",
+            // If you want oni to ignore hidden files, keep this in your ignore
+            ".*"
+        ],
         "oni.bookmarks": [],
 
         "editor.backgroundOpacity": 1.0,
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -48,7 +48,8 @@ export interface IConfigValues {
     "oni.hideMenu": boolean
 
     // glob pattern of files to exclude from fuzzy finder (Ctrl-P)
-    "oni.exclude": string[]
+    // "oni.exclude": string[]
+    "oni.ignore": string[]
 
     // bookmarks to open if opened in install dir
     "oni.bookmarks": string[]
@@ -133,7 +134,8 @@ export class Config extends EventEmitter {
 
         "oni.hideMenu": false,
 
-        "oni.exclude": ["**/node_modules/**"],
+        // "oni.exclude": ["**/node_modules/**"],
+        "oni.ignore" : [".git", ".svn" ],
         "oni.bookmarks": [],
 
         "editor.backgroundOpacity": 1.0,
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -137,7 +137,7 @@ export class Config extends EventEmitter {
         "oni.ignore" : [
             ".git", ".svn",
             // If you want oni to ignore hidden files, keep this in your ignore
-            ".*"
+            ".*",
         ],
         "oni.bookmarks": [],
 
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -250,6 +250,12 @@ export class NeovimEditor implements IEditor {
                     UI.Actions.hidePopupMenu()
                 } else if (key === "<enter>") {
                     UI.Actions.selectMenuItem("e")
+                } else if (key === "<C-i>") {
+                    UI.Actions.incrementMenuItem()
+                    UI.Actions.nextMenuItem()
+                } else if (key === "<C-I>") {
+                    UI.Actions.incrementMenuItem()
+                    UI.Actions.previousMenuItem()
                 } else if (key === "<C-v>") {
                     UI.Actions.selectMenuItem("vsp")
                 } else if (key === "<C-s>") {
@@ -307,8 +313,14 @@ export class NeovimEditor implements IEditor {
                 this._commandManager.executeCommand("oni.editor.gotoDefinition", null)
             } else if (key === "<C-p>" && this._screen.mode === "normal") {
                 quickOpen.show()
+            } else if (key === "<C-B>" && this._screen.mode === "normal") {
+                quickOpen.show(true)
             } else if (key === "<C-/>" && this._screen.mode === "normal") {
                 quickOpen.showBufferLines()
+            } else if (key === "<C-0>" && this._screen.mode === "normal") {
+                quickOpen.showColors()
+            } else if (key === "<C-+>" && this._screen.mode === "normal") {
+                quickOpen.showBuffers()
             } else if (key === "<C-P>" && this._screen.mode === "normal") {
                 this._tasks.show()
             } else if (key === "<C-pageup>") {
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -251,19 +251,30 @@ export class NeovimEditor implements IEditor {
                 } else if (key === "<enter>") {
                     UI.Actions.selectMenuItem("e")
                 } else if (key === "<C-i>") {
-                    UI.Actions.incrementMenuItem()
-                    UI.Actions.nextMenuItem()
+                    if (QuickOpen.isIncrementable) {
+                        UI.Actions.incrementMenuItem()
+                        UI.Actions.nextMenuItem()
+                    }
                 } else if (key === "<C-I>") {
-                    UI.Actions.incrementMenuItem()
-                    UI.Actions.previousMenuItem()
+                    if (QuickOpen.isIncrementable) {
+                        UI.Actions.incrementMenuItem()
+                        UI.Actions.previousMenuItem()
+                    }
                 } else if (key === "<C-v>") {
                     UI.Actions.selectMenuItem("vsp")
                 } else if (key === "<C-s>") {
                     UI.Actions.selectMenuItem("sp")
                 } else if (key === "<C-n>") {
                     UI.Actions.nextMenuItem()
+                    if (QuickOpen.isColor) {
+                        UI.Actions.selectMenuItem("", false)
+                    }
                 } else if (key === "<C-p>") {
                     UI.Actions.previousMenuItem()
+
+                    if (QuickOpen.isColor) {
+                        UI.Actions.selectMenuItem("", false)
+                    }
                 }
 
                 return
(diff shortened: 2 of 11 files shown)


extr0py:
Wow, really cool stuff in the gif! I like the colorscheme switching a lot 😄 

And cool list - it's neat to see the progress indication on the PR.

>  Make editing shortcuts for ease

One thing I was thinking about this - is I think there are going to be a few places in Oni where we will have a 'lightweight' edit experience (like a small textbox). A few scenarios I had in mind for this:
- QuickOpen (obviously)
- Rename experience in the file explorer (press 'F2' or some bound key to trigger a rename, and a small textbox opens over the folder)
- Snippet editing - tracked by #55 

But for all these cases, it'd be nice to have a the full set of neovim key bindings I have configured (<C-w>, <C-e>, etc). I was thinking, potentially, of having these textboxes essentially be tiny neovim windows - so that we could use neovim to process the keybindings. 

That would be a bigger chunk of work, and depend on some of the initial multiplexing changes, so it might make sense to defer that piece (if that approach makes sense).

>  Scrollbar... this is going to be fairly complex due to us only using 10 items right now, but... it shouldn't be that bad.

We might be able to remove the '10-item' limit and use something like https://github.com/bvaughn/react-virtualized to only render a subset.

There were a couple of other things I was thinking top-of-mind too:
- Having the Menu use the current color schemes background / foreground colors - I noticed I had to darken up the background fade a bit, but it looked really nice - and might look even cooler with your colorscheme switcher!
- Have an incremental-loading experience. Like right now, if there is a big folder, we show "Loading..." but then don't show any results until everything is available. Most fuzzy-finders start showing files/results as soon as they can - that would be a huge experience improvement.

Looking awesome!

cyansprite:
>And cool list - it's neat to see the progress indication on the PR.

Didn't realized it did progress I just wanted a nice list :).

> But for all these cases, it'd be nice to have a the full set of neovim key bindings I have configured (, , etc). I was thinking, potentially, of having these textboxes essentially be tiny neovim windows - so that we could use neovim to process the keybindings.
That would be a bigger chunk of work, and depend on some of the initial multiplexing changes, so it might make sense to defer that piece (if that approach makes sense).

If we use the single neovim instance strategy we could use a temporary hidden buffer that the user doesn't have to know about (i.e. we don't show hidden buffers and unless they open it with buffers! they won't ever know, if they close it we can just open it again...) and so we can use this temporary buffer, use setline()/getline() and that will be our input field, this way we don't have to wait for a new spin each and every time, we can just edit the buffer as the text field tied to it, and then we can edit just like any other neovim buffer... it's a bit complex but it's a wildy awesome idea and I would enjoy this a lot, but I don't even know how it's hooked up to begin with but, yeah.

>We might be able to remove the '10-item' limit and use something like https://github.com/bvaughn/react-virtualized to only render a subset.

I will look into that.

>    Having the Menu use the current color schemes background / foreground colors - I noticed I had to darken up the background fade a bit, but it looked really nice - and might look even cooler with your colorscheme switcher!

I plan on adding a bg/fg color to the state by using &bg/&fg, if these are null then use default (i'm not sure what this is in nvim I've never looked at it honestly without a color for normal...)
On source/color scheme loaded recache these values, that way we can use them for our menus.

>    Have an incremental-loading experience. Like right now, if there is a big folder, we show "Loading..." but then don't show any results until everything is available. Most fuzzy-finders start showing files/results as soon as they can - that would be a huge experience improvement.

We need a solution, I can use my binary to do this but there is not a good way in typescript without writing an algorithm or porting a library and their dependencies, however I think a solution implemented in typescript may not be the best solution but using something like ag/ripgrep/etc that way we can stream async and we don't use a scripting language that can't optimize as other coding bases.

cyansprite:
https://github.com/gjtorikian/nak

cyansprite:
currently writing a manual find, it's working pretty nice, almost done.

cyansprite:
typescript, only dependency is fs and ignore

cyansprite:
going to add support for a config ignore and config use global ignore ~/.gitignore

keforbes:
These will be awesome enhancements, but could you add "update README" to your list of todo's?  We need to be sure to track all the various key combos we're introducing, especially when there isn't a (neo)vim equivalent.

cyansprite:
> These will be awesome enhancements, but could you add "update README" to your list of todo's? We need to be sure to track all the various key combos we're introducing, especially when there isn't a (neo)vim equivalent.

Once I have a concrete list :)

extr0py:
Sorry I'm a bit late to the party here!  Was (finally) reading through the comments 😄 

I really liked this:
> If we use the single neovim instance strategy we could use a temporary hidden buffer that the user doesn't have to know about (i.e. we don't show hidden buffers and unless they open it with buffers! they won't ever know, if they close it we can just open it again...) and so we can use this temporary buffer, use setline()/getline() and that will be our input field, this way we don't have to wait for a new spin each and every time, we can just edit the buffer as the text field tied to it, and then we can edit just like any other neovim buffer... it's a bit complex but it's a wildy awesome idea and I would enjoy this a lot, but I don't even know how it's hooked up to begin with but, yeah.
Optionally we can have an entire nvim instance to itself... and other things we might want nvim to use but the user never interact directly, that might be an even better plan.
We could have nvim multiplexing or a neovim singleton, and then a nvim singleton for temporary operations like text inputs.
This kind of goes inline with #633 

Sounds like we're exactly on the same page here! That's exactly how I was thinking of using it. I forgot to call out one use case I want to pursue also - I'd like to have an _embedded browser_ in Oni (which is basically free, because Electron), and then use Neovim to manipulate text inputs. I think it'd be really helpful, at least for me, to be able to pop open github and use the keyboard to navigate around  (#361) and use all the neovim-power I have to write text in the issues. Lots of cool possibilities!

cyansprite:
> Sounds like we're exactly on the same page here! That's exactly how I was thinking of using it. I forgot to call out one use case I want to pursue also - I'd like to have an embedded browser in Oni (which is basically free, because Electron), and then use Neovim to manipulate text inputs. I think it'd be really helpful, at least for me, to be able to pop open github and use the keyboard to navigate around (#361) and use all the neovim-power I have to write text in the issues. Lots of cool possibilities!

Glad we agree! Getting it to work I'm not sure, that's probably better for you to do because I haven't tried connecting to neovim and since you spun it it's make sense but I can give it a go after I'm done with Fuzzy finder, a lot of people have been negatively affected by the slowness so I am killing that by caching and our own manual walker.

If you don't care test it out, it works great for me, however like my commit says the overridden command doesn't seem to work.

cyansprite:
@extr0py 
Also, this is the plugin I've been working on.
It might be buggy, if you find anything please let me know!
https://github.com/cyansprite/Extract

cyansprite:
Nice gif for progress being made

<a href="http://imgur.com/FAEwMU2"></a>

extr0py:
Wow, that's awesome! Can't wait to have that integrated. It's so cool to see the changes to setting & configuration made real-time

extr0py:
And thanks for linking the plugin, I'll check it out in a bit! I read through the README and it sounds interesting. I need to test it out / see how to map in my workflow. Yanking into a list and being able to grab from the list in a convenient manner sounds interesting, especially when you are juggling multiple items you want to store.

cyansprite:
> Wow, that's awesome! Can't wait to have that integrated. It's so cool to see the changes to setting & configuration made real-time

Yeah I was super happy with the color-scheme outcome.

> And thanks for linking the plugin, I'll check it out in a bit! I read through the README and it sounds interesting. I need to test it out / see how to map in my workflow. Yanking into a list and being able to grab from the list in a convenient manner sounds interesting, especially when you are juggling multiple items you want to store.

Thanks, from when I use it it's pretty useful, I haven't found any bugs since the initial commits ( I think 14 while demoing? )
However, there are a few improvements I can think of, let me know how it goes!
As far as I can tell it works with Oni, as it should since it works in vanilla neovim as well.

Bretley:
@cyansprite: I don't know how familiar you are with helm, or if this is within your skillset/goal here, but do you think you could generalize the quickOpen?

I could imagine this replacing wildmenu/cmdline pretty easily, all we'd have to do is have the user give us an activation action, a list of things to choose from, and a final behavior

just a thought

cyansprite:
@Bretley Feel free to take it over, I'm not working on this anymore, it's more of just staying here until @bryphe decides it's no longer necessary, I don't have time to work on Oni anymore fortunately.

Bretley:
@bryphe since I haven't been here for a while, is this a Proof Of Concept or is this something wand merged in? I'm not saying I'd be able to do it, but I could try!

bryphe:
@Bretley I think there are a lot of really great ideas in this PR from @cyansprite  (and great discussion too!), so I've been using it as a reference. There are a few things that are top-of-mind for me in the menu space:
- Asynchronous loading + general menu API (I started doing some work on this in #764)
- Using a binary by default (both the default strategies can be replaced by `ripgrep`, so I have that prototyped in #762)

The most important gap we have today, though (in my opinion), is the 'first-run' experience - the first time you open Oni up, and open the quick open experience - it's a bit confusing and unintuitive. I like the experience VSCode has, where you can type `?` and see some of the available options - figuring out a way to ease the user in and show them the capabilities in an intuitive way is important here. Having the command prefixes would certainly help a lot. I think there's still some design / thinking we need to do here.

Long-term, too, I'd like to see if it is possible to consolidate the Oni commands & Vim `ex` commands together too (I think you mentioned this a bit in #776). It'd be nice if users didn't have to think too hard about whether they need to `:` or `<C+S+P>` to get where they need to go 😄 

It might be a challenge to bring this PR in as-is giving that the menu API surface is changing, but if there are any parts that interest you that could be split out, I'm definitely open to that! Some ideas for chunks to split out could be the multi-selection (extending the new menu API to handle that), or create the colorscheme chooser - that was a really cool part of this PR too 👍 

Otherwise, thinking about how we can smooth out that first-open experience would be awesome.

bryphe:
Will close this out as it's pretty old, and some of the bits of functionality have migrated to other PRs.

PR #637, opened 2017-08-21
Initial Window Management

bryphe:
# Window Management

## Description
This is a rough prototype for a new API - `Oni.windows` for working with splits and window layout in Oni. Wanted to put some thoughts down because I think this is related to thinking @hoschi is doing.

### API

Right now, this adds the following API methods:
- `Oni.windows.split(direction, editor)` - create a new window split , and only supports a single split.
- `Oni.windows.onSplitChanged` - event that is dispatched when the split configuration is changed

And creates a component for rendering the splits. 

This could be used in concert with some of the work for the file explorer and alternate editor concepts, to use them in the canvas. The first item I'd like to tackle with this is creating a markdown-preview experience. It would be very simple at first (just a live preview, not backed by a Neovim instance), but in the future it'd be nice to have a Neovim instance backing it.

### Vim Integration

This also brings in some ideas around navigating multiple neovim instances in terms of Window management - in particular, detecting in a neovim instance when we hit the edge of window navigation, and directing that upward to be handled by the Oni window manager.

In this branch, calling `Oni.windows.split(1, null)` from the console will create a lateral split, with an element controllable by some `Oni.Editor` instance:

image

### Additional API Methods

These are some of the API methods I would like to add:
- `Oni.windows.showSideBar(editor)`/`closeSideBar` - show a sidebar. I'm debating if this should live on WindowManager or somewhere else.
- `Oni.windows.moveUp/moveDown/moveLeft/moveRight`
- `Oni.windows.close(editor)` - close an editor window


hoschi:
`showSideBar` should be probably `showDock` so these special windows can also be at top/bottom not only at the side.

justjoeyuk:
Rather than show/hide maybe it should be 'toggle'? Just a thought!

bryphe:
This is laying the foundation for some of the interesting UI scenarios, and multiplexing work (but won't actually do anything today unless you do something like call `Oni.windows.split(..)` from the developer tools). The next step here is to get two neovim editors side-by-side - there will be additional challenges and refactoring (there are some 'global' things that are handled by `NeovimEditor`, and focus will have to be handled)

PR #640, opened 2017-08-22
make feature table more readable

Granitosaurus:
Y's and N's are very similar and hard to read. This simply replaces them with __+__ and -


extr0py:
Nice, thanks for doing this, @Granitosaurus ! 

Really appreciate the contribution, there's definitely a ton of room for improvement in the documentation.

Issue #636, opened 2017-08-20
Open from command line


mhartington:
Similar to VSCode, Atom, and VimR, Oni should offer a command line script that starts the editor in the current directory.  I have a simple  one for vscode

```
function code () {
  VSCODE_CWD="$PWD" open -n -b "com.microsoft.VSCode" --args $*
}
```

Or something more complicated like atom 

<details>

```bash
#!/bin/bash

if [ "$(uname)" == 'Darwin' ]; then
  OS='Mac'
elif [ "$(expr substr $(uname -s) 1 5)" == 'Linux' ]; then
  OS='Linux'
else
  echo "Your platform ($(uname -a)) is not supported."
  exit 1
fi

if [ "$(basename $0)" == 'atom-beta' ]; then
  BETA_VERSION=true
else
  BETA_VERSION=
fi

export ATOM_DISABLE_SHELLING_OUT_FOR_ENVIRONMENT=true

while getopts ":wtfvh-:" opt; do
  case "$opt" in
    -)
      case "${OPTARG}" in
        wait)
          WAIT=1
          ;;
        help|version)
          REDIRECT_STDERR=1
          EXPECT_OUTPUT=1
          ;;
        foreground|benchmark|benchmark-test|test)
          EXPECT_OUTPUT=1
          ;;
      esac
      ;;
    w)
      WAIT=1
      ;;
    h|v)
      REDIRECT_STDERR=1
      EXPECT_OUTPUT=1
      ;;
    f|t)
      EXPECT_OUTPUT=1
      ;;
  esac
done

if [ $REDIRECT_STDERR ]; then
  exec 2> /dev/null
fi

if [ $EXPECT_OUTPUT ]; then
  export ELECTRON_ENABLE_LOGGING=1
fi

if [ $OS == 'Mac' ]; then
  if [ -L "$0" ]; then
    SCRIPT="$(readlink "$0")"
  else
    SCRIPT="$0"
  fi
  ATOM_APP="$(dirname "$(dirname "$(dirname "$(dirname "$SCRIPT")")")")"
  if [ "$ATOM_APP" == . ]; then
    unset ATOM_APP
  else
    ATOM_PATH="$(dirname "$ATOM_APP")"
    ATOM_APP_NAME="$(basename "$ATOM_APP")"
  fi

  if [ -n "$BETA_VERSION" ]; then
    ATOM_EXECUTABLE_NAME="Atom Beta"
  else
    ATOM_EXECUTABLE_NAME="Atom"
[...]

Issue #346, opened 2017-04-03
Mac: Create way to add Oni to $PATH


extr0py:
I'm not sure what the best practice on Mac is, but similar to #344 on Windows, there should be an option to add `Oni` to the `$PATH` variable so that it is accessible from the terminal.

VSCode has an interesting way of handling this:
https://code.visualstudio.com/docs/setup/mac
![image](https://cloud.githubusercontent.com/assets/13532591/24614788/2f9257d6-1841-11e7-886c-6512bf83cc8a.png)

By exposing a command in the command palette to install/uninstall the command.

bert88sta:
After npm install, nom run build, and nom link, I discovered by accident that I could run Oni from anywhere in terminal, and even on Spotlight. Is this not the case for other installation methods?

keforbes:
It is the `npm link` command which is adding `oni` to your `$PATH`.  The current installer is not performing an equivalent operation.

keforbes:
For what it's worth, all that command does in vscode is creates a symlink to `/usr/local/bin/code`.  And it only enables that "Shell Command" `if (process.platform === 'darwin') {`.  I guess we could put a similar command in our Command Palette since users don't get to answer any questions during install on a Mac.

Or, on first launch (on Mac) we could pop-up a prompt to the user asking "would you like to install 'oni' in your $PATH?" and set a flag in `config.js` so the question is never asked again on future launches regardless of the answer.  I'm not sure which behavior Mac people would prefer; maybe we should do both, prompt at first launch and add command to palette so they can change it later.

keforbes:
I was going to work on this issue since it seemed pretty straightforward, especially since we can use vscode's implementation as a starting point.  Unfortunately, after looking a bit closer, I found this:
https://github.com/Microsoft/vscode/blob/92ee0d1e067ae70706cddaa6feca9a7659aa4d90/src/vs/workbench/parts/cli/electron-browser/cli.contribution.ts#L101

vscode is running `osascript` to prompt the user for administrator privileges to write to `/usr/local/bin`.  I thought I could just create a symlink, I didn't think about the fact that it would be to a restricted directory.  So I guess someone who actually owns a Mac would need to implement this to make sure they get that osascript right.

If anyone cares, this is what I was going to add after the `commands[]` declaration (line 67) in `Commands.ts`:
```
    if (Platform.isMac()) {
        if (fs.existsSync("/usr/local/bin/oni")) {
[...]

PR #646, opened 2017-08-22
ADDED $PATH insert/remove commands. Tasks can now use await.

justjoeyuk:
I've added the ability to Add/Remove Oni to the $PATH on OSX. It would most likely work on Linux if someone went ahead and added tweaks. I only have access to OSX right now so it's the only fix I can present.

I wanted a way to change the command to Add/Remove depending on a boolean check. This would require me to know when a task has been executed. When a task is executed, an event is now fired. I check that event for either of the two $PATH modifications and reload the commands if it was either of them.

The nature of asking a users permission (sudo) is asynchronous and so kicking off the event would have happened before the user confirmed their password, so I went ahead and made task completions async. That way, each task can now perform awaits.

Let me know if there are any thoughts. Demo:

Demo

### DIFF
--- a/browser/src/Platform.ts
+++ b/browser/src/Platform.ts
@@ -1,6 +1,8 @@
+/// <reference path="./../../definitions/sudo-prompt.d.ts"/>
+
+import * as fs from "fs"
 import * as os from "os"
-import * as fs from 'fs'
-const sudo = require("sudo-prompt")
+import * as sudo from "sudo-prompt"
 
 export const isWindows = () => os.platform() === "win32"
 export const isMac = () => os.platform() === "darwin"
@@ -13,27 +15,27 @@ export const getLinkPath = () => isMac() ? "/usr/local/bin/oni" : "" // TODO: wi
 
 export const pathIsLinked = () => {
   if (isMac()) {
-    try { fs.lstatSync(getLinkPath()) } catch(_) { return false }
+    try { fs.lstatSync(getLinkPath()) } catch (_) { return false }
     return true
   }
 
   return false
 }
-export const removeFromPath = () => isMac () ? fs.unlinkSync(getLinkPath()) : false; // TODO: windows + other
+export const removeFromPath = () => isMac () ? fs.unlinkSync(getLinkPath()) : false // TODO: windows + other
 
-export const addToPath = async() => {
+export const addToPath = async () => {
   if (isMac()) {
-    const appDirectory = '/Applications/Oni.app/Contents/'
-    const options = {name: 'Oni', icns: `${appDirectory}Resources/Oni.icns` };
+    const appDirectory = "/Applications/Oni.app/Contents/"
+    const options = {name: "Oni", icns: `${appDirectory}Resources/Oni.icns` }
     const linkPath = `${appDirectory}MacOS/Oni`
     await _runSudoCommand(`ln -s ${linkPath} ${getLinkPath()}`, options)
   }
 }
 
-const _runSudoCommand = async (command:string, options:any) => {
+const _runSudoCommand = async (command: string, options: any) => {
   return new Promise(resolve => {
     sudo.exec(command, options, (error: Error, stdout: string, stderr: string) => {
       resolve({error, stdout, stderr})
     })
-  });
+  })
 }
--- a/browser/src/Platform.ts
+++ b/browser/src/Platform.ts
@@ -3,6 +3,7 @@
 import * as fs from "fs"
 import * as os from "os"
 import * as sudo from "sudo-prompt"
+import * as path from "path"
 
 export const isWindows = () => os.platform() === "win32"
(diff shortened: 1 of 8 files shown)


justjoeyuk:
This pull request is for issue #346

justjoeyuk:
@extr0py It seems that the require is causing the linter to fail. `sudo-prompt` only allows require it seems and I've not much experience with TypeScript, so not sure how to proceed. I think we should use `sudo-prompt` as it's cross platform and makes it easier for other platforms in the future.

justjoeyuk:
Should all be fixed now :+1:

justjoeyuk:
Updated the pull request. Since we're trying to keep each platforms implementation separate, it checks exclusively for OSX when adding the commands at the moment, with a TODO: for when the linux/windows ones are complete.

extr0py:
Wow, very cool! Thanks for implementing this @justjoeyuk  ! I'll give it a go shortly 👍

browser/src/Services/Commands.ts
```diff
@@ -14,10 +14,13 @@ import * as UI from "./../UI/index"
 
 import { CallbackCommand, CommandManager } from "./CommandManager"
 
+import * as Platform from "./../Platform"
 import { replaceAll } from "./../Utility"
 
 export const registerBuiltInCommands = (commandManager: CommandManager, pluginManager: PluginManager, neovimInstance: INeovimInstance) => {
     const config = Config.instance()
+    commandManager.clearCommands()
```
extr0py: Thanks for going the extra mile here to wire up showing add / remove correctly, that's awesome!

package.json
```diff
@@ -158,6 +158,7 @@
     "sinon": "1.17.6",
     "spectron": "3.6.2",
     "style-loader": "0.18.2",
+    "sudo-prompt": "^7.1.1",
```
extr0py: Can you please clamp the dependency at `7.1.1`? (Remove the `^`)

extr0py:
Looks like this would address both #636 and #346 - quite a first PR 😄

browser/src/Platform.ts
```diff
@@ -1,8 +1,40 @@
...
+    return true
+  }
+
+  return false
+}
+export const removeFromPath = () => isMac () ? fs.unlinkSync(getLinkPath()) : false // TODO: windows + other
+
+export const addToPath = async () => {
+  if (isMac()) {
+    const appDirectory = "/Applications/Oni.app/Contents/"
```
extr0py: Can we derive this from `process.execPath` or `process.helperExecPath`? This logic will work fine in the install-from-dmg case, but I see a decent amount of .zip downloads too. It'd be great if it worked in that case as well.

extr0py:
Just tried this out, worked great! OSX is Oni's most downloaded platform, and this addresses one of the first blockers in the experience - excellent contribution!

I think the only thing that would improve the experience is if we gave some sort of confirmation message when we successfully add (or remove) the path. Unfortunately, there isn't a message / notification UI yet. This can come in a separate PR, though - I added created issue #653 to track that.

So the only remaining items I see are:
- Is it possible to derive the application path from `process.execPath` or `process.helperExecPath`? If so, this would work in the zip case too
- Clamp the version of `sudo-prompt`

Once those are addressed, I'll go ahead and bring it in. This is really awesome, thanks for working on this @justjoeyuk !

justjoeyuk:
@extr0py Added a way for tasks to show messages once completed. It's a start, but ultimately the completion handlers should return whether or not they were successful and we can then decide what to do by the returned value.

Clamped the version for sudo-prompt.

Will look into alternative ways to find where the executable is running from.

justjoeyuk:
@extr0py I've not got the application path at run-time so it's flexible. I've tested it and all appears to be working. Let me know if you have any problems with it.

extr0py:
Excellent, thank you for making the updates @justjoeyuk ! Trying it out now. The code looks good to me overall.

extr0py:
The dialog looks great! I am hitting an issue, though.

I'm getting an error when it hits `_runSudoCommand`:
Error: ENOENT: no such file or directory, open '/Users/extr0py/oni/dist/Mac/Oni.app/Contents/Resources/app../../Resources/Oni.icns'

Looks like perhaps its maybe missing the `/` after `app`?

extr0py:
(I also saw the appveyor build failed, that looks like it was a problem installing electron, not related to the PR - I kicked it off again)

justjoeyuk:
@extr0py All done 🎉

extr0py:
Awesome, thanks @justjoeyuk ! The path handling looks more robust now, and the 'add to path' gesture works great on my OS X machine. Most excellent.

I'll go ahead and bring this in! Thanks again for putting this together. 

I believe this has had a bounty on it for a while against #346  - so make sure to claim that!

PR #650, opened 2017-08-22
Fix config file location in README

nfm:
Support for `config.json` was removed in e7e14356fe6db3cd8ee2f1fd94b7ed1919cdf119.


extr0py:
Good catch, thank you @nfm!

PR #651, opened 2017-08-23
FIXED #642 - inoremap was not working with the shift key

### LINKED ISSUE #642: <S-CR> keybinding isn't working
I use shift-enter instead of enter to insert a new line.  
`inoremap <S-CR> <NL>` is ignored and not working. As a sidenote `inoremap <C-CR> <NL>` works as expected.

justjoeyuk:
Added the missing mappings in Keyboard.ts


justjoeyuk:
References Issue #642

PR #652, opened 2017-08-23
FIXED ISSUE #642 - ADDED missing mappings

justjoeyuk:
inoremap wasn't working with the shift key so added the missing mappings.

### DIFF
--- a/Menu.js
+++ b/Menu.js
@@ -25,72 +25,89 @@ const buildMenu = (mainWindow, loadInit) => {
         files.forEach((fileName) => executeVimCommand(`${command} ${normalizePath(fileName)}`))
     }
 
+    const isWindows = os.platform() == "win32";
+
     let preferences = {
         label: 'Preferences',
         submenu: [
             {
-                label: "Edit Oni config",
-                click: () => executeOniCommand("oni.config.openConfigJs")
+                label: 'Edit Oni config',
+                click (item, focusedWindow) {
+                    executeOniCommand("oni.config.openConfigJs")
+                }
             },
         ]
     }
 
     if (loadInit) {
         preferences.submenu.push(
-        {
-            label: "Edit Neovim config",
-            click: () => executeOniCommand("oni.config.openInitVim")
-        })
+            {
+                label: 'Edit Neovim config',
+                click (item, focusedWindow) {
+                    executeOniCommand("oni.config.openInitVim")
+                }
+            }
+        )
     }
 
-    let firstMenu = os.platform() == "win32" ? 'File' : 'Oni';
-    menu.unshift({
-        label: firstMenu,
+    let reopenWithEncoding = {
+        label: 'Reopen with Encoding',
+        submenu: []
+    }
+
+    // TODO: Maybe better show normal encoding name in submenu?
+    // Encoding list from http://vimdoc.sourceforge.net/htmldoc/mbyte.html#encoding-values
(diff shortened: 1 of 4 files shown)


justjoeyuk:
Messed up the first PR. This should do it. References #642

extr0py:
Just tried this out - unfortunately it looks like there is a regression. The <C-P> detection to open the command palette no longer works (since now we are adding the `S` in). Vim seems to treat `<C-P>`, `<C-S-p>`, and `<C-S-P>` the same as far as I can tell, but Oni sees them as different currently.

It seems like there are a couple possibilities to fix this:
- Only append `S-` if the character is longer than 1 (a compound key, like `space`)
- Change the hardcoded key binding in `NeovimEditor.tsx` to be recognize the new key chord (`<C-S-P>` instead of  `<C-P>`. Ideally, our input recognition would treat these equally, but that would be a more in-depth fix.

cyansprite:
> Vim seems to treat <C-P>, <C-S-p>, and <C-S-P> the same as far as I can tell, but Oni sees them as different currently.

As far as I know neovim itself doesn't know the difference between <c-p> and <c-P>, <c-s-p> doesn't work either, if it has c- it just looks at the letter not the case, m-p and m-P ARE case sensitive though.

extr0py:
> As far as I know neovim itself doesn't know the difference between <c-p> and <c-P>,<c-s-p> doesn't work either, if it has c- it just looks at the letter not the case, m-p and m-P ARE case sensitive though.

Hmm, I just tested this (on Windows):
- `nnoremap <C-a> :echo '<C-a>'<CR>`

And I tried it for <C-A> - at least in my testing, Neovim seemed to differentiate between the `<C-A>` and `<C-a>` case.

justjoeyuk:
@extr0py I've had to use <S-C-P> as the mapping prepends "S" onto the command.

extr0py:
Thanks @justjoeyuk !

I just tested this out - the 'Command Palette' (tasks) opens up fine now, but unfortunately I can't enter uppercase numbers into it. I think this is due to the `evt.preventDefault` we do with the shift key - could we remove that `preventDefault` for the time being? We use it for the `alt`/`ctrl` keys to not conflict with global keyboard shortcuts/accelerators, but I don't think we need it for the shift key (and it breaks that 'Command Palette' case).

justjoeyuk:
@extr0py Awesome, sorry for the ignorant attempts, this is all a learning experience for me - first time with TypeScript (and neovim, for that matter!)

extr0py:
> Awesome, sorry for the ignorant attempts, this is all a learning experience for me - first time with TypeScript (and neovim, for that matter!)

Oh no problem! Jumping in and pushing up code is the best way to learn (IMO). Thanks for taking time to dig into the codebase 😄 I really appreciate the contributions.

I think the fix here should be relatively straightforward - we can just remove the `evt.preventDefault` in the shift case, and that should address the remaining issue. After that, should be good to go in!

justjoeyuk:
@extr0py Sorry for the delay, long weekend! Done.

extr0py:
> @extr0py Sorry for the delay, long weekend! Done.

Haha no worries, thanks for updating it! I just tried it out and it's working great. I'll bring it in once the build is complete-

extr0py:
Merged! Thanks for the contribution, @justjoeyuk !

PR #656, opened 2017-08-23
Refactoring Menu.js

catsAND:
Its my first PR to this good project.

Refactor menu list like sublime/atom

Need more testing. I testing only on Win764b.

Any suggest?

### DIFF
--- a/Menu.js
+++ b/Menu.js
@@ -25,72 +25,89 @@ const buildMenu = (mainWindow, loadInit) => {
         files.forEach((fileName) => executeVimCommand(`${command} ${normalizePath(fileName)}`))
     }
 
+    const isWindows = os.platform() == "win32";
+
     let preferences = {
         label: 'Preferences',
         submenu: [
             {
-                label: "Edit Oni config",
-                click: () => executeOniCommand("oni.config.openConfigJs")
+                label: 'Edit Oni config',
+                click (item, focusedWindow) {
+                    executeOniCommand("oni.config.openConfigJs")
+                }
             },
         ]
     }
 
     if (loadInit) {
         preferences.submenu.push(
-        {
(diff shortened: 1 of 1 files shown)


extr0py:
Wow, very cool! I'm excited to try test it out - I'll look at it soon. Thank you for making this contribution, @catsAND !

extr0py:
Just tried it out... most excellent! I liked the commands to insert the date/time and filename. This is really cool and definitely fits in with the 'ease learning curve' goal. 

I want to pursue adding the shortcuts/commands in the menu as hint-text, to show the user how they can do this via keyboard, but that can come later.

The only one that didn't quite work as I expected was 'indent' and 'unindent' - I wonder if they should be `>>` and `<<` instead? I'll go ahead and bring this in, though - we can tackle that as a next step.

Thanks again for putting this together @catsAND !

catsAND:
Thanks for merge.

> in the menu as hint-text

I tried find how do this in electron documentation, but only found sublabel.

>I wonder if they should be `>>` and `<<` instead?

Maybe. I'm noob in vim. Most command i found on vim wikia.

josemarluedke:
This is awesome. One thing to consider adding is to show what will be executed, similar to what other apps show what the short cut is, but here, how to do that in vim.

keforbes:
> This is awesome. One thing to consider adding is to show what will be executed, similar to what other apps show what the short cut is, but here, how to do that in vim.

I had a similar idea when I wrote the first (lame) attempt at a menu.  Unfortunately, Electron forces all shortcuts to be in ALL CAPS, which is dangerously misleading when we're talking about Vim commands.  If anyone can find a way to trick Electron into showing shortcuts exactly as we define them then I completely agree we should show the equivalent Vim command.

PR #667, opened 2017-08-28
Allow moving through Ctrl-P menu with arrow keys

MasonAsh:
Even though using arrow keys is probably a bad habit it should still be allowed IMO.


extr0py:
Thanks for the PR @MasonAsh ! Agreed, this makes sense to be a reasonable default.

extr0py:
Merged in, thanks for the contribution! 😄

PR #672, opened 2017-08-28
Fix #671 - Bump Electron version to 1.7.5

### LINKED ISSUE #671: electron segfaults on startup
Oni ran fine till the first segfault:
```bash
Aug 28 11:04:35 maxxgubbl kernel: electron[26855]: segfault at 208 ip 00007f2c18421ac5 sp 00007ffe196eeed0 error 4 in libnode.so[7f2c175b1000+12b3000]
Aug 28 11:04:35 maxxgubbl systemd-coredump[26870]: Process 26855 (electron) of user 1000 dumped core.
                                                   
                                                   Stack trace of thread 26855:
                                                   #0  0x00007f2c18421ac5 uv_async_send (libnode.so)
                                                   #1  0x0000000003bb156e _ZN4atom12NodeBindingsD1Ev (electron)
                                                   #2  0x0000000003bb2909 _ZN4atom17NodeBindingsLinuxD0Ev (electron)
                                                   #3  0x0000000003bb9681 _ZN4atom18AtomRendererClientD1Ev (electron)
                                                   #4  0x0000000003bb96c9 _ZN4atom18AtomRendererClientD0Ev (electron)
                                                   #5  0x0000000003b88d09 _ZN4atom16AtomMainDelegateD2Ev (electron)
                                                   #6  0x0000000003b88c3f main (electron)
                                                   #7  0x00007f2c106074ca __libc_start_main (libc.so.6)
                                                   #8  0x000000000059fd79 _start (electron)
```
Then on every start it segfaults ... but I can use it as always. I switched to latest master with same problem. Also I installed latest node LTS version 6.11.2, also no changes. I now updated electron which seems to do the trick:

2017-08-28-111235_570x322_scrot

extr0py:
- [ ] Verify OSX regression is fixed (or find workaround)


extr0py:
It looks like there is still a regression with 1.7.5 on OSX: electron/electron#9965 and electron/electron#9880

Opening multiple instances causes problems - it seems that `app.makeSingleInstance` callback is delayed before being called, which is a poor experience (makes startup feel _very_ slow)

extr0py:
Blocked by #679

extr0py:
Will include this in the fix for #679

PR #673, opened 2017-08-29
Move documentation to wiki pages

keforbes:
I took the "documentation" sections of the README and moved them into the wiki.  That way we can grow the documentation as needed without bloating the README.  I also moved CONTRIBUTING.md and ROADMAP.md into the wiki.

I kept the install/build instructions and pretty example images of Oni so the README is still useful for new users.  I figure anything beyond enticing users to install Oni can be moved to the wiki.


extr0py:
>  I figure anything beyond enticing users to install Oni can be moved to the wiki.

Awesome! Yes, the README is way too overwhelming right now - thanks for doing this! Wiki looks great too 👍

PR #674, opened 2017-08-30
Fix #662: Handle undefined files array and handle empty git directories in Quic…

### LINKED ISSUE #662: Fuzzy Finding stalls when opening Oni from the dock (Mac OS)
Reading through the docks I know Oni's fuzzy finder searches a directory using `git ls-files` and "falls back to other methods" when a git directory is not found. I'm not sure what mechanism this is, but in Mac OS (Sierra: 10.12.5) the fuzzy finder never stops loading after I've opened it from my dock:

screen shot 2017-08-25 at 10 46 40 am

Using the `v0.2.8` release of Oni

someguynamedmatt:
…kOpen.ts

Based on [Issue 662](https://github.com/extr0py/oni/issues/662):
> The issue was also found in Linux and seems to popup when either a directory with inaccessible sub-dirs is `<C-p>`'d on _or_ an empty git directory is `<C-p>`'d on
 
- "Infinite loading" was/is a result of no data coming in from the `spawn` process. This is fixed by checking the exit code of `filer` and checking the array length of `this._loadedItems`
    - If both are `0` push the default "Bookmark" and "Open a folder" options to the menu

- Moved some logic in `show()` into its own private function because that code is now being used in two other places

- Check the array length for the non-git option (`filer.stderr`) for directory parsing. It's a simple `undefined` check then pushing the default menu options to the `this._loadedItems` array if the `files` array isn't present.

I hope this helps some. I noticed the team has commented on removing some of the areas of code that I touched in favor of async functions. I'd like to work on this too, at some point, but I think this PR gives a decent/quick solution to [#662](https://github.com/extr0py/oni/issues/662).

Hope this helps!

### DIFF
--- a/browser/src/Services/QuickOpen.ts
+++ b/browser/src/Services/QuickOpen.ts
@@ -97,7 +97,6 @@ export class QuickOpen {
     //             })
     //         })
     // }
-
     public async show() {
         // reset list and show loading indicator
         this._loadedItems = []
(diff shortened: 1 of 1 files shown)


extr0py:
Awesome, thanks for the PR, @someguynamedmatt ! Handling the exit code makes sense to me - seems like we can detect several of the cases where we'd stick on "Loading..." with that.

> Moved some logic in show() into its own private function because that code is now being used in two other places

Excellent!

browser/src/Services/QuickOpen.ts
```diff
@@ -195,12 +164,26 @@ export class QuickOpen {
                 nodir: true,
                 ignore: Config.instance().getValue("oni.exclude"),
             }, (_err: any, files: string[]) => {
+                Log.error(_err)
+                if (!files) {
+                    this._loadDefaultMenuItems()
+                    this._showMenuFromQuickOpenItems(this._loadedItems)
+                }
+
                 files.forEach( (f: string) => {
```
extr0py: Is it possible that `files` can be null/undefined here? Or is it always an empty array? I'm wondering if this should be in an `else` block, otherwise seems like we could get an exception on the `forEach` call.

someguynamedmatt: That's a good catch @extr0py . As I was messing around with that block, I never saw the case of an error being thrown if the `!files` block was entered. Nonetheless, I think you're right. It would be better to be sure of that happening with an `else` check. I'll update the PR later today (LA time) and resubmit.

someguynamedmatt:
@extr0py Changes made!

extr0py:
Excelllent! Just tried it out. Looks great, thanks for making the changes & the contribution, @someguynamedmatt - merging in now!

PR #677, opened 2017-08-31
Make status bar link

triplepointfive:
Other screenshots have a link to a corresponding section, but status bar doesn't.


extr0py:
Thanks for catching this and fixing it, @triplepointfive ! 

I just saw that @keforbes updated this in PR #673 - so looks like this isn't needed anymore. Appreciate you making the change though! I'll close this out as it is addressed.

Issue #663, opened 2017-08-26
Finder window closes immediately after selecting Open Folder with Fuzzy Finder


someguynamedmatt:
I found this when messing with [#662](https://github.com/extr0py/oni/issues/662)...

Release: `0.2.8`
OS: Mac Sierra: 10.12.5

Steps to recreate: 
> assuming issue [#662](https://github.com/extr0py/oni/issues/662) applies to the user
* select directory button (?) in lower-left corner of Oni window
* select home directory (i.e. `/Users/me` or `~/me`), click `Open`
> home directory should be visible in the lower-left now
![screen shot 2017-08-25 at 5 42 50 pm](https://user-images.githubusercontent.com/17413539/29737065-daf277fe-89bc-11e7-9c60-5b7619ca38cb.png)
* open fuzzy finder `<C-P>`
* select `Open Folder` (`<C-n>` then `Enter`)

Now the finder will open up and immediately close. I'm unsure if this and [#662](https://github.com/extr0py/oni/issues/662) are related.

cyansprite:
Would love to help but unfortunately I don't have a Mac and Linux and Windows works perfect for me.
Does this happen when opening a folder from the menu as well? I think I used the same exact logic with QuickOpen.

extr0py:
Interesting, thanks for logging the issue, @someguynamedmatt ! I'm able to reproduce this too with the step you provided. Very helpful, it was interesting to note the different experiences between clicking the path vs 'Open Folder'.

I also noticed that, in the Command Palette (`<C-S-P>`) _clicking_ on the 'Open Folder' option directly works, but pressing enter fails as you describe. I'm suspecting that the 'enter' key is actually picked up by OSX's Open File/Folder dialog, and dismisses it immediately. Going to test that out and see if that's the issue...

extr0py:
> Would love to help but unfortunately I don't have a Mac and Linux and Windows works perfect for me.

Ah ya, unfortunately this looks a quirk specific to Mac - appreciate the offer though!

someguynamedmatt:
@extr0py What files should I check out to see where this issue might be coming from?

extr0py:
Hmm, I haven't been able to dig into this too much - but I suspect what's happening here is that we OSX's platform open-folder modal on the _keydown_ event, and that experience actually listens for a _keyup_ event!

A few options here:
[...]

PR #681, opened 2017-09-01
#663 Fix: preventDefault on keyCode 13

someguynamedmatt:
As noted in #663, this only pops up in Mac OS. It seems that when opening the Finder the `keydown` event makes Oni lose focus (which is deferred to the Finder window), and the accompanying `keyup` event is asserted in the Finder as a user click. It didn't seem possible to `evt.preventDefault()` on the `keyup` from within the Oni instance as focus was lost at this point. To avoid this issue...

* in `Input/Keyboard.ts`:
    - `document.addEventListener("keydown"...` check if `keyCode` is 13, if so `preventDefault`

This seems counterproductive at first, but Oni handles `Enter` events on its own so the event itself doesn't need to propagate outward. This prevents Mac's Finder from swallowing the `keyup` event when Oni loses focus.

I tested this on Linux to make sure it didn't have any weird side effects. None were found. I don't have an easy way to test on Windows unfortunately.

### DIFF
--- a/browser/src/Input/Keyboard.ts
+++ b/browser/src/Input/Keyboard.ts
@@ -5,6 +5,14 @@ export class Keyboard extends EventEmitter {
         super()
 
         document.addEventListener("keydown", (evt) => {
+            /*
+             * This prevents the opening and immediate
+             * (unwanted) closing of external windows.
+             * This problem seems to only exist in Mac OS.
+             */
+            if (evt.keyCode === 13) {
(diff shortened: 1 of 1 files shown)


extr0py:
I tried it out on Windows and it seemed like it worked fine too! I was worried about integrations with native UX (like alert dialogs, system menus) and those were fine.

Thanks for investigating and putting together the fix, @someguynamedmatt ! Merging now 😄

PR #684, opened 2017-09-02
Normalize path for "Save As".

CrossR:
"Save As" isn't working for me on Windows, as the path isn't normalised so forward slashes are lost.
Using the existing `normalisePath` function seems to have fixed this for me.


extr0py:
Change looks great! I'll merge it in when the build completes. Thanks for the fix @CrossR !

CrossR:
Cheers! Think that was the fastest Issue > PR > Finished PR I've done!

extr0py:
Haha ya, as soon as I saw there was the issue, there was already a PR fixing it 😄 Awesome!

PR #687, opened 2017-09-02
Fix #628 - Paste fixes

### LINKED ISSUE #628: Clipboard pastes per line
Copy a lot of lines of text and use c-v, it will go through each line an insert it slowly.
```ts
const pasteContents = (neovimInstance: INeovimInstance) => {
    const textToPaste = clipboard.readText()
    const sanitizedText = replaceAll(textToPaste, { "<": "<lt>" })
    neovimInstance.input(sanitizedText)
}
```

I think it might be better to use norm! "+p?

extr0py:
__Issue:__ The current paste behavior with `editor.clipboard.enabled` is slow, is impacted by auto-indent, and adds extra newlines.

__Fix:__ Vim provides a setting for this - `paste`/`nopaste`, so we should be leveraging that prior to bringing in pasted text. In addition, the newlines should be sanitized to `<cr>`.


extr0py:
Fixes #628

PR #695, opened 2017-09-07
FIX #679 - Workaround for `makeSingleInstance` issue

### LINKED ISSUE #679: Investigate alternative to `makeSingleInstance`
I was investigating the `makeSingleInstance` behavior, which looks like was regressed in Electron in 1.7.x+:  electron/electron#9880. This is very important for OSX to behave like other native apps - as calling 'oni' again in the terminal should focus the existing instance.

I was looking at atom's way of handling this:
https://github.com/atom/atom/blob/2f2d62ca51d91c4b8730bce0a0d0c1b12991704c/src/main-process/atom-application.coffee

It looks like they create a shared socket at a canonical path (looking at where `socketPath`) is used, and use that to detect when a new instance pops up...

Ideally, we could leverage the `makeSingleInstance` API, but it might be worth investigating alternatives.

extr0py:
_Issue:_ Incorporate workaround to handle the `makeSingleInstance` behavior, and upgrade to latest version of electron. This is important for unblocking other OSX fixes, like #636 .

This uses the same strategy as atom uses - a file-as-socket on OSX/Linux, and a named pipe on Windows, using a canonical name, to facilitate the single-instance behavior.

This strategy also allows more flexibility in passing context between instances.

A few remaining items:
- [x] Fix two windows opening on load
- [x] Pass configuration values
- [x] Clean up code


extr0py:
Fixes #679 
Fixes #671 (by bumping Electron version to 1.7.7)

PR #698, opened 2017-09-09
[WIP] Fix #649 - Allow setting vim globals/settings from Oni's config.js

### LINKED ISSUE #649: Can't use Oni b/c of Python in non-standard path
I've just loaded Oni and I wanted to edit the settings but I can't. I get flooded with the messages from the screenshot on every interaction I try, essentially putting Oni in gridlock.

screen shot 2017-08-22 at 23 03 52

The reason this seems to happen is b/c Oni needs Python but my Pythons are in a non-standard place because I maintain a separate virtual env for my editor Python. In neovim I account for that by setting `g:python_host_prog` and `g:python3_host_prog` to the right locations, but that setting isn't exposed by Oni. Though I can make Oni read my `.config/nvim/init.vim` that'll probably break a lot more due to the amount of customisation in there.

Since Oni appears to require Python to function I think those options should be configurable (and the absence of it handled slightly more elegantly). Alternatively, Python could be bundled?

bryphe:


browser/src/Editor/NeovimEditor.tsx
```diff
@@ -361,6 +368,11 @@ export class NeovimEditor implements IEditor {
 
     private _onConfigChanged(): void {
```
bryphe: Update onConfigChanged to pass in the delta configuration (for convenience for subscribes)

Issue #699, opened 2017-09-10
Show Log Enhancements


extr0py:
For the 'Show Log' functionality, I'd like to revisit and add the following capabilities:
- Use common window management (Like #637)
- Use keyboard accessibility common to Vim (Like #637)

I am pulling it out temporarily for performance (the throughput in the redux store caused problems, as additional logging was added, exacerbating existing render issues for some components), and to fit this new UI extensibility model, but would like to bring it back. 

In the interim, logs get plumbed to the dev tools (`Control+Shift+P` -> `Open Dev Tools`)

Issue #588, opened 2017-08-09
Migrate 'main process' Javascript files to TypeScript


extr0py:
- Move `main.js`, `Menu.js`, `installDevTools.js` to a `main` folder
- Set up `tsconfig.json` in `oni\main`
- Hook up build commands

Would be nice to get the main process code in TypeScript

PR #701, opened 2017-09-10
[WIP] Temporarily remove show logs for #699

extr0py:
Temporarily remove show logs functionality, to revisit in #699


extr0py:
This removes a performance regression introduced in  #588 (which introduced more logging, which hit the redux store more frequently). Will revisit in #699 along with other enhancements (like handling window navigation and vim keybindings to navigate through errors).

Issue #638, opened 2017-08-22
Can't input special characters on international keyboard layout


lbebber:
I can't input characters such as `~`, `"`, etc., on my computer (macOS 10.12).

My keyboard is the regular Apple US Layout keyboard, but it's configured with the U.S. International layout, so in order to type, say, a ~ character, I type `~` + `<space>`. Oni seems to simply ignore the first key.

extr0py:
Thanks for logging the issue, @lbebber ! Looks like there are a few issues to resolve in terms of keyboard layouts - we also have #396 which is similiar (but for Windows IME).

The way we are listening to key events and translating them to vim-keys in `Keyboard.ts` might not be sufficient in this case.

lbebber:
If I may suggest, you could look into what https://github.com/zeit/hyper does for text input—if I'm not mistaken it had the same problem once, and it's also an Electron app.

justjoeyuk:
@extr0py Specifically, this file seems of interest: https://github.com/zeit/hyper/blob/7a08b1dc3e07ae552debfe7e62c48b0a5a028ff9/lib/utils/key-code.js

bryphe:
Fixed in PR #716 and in the latest release [v0.2.10](https://github.com/bryphe/oni/releases/tag/v0.2.10-recovery), so I'll close this. Let me know if you're still seeing any issues, though!

Issue #542, opened 2017-07-17
Can't type backticks (`)


badosu:
For some reason I can't type backticks, using oni from Arch User Repositories (`aur/oni 0.2.5-1`).

The input works on any other X window, just on Oni it does not work.

Let me know if additional info is required.

extr0py:
Thanks for logging the issue! Sorry you hit this - that's bizarre (and frustrating)!

Tried this out, and I'm unable to reproduce on Ubuntu. I'll see if I can get an Arch Linux box to test it out there.

A couple of follow-up questions to narrow it down:
- What does your `config.js` look like? Are you using a custom `init.vim`?
- What keyboard layout are you using? 

Thanks!

badosu:
Hello @extr0py, thanks for the helpful response!

My layout is set by: `setxkbmap -model pc105 -layout br`.

I am using a custom `init.vim`, however if I use a vanilla Oni setup:

```javascript
module.exports = {
  //add custom config here, such as
  "oni.useDefaultConfig": true,
  "oni.loadInitVim": false,
  "oni.hideMenu": true,
  "editor.fontSize": "16px",
  //"editor.fontFamily": "Monaco"
}
```

The issue persists.

Gahen:
Hi! I'm having the same issue with a spanish layout. It happens with several special characters (I first noticed it with `~` as `right-alt + ¡`), they are not written after pressing the same keys that work in the rest of my OS.

I'm using an updated Arch and kde.

someguynamedmatt:
@badosu Is this issue still present for you? I'm on Arch and I don't see it.

extr0py:
It looks like right now we are not handling 'Dead Keys' correctly, which might be related here. I see similiar issues with other keyboard layouts that have dead keys (like `~` in ENG INTL)

badosu:
@someguynamedmatt Yes, it's present. On a machine with a standard US keyboard layout it works though.

Gahen:
Just for the laughs, a few hours ago oni was ignoring my caps lock key. It gets the price the most weird bug I ever seen. Latam or spanish keyboard layout.

extr0py:
Sorry @badosu @Gahen  I know these keyboard / input issues are brutal. I added the `daily-editor-blocker` tag, and these blockers for alternate keyboard layouts and input handling (dead keys, compose keys, AltGr, and IMEs) is top of my list to work on. 

[...]

PR #716, opened 2017-09-26
Fix #396, #542 - IME and Dead Key support

### LINKED ISSUE #396: Can not switch non-English input method, such as Chinese input software
Oni is very outstanding, Thank you for great work.

I use Oni on windows 7, programming and write markdown.

I found it can not switch non-English input method, such as Chinese input software.

I'm coming from China, so I need input some Chinese on write markdown.

Can you help me?

bryphe:
__Issue:__ Oni today does not support 'dead keys' or IME on all platforms, and this is a blocker for several users using non-US keyboard layouts. 

__Defect:__ The browser (and Electron by extension) _does_ support dead keys and IME out of the box - for form fields, text areas, etc. However, Oni's current input strategy just tacks a `keydown` handler on the document body. This is sufficient for US keyboard layouts, but for IME / dead-keys, this isn't handled correctly on all platforms.

__Fix:__ Use a hidden input field to support the IME and dead-key scenario. In non-IME/dead key cases, the input will pass-through as it does today (resolving via the strategy in `Keyboard.ts`). When we detect dead key or IME composition, the text field will become visible, to give feedback on the composition.

Once completed, this would address a class of the issues around keyboard input - like #542, #396, and #638


badosu:
I am confused as to why this needs careful treatment, but maybe it's because I'm not familiar with electron and how it implements the key* events.

Is it possible that `keypress` could be a better fit?

Also, I found these events that may be helpful to avoid having to store the composition state:

- https://developer.mozilla.org/en-US/docs/Web/Events/compositionstart
- https://developer.mozilla.org/en-US/docs/Web/Events/compositionupdate
- https://developer.mozilla.org/en-US/docs/Web/Events/compositionend

This my daily editor today so I'm interested in helping to solve these daily editor issues, unfortunately my time is very constrained these days.

badosu:
Excited to see this being worked on, thanks!

browser/src/Editor/NeovimInput.tsx
```diff
@@ -104,45 +98,45 @@ export class KeyboardInputView extends React.PureComponent<IKeyboardInputViewPro
...
+    }
 
-                if (evt.key === "Dead") {
-                    this.setState({
-                        isDeadKey: true,
-                    })
-                    return
-                }
+    private _onKeyDown(evt: React.KeyboardEvent<HTMLInputElement>) {
+        // 'Process' means hand-off to the IME -
```
badosu: 👏 👏 👏

bryphe: 😅

Issue #644, opened 2017-08-22
Can't type square brackets [ ]


defaltus:
For some reason I can't type square brackets using Oni 0.2.8 on Windows 7.
The input works on any other window, just inside Oni it does not work.
It only switches from `insert` to `normal` mode when I try to type `[`. 
When typing `]` Oni does nothing.

I use the standard german keyboard layout (in Neovim without Oni everything works fine).

I am currently not using a custom `init.vim`, my `config.js` looks like this
```javascript
module.exports = {
  //add custom config here, such as
  "oni.useDefaultConfig": true,
  "oni.loadInitVim": false,
  "oni.hideMenu": true,
  "editor.fontSize": "14px",
  "editor.fontFamily": "Roboto Mono"
}
```

extr0py:
Thanks for logging the issue, @Balagrio , and all the details. I'm guessing it is related to some difference in the german keyboard layout. I think this is an issue with the mapping / conversion logic in `Keyboard.ts` - perhaps it is mistaking the `[` keycode for Escape.

justjoeyuk:
@extr0py I don't think so. The global code for ESC is 27, which is the keycode mapping. I tried doing a German keyboard layout and testing different representation of the keycode in JS. It all printed out as ESC being pressed. Can't really see a solid solution.

ioncodes:
I'm expiriencing this too. Same problem happens when I paste text while in Insert mode. Maybe it doesn't handle CRLF/LF correctly?

extr0py:
I think this is because we aren't handling the `AltGr` key correctly - depending on the platform it either comes to our input layer as `C+[` or `C+A+[`, so we need to handle this case.

riedel:
I do not quite get the root of the problem. On the german keyboard  `{[]}\` are AltGr combinations of `7890`. Interestingly `{}` work (!), but not `[]\` which all emit the '^\' escape sequence of ESC, so what is the difference here? BTW also `|` and '~' work, but not `²³`. Trying to type e.g. latex without `[]\` on a German keyboard is pretty much a show stopper.

bryphe:
This was fixed in PR #717 (which is in the v0.2.10 release), the `AltGr` keys should behave as expected now for German keyboards. I'll close this, but let me know if you're still seeing any issues.

soratobukuroneko:
[...]

Issue #711, opened 2017-09-16
Full support for special characters


gruhn:
Oni has some weird behavior, when dealing with special characters. Typing `öäüß` in INSERT mode works but typing their upper-case counterparts `ÖÄÜ` has no effect.

If files, contain these characters, they can be displayed. Looks like the upper-case `Ö` got converted to a reqular `O` though. Not sure, I guess I need glasses...

![bildschirmfoto vom 2017-09-16 10-37-25](https://user-images.githubusercontent.com/26570572/30510803-623a9088-9acb-11e7-8447-ea0e6badaeec.png)

File content:

```
ö : Ö
ä : Ä
ü : Ü
ß
```

In fuzzy finder I'm able to type `ÖÄÜ` but encoding breaks for files with those characters:  

![bildschirmfoto vom 2017-09-16 10-57-29](https://user-images.githubusercontent.com/26570572/30510906-f3911d8e-9acd-11e7-9504-865898581a79.png)

The file is actually called `Überblick.md`

extr0py:
Thanks for the issue, @gruhn ! The special character issue seems like it is the same as #644. The fuzzy finder issue hasn't been tracked yet, it looks like there is a problem with encoding the process output. 

A couple of question that will help once I start looking at this:
- Which OS is this?
- Is the `Überblick.md` file in a git repo? This can impact the behavior of the fuzzy-find.

gruhn:
You're right. This file is in a git repo. I quickly checked in a regular folder and fuzzy-finding the file was not a problem. This is what I experienced on my laptop running  __Linux Mint 18.1__ with  __Oni 0.2.9__ installed. 

But now that you mentioned it, I also tested Oni on another machine of mine running __Linux Mint 17.2__ and __Oni 0.2.8__. Now checkout this:

![out-1-1-1](https://user-images.githubusercontent.com/26570572/30826014-2ef33826-a235-11e7-9eba-ca79c0c022ff.gif)

Lower-case characters aren't an issue here either. `ÜÖ` is replaced with `UO` but only in the first line. In the second line and below, all of them work but the dots are still visible, even when the character and the hole line is deleted. They disappear when I delete the last character of the first line though.

I then upgraded to 0.2.9 on this computer and as far as I can tell, it's the same behavior now as on my laptop, so I can't type `ÜÄÖ` at all anymore.

PR #717, opened 2017-09-26
Fix AltGr and broken shift cases in keyboard input

extr0py:
__Issue:__ There are still several cases that are broken with `AltGr` (and even the `Shift` keys).

__Defect:__ The meta-key resolution (determining if the `c-`, `s-`, and `m-` key prefixes should be sent up) was hacky. For the shift key, we had a hardcoded list that was not internationalized. For the `c-` and `a-` cases, we would always append this if the ctrlKey or altKey was pressed - this meant for all `AltGr` keys, we'd prefix it like `<a-(key)>`, which would fail to be resolved. On Windows, for `AltGr`, we'd append both `c-` and `a-`.

__Fix:__ Leverage the keyboard-layout module added in #715, as that can tell us the various resultant charactesr from a key press - the shift case, the altgr case, and the shift+altgr case. We can use that information to determine if we should append the prefixes (for example, if shift+b is pressed, the mapped keys will show "B", but if shift+enter is pressed, the mapping will show nothing - so we know for shift+b to ignore the `s-` prefix, but for shift+enter to prepend `s-`).


extr0py:
This fixes #644 and helps address #711 (capitalization of keys like ö to Ö work correctly)

PR #728, opened 2017-09-28
Refactor handling of arguments to child process

### LINKED ISSUE #725: Oni does not respect arguments on linux
Currently oni does not open a file when called with the file path as an argument on linux.

I'll try to take a look into this.

@bryphe Can you provide the specification on how the arguments should be parsed?

badosu:
Use the `LOCAL_ONI` environment variable instead of checking the platform and file name of the electron executable.

Also diminishes the surface area of places where there's need to check the running platorm, making cross-platform support easier.

Fix #725

### DIFF
--- a/cli/oni
+++ b/cli/oni
@@ -8,6 +8,7 @@ var fs = require("fs")
 var processOptions = {}
 processOptions.detached = true
 processOptions.stdio = ["ignore", null, null]
+processOptions.env = Object.assign({}, process.env, { "LOCAL_ONI": 1 })
 
 var isWindows = os.platform() === "win32";
 
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -157,13 +157,8 @@ function focusNextInstance(direction) {
 }
 
 function loadFileFromArguments(platform, args, workingDirectory) {
-    const windowsOpenWith = platform === "win32" &&
-                            args[0].split("\\").pop() === "Oni.exe"
-
-    const macOpenWith = platform === "darwin" &&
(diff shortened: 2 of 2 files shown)


bryphe:
Thanks for submitting the PR, @badosu !

I just checked the appveyor build - looks like it failed due to these lint errors:
```
ERROR: C:/projects/oni/main/src/main.ts[160, 22]: object access via string literals is disallowed
ERROR: C:/projects/oni/main/src/main.ts[163, 5]: misplaced 'else'
```

For the first one, you'll need to suppress it by adding `// tslint:disable-line no-string-literal` to the end of the line.

Other than that, looks great! Once those are addressed, I'll bring this in.

badosu:
I don't understand the use case for this rule.

Anyways, I changed it to use a variable access as it makes the code more readable instead of using the comments for the lint.

bryphe:
> I don't understand the use case for this rule.

Here's some info on it: https://stackoverflow.com/questions/36695645/what-is-no-string-literal-rule-in-tslint

I think _in general_ it's to prevent doing things like `document["body"]` (instead of `document.body`), or `window["setTimeout"]` vs `window.setTimeout`, because the former cases are more fragile since a typo can break them. In this case though it is noisy. It'd be nice if it was smart enough to know to skip this for objects that are essentially dictionaries (like `process.env`...).

Thanks for making the changes, I'll bring this in once the builds are done!

badosu:
:+1:

bryphe:
Thanks for the contribution, @badosu ! 🥂

Issue #162, opened 2017-01-13
Underscores aren't being displayed


keforbes:
![underscore](https://cloud.githubusercontent.com/assets/817509/21945116/c054ef70-d996-11e6-80cb-fd859939ed94.png)

In the screenshot above, notice that the QuickInfo shows `_pluginManager` yet the word appears to be `pluginManager` with a space in front of it.  For all those variables, there's supposed to be a `_` between `this.` and the name.  If you try typing a `_` it looks like a space.

extr0py:
Weird! Definitely a bug, thanks for the screenshot:

1) What OS is this?
2) Did you customize the font / font size? If so, what are the settings for that?

extr0py:
Also, I wonder if the DOM rendering PR (#140) fixes this

keforbes:
This is Linux, no customization to any fonts.  Are you not able to reproduce it?  Is this only me?

extr0py:
Wasn't able to repro it on Windows! But I just tried on an Ubuntu Linux box and was able to repro, same behavior

Rafi993:
I tried typing _ on oni and saved it and opened it in another editor it seems to be showing up in the file.

Rafi993:
It is mostly electron rendering issue there was a simiar issue https://github.com/atom/atom/issues/8026 some time ago

extr0py:
Looks like the DOM rendering PR i'm working on in #140 helps this

keforbes:
Just updated and things look much better.  I'm now only seeing the occasional issue where the underscore disappears when the cursor is on the line **below** the underscore.  This causes underscores to flicker as I `jjjjjj` in a file.  Test with the private variables declared at the beginning of the `NeovimInstance` class declaration.

![underscore](https://cloud.githubusercontent.com/assets/817509/22030696/adeef29c-dc9b-11e6-8828-f3cab3d09553.png)

keforbes:
I'm not seeing the issue in my previous comment anymore.  I don't know if it was due to upgrading Electron or some other change but that final issue is fixed.

extr0py:
Awesome! It's possible the electron upgrade could've helped here, as they bumped up the renderer version and picked up the latest fixes from Chrome.

hoschi:
I have the same thing, but with latest release v0.2.7
![2017-08-07-155759_609x54_scrot](https://user-images.githubusercontent.com/163128/29029834-4526ad08-7b89-11e7-8bcf-e7c50e061271.png)
[...]

PR #731, opened 2017-09-28
Fix #173 - Add 'editor.linePadding' setting

### LINKED ISSUE #173: Feature: set line height
Hi all, I know this might be complicated due to the current Grid implementation, but I think it would be really nice if we could set a line height option to give more air to text.

Maybe [this PR on NyaoVim](https://github.com/rhysd/NyaoVim/issues/19) could give some hints.

Thanks!

bryphe:
Add `editor.linePadding` setting to control padding between lines. Some fonts extend right to the bounds, and having a default value here of '1px' will help give some breathing room (especially for issues like #711 and #162)


browser/src/Renderer/CanvasRenderer.ts
```diff
@@ -214,18 +216,21 @@ export class CanvasRenderer implements INeovimRenderer {
...
+            this._canvasContext.fillRect(boundsStartX, y * fontHeight, boundsWidth, fontHeight)
         } else {
-            this._canvasContext.clearRect(startX * fontWidth, y * fontHeight, state.width * fontWidth, fontHeight)
+            this._canvasContext.clearRect(boundsStartX, y * fontHeight, boundsWidth, fontHeight)
         }
 
         if (!state.isWhitespace) {
             this._canvasContext.fillStyle = foregroundColor
-            this._canvasContext.fillText(text, startX * fontWidth, y * fontHeight)
+            this._canvasContext.fillText(text, startX * fontWidth, y * fontHeight + Config.instance().getValue("editor.linePadding") / 2)
```
bryphe: This should come from the screen, like the other values

browser/src/Renderer/CanvasRenderer.ts
```diff
@@ -214,18 +216,21 @@ export class CanvasRenderer implements INeovimRenderer {
         const fontWidth = screenInfo.fontWidthInPixels
         const fontHeight = screenInfo.fontHeightInPixels
 
+        const boundsStartX = Math.floor(startX * fontWidth)
+        const boundsWidth = Math.ceil(state.width * fontWidth)
```
bryphe: There are still some boundaries / gaps present here

browser/src/neovim/NeovimInstance.ts
```diff
@@ -258,16 +258,18 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
...
 
-    public setFont(fontFamily: string, fontSize: string): void {
+    public setFont(fontFamily: string, fontSize: string, linePadding: number): void {
         this._fontFamily = fontFamily
         this._fontSize = fontSize
 
-        const { width, height } = measureFont(this._fontFamily, this._fontSize)
+        const { width, height, kerning } = measureFont(this._fontFamily, this._fontSize)
+
+        document.body.style.letterSpacing = kerning.toString() + "px"
```
bryphe: This should only be applied to the root editor element

bryphe: Remove the kerning (it should really be letter-spacing anyway)

PR #737, opened 2017-09-30
Fix #711 - update git ls-files strategy to handle unicode characters …

### LINKED ISSUE #711: Full support for special characters

bryphe:
…correctly in UI

### DIFF
--- a/browser/src/Services/QuickOpen.ts
+++ b/browser/src/Services/QuickOpen.ts
@@ -125,7 +125,8 @@ export class QuickOpen {
         }
 
         // Default strategy
-        this.loadMenu("git", ["ls-files", "--others", "--exclude-standard", "--cached"])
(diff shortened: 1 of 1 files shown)


bryphe:
Last issue in #711

Issue #416, opened 2017-04-29
Add ability to record videos & screenshots directly in Oni


bryphe:
Based on the Electron desktopCapture APIs, and gif creation libraries, it should be relatively straightforward to add the ability within Oni to capture animated gifs. This would be really nice for bug repros, and to leverage with Spectron/automation to create the demo gifs automatically.

bryphe:
Gif shot looks like a cool library for this

bryphe:
Testing using `gifshot` resulted in poor quality, unfortunately. For now, I'll constrain this to writing video (.webm) and screenshots (.png).

keforbes:
Will you be creating an animated gif demonstrating this behavior for the README/wiki ?

bryphe:
Yep, I'll add something in Wiki (in the configuration section). I was thinking of adding a "Tips" or "Recipes" section, and  show how to bind a key to take a screenshot. 

I'm planning on pushing a new release out Monday as well - so I'll call out it there in the release notes too.

bryphe:
I'll also add some documentation around these new APIs:
`oni.recorder.startRecording()`
`oni.recorder.stopRecording()`
`oni.recorder.takeScreenshot()`

And new configuration options:
`recorder.outputPath`
`recorder.copyScreenshotToClipboard`

bryphe:
Added some API documentation here:
https://github.com/bryphe/oni/wiki/API

And an example of binding a key to use the `takeScreenshot` method here:
https://github.com/bryphe/oni/wiki/Configuration

bryphe:
I'll close this out, but the documentation is still pretty rough (and it would be nice to have some examples or tips/tricks)

PR #739, opened 2017-09-30
[WIP] #416 - Record videos / screenshots

bryphe:
Several open items / ideas for improving:
- [x] remove gifshot, the quality is not sufficient - so this should really be retitled to 'record videos / screenshots'
- [x] Add `recorder.outputPath` to specify where recorded videos should go. Default would be the ~templ folder

Deferring these items to later. The completion modal would make more sense to show in a separate window (once that capability is added..)
- [ ] Add 'countdown' experience
- [ ] Add statusbar item
- [ ] Add 'completion' modal that shows info, allow copying path/file/etc (or open in separate window..)


bryphe:
Fix #416

Issue #490, opened 2017-06-19
Oni frequent crash on Windows 7


saibing:
Oni frequent crash on Windows 7,  such as the following scenario will crash:

*  enter dd  consecutively to delete text line.

* quick open file.

Oni version:
```sh
* fa8a529 (HEAD -> master, origin/master, origin/HEAD) Fix #407, use editor.quickOpen.execCommand to filter files in qui
ckOpen (#489)
```

extr0py:
Thanks for reporting this, @saibing !

I've seen random crashes on Windows 10 as well. It looks like that the Neovim process is actually closing, and then Oni quits in response to that.

It might be related to a couple of crashes reported in Neovim:
- neovim/neovim#6248
- neovim/neovim#4796

There's also the possibility that some of the work Oni is doing (like incremental buffer updates) is exacerbating an existing crash in Neovim. I need to see if there is more info I can get from the actual crash in Neovim.

A couple questions though that might help narrow this down:
- What does your `config.js` look like? Have you modified it at all?
- Are you using any vim plugins?

saibing:
@extr0py 

my oni config's content as follow:

```js
module.exports = {
  //add custom config here, such as
  //"oni.useDefaultConfig": true,
  "oni.loadInitVim": false,
  "editor.fontSize": "15px",
  "editor.fontFamily": "Source Code Pro",
  "editor.quickOpen.execCommand": "fzf -f ${search}"
}
```

badosu:
I think this must be windows related as I have used Oni for a month now and did not have a single crash.

bryphe:
The good news here is I've been testing out a new nightly build and it seems much more stable on Windows (no crashes). I'll plan on upgrading Oni to use 0.2.1 as soon as it is out.

bryphe:
Upgraded to neovim 0.2.1 addresses this - #875 (at least, haven't seen any crashes!).

Feel free to open a new issue if you are still seeing this, though. I'm going to cut a new release shortly with neovim 0.2.1 binaries.

PR #742, opened 2017-10-02
#490 - Part 1: Add ability to override neovim path

bryphe:
I suspect that the Windows crash is fixed in the latest neovim build, so I want to test it out locally before bumping the neovim version. This change adds the ability to point to a local version of neovim.

The intention is to use this primarily for troubleshooting between versions of Neovim, but not something to be used day-to-day. I'd like to package neovim builds so that it is a friction-free setup (and all the features work out of the box), and potentially look at leveraging a WebAssembly build of Neovim later.

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -24,6 +24,9 @@ export interface IConfigValues {
         columns: number,
     } | null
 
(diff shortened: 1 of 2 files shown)


bryphe:
Troubleshooting for #490  - need to validate if moving to a different Neovim version addresses the crash

PR #744, opened 2017-10-03
#490 - Part 2 - Automated stability test

bryphe:
This is part 2 of the fixes for #490 - need a way to reproduce the crash to validate if it is fixed or not.

This test runs 50 times and opens the built bundle, and on my machine, tends to fail 1/50 times. 

Open issues:
- [ ] Reduce duplication between `ci` and `stability` tests (refactoring to `common` folder)


bryphe:
The stability test can be run via `npm run test:stability`. This will be helpful in verify that #490 is actually fixed...

PR #749, opened 2017-10-03
Fix #714 - Menu styles + font sizing issues

### LINKED ISSUE #714: Provide an option to configure the command palette appearance
This is how it's displayed on a high DPI high, font size screen:

screenshot from 2017-09-18 17-32-07

bryphe:
Address a couple of issues with the menu:
- Use the `foregroundColor` and `backgroundColor` from the current color theme, and no hardcoded colors.
- Use `em` instead of `px` for font-sizes here, since `px` doesn't work correctly for HDPI displays. We should use `pt` as the font sizes.

Remaining issues:
- [x] Update default font size to `pt` on Linux/Mac

### DIFF
--- a/browser/src/Config.ts
+++ b/browser/src/Config.ts
@@ -182,7 +182,7 @@ export class Config {
         "environment.additionalPaths": [],
 
         "statusbar.enabled": true,
-        "statusbar.fontSize": "11px",
+        "statusbar.fontSize": "0.9em",
 
         "tabs.enabled": true,
         "tabs.showVimTabs": false,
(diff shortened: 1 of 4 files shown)


bryphe:
Fixes #714

bryphe:
New look against the default `onedark` colorscheme:
image

And in `solarized8_light`:
image

Not perfect, but it is better than before, and an incremental step. Once there is more thorough sync'ing of the colorschemes, we'll be able to do a bit better with the highlight colors.

bryphe:
(Moving remaining fix - updating default font sizes - to a separate fix to improve font rendering).

PR #752, opened 2017-10-03
Fix #746 - Fix dead keys in command line / normal mode

### LINKED ISSUE #746: Dead keys do not work on command mode
The [dead key issue](https://github.com/bryphe/oni/issues/638) seems to have been fixed for the text editor (great job, by the way), but I still can't input them on `:` commands, which is a blocker for me.

bryphe:
Previously, we'd only enable the IME / composition when in insert mode. Looking at how other platforms behave (Iike vim with Hiragana on OSX), it seems the composition / IME is always enabled across modes.

This simplifies the logic, while addressing the issue with #746.

This also addresses a separate issue, where when typing near the far right edge of the screen, the entire app body would scroll - due to the input overflowing. This was addressed in 4ebe8d2.

### DIFF
--- a/browser/src/Editor/KeyboardInput.tsx
+++ b/browser/src/Editor/KeyboardInput.tsx
@@ -30,6 +30,12 @@ interface IKeyboardInputViewProps {
 }
 
 interface IKeyboardInputViewState {
+
+    /**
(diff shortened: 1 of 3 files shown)


bryphe:
Fix #746

PR #753, opened 2017-10-03
[WIP] Fix #541 - Font rendering improvements

### LINKED ISSUE #541: Display issue with status bar when changing font-size
Changing the font size on `config.js` refreshes the screen with a visual glitch in the status bar:

2017-07-16-224658_1356x85_scrot

It does not persist after closing and reopening the application.

bryphe:
There are still cases where the text is blurry during rendering, due to subpixel anti-aliasing. Setting the `{alpha: false}` flag helps in this case, but has the downside that we can no longer have a transparent background.

This is some experimentation around rendering improvements, if we deprecate the background.

This also opens up the ability  to cache words / letters as images, and use `context.drawImage` as opposed to `drawText` for frequently used tokens, which could potentially speed up rendering later. That strategy isn't feasible currently, because the text is blurry.


bryphe:
This also fixes #541

PR #754, opened 2017-10-04
Fix `ERR_FILE_NOT_FOUND` on startup with no background image

bryphe:
__Issue:__ There is a console error on startup when a background image isn't specified - this is because we still create a `div` with `background-image` set to null. This causes problems during debugging because it can get mistaken for an actual error, so fixing it will reduce noise.

__Fix:__ Only render the element with `background-image` if a background is actually specified.


TalAmuyal:
Thanks, that one bothered me a lot :P

PR #757, opened 2017-10-04
[WIP] Support `clangd` as an LSP client

TalAmuyal:
This is still a work in progress.

`clangd` works with the changes in this PR.
However, there are some issues:

# Completion completes too much (probably `clangd`-s fault)

For instance, when writing `namesp`, the first completion option is `namespace name`.
The `name` part is a place holder for the name of the namespace.
The issue here, is that selecting that option for completion adds the word `name`.
I suspect that this is due to a misuse of the `label` LSP field on the server side.

# Stdout used for debugging

`clangd` outputs a copy of the input and output as logs into stdout.
The format is more or less `HTTP-HEADER <-- request-msg reply-msg -->`.
As a temporary hack, a new flag is introduced that instructs which behavior should be used with the stderr of the specific LSP server (either log it as error or as debug).

Which brings me the the next issue: IDK TypeScript, so there is a run-time error with this flag.
It is always evaluated as undefined so I or'ed it with `true` until I figure out what's wrong (hints will be appreciated).

### DIFF
--- a/vim/core/oni-plugin-cpp/lib/index.js
+++ b/vim/core/oni-plugin-cpp/lib/index.js
@@ -8,6 +8,7 @@ const activate = (Oni) => {
 
     const serverOptions = {
         command,
+        stderrAsLog: true,
     }
 
     const getInitializationOptionsAsync = (filePath) => {
@@ -19,7 +20,6 @@ const activate = (Oni) => {
     }
 
     const client = Oni.createLanguageClient(serverOptions, getInitializationOptionsAsync)
-    const stderrAsLog = true
 }
 
 module.exports = {
--- a/vim/core/oni-plugin-cpp/lib/index.js
+++ b/vim/core/oni-plugin-cpp/lib/index.js
@@ -1,0 +1,27 @@
+const fs = require("fs")
+const path = require("path")
+const childProcess = require("child_process")
+
+const activate = (Oni) => {
+
+    const command = Oni.configuration.getValue("cpp.langServerCommand", "clangd")
+
+    const serverOptions = {
+        command,
+    }
+
+    const getInitializationOptionsAsync = (filePath) => {
+        return Promise.resolve({
+            clientName: "cpp",
+            disableDocumentSymbol: false,
+            rootPath: "file:///" + filePath,
+        })
+    }
+
+    const client = Oni.createLanguageClient(serverOptions, getInitializationOptionsAsync)
+    const stderrAsLog = true
+}
+
+module.exports = {
+    activate,
+}
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,7 @@ yarn.lock
 $LOCALAPPDATA
 
 # Output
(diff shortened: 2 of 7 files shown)


bryphe:
Wow, excellent! Thanks for testing out the `clangd` integration, and opening the PR. I was just working on some LSP stuff so it's helpful to know about the flag you needed to add to get it working.

> For instance, when writing `namesp`, the first completion option is `namespace name`.

Interesting - would it be possible to see the the value we get for the completion? If you build One with `npm run build-debug` and put a breakpoint here: 
https://github.com/bryphe/oni/blob/33c0350cab8a6661c6600d09bb8a2f36ba65137e/browser/src/Plugins/Api/LanguageClient/LanguageClient.ts#L278

It would be helpful to see the value of `items` in this case.

> Which brings me the the next issue: IDK TypeScript, so there is a run-time error with this flag.

Hmm, the flag looks ok (and the code using it), but I'm wondering how it gets passed in? It looks like there might be a file missing (the `lib\index.js` that registers the language client). Could you please add and push that up? I can help you troubleshoot for sure.

vim/core/oni-plugin-cpp/lib/index.js
```diff
@@ -0,0 +1,27 @@
+const fs = require("fs")
+const path = require("path")
+const childProcess = require("child_process")
+
+const activate = (Oni) => {
+
+    const command = Oni.configuration.getValue("cpp.langServerCommand", "clangd")
+
+    const serverOptions = {
+        command,
```
bryphe: Regarding the `stdErrAsLog` issue - it looks like it's picked up as part of these options, so you'll need to include it as part of this object:

```
const serverOptions = {
   command,
   stdErrAsLog: true,
}
```

I think that should address the issue you had where it wasn't getting set / defined in the language client code

bryphe:
Just added a comment on passing the `stdErrAsLog` variable through... let me know if it's still not working. Once that's set, I'll bring this in!

TalAmuyal:
Re-based on master, now can't build Oni :(
https://github.com/bryphe/oni/issues/765

TalAmuyal:
Since #765 is not always in the way, I could resume :D 

The flag now works as expected, thanks!

Regarding the completion issue, I took a deeper look at it and the issue is on Oni's side.

The protocol states that `label` should be used unless `insertText` is present.

The first entry in the response that `clangd` sends is: `{"label":"namespace name = namespace","sortText":"00040namespace","filterText":"namespace","insertText":"namespace"}`

So I made a few changes and now it works.

**Please review the changes**

bryphe:
Looks great @TalAmuyal ! Thanks for debugging into the completion issue and fixing the root cause 👍 

Tested it against another language provider (the omnisharp/C# one), and it worked great. Bringing this in now.

bryphe:
Thanks for the contribution, @TalAmuyal ! I'll create a new release today including this change.

PR #760, opened 2017-10-05
[WIP] Add ability to specify language servers from configuration

bryphe:
This change starts adding the ability to specify language servers in the following form:
ie, `'language.go.languageServer.command': ...`

This is just preliminary scaffolding to load the clients from the configuration. I'm also exploring modularizing the LSP interface, so that things like completion/quickinfo can be separate modules and not so tightly coupled (and move away from the weight of the `Channel` and `PluginManager`...)

The current LSP client also has an issue where it depends on `Oni.on` events, which will be deprecated. 

This will also expose the language protocol interface through the Oni API (which means it can be debugged and tested in a REPL, which will make it easier to iterate and add new functionality).


vim/core/oni-plugin-python/lib/index.js
```diff
@@ -13,11 +13,7 @@ const activate = (Oni) => {
     const getInitializationOptionsAsync = (filePath) => {
         return Promise.resolve({
             clientName: "python",
-            rootPath: "file:///" + filePath,
-
-            // workaround for:
-            // https://github.com/palantir/python-language-server/issues/73
-            disableDocumentSymbol: true,
+            rootPath: "file:///" + path.dirname(filePath).split("\\").join("/"),
```
bryphe: need to revert this

PR #769, opened 2017-10-07
[WIP] Fix #484 - Update to use asynchronous strategy for menu window

### LINKED ISSUE #484: can't open a new window after quit (macos)

bryphe:
__Issue:__ Many of the menu options rely on there being an Oni window active - which isn't always the case in OS X. 

__Fix:__ Extract out an asynchronous strategy for fetching the current window. This is basically a `getOrCreate` - if there is an existing window, we use that, otherwise we create a new one.

There is an open issue though - there are some menu options that make sense to have when there is no window open (like new file, or open file), but there are also some that _only_ make sense in the context of a window or active editor context - like saving a file. As part of this, I'm investigating having aspects of the menu be controlled by the active editor - so in the case where a window is not active, we'd show a limited set of hardcoded options. However, I'd like to integrate this with the set of commands, and have the active editor "push up" available commands (so that even plugins could register their own commands). I think it makes sense to have the browser window be able to augment this set of commands, by exposing an `ApplicationMenu` concept on the renderer side, and hooking in the command manager to it (so as part of the metadata of a command, the menu path could also be specified..)

### DIFF
--- a/main/src/menu.ts
+++ b/main/src/menu.ts
@@ -2,7 +2,9 @@ import * as os from "os"
 
 import { app, dialog, Menu, shell } from "electron"
 
-export const buildMenu = (mainWindow, loadInit) => {
+export type BrowserWindowFunction = () => Promise<any>
+
+export const buildMenu = (mainWindowFunction: BrowserWindowFunction, loadInit) => {
     const menu = []
 
     // On Windows, both the forward slash `/` and the backward slash `\` are accepted as path delimiters.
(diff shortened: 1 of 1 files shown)


bryphe:
This is addressed by #1246 now - closing it out!

Issue #704, opened 2017-09-12
Improved error experience for configuration errors


extr0py:

extr0py:
Emacs potentially has similiar concerns - I wonder how they smooth this out?

bryphe:
Per discussion in #771 - there are cases where configuration options may be in conflict, and it would be helpful to surface them in this UI. An example is the `editor.fullScreenOnStart` and `editor.maximizeScreenOnStart` options - the `editor.fullScreenOnStart` takes precedence, so it may be helpful to let the user know that `editor.maximizeScreenOnStart` would take precedence in this case.

PR #771, opened 2017-10-07
Add option to maximize Oni on startup.

CrossR:
The current `editor.fullScreenOnStart` is useful, but other times I just want Oni maximised, rather than fullscreen, so I've added an option for that.

Simple change so shouldn't need much changing, I'm just interested in how it should interact with `editor.fullScreenOnStart`.

Maximising then going full-screen is a valid operation, whereas going full-screen and then maximising the window will have no effect. So I've put full-screen second for now, and as such it will override maximising the screen.

In reality, you'd probably want some mechanism in the config file to warn you that you have both?

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -57,6 +57,7 @@ const BaseConfiguration: IConfigurationValues = {
     "editor.scrollBar.visible": true,
 
     "editor.fullScreenOnStart": false,
+    "editor.maximizeScreenOnStart": false,
 
     "editor.cursorLine": true,
     "editor.cursorLineOpacity": 0.1,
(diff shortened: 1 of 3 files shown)


badosu:
I agree with this change (as mentioned on https://github.com/bryphe/oni/issues/452#issuecomment-332699733), it's a better use case for many window managers as opposed to full screen,

bryphe:
Excellent, this looks great! I prefer the `maximized` vs `fullscreen` state too in some environments (like OSX),... I'll be able to use this in my config there.

> Maximising then going full-screen is a valid operation, whereas going full-screen and then maximising the window will have no effect. So I've put full-screen second for now, and as such it will override maximising the screen.

This sounds reasonable. If both options are set, I would expect the full-screen would take precedent.

> In reality, you'd probably want some mechanism in the config file to warn you that you have both?

I like this idea. I added some notes in the context of #704 - which I opened to track some better experiences around troubleshooting configuration errors. This would be a natural place to surface up 'warnings' or conflicts like these.

bryphe:
Just tested; works great! Bringing it in now. Thanks for the contribution, @CrossR !

I'm planning on doing a release in the afternoon w/ this included.

PR #777, opened 2017-10-10
Fix #473: Open binaries crashes Oni

### LINKED ISSUE #473: Oni open binaries will cause crash

bryphe:
__Issue:__ When opening binaries (or any large file), Neovim crashes on windows. 

__Defect:__ When entering the buffer, we send a massive payload (the entire buffer contents) from Neovim -> Oni. This causes crashes in the large binary case, and may be a contributing factor to #490 as well.

__Fix:__ 
- Instead of a _push_ model, where Neovim always pushes buffer updates, switch to a _pull_ model, where Oni requests the buffer contents when necessary.
- Since Oni now decides when to _pull_ the buffer contents, we can put a hard limit on it - in this case, we'll only take in buffer updates if the file is <5000 lines.

> __NOTE:__ For incremental updates, we always send the individual line - that is still a _push_, but the data payload is much smaller. Since the latency is important (to show suggestion results quickly as the users types), it seemed reasonable to keep the incremental updates as-is.

### DIFF
--- a/browser/src/Services/BufferUpdates.ts
+++ b/browser/src/Services/BufferUpdates.ts
@@ -5,7 +5,7 @@
  * to the plugins. Sanitizes and manages incrementental state.
  */
 
-import { IIncrementalBufferUpdateEvent, IFullBufferUpdateEvent, INeovimInstance } from "./../neovim"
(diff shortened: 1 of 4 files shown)


bryphe:
Fixes #473

PR #779, opened 2017-10-10
Add link to syntax/compilation errors section of wiki

keforbes:
It seemed weird for 'Syntax / Compilation Errors' to be the only Feature without a corresponding wiki link so I added a section to the wiki and added a link in the README.


bryphe:
Nice, thanks for adding this @keforbes ! Looks much better now.

PR #780, opened 2017-10-10
Fix #713: Replace 'px' with 'em' for statusbar

### LINKED ISSUE #713: Adjust height of the statusBar
On a high DPI display the statusbar is almost illegible. See:

screenshot from 2017-09-18 14-28-51

bryphe:
__Issue:__ Previous fixes replaced several cases where we were using `px` instead of `em` (which cause problem in HDPI displays). However, the actual height of the statusbar was still using `px`, so even though the font height was fixed, the status bar height wasn't.

__Fix:__ Replace the `28px` with an `em` value.

### DIFF
--- a/browser/src/UI/components/StatusBar.less
+++ b/browser/src/UI/components/StatusBar.less
@@ -5,7 +5,7 @@
 
     color: @text-color;
     background-color: @background-color;
-    height: 28px;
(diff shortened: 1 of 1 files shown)


bryphe:
Fixes #713

Issue #436, opened 2017-05-02
init.vim on Windows


gglanzani:
Hi,

I noticed that on Windows `Oni.exe` looks for the init file by looking at  `~\oni\$MYVIMRC` instead of `~\oni\init.vim` (assuming `Oni.exe` lives in `~\oni`. Is this know behavior to scare users off? :)

extr0py:
Hi @gglanzani ,

Thanks for the issue! Ah yes, this is related to #332 .

If you want to use an `init.vim` file with Oni, you need to set the `oni.loadInitVim` value to true. You can do this as follows:
1) Open Oni
2) File -> Preferences - Edit Oni Config
3) Add an entry as follows:
```
...
"oni.loadInitVim": true,
...
```

Without this setting, `$MYVIMRC` isn't populated correctly. 

Hope that helps!

extr0py:
It'd be nice to streamline this on Windows in the feature, by detecting this case and showing some actionable message to the user.

gglanzani:
@extr0py That is set :)

And when that is set, and I click on Edit Neovim config, it opens `~\oni\$MYVIMRC` instead of `~\oni\init.vim`

extr0py:
Weird! With that flag set, it should be able to find $MYVIMRC.

I just updated the binary for Neovim to 0.2.0 in the repo - you might want to try from the latest master. That fixes several Windows issues.

Are you using a released build, or did you build from source?

gglanzani:
@extr0py I've overwritten the neovim files (bin, share, and everything below that) with the 0.2.0 release :)

keforbes:
> @extr0py I've overwritten the neovim files (bin, share, and everything below that) with the 0.2.0 release :)

Did that resolve the issue or is `$MYVIMRC` still not set for you?

gglanzani:
@keforbes Yes, that worked (apparently)

Congee:
I still have the issue with `"oni.useDefaultConfig": false, "oni.loadInitVim": true`
BTW, I use the latest v0.2.5 release, with the binary installer `Oni-0.2.5-ia32-win.exe`.
:/

extr0py:
Thanks for the report, @Congee . Bummer - it looks like there are still cases where somehow the `$MYVIMRC` variable isn't set. It might be more reliable for Oni to directly open the `init.vim` directly instead of relying on that variable.

Congee:
@extr0py Or make the path customizable. 🤔

[...]

PR #784, opened 2017-10-11
[WIP] Allow the passing of a custom init.vim path.

CrossR:
This PR should allow the defining of a custom init.vim path in the Oni config file.

I need to check into some error checking, ie if an incorrect path is given, as well as some testing outside of Windows (Mac or Linux users would be useful, since I've only got Window + a few Linux VMs).

Also, is it acceptable to assume the user will sufficiently escape the path they provide?
Ie if they are on Windows and their init.vim is at `"C:\Users\CrossR\init.vim"`, they would need to set it with escaped back slashes as follows:

`"oni.customInitVimPath" : "C:\\Users\\CrossR\\init.vim"`.

Also since this is applied as an argument at startup, setting a custom path isn't applied until restarting oni currently. Is it worth looking into applying it straight away, or should that be a separate PR?

### DIFF
--- a/browser/src/Services/Commands.ts
+++ b/browser/src/Services/Commands.ts
@@ -53,7 +53,7 @@ export const registerBuiltInCommands = (commandManager: CommandManager, pluginMa
         // Menu commands
         new CallbackCommand("oni.config.openConfigJs", "Edit Oni Config", "Edit configuration file ('config.js') for Oni", () => openDefaultConfig(neovimInstance)),
 
-        new CallbackCommand("oni.config.openInitVim", "Edit Neovim Config", "Edit configuration file ('init.vim') for Neovim", () => neovimInstance.open("$MYVIMRC")),
+        new CallbackCommand("oni.config.openInitVim", "Edit Neovim Config", "Edit configuration file ('init.vim') for Neovim", () => neovimInstance.openInitVim()),
 
         new CallbackCommand("oni.openFolder", "Open Folder", "Set a folder as the working directory for Oni", () => openFolder(neovimInstance)),
 
--- a/browser/src/neovim/NeovimInstance.ts
+++ b/browser/src/neovim/NeovimInstance.ts
@@ -326,12 +326,12 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
     }
 
     public openInitVim(): Promise<void> {
-        const customInitVimPath = configuration.getValue("oni.customInitVimPath")
+        const loadInitVim = configuration.getValue("oni.loadInitVim")
 
-        if (customInitVimPath) {
-            return this.command(`e! ${customInitVimPath}`)
+        if (typeof(loadInitVim) === 'string') {
+            return this.open(loadInitVim)
         } else {
-            return this.command(`e! $MYVIMRC`)
+            return this.open('$MYVIMRC')
         }
     }
 
--- a/browser/src/neovim/NeovimInstance.ts
+++ b/browser/src/neovim/NeovimInstance.ts
@@ -109,6 +109,7 @@ export interface INeovimInstance {
     getApiVersion(): Promise<INeovimApiVersion>
 
     open(fileName: string): Promise<void>
+    openInitVim(): Promise<void>
 
     executeAutoCommand(autoCommand: string): Promise<void>
 }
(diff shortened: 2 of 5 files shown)


CrossR:
This could help with #436.

browser/src/neovim/NeovimInstance.ts
```diff
@@ -324,6 +325,16 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
         return this.command(`e! ${fileName}`)
     }
 
+    public openInitVim(): Promise<void> {
+        const customInitVimPath = configuration.getValue("oni.customInitVimPath")
+
+        if (customInitVimPath) {
+            return this.command(`e! ${customInitVimPath}`)
```
CrossR: I should probably refactor this to just use the open command, so the logic isn't duplicated.

browser/src/Services/Commands.ts
```diff
@@ -53,7 +53,7 @@ export const registerBuiltInCommands = (commandManager: CommandManager, pluginMa
         // Menu commands
         new CallbackCommand("oni.config.openConfigJs", "Edit Oni Config", "Edit configuration file ('config.js') for Oni", () => openDefaultConfig(neovimInstance)),
 
-        new CallbackCommand("oni.config.openInitVim", "Edit Neovim Config", "Edit configuration file ('init.vim') for Neovim", () => neovimInstance.open("$MYVIMRC")),
+        new CallbackCommand("oni.config.openInitVim", "Edit Neovim Config", "Edit configuration file ('init.vim') for Neovim", () => neovimInstance.openInitVim()),
```
bryphe: I like that you factored this out to a separate method, thanks!

browser/src/neovim/NeovimInstance.ts
bryphe: Good idea 👍

bryphe:
Cool, thanks for submitting this @CrossR ! I definitely think it makes sense to be able to override the `init.vim` path. There are some cases where users would prefer to have a completely separate `init.vim` between Oni/Neovim, so this helps accomplish that.

The only concern I have is that the configuration settings we have today are already confusing - `oni.loadInitVim`, `oni.useDefaultConfig`, and `oni.customInitVimPath`. I'd like to see if we could consolidate the `oni.loadInitVim` and `oni.customInitVimPath` configuration settings, if possible (since setting `oni.customInitVimPath` would imply `oni.loadInitVim` is `true`).

A couple of options I'm thinking about:
- _Change `oni.loadInitVim` to support `string | boolean`_. If it is `true`, we would just load `init.vim` from its usual place. If it is a `string`, we'd load it from the custom path.
- _Only support custom paths_ - We could make `oni.customInitVimPath` be the way to using an `init.vim`, whether it is in the default place or not, and deprecate `oni.loadInitVim`. The downside is it would hose users that use `oni.loadInitVim`, we'd have to make it clear we are deprecating that value and how to upgrade - it would also be a bit more inconvenient for users who pick up their config in the default place, because they might have to look up where that is.

I'd lean towards the first one, just because it is simplest... but I'm sure there are other ways we could consolidate this and simplify the relationship between `oni.loadInitVim`/`oni.customInitVimPath`. Let me know what you think!

CrossR:
> The only concern I have is that the configuration settings we have today are already confusing - oni.loadInitVim, oni.useDefaultConfig, and oni.customInitVimPath. I'd like to see if we could consolidate the oni.loadInitVim and oni.customInitVimPath configuration settings, if possible (since setting oni.customInitVimPath would imply oni.loadInitVim is true).

Yep, totally agree!
I did find myself thinking this as I was coding it, but wasn't too sure if it was the right way to go, though I hadn't considered letting `oni.loadInitVim` accept either a string or a boolean. That makes the logic a lot cleaner, and clears up the user config.

I'll have a look at implementing it now.

CrossR:
Consolidated those two options together now and seems to work well!
Also swapped to using the built in open where I can, and updated some variable names to be more appropriate.

There is still some oddness with the interactions with `oni.useDefaultConfig`, but only the same that is in the live version, and this is outlined in the Wiki, so its possibly an issue to raise later.

bryphe:
Excellent, looks great! I just tried it and it works well - thanks for making the changes!

> There is still some oddness with the interactions with oni.useDefaultConfig, but only the same that is in the live version, and this is outlined in the Wiki, so its possibly an issue to raise later.

Definitely, I'd like to see how to consolidate this at some point. But as you mentioned at least this change doesn't introduce any new issues, the problems are pre-existing 😄 

I'll go ahead and bring this in. I made a quick edit to the [wiki](https://github.com/bryphe/oni/wiki/Configuration#oni) to detail that a `string` path can be accepted, but feel free to revise.

PR #787, opened 2017-10-12
Change quit to quit app instead of close current window

MikaAK:
Quit was previously only closing the current window, this meant that the application didn't quit as expected when on OSX. This fixes that issue and causes the whole application to quit.


bryphe:
Looks great, @MikaAK ! Thanks for investigating this and fixing it 👍 I'll bring it in now.

PR #789, opened 2017-10-12
Make QuickInfo standout on dark themes

josemarluedke:
I found difficult to read the quick info box on a dark vim color scheme, so I added a border color to it.
This PR is more a suggestion then the "final product", not sure if you have something else in mind to solve this problem.

## Now 
screen shot 2017-10-12 at 8 39 25 am

## Before
screen shot 2017-10-12 at 8 45 39 am

### DIFF
--- a/browser/src/UI/components/QuickInfo.less
+++ b/browser/src/UI/components/QuickInfo.less
@@ -9,6 +9,7 @@
 
     .quickinfo {
         .box-shadow;
+        border: 1px solid #333333;
 
         text-overflow: ellipsis;
         overflow: hidden;
@@ -21,6 +22,7 @@
 
         .documentation {
             .box-shadow-inset;
+            border-top: 1px solid #222222;
             padding: 8px;
             font-size: @font-size-small;
             min-width: 300px;


bryphe:
Hi @josemarluedke ,

Yes, good point, now that the background color of the quick info box matches the background, it's tough to see. I like your 'After' a lot - I think it does make sense to bring in the border. Definitely easier on the eyes! Thank you for proposing this and including the before/after.

The only thing I think we need to consider is how it looks against other colorschemes with the hardcoded color:
image

I wonder if we could use some variant of the `foreground` color as the border color? Or use a library like this https://www.npmjs.com/package/color - we could look at the contrast of the background color and pick between a light/dark color, for example? 

Let me know if you have any ideas here - otherwise I might experiment with a few things to see if I can get a border that looks good in both light/dark colorschemes.

bryphe:
Thanks for the idea & proposal, @josemarluedke ! I adapted for other colorschemes and brought it in with #798 , I'll go ahead and close this. 

BTW, which colorscheme do you use? Seems like it'd be a good test case for me to include.

josemarluedke:
@bryphe Thank you for working on this. I appreciate it. I'm using this color scheme: https://github.com/tomasiser/vim-code-dark

PR #794, opened 2017-10-12
Make notification handler be more immutable

cdelahousse:
This patch creates a new array array instead of mutating the existing array `a`.

While playing and debugging the neovim Session code, I noticed that a lot of the data I was consoling out from the notification event handler was missing the command name. It is because the arguments array returned from the notification event was being mutated in place by this code. Objects in the console are late bound, so they don't know their values at the time of logging.


bryphe:
Hi @cdelahousse ,

Thanks for the contribution. Seems reasonable - it might be worth looking at bringing your logging in too (if you think it would be useful in the general case). This whole function / paradigm could use some  refactoring to move it to a more functional paradigm (most of the `actions` that are emitted go to the `Screen` class, which updates its state in a mutable way). I'll go ahead and bring this in.

PR #798, opened 2017-10-14
Add border for QuickInfo UI

bryphe:
This was inspired by @josemarluedke 's PR #789 to make the QuickInfo UX more apparent against the background. I took a try at generalizing the border for different schemes (bringing in the `color` library to help derive a border color), and also cleaned up the arrow / beak a bit.

Here's how it looks with the changes on various colorschemes:
image
delek
solarized dark

I also made the 'beak' smaller and added a small entrance animation, to be in sync with the QuickInfo container (which has a 0.5s entrance animation).

### DIFF
--- a/browser/src/UI/components/Arrow.less
+++ b/browser/src/UI/components/Arrow.less
@@ -1,0 +1,25 @@
+@import (reference) "./common.less";
+
+.arrow {
+    animation-duration: 0.3s;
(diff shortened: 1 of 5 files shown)


bryphe:
Thanks for looking at this, @josemarluedke ! Bringing it in for 0.2.13

PR #799, opened 2017-10-14
Fix #436 - init.vim on Windows

### LINKED ISSUE #436: init.vim on Windows

bryphe:
This builds on the PR @CrossR put together for custom init.vim paths (#784).

Instead of editing `$MYVIMRC` which seems to have inconsistent results on all platform, I baked in the paths based on the Neovim documentation. This should give more reliable behavior. I also create the directory if it hasn't been created yet.


bryphe:
Fixes #436

PR #800, opened 2017-10-14
LSP: Add quick info and buffer updates to config-based strategy

bryphe:
This continues to build on the PR from #760, and expands the capabilities of config-driven language clients.  In particular, this adds `Find all References`, `Hover` (QuickInfo), diagnostics, and sends buffer updates.

This is putting parallel scaffolding in place for the existing language client and plugin manager, but the goal is once there is parity (completion is the only outstanding item), we can flip to using this strategy exclusively. 

This also modularizes the language capabilities, so it is easier to add and support new features in the language server protocol, as well as making it available in the API surface.

This also adds an `activeEditor.activeBuffer` API as well as an `onBufferChanged` event.


browser/src/Editor/NeovimEditor.tsx
```diff
@@ -66,11 +81,18 @@ export class NeovimEditor implements IEditor {
...
 
     private _isFirstRender: boolean = true
+    private _currentBuffer: Oni.Buffer = null
 
     public get mode(): string {
         return this._currentMode
     }
 
+    public get activeBuffer(): Oni.Buffer {
+
```
bryphe: Clean up whitespace here

browser/src/Services/Language/FindAllReferences.ts
```diff
@@ -0,0 +1,99 @@
...
+        const references = convertLocationToReferences(result)
+        showReferencesInQuickFix(references, activeEditor.neovim as any)
+    } else {
+        pluginManager.findAllReferences()
+    }
+}
+
+const convertLocationToReferences = (result: types.Location[]): Oni.Plugin.ReferencesResult => {
+
+    // TODO: Get symbol under cursor
```
bryphe: Bring this back, now that the buffer updates are available

Issue #395, opened 2017-04-20
Weird scrolling behavior when `scrolloff` is set to a non-zero value


ohle:
### To reproduce:
 - In a buffer with more lines than fit in the window, `:set scrolloff=1`
 - Move using <kbd>j</kbd> or <kbd>k</kbd>

### Expected behavior:
The cursor moves down or up, but the content of the window doesn't change until the cursor reaches the penultimate line when moving down or the second line when moving up

### Observed behavior:
When starting from the top, as soon as the cursor is on the second line, the content starts to scroll with every move. When the cursor is in a lower line at the time the setting is changed, moving up also causes weird scrolling

This is using a current checkout (d9e6571) and neovim 0.1.7

keforbes:
Yep, I can reproduce this.  That's some weird behavior.

LER0ever:
I can reproduce this too. I have scrolloff 3 in my vimrc and oni just moves the content at every 5 j scrolldown.

LER0ever:
Actually, if scrolloff is not zero, another weird scrolling occurs when you scroll from the bottom of the file up using mousewheel. The content just keeps glitching while going up. @ohle Can you reproduce this one? or it's just my configuration problem?
I've temporarily [disabled scrolloff](https://github.com/LER0ever/EverVim/commit/d51c89814d3da74af9f79ca0f0f1449cb9f9cf56) if Oni is detected.

josemarluedke:
I'm also having this problem. Took me a while to find this issue here to see what the problem was. Looking forward to have this fixed.

extr0py:
Put a bounty on this in case someone is interested in checking it out, as it sounds like it is impacting several people:
https://www.bountysource.com/issues/44356173-weird-scrolling-behavior-when-scrolloff-is-set-to-a-non-zero-value

There are a few places to look in terms of investigation:
- `browser\src\neovim\NeovimInstance.ts` - check out the `msgpack-rpc` messages we receive from Neovim in this scenario
- `browser\src\Screen.ts` - manages the state of the screen, and it can actually be queried at runtime
- `browser\src\Renderer\CanvasRenderer.ts` - if the state of the screen is updating correctly, then it's likely a problem with the render strategy.

CrossR:
I'm going to have a bit of a look into this, and see what is causing it.

[...]

PR #802, opened 2017-10-16
Fix ScrollOff Issue.

CrossR:
There looks to be some form of odd issue with using `winline()` in the Oni interop plugin for the Window  display map.
It causes some very odd scrolling, as outlined in #395.

Updating it to get the current pos and use that line instead seems to work fine, but there may be a better way of doing it. 

This change brings the experience inline with NVIM-QT, at least as far as I can tell.
Anyone who uses scrolloff more than me should probably test this.

### DIFF
--- a/vim/core/oni-core-interop/plugin/init.vim
+++ b/vim/core/oni-core-interop/plugin/init.vim
@@ -139,7 +139,9 @@ function OniUpdateWindowDisplayMap(shouldMeasure)
     while(cursor <= windowEndLine)
         call setpos(".", [bufNum, cursor, 0])
         let cursorString = string(cursor)
-        let mapping[cursorString] = winline()
+        let newPos = getpos(".")
+        let newLine = newPos[1]
+        let mapping[cursorString] = newLine
         let cursor = cursor+1
     endwhile
 


bryphe:
Hi @CrossR ,

Thanks for submitting the PR! That's really helpful that you narrowed down the `scrolloff` issue as being related to the `winline` call. I just looked at `:help winline` and didn't realize it has side effects... but it definitely does and can impact scrolling. 

The way this code works is it creates a mapping of `buffer line` -> `window line`. The `cursorString` is the buffer line, and the `winline()` call gave us the window line. This mapping is then used by the `WindowContext` to help map 'buffer space' coordinates to 'screen space' coordinates.

Unfortunately, changing this to `getpos(".")` breaks this - the mapping won't work correctly in some cases. In particular, one case that is broken is rendering the error squiggles if you've scrolled down in a file - these will never be drawn because the coordinates will always be outside of 'screen space' without the `winline()` mapping.

This whole pipeline is pretty hacky and convoluted - really what this is building for us is one function - a function that maps _buffer space_ (like a `line` and `column` in a buffer) into _screen space_ - and I suspect there is a way to simplify this without all the back-and-forth events and crazy iteration in `init.vim`.

I'm trying to explore right now if there is an alternative fix - instead of needing to have this `OniUpdateWindowDisplayMap` in `init.vim`, perhaps we can do a `pull` instead from the Oni side, and based on the info we already have in the event context, and the state of the buffer, maybe we can simplify and build up that `bufferSpaceToScreenSpace` function. If we go this route, we could get rid of the hacky iteration + `winline` code that is causing issues with `scrolloff` (and I suspect it is causing some other problems as well), and have a more robust solution in general.

CrossR:
Seems like that would be a much better implementation, it was awkward to decipher the code in the init.vim, plus as its over in the neovim side of thing, it was awkward to follow the calls back and debug the issue as a whole. At least we have an idea of what seems to be causing the issue!

I'll close up this PR and move some of the discussion over into the issue for this(#395), so that its still easily accessible.

PR #804, opened 2017-10-17
Fix #782 - Fix overlapping errors

### LINKED ISSUE #782: Overlapping syntax / compilation errors
I'm sure it is a known issue, but I think it should be logged.

When having more then one syntax/compilation errors on the same line (and the cursor is on that line), the tool-tips usually overlap which makes it really hard to read.

bryphe:
WIP for #782


bryphe:
Here's what the updated UI looks like: 
image

Basically integrated into the QuickInfo pane

TalAmuyal:
Looks great!

Issue #543, opened 2017-07-17
Scrollbar not updated after cursor movements (zz, zt, zb)


badosu:
After using one of the cursor movements (zz, zt or zb), the scrollbar on the right side is not visually updated.

See:

![2017-07-17-133036_1365x228_scrot](https://user-images.githubusercontent.com/347552/28278643-26381eb8-6af4-11e7-8962-cec5dbb60e69.png)

extr0py:
Nice catch! I'm able to repro this.

It looks like most of the autocmds we hook watch for cursor being moved - but in the case where the window scrolls, and the cursor doesn't move, the scrollbar isn't updated.

A potential fix is to update the scroll state when we get the `scroll` action during `redraw`.

cyansprite:
It also doesn't update the linting errors.

extr0py:
I have a partial fix - but it turns out there isn't a great way currently to detect this case. The autocmds we listen to work fine if the cursor moves, but in the case where the screen scrolls but the cursor _does not_ move, it's a bit less reliable.

Ideally, there'd be a `Scroll` autocmd we could listen to in order to give us this info. There's a tracking issue in neovim/neovim#4322

cyansprite:
An idea...
it's not a fix but more of a hack, but if we get the keys pressed to do with scrolling, we can catch those keys, then send them to vim (like we do with all the other keys) and then after wards, redraw the screen.

extr0py:
> An idea...
> it's not a fix but more of a hack, but if we get the keys pressed to do with scrolling, we can catch those  keys, then send them to vim (like we do with all the other keys) and then after wards, redraw the screen.

Ya, in a similiar vein, we could do something like:
``` nnoremap zt :Call ExecAndNotifyOni('zt')<CR>```

Like run the command on the Vim side, and then notify Oni that there was an update

bryphe:
This was fixed by #805

wellle:
@bryphe: I believe this is still not fully fixed, unfortunately.

To reproduce open a sufficiently large file and go to the middle of it by typing `50%`. Try alternate typing `zt` and `zb`. The visible lines in the buffer update accordingly, but the scroll bar only updates a single time. Interestingly if you type `zz` it updates both the visible lines and the scrollbar properly, and after that the next `zt` or `zb` works too. But if you start alternating them again they won't update the scrollbar anymore.

I tried it on `oni version 0.3.9` which seems to be the latest.

PR #805, opened 2017-10-18
[WIP] Fix #395 & #543 - Factor window management out of `init.vim`

### LINKED ISSUE #395: Weird scrolling behavior when `scrolloff` is set to a non-zero value

bryphe:
__Issue:__ There were several issues relating to window management - #395 being a really strange issue, where the `scrolloff` setting wouldn't behave as expected.

__Defect:__ For Oni's rich-UI integration, we need to often map _buffer_ positions to _screen_ and _pixel_ positions. The _buffer_ -> _screen_ mapping is dependent on the size of the window, where the window/viewport is scrolled, etc. 

The way we previously calculated this was in the `OniUpdateWindowDisplayMap` function in `init.vim`. This would iterate through each _window line_ and record a mapping of _window line_ -> _buffer_. With this dictionary, we could then derive the _buffer_ -> _screen_ mapping. 

This method was problematic, because it depended on the `winline` function, which has side effects and can impact scrolling. Doing this iteration frequently could potentially have performance impact, as well.

__Fix:__ The main item this fix achieves is doing the majority of this mapping on the Oni client side, deriving the function based on the window dimensions, the top/bottom buffer line, and the cursor position. There are several pieces in particular:
- __Remove__ the `OniUpdateWindowDisplayMap` from `init.vim` and all the back-and-forth events.
- __Update__ `NeovimWindowManager` to listen to certain events and recalculate the mapping functions based on the dimensions.
- __Formalize__ the relationship between `buffer space`, `screen space`, and `pixel space` in `Coordinates.ts`, and use strong-types to assist.
- __Remove__ the `WindowContext` object that was over-complicated and required us storing additional state - this can be completely replaced with the `bufferToScreen` and `screenToPixel` transforms.
- __Refactor__ the `Error` component to use the new transform functions, as opposed to WindowContext
- __Refactor__ the `BufferScrollBar` component to use the new transform functions
- Add `NeovimAutoCommands` to make it easy for consumers of `NeovimInstance` to suscribe to relevant autocommands
- Add an `onScroll` event to help with #543 

One item that will need to be addressed is that the strategy we use to build up the buffer -> screen map doesn't handle multibyte characters.


bryphe:
Fixes #395 and #543

PR #807, opened 2017-10-19
UX: Autocompletion styling

bryphe:
These are some tweaks to autocompletion, to make it more consistent with the work done for the Menu, QuickInfo, and Signature help - this leverages the foreground / background colors, so that it doesn't look so out of place:

image

And it meshes better with other color schemes:
image


jordwalke:
Looks much more consistent!

Issue #214, opened 2017-02-12
Completion Improvement: Reduce chatter with the plugin


bryphe:
Right now, autocompletion is invoked *for every key press*. This is really expensive, especially since there are usually only a few places we need to get the 'full list'. We can make this less chatty by only grabbing completions at particularly places - for example, the first letter pressed after a space, and the `.` character (or whatever other characters are called out by the service).  When we see those, we can request completions there.

Then, as we are typing, we look back to see if the word meets one of the existing completions, and filter those completions based on the substring.

For this, we'd need the following:
- Add a concept in the language service meta about the characters that should trigger completion (`.` in typescript, `->`, etc)
- Store the last completion we received for that position (in redux store). It should include the buffer, prefix, and matching string.
- As we type, the `AutocompletionService` makes requests to the plugin language service
- As the `AutocompletionService` gets responses, it dispatches actions to the UI to store autocompletion entries at points.
- The completion details would still work as-is - we'd invoke that as items are selected.

bryphe:
This was addressed by #819 - we now only query at the meet, and only if the meet has changed.

PR #810, opened 2017-10-19
[WIP] Reduce chatter in completion (#214) and enable for new language clients

bryphe:


bryphe:
Merged into `bryphe/features/lsp-v2`

PR #818, opened 2017-10-21
Fix #817 - Scroll bar / cursorline not in correct window

### LINKED ISSUE #817: OSX - Not scroll bar or cursorline in correct window
screen shot 2017-10-20 at 10 59 46 pm

The scrollbar just doesn't exist in the right window. It works fine in the left window.

Also the cursorline doesn't go to the right window.

There seems to be a ton of weirdness with splits that I can't document right now

bryphe:
__Issue:__ Window splits don't show up Oni's UI layers correctly (like the scroll bar, cursor line, error squiggles), they either don't show up at all or are in the wrong place.

__Defect:__ These are regressions from #805 , which changed the way we measure the window dimensions. There were a couple of issues:
- We were interpreting the `nvim_win_get_position` result as `[x,  y]` instead of `[row, col]` (which is inversed).
- There was a race condition - we would force a remeasure of the Windows when we detected a resize event on the Electron side, but there were cases where Vim hadn't actually completed the resize yet - so we'd get a stale value. 

__Fix:__ 
- Be more explicit about the return value of `nvim_win_get_position`, grabbing the `[row, col]` directly
- Wire up the `VimResized` event and use that in the `NeovimWindowManager`
- Listen to changes in buffer width / height
- Remove a few now-unused code paths


bryphe:
Fixes #817

Issue #820, opened 2017-10-23
"Edit with Oni" menu item


topher96:
Vim on Windows installs an "Edit with Vim" right click menu I always found very useful as it was on the top level menu and applies to every filetype.  Could you add a similar menu for Oni?

badosu:
I would just label this as specific to windows or mac if not available there, as the specification for the `.desktop` file we have for linux (on Arch at least) knows the MIME types oni can handle.

CrossR:
I've submitted a PR for adding this for Windows.

A second point I realised is that we could potentially also have this work for Folders as well as files.
Then you could "open a folder" with Oni, and it would open oni with the current folder set to that one.

keforbes:
This was resolved in #823, which has been merged in.  Closing issue.

bryphe:
Thanks @keforbes !

PR #823, opened 2017-10-23
Added optional task for adding Oni to right click menu on Windows

CrossR:
Updated the Windows Registry keys so that optionally Oni can be used from the right click menu for all files. 

Currently, I've only added the registry keys for opening a file, so you can right click any file and there will be an option to open it with Oni. 

We could look at also adding registry keys for opening a folder? I think code does this.
Then you would be able to Right click and open a folder with Oni and it would default the current directory to that folder.

Fixes the Windows portion of #820.

### DIFF
--- a/build/BuildSetupTemplate.js
+++ b/build/BuildSetupTemplate.js
@@ -42,7 +42,12 @@ const fileExtensions = {
 }
 
 const addToEnv = `
-Root: HKCU; Subkey: "Environment"; ValueType: expandsz; ValueName: "Path"; ValueData: "{olddata};{app}"; Tasks: addtopath; Check: NeedsAddPath(ExpandConstant('{app}'))`
+Root: HKCU; Subkey: "Environment"; ValueType: expandsz; ValueName: "Path"; ValueData: "{olddata};{app}"; Tasks: addtopath; Check: NeedsAddPath(ExpandConstant('{app}'))
+
+Root: HKCU; Subkey: "SOFTWARE\\Classes\\*\\shell\\${prodName}"; ValueType: expandsz; ValueName: ""; ValueData: "Open with ${prodName}"; Tasks: addToRightClickMenu; Flags: uninsdeletekey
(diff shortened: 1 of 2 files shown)


bryphe:
@CrossR - awesome, thanks for putting this PR together! Code looks good to me - I'll test it out in the morning tomorrow 👍 

The OSX build failure isn't due to your code - it seems like the `vscode-ripgrep` library gets rate-limited on PRs. I'll take a look at a fix for this, but in the meantime we can safely ignore that failure.

bryphe:
Awesome! I just built it locally and it is very cool to see it in the context menu:

CrossR:
Yeah thats fair enough, I wasn't sure what the best wording for it was.

VS Code does it as follows: 
image

I think that VS Code's wording is more clear, since "for text files" could be interpreted as only working with `.txt` files when really it works with `.txt, .js, .ts, .md`... and so on.

So maybe:

* Add Oni to %PATH%
* Register Oni as the default editor for supported file types
* Add Oni to the right-click menu for all files

As for combining them? Personally, I would prefer the choice even though I'll tick them all. It does sometimes annoy me how many programs you end up with on the right click menus that never asked permission to be there. I'd rather spend an extra second during an install and have that extra control.

bryphe:
Cool, thanks for sharing out the screenshot @CrossR ! Yes, seems like there is a precedent for setting up the options that way - your proposal sounds good to me:

> - Add Oni to %PATH%
> - Register Oni as the default editor for supported file types
> - Add Oni to the right-click menu for all files

> As for combining them? Personally, I would prefer the choice even though I'll tick them all. It does sometimes annoy me how many programs you end up with on the right click menus that never asked permission to be there. I'd rather spend an extra second during an install and have that extra control.

Agreed, combining them doesn't seem like a very good option - I like your latest suggestion the best.

CrossR:
I actually went with "Register Oni as an editor for all supported file types" in the end (the same as VS Code), as I realised since Windows 10 there is no way to force the default program, so saying "as the default editor" could be wrong, as if the user already has Notepad/Notepad++/VSCode etc set, we can't undo that. 

Instead when they select "Open With" they then get to check a box to select the new default and then at that point Oni becomes the default.

bryphe:
Sounds good, the text looks solid - thanks @CrossR ! Bringing this in now.

topher96:
Thanks!

PR #824, opened 2017-10-24
Fix #822: Improved OSX defaults for QuickOpen and Command Palette

### LINKED ISSUE #822: Defaults for OSX should allow c-p, c-n to move up down menu.
It's common for OSX editors, (and chrome even) to allow navigating up/down in modal lists via `c-n`/`c-p`. In Oni, the default keybindings to open the various command pallets in Oni is `c-p`, and `c-shift-p`. This means that once the modals are opened, `c-n` can be used to move down, but `c-p` cannot be used to move up because that's the keybinding for opening the overlay in the first place.

I recommend using the Sublime Text default shortcuts, which have inspired other editors such as VSCode.
[Here's](http://docs.sublimetext.info/en/latest/reference/keyboard_shortcuts_osx.html) the OSX keyboard reference for Sublime, which has the overlay windows opening with command+p, and command+shift+p, so that the menu may be navigated with control+p, control+n for down/up respectively.
screen shot 2017-10-22 at 10 25 50 pm

I notice the trend of having `cmd+p`, and `cmd+shift+p` be OSX shortcuts for two types of typeahead overlays, so going by the same pattern in Oni will make people feel instantly comfortable in Oni.

bryphe:
__Issue:__ Most editors on OS X (like VSCode, Sublime, Atom), use `<Command+P>` for file finder and `<Command+Shift+P>` for the command palette. Oni uses the same defaults as Windows, `<Control+P>` and `<Control+Shift+P>`, which is not very intuitive, and has the other downside that `<Control+P>` is double-booked as both previous entry and opening the menu.

__Fix:__ Change the defaults on OS X to use `<Command+P>` and `<Command+Shift+P>`. In addition, there was a bug where keys weren't resolved properly if `Command` and `Shift` were both pressed, so that was fixed to unblock the scenario.

### DIFF
--- a/browser/src/Input/KeyBindings.ts
+++ b/browser/src/Input/KeyBindings.ts
@@ -8,7 +8,6 @@ import * as Platform from "./../Platform"
 import { Configuration } from "./../Services/Configuration"
 
 export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configuration): void => {
-
     const { editors, input } = oni
 
     input.unbindAll()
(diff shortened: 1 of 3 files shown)


bryphe:
Fix #822

Issue #783, opened 2017-10-11
Adjust tab header scaling issues


badosu:
In line with the work being done to scale Oni's layout there's an issue that has been bothering me for some time.

If you use a bigger font size the tab header will not scale well, see:

![screenshot from 2017-10-11 12-27-59](https://user-images.githubusercontent.com/347552/31450472-e3132810-ae7f-11e7-8075-bf3238ff196f.png)

One possible solution is to use scalable units for the height and max-width of the tab element, I can take a stab at this if necessary.

bryphe:
Ah yes, thanks for logging this! It's a similiar issue to what we saw with the status bar - there are a few hardcoded `px` values `Tabs.less`:

https://github.com/bryphe/oni/blob/1ad48df21ea1b62fcb51cd15df8c9a34420fdb4b/browser/src/UI/components/Tabs.less#L44

I think if we change the `max-height`, `min-height`, and `height` values to use `em` instead of `px`, that should solve the problem. 

If you want to give it a go @badosu , that would be great! I'll assign it to you for now - but LMK if you have any questions

bryphe:
I merged in #832 - with the following configuration settings:
- `tabs.height` - set the size of individual tabs
- `tabs.maxWidth` - set the maximum width a tab can expand to
- `tabs.wrap` - set to `true` if you want the tabs to wrap, otherwise there will be a horizontal scrollbar for overflow.

I'll close for now, but keep me posted if they aren't behaving as expected. Thanks for the PR and brainstorming around this, @badosu !

PR #827, opened 2017-10-25
Make tabs scale vertically with font size

badosu:
A partial fix for https://github.com/bryphe/oni/issues/783

As far as I tested looks pretty similar to the previous configuration except that it can handle better bigger font sizes now.

The only reason I say this is a partial fix is because the css for hiding and truncating overflow of text for the tabs is not working for me, even after adding `white-space: nowrap` rule. I'd like to address this issue before marking #783 as fixed.

### DIFF
--- a/browser/src/UI/components/Tabs.less
+++ b/browser/src/UI/components/Tabs.less
@@ -27,8 +27,8 @@
     align-items: flex-end;
 
     width: 100%;
-    max-height: 48px;
-    min-height: 32px;
+    max-height: 3em;
+    min-height: 2em;
     color: #c8c8c8;
     background-color: rgba(0, 0, 0, 0.2);
     overflow-x: auto;
@@ -41,7 +41,7 @@
     justify-content: center;
     cursor: pointer;
 
-    height: 32px;
+    height: 2em;
     flex: 0 0 auto;
 
     max-width: 350px;


bryphe:
I just tried this out and it makes tabs pretty small on my Windows machine:

__Before:__
image

__After:__
image

I prefer having the extra space here. I definitely think it makes sense to use the `em` sizing - could we set it to have parity with the sizing previously? It might be a platform difference too - or if you prefer the smaller size, we could certainly expose a `tabs.height` setting.

badosu:
@bryphe I'll use a em measure similar to the previous one and also expose the option as I prefer the previous setting as it exposes more real estate for code.

badosu:
@bryphe Done, now it should look similar to the previous config. See how it shows on my screen:

screenshot from 2017-10-25 09-15-10

This is to showcase the issue with scaling width (this with fontSize 34px):

screenshot from 2017-10-25 09-15-54

bryphe:
I tried it out and it still looks small for my liking - I think it does make sense for it to be configurable because the aesthetic I'm going for may not appeal to everyone 😄 I put together some options in #832 to make this customizable - let me know what you think.

badosu:
@bryphe Your PR is much better, I'll be able to configure how I like with it and have reference for when I need to add customizable css to Oni in the future.

Thanks!

PR #828, opened 2017-10-25
Adjust statusbar height scaling

badosu:
The previous height was too big for my taste, see how it looked like previously:

screenshot from 2017-10-24 22-39-58

See it with the adjust:

screenshot from 2017-10-24 22-39-42

If this is not acceptable, please let me know if we can make this configurable.


bryphe:
I like it! It's a bit less obtrusive for sure. If we find that people still want to tweak the size, we can add it as a setting at that point.

Thanks for submitting the PR, @badosu ! I'll bring it in now.

PR #831, opened 2017-10-25
Fix #825 - Implement `editor.quickOpen.caseSensitive` setting

### LINKED ISSUE #825: Quick Open Case Sensitivity
As far as I can tell, the quick open window is Case sensitive?

If that is the case, I suggest instead that Oni copy Vims config options for `ignorecase` and `smartcase`.

- Ignore case is fairly self explanatory, we just always ignore the case, IE `a` returns `A` or `a`.

- Smart case is slightly different, you ignore case when a lower case string is given, but if you include upper case letters then it swaps to case-sensitive search.
  - IE `apple` returns `Apple`, `apple`, `APPLE` or any other combo, but searching for `aPPle` will only return `aPPle`.

I'm quite used to using just lower case search terms as I have this set in Vim, so having the quick open window work similarly would be nice.

After having a quick look, it looks like we explicitly pass over a "--case-sensitive" flag to ripgrep, so it would look like I can add a config option or two to pass over "--ignore-case" or "--smart-case" instead?

Assuming that sounds okay, what config options would we want?

bryphe:
__Issue:__ The 'case-sensitivity' behavior of the QuickOpen is not very clear, and it is even less clear on how you can change the behavior to get preferred functionality (more discussion on this in #825).

__Fix:__ This changes adds a new setting - `editor.quickOpen.caseSensitive` that accepts either a `boolean` value or `smart`. This change makes the pre-filter step case-agnostic, and then uses this setting for running the `fusejs` filter step.

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -59,7 +59,6 @@ const BaseConfiguration: IConfigurationValues = {
     "editor.linePadding": 2,
 
     "editor.quickOpen.execCommand": null,
-    "editor.quickOpen.caseSensitive": "smart",
 
     "editor.scrollBar.visible": true,
 
@@ -73,6 +72,8 @@ const BaseConfiguration: IConfigurationValues = {
     "editor.cursorColumnOpacity": 0.1,
 
     "environment.additionalPaths": [],
+
+    "menu.caseSensitive": "smart",
 
     "recorder.copyScreenshotToClipboard": false,
     "recorder.outputPath": os.tmpdir(),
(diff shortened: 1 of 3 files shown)


bryphe:
Fixes #825

bryphe:
One issue I just realized is that the filtering logic is shared between `Quick Open` and the `Command Palette`. I think the behavior is reasonable, but it might influence how we name the configuration option (maybe it makes sense to be more general than just the quickOpen menu)

bryphe:
I decided to change the setting to `menu.caseSensitive`, because this filtering logic is the same across all menus - Quick Open, Command Palette, and (soon) symbol search through language servers.

PR #832, opened 2017-10-25
Proposal for #783 - Configuration options for tabs

bryphe:
Related to #827 - this adds the following customization options:
- `tabs.height`
- `tabs.maxWidth`
- `tabs.wrap`

I think it makes sense to make this customizable, because some people might prefer the Vim-style line sizing. Setting `tabs.height` to `1em` or `1.5em` would be closer to that.

@badosu  - let me know how you feel about these customization settings.


badosu:
This is great, thanks!!

PR #833, opened 2017-10-27
updated keybindings to escape from menu using c-c and c-[

jared-skinner:


bryphe:
Awesome, thanks for the PR, @jared-skinner ! Fix works great.

It looks like there is a tslint issue that is causing the build to fail from the appveyor logs:
```
ERROR: C:/projects/oni/browser/src/Input/KeyBindings.ts[60, 25]: missing whitespace
```

(looks like it wants a space after that first comma). Can you please fix that and push it up? You can run `npm run lint` to check if its fixed locally (and `npm run fix-lint` automatically fixes some of these issues, it might take care of this one).

Once the lint check passes, I'll bring it in. Thanks again!

bryphe:
Thanks for the fix, @jared-skinner !

PR #842, opened 2017-10-31
[WIP] Markdown language server

bryphe:
This is a prototype to include a markdown language server, with some limited capabilities:

- Linting via the `write-good` library
- Spellchecking via `node-spellchecker`
- Go-to definition for following links

At the moment, it's basically just the sample language server, but will be expanded on (and could be extracted out to a separate project / npm package)

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -79,6 +79,8 @@ const BaseConfiguration: IConfigurationValues = {
     "language.cpp.languageServer.command": "clangd",
     "language.c.languageServer.command": "clangd",
 
+    "language.markdown.languageServer.command": path.join(__dirname, "languages", "markdown", "server.js"),
+
     "language.reason.languageServer.command": ocamlLanguageServerPath,
     "language.reason.languageServer.arguments": ["--stdio"],
     "language.reason.languageServer.rootFiles": [".merlin", "bsconfig.json"],
--- a/languages/markdown/server.ts
+++ b/languages/markdown/server.ts
@@ -1,0 +1,151 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+'use strict';
+
+import {
+	StreamMessageReader, StreamMessageWriter, createConnection, IConnection, TextDocuments, TextDocument,
+	Diagnostic, DiagnosticSeverity, InitializeResult, TextDocumentPositionParams, CompletionItem,
+	CompletionItemKind
+} from 'vscode-languageserver';
+
+// Create a connection for the server. The connection uses Node's IPC as a transport
+let connection: IConnection = createConnection(new StreamMessageReader(process.stdin), new StreamMessageWriter(process.stdout));
+
+// Create a simple text document manager. The text document manager
+// supports full document sync only
+let documents: TextDocuments = new TextDocuments();
+// Make the text document manager listen on the connection
+// for open, change and close text document events
+documents.listen(connection);
+
+// After the server has started the client sends an initilize request. The server receives
+// in the passed params the rootPath of the workspace plus the client capabilites.
+let workspaceRoot: string;
+connection.onInitialize((params): InitializeResult => {
+	workspaceRoot = params.rootPath;
+	return {
+		capabilities: {
+			// Tell the client that the server works in FULL text document sync mode
+			textDocumentSync: documents.syncKind,
+			// Tell the client that the server support code complete
+			completionProvider: {
+				resolveProvider: true
+			}
+		}
+	}
+});
+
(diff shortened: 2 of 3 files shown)


TalAmuyal:
@bryphe 
> and could be extracted out to a separate project / npm package

If there is no existing project for a Markdown LSP server, I implore you to make it a separate project so:

1. The non-Oni community could benefit your work.
2. that the LSP server will have a wider audience for incoming feedback/PRs.

Also (and this is completely unrelated to the above), I would like to suggest that this will be implemented in a different language (like Python, Go, Rust or Kotlin) for the following reasons:

- Completely subjective but true honesty: I find it hard to use (and thus contribute to) JS/TS projects
- Wider target audience (I think that running Python or Go applications is much easier for most users on most platforms)
- Performance (AFAIK, Go should yield superior performance over JS/TS, see https://benchmarksgame.alioth.debian.org/u64q/compare.php?lang=go&lang2=node)

Please forgive me if it seems rude, I have only good intentions and am greatly inspired by your work.

bryphe:
> If there is no existing project for a Markdown LSP server, I implore you to make it a separate project so:

Yes, seems reasonable! I actually just created a CSS/LESS LSP server (just extracted from VSCode): https://github.com/onivim/css-language-server

It'd be straightforward to do the same with the markdown server.

>  I would like to suggest that this will be implemented in a different language

Thanks for being candid, I appreciate the feedback! Here are my thoughts on each point:

> Completely subjective but true honesty: I find it hard to use (and thus contribute to) JS/TS projects

Yes, understand this - the JS/TS ecosystem is changing quickly and there are so many paradigms and lots of ramp-up needed to get productive. However, the plus side of this is that there is a _huge_ repository of packages - and this is the major strength for implement a language server. The https://github.com/Microsoft/vscode-languageserver-node does most of the heavy lifting - abstracting the transport layer, whether its socket/IPC/stdio/etc - and there are some great packages for spell and grammar check.

Contrast that with https://github.com/palantir/python-language-server or https://github.com/sourcegraph/go-langserver - they have to re-implement a transport layer and consequently there are lots of issues. Neither server works well on Windows, for example. I think there's an opportunity for both ecosystems to have a package that provides a similiar abstraction to `vscode-languageserver`, but that's out of scope for me 😄 

I did a lot of C++ in school, and then professionally in C# & JS (TS). I do think it's worth learning, as some of the ecosystem - like [React](https://reactjs.org/) / [Redux](https://redux.js.org/) / [RxJS](http://reactivex.io/rxjs/) - have introduced some cool functional-style programming, and I learned a lot just working with those libraries.

> Wider target audience (I think that running Python or Go applications is much easier for most users on most platforms)

I'm not convinced this is the case. I actually researched this a bit prior to working on Oni - looking at things like the [Stack Overflow Developer Survey](https://insights.stackoverflow.com/survey/2017#developer-profile). Specifically, stats like this:
image

In addition, Python and Go are convenient in the OSX and Linux ecosystem - but on Windows, it's actually much worse. In fact, Python is a _huge_ pain in Windows, and when you look at a stat like this:
image

Windows is still a relevant and important platform. Electron essentially provides a node runtime, so it makes sense to leverage that - there doesn't need to be additional install or setup logic to drop a python runtime on Windows. These stats are part of the reason I chose the Electron platform (even for all the hate Electron gets on reddit/HN!), because I believe it has the right set of tradeoffs to address a niche that has a need.

I like Python as a language a lot, but the impracticality on Windows is a deal breaker for me leveraging it more heavily in Oni. A go or rust executable would be potentially more doable, but it'll be quicker to build leveraging the `vscode-languageserver`.

With these projects, I try to balance two factors - _how quickly can I build something_ and _how applicable would it be for others_ - and I think for this case JS + NPM strike the right balance. I have a few months of runway to make Oni sustainable, so I have to make sure I make judicious use of that time.

> Performance (AFAIK, Go should yield superior performance over JS/TS, see https://benchmarksgame.alioth.debian.org/u64q/compare.php?lang=go&lang2=node)

Agreed, I think that a native executable built with Go/Rust/C++ would be more performant. I'm not sure the extra development time would justify it. The language server built with `vscode-languageserver` with JS + Node should only take a couple days to get going, whereas it would be a multiple of that on any of those platforms (at least for me).

 The nice thing about OSS is others are welcome to build a native implementation, too! And I'd be happy to leverage it and scrap my implementation if its better.

> Please forgive me if it seems rude, I have only good intentions and am greatly inspired by your work.

Cheers, not rude at all, I appreciate the questions & feedback - I hope my responses don't come across negative
 😄 It's great to think through these points and they are certainly applicable to the roadmap & future of this project.

As an aside, one language I am excited about is [Reason](https://reasonml.github.io/) and I'm considering trying to write the language server in that - it's an OCaml variant that compiles to JS. I've been meaning to try a real-world project with it, this might be a good fit...

bryphe:
FYI: I started this over at https://github.com/onivim/markdown-language-server

It's really basic but it has `write-good` diagnostics hooked up to it, at least. Next step is to implement `hover` and `go to definition`, and have a meaningful README.md...

I tried out Reason, but it's time consuming to stub out all the imports unfortunately (whereas the definition files are just there for TypeScript)

I'll close this out, and we can track issues over there

PR #844, opened 2017-10-31
Fix #843 - Oni freezes when using Neovim's completion popup menu

### LINKED ISSUE #843: Oni freezes entirely on python autocompletion
Repro:
screen shot 2017-10-31 at 5 30 01 pm

bryphe:
__Issue:__ When using the Neovim popup menu, Oni could crash and not respond to input.

__Defect:__ There was special logic to position the popup menu that wasn't airtight. This was to address a flickering issue - with `position: null`, the completion menu shows relative to the cursor. The problem is, the cursor moves to the echo line during completions, so this can cause a flicker.

__Fix:__ The flickering is better than Oni crashing, so revert that special logic - we can address the flickering separately if is it still an issue.

### DIFF
--- a/browser/src/Editor/NeovimPopupMenu.tsx
+++ b/browser/src/Editor/NeovimPopupMenu.tsx
@@ -6,17 +6,10 @@
 
 import * as React from "react"
 
-import * as types from "vscode-languageserver-types"
-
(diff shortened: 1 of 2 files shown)


bryphe:
Fix #843

PR #848, opened 2017-11-01
[WIP / Prototype] Use TextMate themes for syntax highlighting

bryphe:


jordwalke:
I'm curious how this performs compared to Vim syntax highlighting? Did you consider creating an at-startup-time conversion from a textmate theme to a Vim highlighting scheme as opposed to the more dynamic version?

bryphe:
Unfortunately, Vim syntax highlighting doesn't quite match parity-wise with TextMate's tokenization. Vim syntax highlighting makes some tradeoffs such that it doesn't need to parse the entire file (which makes it faster!) but also more limited. TextMate's tokenization allows carrying a rulestack for each token which gives a semblance of what scopes are available - the VSCode team has a great write-up here: https://code.visualstudio.com/blogs/2017/02/08/syntax-highlighting-optimizations

In other words, AFAIK, there unfortunately isn't a lossless conversion of TextMate tokenization -> vim syntax highlighting.

There are a couple other things I plan on using this for (besides just coloring):
- Querying to see what scope a particular token is in - this is important for things like auto-pairs - maybe we don't want a particular character enabled in comments, for example. It's not straightforward at the moment to ask what syntax a particular character has. Also, we might not want to show hover in comments, for example.
- Alternative visualizations of the code (like a minimap view)

The degenerate case for performance is the bottom of a large file - since you need to evaluate the `ruleStack` for all previous lines, at least the first time you navigate to the bottom. The parsing is relatively quick (`vscode-textmate` is great, and bundles a native parser called `oniguruma`!), but it is synchronous and for large amount of lines it can add up. Batching is possible - but in the worst case you end up with a brief FOUC (well, in this case, it'd be 'flash of partially styled content' - we'd use the vim syntax theme as a base, and only style additional elements on top of that), but better than hurting the responsiveness. This current implementation isn't fully optimized yet. 

Some discussion in Neovim too here: https://github.com/neovim/neovim/issues/6052 (about natively supporting it, and limitations of the current model).

PR #855, opened 2017-11-02
LSP error improvements

bryphe:
- Fix bug where we show the error balloon in insert mode while typing (jarring experience)
- Fix some typings


bryphe:
This will be a nice UX fix, as well as improve the demo flow 😄

PR #861, opened 2017-11-03
Use custom trigger characters when supplied.

CrossR:
When a custom trigger array is given, we should use that for the trigger characters of that language.

Currently, I've set the Typescript and Javascript ones to . / \ which are the ones I've seen used in other editors.

We could also extend the trigger character settings to be more complex.
I think in VSCode, they have . / \ ", with the last 3 all only working in certain cases.
Ie only allow / when on an import line, don't use " if its the end of an already open string.
This would also let us add " which I have excluded for now, as it messes with auto completion when closing import strings.

I can look at adding that as part of this PR, or add an issue and sort it separately, if its decided its needed/useful.

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -86,6 +86,10 @@ const BaseConfiguration: IConfigurationValues = {
     "language.ocaml.languageServer.arguments": ["--stdio"],
     "language.ocaml.languageServer.configuration": ocamlAndReasonConfiguration,
 
+    "language.typescript.completionTriggerCharacters": [".", "/", "\\"],
+
+    "language.javascript.completionTriggerCharacters": [".", "/", "\\"],
+
     "menu.caseSensitive": "smart",
 
     "recorder.copyScreenshotToClipboard": false,
--- a/browser/src/Services/Language/LanguageManager.ts
+++ b/browser/src/Services/Language/LanguageManager.ts
@@ -13,6 +13,7 @@ import { Event } from "./../../Event"
 import { IDisposable } from "./../../IDisposable"
 import * as Log from "./../../Log"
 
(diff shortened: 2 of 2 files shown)


bryphe:
Looks great, thanks @CrossR ! Tried it out and it definitely helps for the import statements.

> We could also extend the trigger character settings to be more complex.
I think in VSCode, they have . / \ ", with the last 3 all only working in certain cases.
Ie only allow / when on an import line, don't use " if its the end of an already open string.
This would also let us add " which I have excluded for now, as it messes with auto completion when closing import strings.

Yes, I think they have some logic around this here: https://github.com/Microsoft/vscode/blob/d2f1fd4763b2312acf879bbcb3e9a2e386c1708b/extensions/typescript/language-configuration.json#L11

We could potentially extend our trigger characters schema to be similiar for that - the only tricky  thing is figuring out _when_ we're in a string/import line, etc. 

I'm exploring using TextMate themes for better syntax highlighting: https://github.com/onivim/oni/pull/848 (because they give more detail about the identifiers and symbols), but it also would give enough info to do that filtering too.

CrossR:
Looks to be some logic for it in here as well: 

https://github.com/Microsoft/vscode/blob/d2f1fd4763b2312acf879bbcb3e9a2e386c1708b/extensions/typescript/src/features/completionItemProvider.ts#L179

bryphe:
Ah interesting, thanks for finding that! There is definitely some specialized logic there beyond the textmate scopes.

It seems we could bake some similiar logic into our TypeScript extension (or figure out a way to generalize expressing that logic in our language configuration). It'd be nice if it was generalized in our configuration in some way.

bryphe:
CI failure is still a token issue, not related to your change - I'll take a look at fixing that today.

I'll bring this in, and I opened #864 to track adding some additional logic / smarts to the completion trigger. Thanks for the PR and ideas, @CrossR !

Issue #326, opened 2017-03-28
Build OSX version of Neovim as part of Oni


bryphe:
Similiar to how VimR works, we should be building Neovim from source instead of including all the binaries.

For now, #325 is a quick-fix to check in the binaries to the enlistment, but we'll need a better solution long-term.

extr0py:
As part of the build, we should be adding the `gui_oni` feature, such that `has("gui_oni")` returns `1`

bryphe:
Since we now have OSX binaries bundled as part of `oni-neovim-binaries`, this is no longer necessary. We can still consider doing a manual build if there is a demand for it to include a `has("gui_oni")` variable.

PR #875, opened 2017-11-07
Use downloaded binaries instead of checking-in Neovim binaries

bryphe:


bryphe:
Will be holding off on this change until the `0.2.1` release, and then wiring it up. It'll be great to get these checked-in binaries out of the repo (and keep it from showing as language `VimL` in github 😄 ).

This will address both #326 and #123 . Later, we might want a forked Neovim so that we can add a GUI flag (or experiment with a web assembly build).

bryphe:
Need to incorporate & test Linux support here, too.

PR #876, opened 2017-11-07
[NEXT] Add UI font for shell UI

bryphe:
__Issue:__ Many of the newer text editors use separate fonts for their _shell_ interface vs the _edit_ experience. This helps improve the look and feel of the editor vs a single-font terminal UX.

__Fix:__ Enable specifying a separate font for shell items+

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -106,6 +106,9 @@ const BaseConfiguration: IConfigurationValues = {
     "tabs.height": "2.5em",
     "tabs.maxWidth": "30em",
     "tabs.wrap": false,
+
+    "ui.fontFamily": "BlinkMacSystemFont, 'Lucide Grande', 'Segoe UI', Ubuntu, Cantarell, sans-seriff",
+    "ui.fontSize": "13px",
 }
 
 const MacConfigOverrides: Partial<IConfigurationValues> = {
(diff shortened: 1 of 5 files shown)


browser/src/Services/Configuration/DefaultConfiguration.ts
```diff
@@ -106,6 +106,9 @@ const BaseConfiguration: IConfigurationValues = {
     "tabs.height": "2.5em",
     "tabs.maxWidth": "30em",
     "tabs.wrap": false,
+
+    "ui.fontFamily": "BlinkMacSystemFont, 'Lucide Grande', 'Segoe UI', Ubuntu, Cantarell, sans-seriff",
```
badosu: @bryphe Just noticed not, perhaps `s/Lucide/Lucida` ?

bryphe: @badosu - wow, nice catch, thank you! I'll update this.

badosu: I am sorry @bryphe just noticed also Cantarell -> Cantarelli

bryphe: Thanks for checking and letting me know!

I looked at this one and I believe `Cantarell` is actually correct: https://en.wikipedia.org/wiki/Cantarell_(typeface)

LMK though if I'm mistaken here

PR #885, opened 2017-11-07
Implement #908 - Integrate `ext_cmdline` and `ext_wildmenu`

bryphe:


Bretley:
If you could make a semi-detailed checklist of what needs done I could probably do this. I stole the changes from this and added the cases to the switch statement, I'm just not sure what to add to State.ts

bryphe:
Sweet, thanks for taking a look at this @Bretley ! Would be great to have help.

Your checklist looks good to me - here are some of my thoughts:

- Set up switch cases - seems like a great starting point! Pushing out the events in the `_handleNotification` method - similiar to how we handle the `popupmenu_select`/`popupmenu_hide`/`popupmenu_show` notifications - we'd want to start handling the cmdline and wildmenu notifications, and `dispatch` them via the events defined in the PR. That will give us a nice, strongly-typed interface to build on.

Regarding the state / action / reducer - if you're not familiar with [Redux](https://redux.js.org/) and [React](https://reactjs.org/docs/hello-world.html), it's worth checking out those docs... This write-up - [Thinking in React](https://reactjs.org/docs/thinking-in-react.html) is helpful too!

I usually start by defining the __state__ - what is the data model we need? Luckily, for a lot of the neovim-externalization features, I think it maps pretty closely to the interfaces they send us via the `msgpack-rpc` protocol.

Once we have the __state__ figured out, we can do two things:
- Figure out how to render our UI, based on the __state__. A helpful way to start is just making a hard-coded state, and get it showing in the UI. Once you have that, it's just about connecting the 'live' state! You could put this hardcoded component in `browser\src\UI\RootComponent` for now (near the `<MenuContainer />`, so you could quickly see it live).
- Figure out how the state can change - this is the responsibility of the __actions__ and __reducers__ - these define what changes can be made, and then the result of those actions. We'd likely at least need ones that map to the notifications from Neovim - a show action, a hide action, and a setCursorPosition. I'm not sure exactly how the text updates, but we might need an action for that too.

I hope that helps at least give some ideas... thanks for your help! Let me know if you have other questions as you get going.

Bretley:
```TypeScript
                case "cmdline_show":
                    const [content, pos, firstc, prompt, indent, level] = a[0]
                    const commandLineShowInfo: INeovimCommandLineShowEvent = {
                        content,
                        pos,
                        firstc,
                        prompt,
                        indent,
                        level,
                    }
                    this._onCommandLineShowEvent.dispatch(commandLineShowInfo)
                    break
                case "cmdline_pos":
                    const [pos, level] = a[0]
                    const commandLinePositionInfo: INeovimCommandLineSetCursorPosition = {
                        pos,
                        level,
                    }
                    this._onCommandLineSetCursorPositionEvent.dispatch(commandLinePositionInfo)
                    break

```
This is modeled after the popupmenu switch case, but I get the error 'cannot redeclare block-scoped variable. Which ones should be const and which shouldn't?

bryphe:
Oh ya, there's weird scoping rules in JavaScript for this - the `case` statements actually aren't their own scope, so it thinks you're trying to redeclare the same variable.

You can avoid this by creating a scope like:
```
      case "cmdline_show":
     {
                    const [content, pos, firstc, prompt, indent, level] = a[0]
                    const commandLineShowInfo: INeovimCommandLineShowEvent = {
                        content,
                        pos,
                        firstc,
                        prompt,
                        indent,
                        level,
                    }
                    this._onCommandLineShowEvent.dispatch(commandLineShowInfo)
                    break
   }
```

Bretley:
Alright here's where I'm stuck: 
I have an action and a reducer and an actioncreator defined, but I'm not sure how to get from neovimInstance.ts to actually changing the state. I see a lot of things calling dispatch, but I'm not sure what rabbit hole that goes down if that makes any sense.

bryphe:
> I have an action and a reducer and an actioncreator defined,

Cool, sounds like great progress!

> But I'm not sure how to get from neovimInstance.ts to actually changing the state. I see a lot of things calling dispatch, but I'm not sure what rabbit hole that goes down if that makes any sense.

Ah ya... the brute force way is to create the action via the action creator, and then pass to the UI store's `dispatch` method.

However, we use a utility called `bindActionCreators` that takes all of the ActionCreators and 'binds' them to the dispatch method for us, and gives us a new method that takes care of that - that's over here:
https://github.com/onivim/oni/blob/e8bdda46f2d079929fee9154b245a3d8a4743bc8/browser/src/UI/index.tsx#L39

(more info on `bindActionCreators` here: https://redux.js.org/docs/api/bindActionCreators.html)

From there, we can call any of them via `UI.Actions`, like here:
https://github.com/onivim/oni/blob/e8bdda46f2d079929fee9154b245a3d8a4743bc8/browser/src/Editor/NeovimEditor.tsx#L158

The pattern we usually have is subscribing to an event from NeovimInstance, so that the NeovimInstance object itself doesn't have to care about the UI - but NeovimEditor is the one that does all the UI stuff. Hope that helps!

Bretley:
Is there a way to get the state in the console?

bryphe:
I'll close this out since @Bretley and @samvv both have better versions 😄

Issue #730, opened 2017-09-28
Modernize Quickfix List


badosu:
This is one of the prospects that excited me for a long time since neovim was released.

Currently we use neovim's own quickfix list (the lower window):

![screenshot from 2017-09-27 22-25-15](https://user-images.githubusercontent.com/347552/30945166-ec1d2a1c-a3d2-11e7-9069-e503322693db.png)

We can leverage the UI possibilities provided by electron to:

- Make the quickfix window a proper window, separated from electron window or integrated, much like a console would.
- Improve the display of each quickfix entry (filename, column and text excerpt). E.g. highlighting the found part and truncating the content of the rest of the line to fit it into a sane layout.
- Improve the capabilities of the quickfix window to enable easier iteration through results.

bryphe:
> This is one of the prospects that excited me for a long time since neovim was released.

Definitely! This is a terrific case to 'break free' of the terminal limitations and build something nice. It's somewhat related to what I was exploring in #622 - which is the ability to have 'custom renderers' on top of a Neovim surface. That then enables us to do all kinds of interesting UI on top of Neovim.... while still allowing keyboard navigation and behavior as you'd expect from Neovim.

PR #886, opened 2017-11-08
[WIP] File Explorer v2

bryphe:


badosu:
Really can't tell too much without testing this, but I was wondering whether we could reuse something for the logic and just stylize the output? If this requires more boilerplate than what we are saving then great (or has fundamental limitations/flaws), if not I would suggest:

- (NerdTree) https://github.com/scrooloose/nerdtree
- (vim-vinegar) https://github.com/tpope/vim-vinegar
- vim's native netrw

bryphe:
Some progress:
image

Still a lot left, though!

bryphe:
> Really can't tell too much without testing this, but I was wondering whether we could reuse something for the logic and just stylize the output?

Thanks for looking at it, @badosu ! I wanted to test a few things as part of this change - one of them brought up in #622 - the idea of having Oni control the render layer, but using Neovim as the input layer (which I believe would work well for any non-modifiable list-type UI, like the quickfix in #730). This allows us to keep things like smooth-scrolling, while getting the benefit of respecting the user's keybindings and customizations.

It'll be a bit more work but I think it'll be worth it in the long run, as it will push us further from terminal limitations and enable us to do some more interesting things in the UX layer. Thinking of features like a version control pane, or a debug pane - it'd be nice to interact with those in a vim-consistent way, while having a UX that is on-par with Atom/VSCode. That's my dream for Oni, anyway 😉

bryphe:
This is almost live (well, behind a feature flag) in #1104 - will track remaining work there.

PR #888, opened 2017-11-08
Subtle cursor motion with react-spring

bryphe:
This uses the `react-motion` library to provide some springy-ness to the cursor as typing or moving (and it also creates some 'tension' when in operator-pending mode).

Needs a little bit of work:
- Refine spring constants
- Add setting to disable it, because some may find it annoying
- Fix text scaling in cursor


bryphe:
Added config value of `ui.animations.enabled` to turn this off. IMO it's nice to have instant feedback (this is one step before the _optimistic updates_ I'd like to add to improve typing feel), but it might be too much for some.

PR #915, opened 2017-11-10
Add escape to leave terminal mode by default

bryphe:
Fix #915  - add default `Esc` binding for leaving terminal mode.

Good info here: https://neovim.io/doc/user/nvim_terminal_emulator.html

### DIFF
--- a/vim/default/bundle/oni-vim-defaults/plugin/init.vim
+++ b/vim/default/bundle/oni-vim-defaults/plugin/init.vim
@@ -34,4 +34,6 @@ inoremap <expr> <C-a> pumvisible() ? "<Esc>A" : "<C-o>A"
 inoremap <expr> <C-b> pumvisible() ? "<Esc>bi" : "<C-o>b"
 inoremap <expr> <C-l> pumvisible() ? "<Esc>la" : "<C-o>a"
 
+tnoremap <Esc> <C-\><C-n>
+
 colorscheme onedark


hkupty:
Please refrain from adding any kind of remap, specially directly in VimL.

Editor options should be toggleable. Specially this one, which is very intrusive.

`<esc>` can be a valid key for terminal applications (and this is precisely why it wasn't adopted as a default by the neovim team).

I strongly advise you to implement such bindings as options toggleable through oni config. Now, to circumvent this undesired behavior I (and probably other people) must undo what this PR does in my config.

While it is pretty simple for me to do, it is not desirable that I must edit my `init.vim` in order to properly use the gui.

Don't get me wrong, I enjoy using oni. It is a beautiful and fast GUI, but there must be a clean separation of concerns.

bryphe:
Thanks for the thoughtful feedback, @hkupty !

@badosu  gave some similar feedback as well in #929 .

> I strongly advise you to implement such bindings as options toggleable through oni config.

> It is a beautiful and fast GUI, but there must be a clean separation of concerns.

This seems reasonable, I do agree with the sentiment. I would like to minimize our surface-area of VimL configuration as much as possible. Sorry that the shortcut we took with the `tnoremap` setting caused extra config work for you - moving it to an Oni-setting does make sense.

FYI, I tried to collect notes from both yours and @badosu 's feedback here in #973 . I believe that the configuration story can be improved significantly - would be great to have your thoughts.

Piping:
@bryphe I agree with @hkupty I used a plugin called FZF and it use terminal inside nvim. It takes escape to close the window. This is just one example that default behavior used by other developers.

PR #922, opened 2017-11-11
Markdown preview

### LINKED ISSUE #49: Feature - Preview: Markdown preview

TalAmuyal:
Based `bryphe/proto/markdown-preview` and merged `origin/master` onto it.
Next up: make something that works.

Should fix #49.


browser/src/UI/RootComponent.tsx
```diff
@@ -8,8 +8,19 @@ import * as WindowManager from "./../Services/WindowManager"
 
 import { Background } from "./components/Background"
 import StatusBar from "./components/StatusBar"
+
 import { WindowSplits } from "./components/WindowSplits"
 
+
+
+
+import * as marked from "marked"
```
bryphe: Oops, I was using this as debugging to see what `marked` returns 😄

vim/core/oni-markdown-preview/src/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+const os = require("os")
+
+import * as React from "react" 
+
+import { MarkdownPreview } from "./MarkdownPreview"
+
+export class MarkdownPreviewEditor implements Oni.Editor {
+
+    constructor(
+        private _oni: any
```
bryphe: A good next step would be to listen to the buffer update events here - we could do:

```
Oni.editors.activeEditor.onBufferChanged.subscribe((evt) => { console.log("Buffer Changed"; debugger; })
```

That way, if you have the debugger open, you can inspect `evt` and see the values for it. We can get the filename + filetype (language) and decide if we should render a preview for it.

If we should, we can get the content via the `contentChanges` array, and then render it out to our component using `marked`.

TalAmuyal:
Got stuck on updating the preview component.

When a Markdown file is opened, a split opens and tries to update the component's state with the loaded file's content.
At that stage, the following error is logged:
```
setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op.

Please check the code for the MarkdownPreviewEditor component.
```

I followed a few tutorials and read some docs, but to no avail.
Implementing:
```
    componentDidMount() {
        console.warn("Mounted")
    }

    componentWillUnmount() {
        console.warn("Unmounted")
    }
```

Only confirmed that something is wrong.

bryphe:
Ah ya, just tried this out - great progress!

 That's one confusing aspect of React initially - is how to instantiate the elements (as well as the [JSX](https://reactjs.org/docs/introducing-jsx.html) syntax and the [Lifecycle](https://reactjs.org/docs/state-and-lifecycle.html))

I made some tweaks to get it working in this branch - some ideas here: https://github.com/onivim/oni/tree/bryphe/markdown-preview-mods 

Specifically [this commit](https://github.com/onivim/oni/commit/70219c4ad796d3a053a9ff60f5d17d13a2d2946d)

I made a few tweaks:
- I split out `MarkdownPreview` into two separate classes - `MarkdownPreviewEditor` and `MarkdownPreview` - the first being something we `new` and pass over to the window split, and the latter being the actual react component that does the rendering. The `MarkdownPreviewEditor` is really just a shim that implements `IWindowSplit` - later it might be cool to have it implement `Oni.Editor`, once we integrate some of the Neovim pieces here.

The way we actually constructor the `MarkdownPreview` component is interesting: 
```
        return <MarkdownPreview bufferEnter={this._oni.editors.activeEditor.onBufferEnter} />
    }
```

That's actually syntactic sugar for instantiating a `MarkdownPreview` object - it actually compiles into this:
```
    render() {
        return React.createElement(MarkdownPreview, { bufferEnter: this._oni.editors.activeEditor.onBufferEnter });
    }
```

We then hand over all the state management stuff to it - let it listen to the buffer enter event and render in response to that.

That solves the lifecycle stuff. It actually kind of rendered with these changes 😄 :
oni-screenshot-1510881127569

Some possible next steps:
- listen to the `onBufferChanged` event to see when changes occur (we might not even have to read from the file system, we could use the contentChanges that gives us instead)
- open / close the split - I think `oni.windows.close(split)` would work, but didn't try it...

I hope that gives you some ideas! Thanks for all your work on this.

TalAmuyal:
Now the following can be set in the configuration file:
```
    oni.input.bind("<f6>", "markdown.openPreview")
    oni.input.bind("<f7>", "markdown.closePreview")
    oni.input.bind("<f8>", "markdown.togglePreview")
```

Next up:
- Fix the styling
- Implement automatic coordination between preview-pane and source-buffer

TalAmuyal:
Tried to implement auto scrolling, but no success so far :(
What I did was setting `overflow-y: scroll` to `<MarkdownPreview />`-s "container" `div` and setting the div's `scrollTop` property.
After setting a value to it, it just remains at 0...

Any tips?

vim/core/oni-markdown-preview/src/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+        this.setState({source: str})
+    }
+
+    public render(): JSX.Element {
+        const containerStyle: React.CSSProperties = {
+            padding: "1em 1em 1em 1em",
+        }
+
+        const html = marked(this.state.source)
+        return <div style={containerStyle} dangerouslySetInnerHTML={{__html: html}}></div>
```
bryphe: There might be a couple classes you want to add here: `className="stack enable-mouse"`. The first one causes it to be sized to fit the editor window (it sets `position: absolute` plus clamps the bounds to the parent), and `enable-mouse` will allow it to pick up mouse events.

vim/core/oni-markdown-preview/src/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+        buffer.getLines().then((lines: string[]) => {this.previewString(lines.join("\n"))})
+    }
+
+    private previewString(str: string): void {
+        this.setState({source: str})
+    }
+
+    public render(): JSX.Element {
+        const containerStyle: React.CSSProperties = {
+            padding: "1em 1em 1em 1em",
```
bryphe: Once you have the classnames below, you should be able to set `overflow-y: auto` on this element, which will cause a scrollbar to show up when it expands across the box (as well as `scrollTop` for auto scrolling)

bryphe:
Just loaded up the branch, looks like great progress!

To get `scrollTop` working, the very first step is to make sure a scrollbar is visible - left a couple comments to help with that. We want the scrollbar to be on the root element rendered by that `<MarkdownPreview />` component.

The next challenge that I'm not exactly sure how to handle, is how to map from the current buffer position -> preview window position? One option would be to try and align it to the nearest heading, or something... The mapping is not super straightforward because something that takes one line in markdown might take a bunch of space in the preview window (like a big image) - will be interested to see how that works!

TalAmuyal:
Tried your suggestions and once again they have worked like magic.
Scrolling is now possible with the mouse wheel, as well as the auto-scrolling I've implemented.

Unfortunately, the auto-scrolling isn't really usable for since I calculate the ratio of the current cursor's line in proportion to the total lines count.
This is bad since it is both jarring and otherwise uncomfortable.

A bit more suitable solution should be using the top-of-screen's line instead of the cursor's line, though it doesn't solve the issue created from the fact that on the rendered pane, the lines' height is not constant.

I think of implementing a logic that identifies the top-most rendered element that is in the current visible part of the buffer and use that instead.
Another possible solution could be implanting an anchor in the input to `marked` and then scroll to it using `scrollToView()`.

In any case, a `bufferVisibleSectionChanged` event is required.

Will report back again when I have more progress with that.

bryphe:
Awesome progress! 👍 

> Unfortunately, the auto-scrolling isn't really usable for since I calculate the ratio of the current cursor's line in proportion to the total lines count.
This is bad since it is both jarring and otherwise uncomfortable.

Makes sense... it seems like this wouldn't be a smooth or continuous mapping.

> I think of implementing a logic that identifies the top-most rendered element that is in the current visible part of the buffer and use that instead.
Another possible solution could be implanting an anchor in the input to marked and then scroll to it using scrollToView().

This sounds good! I was wondering how the mapping would work. Ideally we'd have a function that would be a mapping of `(markdownSourceLine) => markdownPreviewLocationInPixels`, but I'm not sure how that would work. I like the solution of implanting an anchor - that seems pretty straightforward, and would at least let us map a certain set of elements directly via `scrollToView` as you mentioned.

> In any case, a bufferVisibleSectionChanged event is required.

Makes sense. There actually is some code to track scroll events (really, any sort of window size/position change) - here: https://github.com/onivim/oni/blob/d94397e5ad058410133c55a98e34ce90eae7d077/browser/src/neovim/NeovimWindowManager.ts#L174

It's not actually bubbled up anywhere, though. `NeovimEditor` creates an instance of it, but doesn't expose the event. 

I think we could do the following:
- Refactor `NeovimWindowManager` so that, instead of calling `UI.Actions.setWindow...` directly, it exposes an event (like `onWindowScrolled`). We could then bubble that up via an event on `NeovimEditor`, that is available to plugins, like `onBufferScroll`. With that, we should have the context to kick off the scroll sync code.

Really great work here @TalAmuyal - thank you!

TalAmuyal:
Worked some on the auto-scrolling.
I think it works pretty good, though there might be some corner cases which will require special attention.
For instance, the auto scrolling implementation breaks the table rendering (will fix soon).

Aside from that, I think that `gg` and `G` doesn't fire up the scrolling event.
@bryphe, do you have an idea?

Also, waiting on https://github.com/onivim/oni-api/pull/4.

bryphe:
Wow, really cool! I like your strategy for figuring out how to map the lines... It looks like we render each line, wrap it in an anchor/tag we can reference, and then as we auto-scroll, we go to that nearest anchor. Looks solid to me, I'm not really sure of a better way.

I just checked out and tried out the branch, seemed to work pretty well!

> Aside from that, I think that gg and G doesn't fire up the scrolling event.
@bryphe, do you have an idea?

Ah yes, I do actually... The scroll event that we use doesn't have very high fidelity. It turns out that Neovim doesn't have a `ScrollChanged` autocmd or equivalent, so we use the `scroll` event from the redraw msgpack-API to fire that:
https://github.com/onivim/oni/blob/ed10c47f1b8e60a6f1eed6a7b8c773a6ae2c26b4/browser/src/neovim/NeovimInstance.ts#L534

Turns out, this doesn't handle the case where there is as _complete redraw_ (like you use `G` or `gg`).

However, we do actually have another place where we are watching for this - in the `NeovimWindowManager`, which lives in `NeovimEditor`.

This class listens to a bunch of different events:
https://github.com/onivim/oni/blob/ed10c47f1b8e60a6f1eed6a7b8c773a6ae2c26b4/browser/src/neovim/NeovimWindowManager.ts#L36

And lets us know when the window dimensions change (really, whenever the mapping of _buffer space_ to _screen space_ changes).

For the scroll event that `NeovimEditor` exposes, instead of hooking to the `onScroll` event on `NeovimInstance`, we could potentially use the `NeovimWindowManager` instead. We'd want to perhaps expose an event off `NeovimWindowManager`, like `onWindowDimensionsChanged`, and map that to a scroll event on the `NeovimEditor` side. Let me know if that's unclear. But I think if we wire that up, we should get `gg` and `G` handled.

> Also, waiting on onivim/oni-api#4.

This should be set now - I merged in the PR, and updated it so that it publishes to NPM from Releases.  So the NPM build is up to `0.0.8` now. You may need to upgrade the package.json and/or run `yarn upgrade oni-api@0.0.8`. 

It's looking really good!

TalAmuyal:
At last, I succeeded in syncing the colors.
My next and last step (if it is enough for this PR) is to add a enable-disable flag.

I think of something like `experimental.markdownPreview.enabled` which defaults to `false`.

@bryphe, do you think it should be experimental for now? and should it defaults to false? (BTW, waiting for the API PR)

bryphe:
> At last, I succeeded in syncing the colors.

Awesome!!

> @bryphe, do you think it should be experimental for now? and should it defaults to false? (BTW, waiting for the API PR)

That sounds great - `experimental.markdownPreview.enabled` with a default of `false` sounds good for now. But ya, I once we have the flag I'm ready to bring it in - we can then test it out, and it will be available for others to try. From there, we can decide when we're ready to enable it by default.

I'm just wrapping up the 0.2.19 release, and I'm planning on the next release being 0.2.20 on 1/4 - hopefully we can enable it by default for that release 👍 

Oh and I just brought in the `oni-api` change and updated the npm package - thank you for making that change!

TalAmuyal:
I'm done, but unfortunately the PR [fails in CI](https://ci.appveyor.com/project/bryphe/oni/build/1.0.927/job/br9ng4tfydguyesj#L341) on `npm run dist:win` (as well as other dists).
Any idea why or how it can be fixed?

bryphe:
> I'm done, but unfortunately the PR fails in CI on npm run dist:win (as well as other dists).
Any idea why or how it can be fixed?

Cool, it's the home stretch now! Can't wait to get this in!

I'm looking into the failure now.

bryphe:
I just saw a couple issues that are unrelated to the failure, but they might cause tests to failure:
- It looks like there is a large `h q` file at the root, should be removed
- It looks like the submodules in `vim/core` and `vim/default` were removed. This will cause the typescript test to fail (because the filetype won't be picked up).
image

Those will need to be restored.

Still looking at the failure with `npm run dist`/`npm run dist:win`. I see other people have had similiar issues:
https://github.com/electron-userland/electron-builder/issues/2236

But it looks like it might be an actual issue with a dependency. Trying to see what's happening.

bryphe:
I tried setting the `ELECTRON_BUILDER_ALLOW_UNRESOLVED_DEPENDENCIES` and the build completed, and I was able to run the app.

We could try this on the CI machine:
```
    "postinstall": "npm run install:plugins && cross-env ELECTRON_BUILDER_ALLOW_UNRESOLVED_DEPENDENCIES=1 && electron-rebuild && opencollective postinstall"
```

However, once I got that, I tried running the app from the install package, and I see an issue:
image

Seems like the problem there is that `react` is a development dependency (which worked fine before, since the `browser` bundle was the only code that used it, and it bundled it with webpack). For this case, we should move it to a `dependency` as opposed to a `devDependency` at the root.

bryphe:
Once I made that change, it worked in the dist build!
markdown-preview

I pushed up a branch with the full set of fixes I used here: [bryphe/922/fixes](https://github.com/onivim/oni/tree/bryphe/922/fixes)

TalAmuyal:
Doesn't seem to solve it :(

bryphe:
Sorry, looks like I messed up this line:
```
  "postinstall": "npm run install:plugins && cross-env ELECTRON_BUILDER_ALLOW_UNRESOLVED_DEPENDENCIES=1 && electron-rebuild && opencollective postinstall"
```

The `cross-env` call actually has to be before the `npm run install:plugins`

bryphe:
Ah looks like I messed up... I had set the variable in my shell so I was getting a false positive.

There were two problems with my original solution:
- The `cross-env` calls need to be prior to the `build` call in `dist:win` and `pack`, not in the `postinstall` step.
- The `cross-env` calls need to _not_ have a `&&` prior to the call, otherwise it's a no-op

Posted the fixes here in [bryphe/922/move-cross-env-to-build-calls](https://github.com/onivim/oni/compare/bryphe/922/move-cross-env-to-build-calls), and the relevant commit is this one:
https://github.com/onivim/oni/commit/d9ef2323ee77d1f5570d363f38e7e3d6e632d069

Hopefully that's the last issue... fingers crossed...

TalAmuyal:
Now there is a different error >< https://travis-ci.org/onivim/oni/jobs/317490353#L1019

bryphe:
Ah well, at least the windows builds passed... making progress... 😄

bryphe:
Taking a look at the OSX failure... that's strange. Seeing if I can repro it on my OSX machine.

bryphe:
Strange, no issue on my machine. I couldn't find a lot of similiar issues, there were ones like this: https://github.com/electron-userland/electron-builder/issues/1183

But not quite the exact same. Still looking.

bryphe:
I'm wondering if it's having an issue with the nested dependencies in `oni-markdown-preview`. I'm going to test out promoting them to the root to see if that changes anything.

bryphe:
I saw there was also an issue with the build caches - it seems that sometimes our node_modules / yarn caches get out of date 😦 I'll cleared them out, and I got a passing build with #1141 with the change to move the packages to the root.

bryphe:
@TalAmuyal - let me know if you'd just prefer me to merge in #1141 . I wasn't sure if your name is associated with the squashed commit / if you get github activity in that case, so I was holding off. But I'm happy either way. Really excited to get this in 👍 !

TalAmuyal:
Well, I tried a few more things, but no complete success.
It fails a bit further down the road (times-out on every `ci tests * ci test` test).
I don't think the issues originate from this PR.

If you don't mind, please do merge this PR and let's fix any ramifications independently.

bryphe:
> If you don't mind, please do merge this PR and let's fix any ramifications independently.

Sounds good - I'm in the midst of investigating some of the failures, so I'll take a look. I'll merge this in now - thanks for all your work on this, @TalAmuyal ! Congratulations on getting it, this is awesome 💯

PR #924, opened 2017-11-12
bugfix/pass additional paths

Akin909:
I've swapped out the call to `child_process` with a call to the helper function which loads in the additional paths.

I've had trouble adding the `shell-env` module to the project as it seems to cause loads of errors once installed and disrupts typings for other modules like react, I'm not entirely sure what the cause of this is, I tried adding a type definition file but I can't get the app to build with shell-env installed possibly because of a dependency conflict.

I though as an interim measure it would be worth using the config path, I can continue to have a look but I think it's going to prove less straight forward than I anticipated.

### DIFF
--- a/browser/src/Plugins/Api/Process.ts
+++ b/browser/src/Plugins/Api/Process.ts
@@ -1,4 +1,5 @@
 import * as ChildProcess from "child_process"
+import * as fp from "fix-path"
 
 import * as Platform from "./../../Platform"
 import { configuration } from "./../../Services/Configuration"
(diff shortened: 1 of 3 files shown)


bryphe:
Just tested this out - seems like it works great! I see the additional paths from `environment.additionalPaths` when launching from OSX, and running ':term' and then 'echo $PATH'. Thanks @Akin909 for diving in and looking into this!

> I though as an interim measure it would be worth using the config path, I can continue to have a look but I think it's going to prove less straight forward than I anticipated.

Seems reasonable - I think we should get this PR in, because it would address the majority case where `python3` can't be found via `usr/local/bin`. I'll log a new issue to track `shell-env`, which we can integrate separately.

bryphe:
I'm investigating the failed OSX build in #931 , but it's not related to your change.

Fix looks great, I'll bring it in now. Thanks for your investigation & contribution @Akin909 ! 👍 

I'm planning on pushing out a new release on Wednesday (0.2.17) including this fix.

PR #925, opened 2017-11-12
Update ocaml-language-server dependency

freebroccolo:


bryphe:
Thanks @freebroccolo ! Looking forward to trying the new version of the language server.

I'm investigating the build issue in parallel, but it is not related to your change, so I'll bring this in now.

PR #927, opened 2017-11-12
Add a check for the symbol search keybind.

CrossR:
Currently, the LSP symbol search is set to <Ctrl+t> which overlaps with the Quick open, open in new tab keybind.

This should fix that by checking the menu isn't open first, assuming this wasn't by design?
If it was, I can update the PR to remap the open in new tab option, since currently its not usable.\

I'm putting a very early table of the keybinds in the wiki as well, so please fill out around it to add a bit more context! (I also wasn't 100% on the LSP Expand keybind, so I put in a placeholder for now).

Two other quick questions I had:

1. There looks to be a number of keybinds with both a C and M version, such as this PR.
Others just seem to have a Ctrl version only (ie Ctrl+/ for search current buffer), is that by design?
Or more likely, I'm just misunderstanding the keybind system / Mac keybinds in general and somehow it applies to both?
2. I'm having some issues with ripgrep since the move to our own repo, it was very flaky getting it installed so I'd appreciate you testing this PR to check its working as intended. Am I missing something I need to do to get it installed? The neovim download worked perfectly, but ripgrep seemed to not return any results, so I'm assuming it wasn't downloaded properly.

### DIFF
--- a/browser/src/Input/KeyBindings.ts
+++ b/browser/src/Input/KeyBindings.ts
@@ -23,7 +23,7 @@ export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configurati
         input.bind("<m-p>", "quickOpen.show")
         input.bind("<m-s-p>", "commands.show")
         input.bind("<m-enter>", "language.codeAction.expand")
-        input.bind("<m-t>", "language.symbols.workspace")
+        input.bind("<m-t>", "language.symbols.workspace", () => !menu.isMenuOpen())
         input.bind("<s-m-t>", "language.symbols.document")
 
         if (config.getValue("editor.clipboard.enabled")) {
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for fixing this, @CrossR !

> This should fix that by checking the menu isn't open first, assuming this wasn't by design?

Yes, this looks right, it was a regression on my part when I introduced symbol search. Thanks for catching it and fixing it!

> I'm putting a very early table of the keybinds in the wiki as well, so please fill out around it to add a bit more context! (I also wasn't 100% on the LSP Expand keybind, so I put in a placeholder for now).

That's awesome! This will be a huge help. I'll take pass through it and add some more info.

> There looks to be a number of keybinds with both a C and M version, such as this PR.

Yes, there are some differences in behavior between OSX + Windows/Linux. For OSX, it's more tradditional to use the `Meta` (Apple) key for operations - for example, open fuzzy finder is `Control+P` on Sublime Text/Atom/VSCode on Windows, but `Meta+P` for Sublime Text/Atom/VSCode on OSX.

There's a PR (#824) and an Issue (#822) tracking this with some more context, if you're interested. I like the way you formatted the wiki to reflect the different keybindings for OSX + Windows.

> I'm having some issues with ripgrep since the move to our own repo, it was very flaky getting it installed so I'd appreciate you testing this PR to check its working as intended. Am I missing something I need to do to get it installed? The neovim download worked perfectly, but ripgrep seemed to not return any results, so I'm assuming it wasn't downloaded properly.

A couple things to try:
- Remove the `node_modules` directory completely (I like to use `rimraf` on Windows)
- Run `npm install` again

Sometimes, if there are older / out-of-date dependencies there can be problems. This will ensure you're on a clean slate. The next thing to check is if the download succeeded - you can check the `node_modules\oni-ripgrep\bin\rg.exe` folder to see if it is there.

And of course, you'll need to `npm run build` again. LMK if you're still seeing issues after that.

In the meantime I'll test out the change and report back.

bryphe:
Work great for me! Thanks @CrossR . I'll bring this in now.

PR #933, opened 2017-11-13
Auto-closing pairs

bryphe:
I miss the auto-closing pair functionality of VSCode and IDEs, so I wanted something that works and will be compatible with language definitions.

Still need to test this day-to-day (which is why it is gated by the `experimental.autoClosingPairs.enabled` flag), but it seems to work OK in my testing.

A few ideas for next steps:
- Handle `<CR>` behavior when between auto-matched pairs, to create a new line
- Improve perf on the opening pair - it feels a bit sluggish as the cursor position gets re-mapped (due to the back-and-forth currently required)


bryphe:
> Have you considered using PairUtils vim plugin for this (it’s quite good)?

Actually I didn't know about that one... I tried a couple of them (https://github.com/jiangmiao/auto-pairs was the one I was primarily considering, as well as https://github.com/Raimondi/delimitMate). 

One consideration I had while testing is incorporating some of the niceties that VSCode has - in particular, it's smart enough to toggle auto-pairs via language syntax (using the textmate language integration), so I was hoping to bring that in to Oni too. I couldn't find a Vim plugin that did that. It'd be straightforward to add that functionality to Oni once #848 is integrated.

For some reason I couldn't find PairUtils to test out - the only link was to a vimrc in VimBox 😄 
 https://github.com/jordwalke/VimBox/blob/master/dotVim/keysVimRc

I checked [vim-awesome](https://vimawesome.com/?q=pairutils) too and didn't see it - do you happen to have a link?

> It then makes it possible to instantly boot up an interactive Vim editing window (command/ctrl n), while lazily loading all the js code for window/modal creation and async LSP interaction.

The auto-pairs code on the Oni side is relatively small, so it should have a negligible impact on perf, but it is good to consider (so as not to run into a death-by-1000 cuts scenario..). My concern with it at the moment is _responsiveness_, as it has to make at least one round trip to Neovim aside from the actual input.

But I agree for sure with streamlining the startup - I'm working on adding some more discipline/rigor to the startup path (#942) - by calling out exactly what tasks need to happen and parallelizing. We still unfortunately have a bunch of low-hanging fruit in the form of `.sync` calls for loading config / discovering plugins - but that can and should all be parallelized with starting the `nvim` instance. Startup is very ad-hoc today...

You brought up another good point (which #942 won't address) - we have one giant `bundle.js` that has everything - bootstrapping/init code, UI, LSP, etc - we should split that up so that we can lazy load all the non-essentials. Opened #943 to track that - thanks for your thoughts around this!

jordwalke:
I had the name wrong: it’s 'MartinLafreniere/vim-PairTools.

You might seriously consider it. Like you, I’ve tried everything else and they are all terrible. This one is really good, allows per language configuration and even supports jsx pair insertion etc.

bryphe:
Cool, thanks @jordwalke ! I'll check this one out - tracked this in #947

PR #934, opened 2017-11-13
Feature/incorporate shell env package

Akin909:
I had another go adding `shell-env` I previously tried `shell-path` a similar more focused module with no success. I haven't added explicit error handling as from the source code the fn only throw an error if it's passed an incorrect argument (I havent passed it any in which case it always returns a string)

Shell env on the other hand seems to be working, It required a few dependencies which I added. I also added a `shell-env.d.ts` as it required typings though I haven't fleshed this out  (wasn't entirely sure how).


bryphe:
Thanks for the PR, @Akin909 , and trouble shooting the issues with `shell-env` 😄

browser/src/Plugins/Api/Process.ts
```diff
@@ -27,7 +28,8 @@ const mergeSpawnOptions = (originalSpawnOptions: ChildProcess.ExecOptions | Chil
         },
     }
 
-    const existingPath = process.env.Path || process.env.PATH
+    const shellEnvironment = shellEnv.sync()
```
bryphe: Is it possible to use the `async` version of the API? I'm on a crusade to try and minimize `sync` calls since they can sometimes cause stutters in the UX. 

In this case, this would be called when opening up a file, and there are a few other bottlenecks we have at that point.

I'm thinking of a couple alternatives:
- We could switch to using the async version here, but that might be painful, because we'd need to wire up `async` and `Promise` up to the call sites
- We could call the sync version once on startup. I'm looking at how we can streamline our startup (#938), and we could potentially switch to using the async version at that point, and execute in parallel to some of our other startup tasks. But for now, we could just call the sync version on startup to seed the values. It shouldn't change while we're running, AFAIK

browser/src/Plugins/Api/shell-env.d.ts
```diff
@@ -0,0 +1 @@
+declare module "shell-env"
```
bryphe: Ah ya, I was looking to see if there was a DefinitelyTyped definition for this (`npm install @types/shell-env`), but unfortunately there isn't - so we'll have to go with this plan! Thanks for figuring this out.

package.json
```diff
@@ -121,6 +121,10 @@
     "ocaml-language-server": "1.0.12",
     "oni-neovim-binaries": "^0.0.7",
     "oni-ripgrep": "0.0.3",
+    "shell-env": "^0.3.0",
```
bryphe: Could you please set the version numbers to fixed versions? (IE, `0.3.0` instead of `^0.3.0`).

Sometimes we get burned by dependencies updating and breaking us.

browser/src/Plugins/Api/shell-env.d.ts
Akin909: I tried adding some typing to the module declaration but not exactly sure how that's done and  could quickly figure this out from the docs

browser/src/Plugins/Api/Process.ts
Akin909: I opted for option one here and used async await and added returned promise typings up to the call sites hopefully this doesn't clash with #938

bryphe:
Awesome, thanks for going through and `async-ifying` all those code paths!

bryphe:
Sorry about the build issues - was just looking at these.

One of them was potentially concerning:
```

  1) ci tests AutoCompletionTest ci test: AutoCompletionTest:
      AssertionError [ERR_ASSERTION]: false == true
      + expected - actual
      -false
      +true

```

It looked like the AutoCompletionTest potentially failed, which means there might be a regression with the TypeScript language service?

I'm thinking potentially this spot may need to be updated also:
https://github.com/onivim/oni/blob/22f4f4cc9046b1b169e3e5f583d49e109a439a40/vim/core/oni-plugin-typescript/src/TypeScriptServerHost.ts#L42

Akin909:
I'll have a look at resolving that although it seems like that would make it an async call meaning it can't be in the constructor function 😟 , i'll have a go at trying figure out how and where to call that function if it's async. Any suggestions?

bryphe:
Ah ya, that's a pain! Here's an idea of what we could do:
- Create a private variable to hold a promise, like `this._initPromise`
- Refactor out the async call to a helper method, like `startTypeScriptServer`
- In the constructor, kick that off, and cache the result, like:
```
constructor() {
this._initPromise = this._startTypeScriptServer()
}
```

Then, we make `_makeTssRequest` async, and add at the very top:
```
    public _makeTssRequest<T>(commandName: string, args: any): Promise<T> {
     await this._initPromise
...
```

So that it waits until the server is started before it starts processing requests. Hope that helps, thanks for looking at it!

Akin909:
Thanks for the pointer @brype 👍 , hopefully my changes are in the right vein

Akin909:
I've remove the `try-sleep-thread` module which seems to be a subdependency of shell-env or one of its dependencies, the functionality seems unchanged and it seemed to have been causing the build failure on windows. This does result in a warning from `npm` that the module is missing  though 😟

CrossR:
Just a heads up that your last commit / merge from master, the submodules for the vim plugins seem to have been deleted.

I think this is because the PR was setup before the git submodules were swapped over to, then I assume the act of merging master with this change caused some issues. Sorry!

bryphe:
Nice catch, @Akin909 ! Sorry you've had to keep sync'ing this PR.

I pulled in your change on `oni-api`, and published a new version with the updates: `0.0.5`.

So hopefully if you update that version in our `package.json`, we should be all set

Akin909:
@bryphe thanks 👍 just upgraded the package and updated the PR

bryphe:
It's weird, looks like there are failures:

```
  1) ci tests AutoCompletionTest ci test: AutoCompletionTest:
     AssertionError [ERR_ASSERTION]: Validate editor element is present
      at Suite.<anonymous> (lib_test/test/CiTests.js:97:17)
      at fulfilled (lib_test/test/CiTests.js:4:58)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
  2) ci tests BasicEditingTest ci test: BasicEditingTest:
     AssertionError [ERR_ASSERTION]: Validate editor element is present
      at Suite.<anonymous> (lib_test/test/CiTests.js:97:17)
      at fulfilled (lib_test/test/CiTests.js:4:58)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
  3) ci tests QuickOpenTest ci test: QuickOpenTest:
     AssertionError [ERR_ASSERTION]: Validate editor element is present
      at Suite.<anonymous> (lib_test/test/CiTests.js:97:17)
      at fulfilled (lib_test/test/CiTests.js:4:58)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
  4) ci tests NoInstalledNeovim ci test: NoInstalledNeovim:
     AssertionError [ERR_ASSERTION]: Validate editor element is present
      at Suite.<anonymous> (lib_test/test/CiTests.js:97:17)
      at fulfilled (lib_test/test/CiTests.js:4:58)
      at <anonymous>
      at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

I tried re-running the build a couple times and still see them (these are tests run against the published package - you can try it on your machine by running `npm run pack` and `npm run test`). On my Mac they seem to be passing, so I'm going to try deleting the cached modules and see if that helps - maybe an outdated dependency is in the Travis's cache and causing problems.

Akin909:
Tried running `npm run pack` but this command just hangs on my system and I eventually have to kill it, not sure if i'm missing a step, when I run npm test without the pack command which may not be relevant since I haven't generated a new package all I get is the autocompletion test error.

bryphe:
Hmm, strange, where does it hang? Are you running on OSX? It should work as long as you've done `npm run build` and then `npm run pack` - it does take some time though. It's not a bad idea to delete the `dist` folder too if it is around. It also helps to close any instances of Oni that are running.

I tried kicking off the build again (after clearing the cache), but it seems the tests are still failing. I'm not sure why - I wonder if it is related to one of the dependencies brought in? The other branches seem to be passing these tests. My main concern is that usually these tests indicate something is wrong with the package, and when we publish the next release the OSX zip/dmgs would be broken.

It would be helpful to have the built OSX package (like the .zip or .dmg) to try running locally - that might give us enough info. Unfortunately TravisCI doesn't host the build artifacts like AppVeyor does.

Akin909:
@bryphe  I'm running MacOS, high sierra (version 10.13.1), I've tried running `npm run build` then `npm run pack` but this process reaches the point where it says its building the dmg but then stalls (I've left it for at least an hour at some point with no progress), i'll try running the command on a different branch to find out if its related to my system or my branch and then try taking out the dependencies to see if they resolve the issue

Akin909:
@bryphe I've tried running the command on my version of master with the node modules cleared and I still get the same issue, here's a screenshot with the warnings I'm getting.

screen shot 2017-11-23 at 21 53 07

I'm not familiar with `asar` but not sure why or if that'd could be the culprit

bryphe:
Ah interesting.... thanks for looking into that. That's really frustrating!

> I'm not familiar with asar but not sure why or if that'd could be the culprit

The asar issue is actually okay to ignore.

I did see in that thread that one user had success clearing `node_modules` and then using `yarn` to install. It might be worth trying that:
- Install yarn and rimraf (`npm install -g yarn rimraf`)
- Delete node_modules (`rimraf node_modules`)
- Run `yarn` - (`yarn install`)

I'll look into this a bit next week and see if I can upload the OSX package it's building, so we can download and inspect it (and run it locally) - maybe there'd be some clues there. I'd also like to incorporate some more logging output into the tests (like, being able to see the console) - right now it just feels like we're in the dark when there are failures!

bryphe:
My suspicion too is that one of the dependencies is causing a problem with the installed package... so another option would also be to step back and see if we could implement this w/o those additional packages.

It seems like we could potentially run these commands to get the same info:
```
import * as childProcess from "child_process"

// Windows
const paths = childProcess.execSync("echo %PATH%")

// OSX / Linux
const paths = childProcess.spawnSync("echo $PATH")
```

And worst case, if they fail, we'd just fall back to using `environment.additionalPaths` (it would be good `Log.warn` though). But I think this would handle the 99% case. 

Ideally we could switch to the async version and `promisify` them, leveraging all the work you did to make those code paths async 😄

Akin909:
@bryphe I managed to get the app building by upgrading electron-builder to the latest version using `yarn`, which seems to have resolved the issue. Testing against that I now get` "before each" hook for "ci test:` for all CI tests and each test specifically errors with the following message ` Error: ChromeDriver did not start within 5000ms` .

I think it might relate to making the startup async as `Oni.start` in `AutocompletionTest.ts` is sync and meanwhile i think that startup now returns a promise though awaiting it doesn't seem to work. I'll have a closer look (I seem to also get this error with the test on my local master as well). 

Also given that child_process can be used to find the `PATH` at startup would it be preferable to use that and do away with shell env? or if I get it working with shell env shall I leave it in, looking at the source it looks to be a wrapper around using `child_process` with some intermediate steps and error handling thrown in.

Running my local dmg once built works fine btw.

Akin909:
I've refactored this PR to use to `Child_Process.exec` (promisified - using util.promisify) method, great find 👍 @bryphe , also removed shell-env and its dependencies, though my build version is still running into errors with the CI tests due to some async quirk somewhere which i'm investigating

bryphe:
Wow, nice work @Akin909 ! Really appreciate all the effort you put in this PR.

The change looks great - I didn't know there was a `promisify` method available in `util`, that's really helpful 👍 

The build looks like it is in much better shape, as the other 3 integration tests are passing. I ran your change locally on OSX and I'm not seeing the failure, either, which makes me suspect it's a test issue.

bryphe:
I'll debug the test some more, as it is also blocking my PR - #848 . I don't believe this failure is related to your change, so I'm good to bring this in (I'll investigate the failure anyway as part of #848 / #1015 , and set up our infrastructure so we're more robust in the future, and get more actionable logs).

But I'm good to bring this in if you're all set - don't want you to be blocked on this test issue. 👍

Akin909:
@bryphe 👍 yep all set I've been running this version of Oni locally and haven't had any issues I am getting a fluctuating set of results with the tests which was probably due to not running `build` and `pack` before the integration tests, now I'm getting the autocompletion error as before, did a bit of digging couldn't quite find what's causing it to fail

bryphe:
Awesome, thanks @Akin909 ! I tested it out locally and looks like it worked great for me.

> I am getting a fluctuating set of results with the tests which was probably due to not running build and pack before the integration tests

I hit this sometimes if I didn't run `npm run build`+ `npm run pack`as you suggested - since the integration test run on the last 'packed' build (in order to test the release packages), it can be confusing. Sometimes I'll run just `npm run build` and then `npm run test` and wonder why those are still failing.

>  now I'm getting the autocompletion error as before, did a bit of digging couldn't quite find what's causing it to fail

One case I hit locally that caused the test to fail was that I hadn't run `git submodule update --init --recursive`. Some of our vim plugins (like `typescript.vim`) are in submodules now, and the autocompletion test uses a `.ts` file - without those plugins, Vim actually thinks it is an XML file and thus we don't load the language server (causing the test to fail). Doesn't seem like the same issue as on the test machines unfortunately..

Anyway I'll go ahead and bring this in now - thanks for the contribution and all the work you put into this! 👍

Akin909:
Thanks @bryphe 👍

PR #935, opened 2017-11-13
Fix #923: Not loading `after` folder

### LINKED ISSUE #923: configuration in after folder is not recognized
A have certain configuration in the folder `.config/nvim/after` that is not being recognized by oni. Usually a have some extra syntax highlighting in this folder. Is there any reason why oni does not take into account this folder? How can I make use of oni taking into account this additional configuration in the folder `after`? 

My config.js is as follows:

```
const activate = (oni) => {
    console.log("config activated")
    oni.input.bind("<c-enter>", () => console.log("Control+Enter was pressed"))
}

const deactivate = () => {
    console.log("config deactivated")
}

module.exports = {
    activate,
    deactivate,
   "oni.useDefaultConfig": false, // default is true
   "oni.loadInitVim": true,
    "statusbar.enabled": true, // yes by default
    "editor.fontSize": "16px",
}
```

bryphe:
__Issue:__ We were not properly concatenating Oni's plugin runtime paths with Neovim's runtime paths. This could cause unpredictable issues with other plugins / Oni's built in plug-ins, depending on the ordering.

image

__Fix:__ Add a comma explicitly when setting the runtime paths. This SO post had good info too: https://superuser.com/questions/806595/why-the-runtimepath-in-vim-cannot-be-set-as-a-variable

### DIFF
--- a/browser/src/neovim/NeovimProcessSpawner.ts
+++ b/browser/src/neovim/NeovimProcessSpawner.ts
@@ -48,7 +48,7 @@ export const startNeovim = (runtimePaths: string[], args: string[]): Session => 
     }
 
     const argsToPass = initVimArg
-        .concat(["--cmd", `let &rtp.='${joinedRuntimePaths}'`, "--cmd", "let g:gui_oni = 1", "-N", "--embed", "--"])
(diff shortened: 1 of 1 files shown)


bryphe:
Fixes #923

Issue #900, opened 2017-11-09
Doesn't run on Ubuntu


vnikk:
The window opens and closes immediately.
The run script in electron ./app/oni.sh contains "TODO Enable Linux"

TalAmuyal:
I just tried 0.2.16 on Ubuntu 17.10 and it works for me.
Can you describe what actions you've taken? (I installed using `sudo dpkg -i Oni-0.2.16-amd64-linux.deb`)

keforbes:
The `oni.sh` file is only intended to be used by Mac.  I think the command to launch on Linux is in `./cli/oni`.  Regardless, if you're compiling from source it's best to just run `sudo npm link` so the `oni` command will be added to your path.

bryphe:
Do you have `nvim` installed and available in your path?

Unfortunately, we are not robust in the case where `nvim` is not found, and we will exit immediately as you describe. We need better error handling in that situation for sure!

bryphe:
The issue with the app crashing on Linux if Neovim isn't installed has been addressed with PR #937, and is now available in [0.2.17](https://github.com/onivim/oni/releases/tag/v0.2.17). I'll close this out, but feel free to reopen if you're still seeing issues.

PR #937, opened 2017-11-14
Fix #911 - Handle case when 'nvim' is not available on path

### LINKED ISSUE #911: Fix / Implement UI when neovim is not installed
There are several issues where Oni is behaving poorly in cases where `nvim` is not available, or the wrong version.

This needs to be made simpler and clearer (no crashing on start or black screen - just a UX stating that Neovim is not available, or is the wrong version.

bryphe:
Currently, the behavior is very poor when `nvim` is not available on the path. This occurs in Linux, as on OS X and Windows we package the binaries.

Right now, if `nvim` is not available (which is common in Linux), Oni crashes and burns in a bad way - like in #900 (just shows a white screen and exits).

This change adds a UI to guide users to install Neovim:
image

It's not perfect or ideal, but better than a flashing white screen and exiting instantly...

### DIFF
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -5,7 +5,6 @@
  */
 
 import * as React from "react"
-import * as ReactDOM from "react-dom"
 
 import "rxjs/add/observable/defer"
 import "rxjs/add/observable/merge"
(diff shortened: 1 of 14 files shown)


TalAmuyal:
Looks great! :+1:

Issue #938, opened 2017-11-14
Version to Version Upgrade (Backwards Compatibility)


bryphe:
## Version-to-Version Upgrade (Backwards Compatibility)

Currently, the statement on our README is to consider Oni 'unstable' and that updates may break your configuration, API may not be compatible, etc... and we've certainly been exercising that 😄 We've been in the "move-fast-and-break-things-mode"

I'd like to start investigate ways for us to provide a better experience in version-to-version upgrade, and incrementally improve. The ideal end-goal for Oni (for v1) is that we gracefully handle version-to-version upgrades, and users can be confident that updates won't break them or their config. This is especially important for auto-updates.

IMO there are two separate pain points today: _regressions_ (like the `Control+T` in QuickOpen) and API changes or other user data changes. This issue is primarily addressing the latter (I hope that we can build a better 'safety net' to prevent regressions via tests).

A great litmus test for this is the config migration work - ideally, when users open up a version of Oni from the previous one, we can know to migrate the old schema to the new schema.

I'm thinking as __high-level architecture__, we could have the following.

- The `version` of our data schema would be a monotonically increasing integer (I don't think we'd need the complexity of semver). We'll call this `ONI_DATA_VERSION`.
- We would keep a _counter_ of our data version. This would be _persisted locally_ on the machine (either a file, like `oni.version` or via `localStorage`). 

The `ONI_DATA_VERSION` would be incremented when something in our data schema changes (essentially, something that needs upgrade logic). This will probably be incremented much less often than our actual releases, and will be decoupled from the version in `package.json`. 

- We would have a `VersionToVersionUpgrader` that runs on startup of the app. This would check the local value of `ONI_DATA_VERSION` and compare it with the build's `ONI_DATA_VERSION`. If the builds `ONI_DATA_VERSION` is > than the local `ONI_DATA_VERSION`

- The actual work of the upgrade will be handled by an `IUpgrader`, which would look like this:

```
interface IUpgrader {
   fromVersion: number
   toVersion: number

   // Actual upgrade logic, like copy config to new location
   upgrade: Promise<void>
}
```

And the idea is we'd run each `IUpgrader` in a chain until we get to the latest `ONI_DATA_VERSION`. So it might be the case that a user has a config at `ONI_DATA_VERSION` 1, didn't use the app for a long time, and then starts using it again after upgrading, and now the build is at `ONI_DATA_VERSION` 5 - we'd run the 1- > 2 upgrader, then the 2->3 upgrader, etc.

[...]

PR #939, opened 2017-11-14
Fix #903 - Move Windows Config and Create Config folder if needed.

### LINKED ISSUE #903: Windows, config.js, E212: Can't open file for writing: no such file or directory
- Oni Version 0.2.16
- Winodws 10

1 - "File -> Preferences -> Edit Oni Config"
2 - add a configuration line
3- Enter `:w`, get the following error:  `E212: Can't open file for writing: no such file or directory:
image

When I execute `pwd`, I get the following: `C:\Program Files (x86)\Oni`

CrossR:
This PR moves the Windows config file from from `C:\Users\XXX\.oni\config.js` to `%APPDATA%\oni\config.js`. It will also make the relevant folder if needed on any platform.

Currently, I'm using `mkdirp.sync()` since the package was already there, but should I instead be using an Async version, and if so any suggestions what?

Since this PR is a breaking change, it may need doing alongside some of the ideas in #938.

### DIFF
--- a/browser/src/Platform.ts
+++ b/browser/src/Platform.ts
@@ -10,7 +10,7 @@ export const isMac = () => os.platform() === "darwin"
 export const isLinux = () => os.platform() === "linux"
 
 export const getUserHome = () =>
-    isWindows() ? process.env["USERPROFILE"] : process.env["HOME"] // tslint:disable-line no-string-literal
+    isWindows() ? process.env["APPDATA"] : process.env["HOME"] // tslint:disable-line no-string-literal
 
 export const getLinkPath = () => isMac() ? "/usr/local/bin/oni" : "" // TODO: windows + linux
 export const isAddedToPath = () => {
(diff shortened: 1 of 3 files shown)


bryphe:
Thanks @CrossR for fixing this! Change looks good, smoothing out the configuration saving will be a big win!

It looks like there are a couple issues with the build:
- Minor lint error (`npm run fix-lint` might take care of it) 
- I think the new automation tests that I added to verify the no-`nvim` case will fail, because they make assumptions about the config path:
https://github.com/onivim/oni/blob/744fd5765131be8c93428a64b574fa7c03e3ffdb/test/CiTests.ts#L51

So that place might have to be updated too for the tests to pass.

Once that is good, I'll bring this in. I'll call it out as a __BREAKING CHANGE__ in the next release's notes.

bryphe:
Looks good - the OSX failure is due to the token issue, github downloads are throttled. I'll go ahead and bring this in so we can include it in tomorrow's release - thanks @CrossR !

PR #942, opened 2017-11-14
[WIP] Fix #355 - Improve startup experience

### LINKED ISSUE #355: Improve Startup UX & perceived performance
The "startup" experience of Oni is very choppy and disjointed. The experience today is:
- ...some delta between running 'oni' and having a browser window show up
- A blank, white screen is shown
- A black screen with a green cursor in the top left is shown
- The editor finally appears

I'd like to address it by the following:
- Show a splash screen as quickly as possible
- Show a black screen with a spinning 'oni' logo in the center of the screen immediately
- Once Oni is loaded, have the editor fade in and the 'oni' logo disappear

Once we have 'spectron' tests, we can start having a benchmark for the load time for Oni across different platforms.

bryphe:
This change formalizes the startup tasks we require on load, and ideally will improve on the UX.

This opens the door for us to move away from some of the synchronous loading, specifically:
- We should refactor config loading to be asychronous
- We should refactor plugin discovery to be asynchronous

In addition, I'd like to have a transition-in effect for the Oni UI

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,4 @@
+- Animate startup
+- Add loading spinner
+- Add transition-out for loading screen
+- Add transition-in for tabs + statusbar
--- a/browser/src/Lifecycle/Startup.ts
+++ b/browser/src/Lifecycle/Startup.ts
@@ -1,0 +1,35 @@
+/**
+ * Startup.ts
+ *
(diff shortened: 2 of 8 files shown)


bryphe:
Switching to #1085

PR #944, opened 2017-11-14
[WIP] Swap Bundled plugins to Submodules

CrossR:
The bundled plugins we have are coming up on 1 year old, so they need updating in some fashion.

I've added them as submodules, partly to reduce the code clutter in the repo, and also so they are more easily kept up to date.

Few other changes to consider before this is potentially merged.
* README / Wiki updates
* CI Updates potentially?


bryphe:
Cool, looks like `travis-ci` adds in the `git submodule update --init --recursive` for us:
image

The failure on OS X is due to the github throttling the downloads.

I'm not sure why the appveyor build didn't kick off for this - I believe for that we will need to explicitly add the command to update submodules there: https://github.com/appveyor/ci/issues/899

CrossR:
Not actually changed any of the AppVeyor stuff before, would this just be adding `git submodule update --init --recursive` into the  `install:` section of `appveyor.yml`? Looks like that is the part ran straight after cloning.

bryphe:
> would this just be adding git submodule update --init --recursive into the install: section of appveyor.yml? 

Yes, I believe that should be sufficient. We'll be able to look at the appveyor logs and see if it worked

CrossR:
The x64 build is still ongoing, but the x86 test has passed fine, and the logs show that it is correctly cloning the submodules now, by the looks of things.

CrossR:
Looks like both the Appveyor builds are fine now and I'm assuming the Travis failure is due to #971.

bryphe:
Cool, ya, looks like the build is good!  Thanks for setting everything up. 

Yes, I'm hoping #971 finally addresses these throttling issues we get when downloading from github ☹️ 

I'll bring this in - I updated the wiki here with the new step, feel free to edit: https://github.com/onivim/oni/wiki/Development

bryphe:
Just saw you fixed the repo URL in the wiki, thanks!

PR #946, opened 2017-11-15
Optimistic typing updates

bryphe:
One consequence of the architecture we have - Electron talking to Neovim via `msgpack-rpc` - is that input has a non-trivial round-trip latency, regardless of the configuration.

For an editor, ideally you'd want an avg responsiveness of <10ms from keypress to rendering for a desirable experience (ex, [Typing With Pleasure](https://pavelfatin.com/typing-with-pleasure)). Avg of 15ms is pretty good.

At least based on the chrome profiler, we aren't there yet - and we spend a significant amount of time idling until we get a response from Neovim, in which case a bunch of stuff happens - lots of redux actions in response to mouse-move, canvas re-rendering, etc.

This change implements 'optimistic updates' - similiar to the IntelliJ IDEA zero-latency (it also has parallels to [Client-side network prediction](https://developer.valvesoftware.com/wiki/Source_Multiplayer_Networking) models used in multiplayer games, just a really simple version!).

For insert mode, we can predict fairly reliable on the cursor position + character, so we should render that _immediately_.

Not only does this mean that we achieve better typing latency, but it also amortizes the workload in a more efficient way (since the cursor already moved, and lots of components get re-rendered in response to that, the canvas rendering can occur more quickly too).


jordwalke:
What are some of the things that can go wrong? How does this work with auto wrap mode?

jordwalke:
How long does a neovim round trip take? Is there a way to bump the neovim process priority to help improve it?
Since key strokes are highly interactive, it would be very helpful to align the keyword interactions with display refresh events, such that the UI is updated by the next frame update deadline. Would it be possible to run neovim in essentially an embeded library mode (even if it goes over serialized rpc - it could exist in the same process so that we are not at the mercy of the operating system scheduler’s opinion about how important neovim is to Oni.)

cc @justinmk

justinmk:
Vimr uses nvim as an in-process lib.
https://github.com/qvacua/vimr

But this doesn't sound like RPC is the problem :

> we spend a significant amount of time idling until we get a response from Neovim, 

And the approach proposed by @bryphe   here is very interesting. Really interested to see how it works out.

bryphe:
Thanks for the feedback, @justinmk  & @jordwalke !

> What are some of the things that can go wrong? How does this work with auto wrap mode?

This is interesting because I called out the source networking article / multiplayer games as an example, and this has parallels 😄  Most modern multiplayer games predict client input - so when you press `forward` you move forward instantly, without needing a round-trip from the server. The thing is, maybe when the server reconciles the motion, it realizes there was another player in front of you - so you predicted wrong. This is non-trivial to solve because it can be pretty jarring - you can teleport the user, you can `lerp` their position, etc...

Our case is much simpler but we still get a few of those error cases:
- __Line wrap__ - our logic is simple and we don't try and predict when the line will wrap
- __Bindings__ - some users have mappings like `jj` that wouldn't actually insert two `j`s, but go back to normal mode - we predict wrong...

However, the simple logic we have actually works well for these cases. We have it much easier than say Overwatch in this case, because both the time-to-correction and the degree of error in the prediction is much lower.

In both cases where we predict wrong, there isn't much 'badness' - both cases implement an _overshoot_ animation which conveys energy. The __binding__ case is kind of fun because it feels like the cursor gets pulled back into normal mode, so IMO there isn't any special handling needed here. (I will gate it with a configuration variable though!)

> How long does a neovim round trip take? Is there a way to bump the neovim process priority to help improve it?

Grabbed some traces to provide more details here. From my basic research so far, Windows has the highest latency (`nvim` gives us lots of options to connect, so #940 is looking at using named pipes as an alternative to stdio).

But here's the current state of behavior on `master`. This trace is a _single keypress_ on `README.md`:
image

- The interaction starts are _~3890ms_
- The response from `nvim` occurs at _~3910ms_ (__green arrow__)
- The paint happens at the end of the frame _~3940ms_ (__blue arrow__)

This means the approximate latency for that keystroke is _~50ms_, which isn't really ideal.

The trace for the 'after' case is similiar right now (I haven't hooked up the predicted text to all the elements that use cursor position, which would fill up some of that idle space more effectively), with one major exception:

image - the spot where that __black arrow__ is where we render the predicted text - that means we show the typed character in _~10ms_ from the start of the interaction, which is a huge improvement!

For debugging, I highlight the 'predicted' characters - here's a gif:
predicted-typing

Each of those cases where you see a red-background character is a case where there would've been a _subtle but perceptible delay_ in the typing interaction. (One note - there's actually more than the gif reflects, because my gif recorder maxed out at 40fps/25ms per frame, so it didn't catch all of them, but I think it gives an idea)

bryphe:
A few pieces left:
- Hook up other elements that rely on cursor position to the predicted position, to amortize render time
- Clear predictions using an array instead of calling the action multiple times, to avoid unnecessary re-renders
- Add a configuration variable to enable/disable this feature
- Add a debug setting to show the predicted characters

jordwalke:
Very interesting findings.

One other potentially jarring experience I'm curious about:
Suppose you have a syntax highlighting rule that highlights `let`. If you type `le` it will remain white text, but then when you hit the final `t`, today, Vim will instantly highlight the `let` keyword. Does predictive editing allow for a moment where `let` is white, but then asynchronously highlighted to be the correct color? If so, this should probably be added to the list of error cases even though I think this may not be bad at all - compare to Atom's very distracting FOUC (Flash Of Unstyled Content) when opening an entire file.

 I've always wished that Vim had *built in* support for predictive edits. The approach you're taking is a great way to explore the benefits even if not baked into Vim/NeoVIm. Along with every UI update command, I would like Vim to report back a set of possible (and commonly executed) key combinations that might be entered next, and how they will affect the UI. I imagine something very minimal like:
```
alphanumeric: [insert_char_at(x=34, y=98, bufferID)]
esc: [leave_insert_mode, block_cursor, move_cursor(x=.., y=.., bufferID)]
```
If baked into NeoVim, it could account for line wrapping as well.

> This means the approximate latency for that keystroke is ~50ms, which isn't really ideal.
> ...
>  the spot where that black arrow is where we render the predicted text - that means we show the typed character in ~10ms from the start of the interaction, which is a huge improvement!

That's really great!

I'd be interested to do the experiment with a completely empty vim configuration and no syntax highlighting to try to isolate what is purely NeoVim + communication overhead.

I would highly recommend recording the interaction with a high speed camera because the instrumentation can lie sometimes, and the act of having the instrumentation active/recording can (in my experience) disrupt the measurement itself and provide misleading results. The iPhone 6/7Plus has a 240fps mode which allows you to time with a granularity of 4ms.
Here's what I do: I record a video of hitting a key very quickly, and then isolate the part of the video where I can hear the key enter - and begin counting the number of frames from that point until the time when the screen updates and multiple by `4ms`. If you don't have an iPhone Plus, I can perform this experiment if you make a release of Oni that has predictive edits enabled (the red flashing is also helpful).

bryphe:
> One other potentially jarring experience I'm curious about:
Suppose you have a syntax highlighting rule that highlights let. If you type le it will remain white text, but then when you hit the final t, today, Vim will instantly highlight the let keyword.

Yes, this is a good one to call out too. The predictive editing would use the default color, and then it would snap once the highlight is picked up. As you're typing `l`, `le`, and `let`, the `l` and `le` would be the default color anyway, there'd be a split second where `let` would be the default color, and then the highlighting would snap-in. As you mentioned it should be much less distracting then the lack of syntax highlighting when opening a file 😄 

> I've always wished that Vim had built in support for predictive edits.

This would be really interesting! It would also help in the case of 'lowering the bar' - we could do a better job of helping new vim users know what sort of key combinations they may wish to try, and what the outcomes would be. 

> I'd be interested to do the experiment with a completely empty vim configuration and no syntax highlighting to try to isolate what is purely NeoVim + communication overhead.

Definitely - there is also some bottleneck in our `init.vim` we load for Oni to listen for autocommands.

> I would highly recommend recording the interaction with a high speed camera because the instrumentation can lie sometimes, and the act of having the instrumentation active/recording can (in my experience) disrupt the measurement itself and provide misleading results.

This would be a good next step! There are some additional bottlenecks that can show up here too (keyboard latency, usb latency, display latency, etc). In the interim, I did some quick profiling with [Typometer](https://pavelfatin.com/typometer/) just to get ballpark, and the results so far are very promising:

image

I just did the empty file benchmark, on Windows, and Oni's average response time went from __47ms__ to __29ms__. That essentially takes us from the ballpark of Atom & VSCode to Sublime territory! 😄 (well, at least for this particular benchmark - will be interesting to see OSX / Linux performance, files with syntax highlighting, etc...).

jordwalke:
Really exciting!

bryphe:
FYI - I tested on master against some of the `intl` scenarios (dead keys, IME), and it seems to be working well. There was one case that had a bug, `:inoremap jj <ESC>`, where the predicted characters would stick around, but just pushed up a fix.

I'll bring this into master behind an `experimental` config variable - these are some config values to play with:
- `experimental.editor.typingPrediction` - set to `true` to enable prediction
- `debug.showTypingPrediction` - set to `true` to enable highlighting for the predicted characters

I also added another value that was helpful for debugging some cases:
- `debug.fakeLag.neovimInput` - set to number (milliseconds), like `100` - this causes us to add extra latency to the input path to neovim. Helpful for testing cases where there are lots of predicted values, wrapping, etc. It's also interesting to see how it changes the feel of typing if you set this value w/o prediction... Of course, make sure to disable this when you're done, otherwise Oni will feel super slow 😄

So it'll be off by default for but can be enabled by setting `experimental.editor.typingPrediction` to `true`. If things look good I'll move it out of the `experimental` bucket and turn it on by default for the next release.

PR #960, opened 2017-11-16
Fix #951: Sass support

### LINKED ISSUE #951: Add default configuration for sass to use `css-language-server`
The css-language-server supports sass, so it should be straightforward to add a default configuration that uses this. Just need to add + test it.

bryphe:
__Issue:__ Sass files (`.scss`) weren't being picked up by the bundled `css-language-server`.

__Defect:__ The filetype is `scss` instead of `sass`

__Fix:__ Use correct filetype in default configuration


bryphe:
Fixes #951

PR #965, opened 2017-11-16
Fix #957 - Remove results-caching logic in MenuReducer

### LINKED ISSUE #957: QuickOpen is scoped to single folders/files/exact file path
Suppose you have a file with the same name on two different folders, for instance:

- `signup/index`
- `signin/index`

Usually I would type something like `upindex` or `inindex` to open that. This is a very simple example but for deeply nested files is incredibly useful, e.g. for a Rails project:

- `app/models/user`
- `app/controllers/users_controller`

I would type `amouser` and `acuser` to open each respectively.

It seems the current fuzzy finder is restricted to a single folder/file or the exact file path, so if I type `models` I get all model files, but if I type `modelsu` I get an empty list.

Not sure if this is hardcoded by ripgrep but would help me a lot if could be configured (or enabled by default), and one of the reasons I wanted to use a different fuzzy finder (apart from the previous case sensitivity issue, that was fixed).

I know for instance that this works as I expected on VSCode and all fuzzy finder plugins for vim I have used so far.

bryphe:
__Issue:__ When typing quickly in the Menu, not all expected results are available.

__Defect:__ We had some code that uses the 'cached' results for the menu when searching. This logic worked fine in the case where we acquired all results at once, but in the 'streaming' case, we'll lose any results that come in after the first filter.

__Fix:__ Remove the caching logic, and always filter against the set of items.

The only concern I have with this fix is it might cause a degradation in performance in large directories. But we should be accurate in the results - it's hard to have confidence in a fuzzy finder that doesn't find everything... We can find an alternative fix to performance if necessary.

### DIFF
--- a/browser/src/Services/Menu/MenuReducer.ts
+++ b/browser/src/Services/Menu/MenuReducer.ts
@@ -96,10 +96,8 @@ export function createReducer<T, FilteredT extends T>(filterFunc: MenuFilterFunc
                 if (!s) {
                     return s
                 }
-                // If we already had search results, and this search is a superset of the previous,
-                // just filter the already-pruned subset
-                const optionsToSearch = a.payload.filter.indexOf(s.filter) === 0 ? s.filteredOptions : s.options
-                const filteredOptionsSorted = filterFunc(optionsToSearch, a.payload.filter)
+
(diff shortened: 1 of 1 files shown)


bryphe:
Fixes #957

Issue #961, opened 2017-11-16
Filetype wheel continuously spinning


bpierce1:
I'm seeing a filetype wheel that continuously spins on any file type. Here's an example:

![image](https://user-images.githubusercontent.com/2482596/32906299-5a880792-cab1-11e7-9898-982639b69dcd.png)

It's happening at least on Ubuntu 14.04 and OS X with the latest version 0.2.17 (and some previous versions as well). 

I notice the CPU usage is also much higher while this is happening. Have I misconfigured something?

bryphe:
Thanks for logging the issue, @bpierce1 !

I believe there are two problems here:
- `javascript.jsx` not supported - we should be bringing in the out-of-box javascript language service (#882)
- In general, if the filetype is not recognized / no language server shown, we are showing that spinner (even in places like the quickfix list).

bpierce1:
Yes that was an example, but I see it on all filetypes. But you're right - I currently don't have a language server. Is there a way to disable the need for the language server so that the CPU utilization isn't so high?

By the way, I noticed the spinning began in 0.2.15, and isn't there in 0.2.14.

Metamist:
In my developer tools, I am getting this error:
```
Error: No content available.
    at TypeScriptServerHost._parseResponse (/home/metamist/apps/oni/resources/app/vim/core/oni-plugin-typescript/lib/TypeScriptServerHost.js:239:49)
    at Interface.<anonymous> (/home/metamist/apps/oni/resources/app/vim/core/oni-plugin-typescript/lib/TypeScriptServerHost.js:60:23)
    at emitOne (events.js:115:13)
    at Interface.emit (events.js:210:7)
    at Interface._onLine (readline.js:278:10)
    at Interface._normalWrite (readline.js:420:12)
    at Socket.ondata (readline.js:138:10)
    at emitOne (events.js:115:13)
    at Socket.emit (events.js:210:7)
    at addChunk (_stream_readable.js:252:12)
```
Is it trying to use typescript inside `javascript.jsx` files?

CrossR:
I used to notice this, but in the latest release, it looks like its been fixed.
Can this be closed now?

bryphe:
Cool, ya, it should be fixed now (at least the latest release had some improvements around the statusbar state). Thanks @CrossR !

PR #966, opened 2017-11-16
Fix #882 - Handle 'javascript.jsx' filetype

### LINKED ISSUE #882: Dev tools says "No supported language" for "javascript.jsx"
It seems that all my `.js` files are opened as `javascript.jsx` language (says so in the status bar). This causes autocomplete to not work at all, and I can't figure out how to switch to `javascript` language.
`set syntax=javascript` does not do anything except change colors.

Edit: Seems like `set filetype=javascript` works to get autocomplete to work again, however the status bar does not update unless you switch to another tab and then back.
I do not want to use filetype javascript since I am working with `.jsx` files, not `.js` (however, all my `.js` files open as `javascript.jsx`, so `.js` files does not get autocompletion either), and the autocomplete should work the same for both.

bryphe:
__Issue:__ With `vim-polyglot`, which is a popular plugin, JSX files show up as `javascript.jsx`, and no completion is available.

__Defect:__ Our built-in javascript/typescript integration expect a filetype of `typescript` or `javascript`.

__Fix:__ Disregard any `.` in the language name and grab the first part when looking for an available language server. If it turns out there are more generalized mappings that are interesting, we could add a configuration value to map filetype -> language server.

### DIFF
--- a/browser/src/Services/Language/LanguageManager.ts
+++ b/browser/src/Services/Language/LanguageManager.ts
@@ -254,7 +254,15 @@ export class LanguageManager {
     }
 
     private _getLanguageClient(language: string): ILanguageClient {
-        return this._languageServerInfo[language]
+        if (!language) {
+            return null
+        }
+
+        // Fix for #882 - handle cases like `javascript.jsx` where there is
+        // some scoping to the filetype / language name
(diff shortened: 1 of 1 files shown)


bryphe:
Fix #882 and part of #961

PR #967, opened 2017-11-17
Fix #961 - Don't show a spinner if no language server is available

### LINKED ISSUE #961: Filetype wheel continuously spinning

bryphe:
__Issue:__ In the case where we don't have a language server for a filetype, we get stuck showing a spinner. This even happens for buffers like `qf` (quickfix).

__Defect:__ We'd always set the statusbar state to initializing but not turn it off. 

__Fix:__ For the case where we don't have a language server, we should set the state to `NotAvailable`.

### DIFF
--- a/browser/src/Services/Language/LanguageManager.ts
+++ b/browser/src/Services/Language/LanguageManager.ts
@@ -37,7 +37,7 @@ export interface ILanguageServerNotificationResponse {
 export class LanguageManager {
 
     private _languageServerInfo: { [language: string]: ILanguageClient } = {}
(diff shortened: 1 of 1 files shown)


badosu:
Thanks, this was happening with me for Ruby

PR #969, opened 2017-11-17
Fix "Edit NeoVim config" command.

CrossR:
Looks like I broke the init.vim edit command when I updated the `getUserHome()` command for Windows.
I've updated it to use the environment variable instead now.


bryphe:
Thanks for catching this, @CrossR!

PR #974, opened 2017-11-18
[WIP] LSP - Code Action support

bryphe:
This is a WIP to add `Code Action` support. For now, this will be available in `visual` and `normal` mode. In addition, we should support querying code actions based on diagnostics.

I still need to figure out the UX to show when code actions are available - I'm considering showing a `lightbulb` icon in the right margin, sort of where the errors are. This would give a heads-up if there is a code action that can fix an error, as well as show that actions are available when there are no errors.  Need to experiment a bit.

But when there are code actions available, you can press `Alt-Enter` and get a context menu with the available options:
image

The TypeScript server supports a few different refactorings, which is pretty cool... And the ocaml-language-server supports some scoped fixes for errors, which is also cool.


bryphe:
Closing this in lieu of #1326 , which is updated against the latest `master` .

Issue #950, opened 2017-11-15
Configuration: Add `vim.colorscheme` setting


bryphe:
Add `vim.colorscheme` setting, so that it is easy to change the colorscheme without having to touch `init.vim`.

Later, I'd like to expose custom theming for Oni, which would be a vim colorscheme + styles for the UI (and potentially a CSS stylesheet, #412).

bryphe:
PR #975 adds this as `'ui.colorscheme'`

bryphe:
Added a note in the wiki: https://github.com/onivim/oni/wiki/Configuration#ui

PR #975, opened 2017-11-18
#950 - `ui.colorscheme` setting

bryphe:


browser/src/Services/Configuration/DefaultConfiguration.ts
```diff
@@ -138,6 +138,8 @@ const BaseConfiguration: IConfigurationValues = {
     "ui.animations.enabled": true,
     "ui.fontFamily": "BlinkMacSystemFont, 'Lucida Grande', 'Segoe UI', Ubuntu, Cantarell, sans-serif",
     "ui.fontSize": "13px",
+
+    "vim.colorscheme": "onedark",
```
badosu: What do you think about also exposing this on `config.js`?

bryphe: 👍

Issue #977, opened 2017-11-18
Tab dirty state doesn't show in tab


hoschi:
![2017-11-18-110922_2058x1439_scrot](https://user-images.githubusercontent.com/163128/32979204-ffba2402-cc50-11e7-8383-87fff77c1372.png)

Window title shows it, git gutter plugin shows it also, but tab doesn't have the circle icon. I'm on release 0.2.17

config:
```js
module.exports = {
  "statusbar.fontSize": "12px",
  "editor.fontSize": "14px",
  //"editor.fontLigatures": true,
  "oni.useDefaultConfig": false,
  "editor.scrollBar.visible": false,
  "editor.quickInfo.enabled": false,
  "editor.cursorLine": false,
  "editor.cursorColumn": false,
  "editor.clipboart.enabled": false,
  "editor.quickInfo.delay": 250,
  //"editor.cursorColumnOpacity": 0.1,
  "tabs.showVimTabs": true,
  "tabs.wrap": true,
}
```

CrossR:
Worked fine for me at first, then when I enabled `"tabs.showVimTabs": true,`, it stopped working.
Maybe the  logic is buffer specific and not considering tabs?

bryphe:
Yes,  I could be wrong, but it seems t's not very straightforward to get the dirty state for _all buffers in a tab_, so in `vim.showVimTabs` mode, we simply don't show the dirty state.

It would be a nice feature to have, though

hoschi:
Hm I don't think I can live without that. How do I get the vim tabs back? `tabs.enabled=false` gives me no tabs at all ...

bryphe:
Yes, it seems even if `tabs.enabled` is set to `false`, we always set `ext_tabline` to `true` (which tells Neovim to let us handle the rendering of tabs, and not to worry about it).

https://github.com/onivim/oni/blob/af12e7e847b02f6b0f702cfb055f016b2ef9ec9f/browser/src/neovim/NeovimInstance.ts#L643

To get the original tab line back, we should also add a configuration option to disable this. It seems like there are three scenarios we have:
- Oni's default tabs (buffers, `tabs.showVimTabs` is `false`)
- Oni showing vim tabs (tabs, `tabs.showVimTabs` is `true`).
- Oni not showing tabs at all, and deferring that to neovim

[...]

Issue #973, opened 2017-11-18
Examine / improve Oni's default VimL settings


bryphe:
Both @badosu & @hkupty gave good feedback on our VimL settings [here](https://github.com/onivim/oni/pull/915#issuecomment-345285924) and [here](https://github.com/onivim/oni/issues/929#issuecomment-345294294)

Some common feedback:
- Settings in VimL are causing pain - the defaults (especially the ones in `oni.useDefaultPlugins` cause issues). The ones in particular here are the tabstop settings + the tnoremap settings.

The main goals I have for Oni are that it _works out of the box_, with zero configuration - so when someone downloads it and runs it, they can get the complete experience. For example, trying to save people from the pain I went through trying to set up YCM on Windows 😃 

In addition, it should be easy to bring in your `init.vim` settings - that is, after all, a big selling point of a Neovim-based UI - without having to dig too deep into Oni settings.

One way to simplify this is to remove the (opinionated) defaults we have, and set the ones that are truly needed as default settings. This would mean pruning the `init.vim` that is used when `oni.useDefaultPlugins` is `true`: https://github.com/onivim/oni/blob/master/vim/default/bundle/oni-vim-defaults/plugin/init.vim

Settings that we __can definitely remove__ (these belong in my `init.vim`, not in Oni...):
```
" Helpers for command mode
" %% for current buffer file name
" :: for current buffer file path
cnoremap %% <C-R>=fnameescape(expand('%'))<CR>
cnoremap :: <C-R>=fnameescape(expand('%:p:h'))<CR>/

" Make Control+nav keys functionality in insert mode
inoremap <expr> <C-a> pumvisible() ? "<Esc>A" : "<C-o>A"
inoremap <expr> <C-b> pumvisible() ? "<Esc>bi" : "<C-o>b"
inoremap <expr> <C-l> pumvisible() ? "<Esc>la" : "<C-o>a"
```

Settings that can be __replaced with an Oni setting__ (configured via `config.js`):

##### Proposed: Add a setting for `terminal.escToExit`
```
tnoremap <Esc> <C-\><C-n>
```

##### Proposed: Add a `vim.colorscheme` setting (#950)

```
colorscheme onedark
```

##### Proposed: Create Oni settings for tabstops (inspired by VSCode's settings: https://stackoverflow.com/questions/29972396/how-to-set-tab-space-style)
- `editor.detectIndentation`
- `editor.insertSpaces`
- `editor.tabSize`
```
set tabstop=4
[...]

PR #992, opened 2017-11-20
Add native tab option

CrossR:
Adds a native tab option, to render vim tabs instead of Oni tabs, as well as fixing the `tabs.enabled` option. 
A partial fix for #977. Next Oni needs to be able to check the modification status of all buffers in a tab and show a dirty state if any are modified.

Whilst this works, if the user changes to "native" they will lose their tabs, and need to reload Oni to see the native tabs, which is because we pass the command when we start communicating with neovim. This means the user must restart for this to take place, and they are left in an awkward state until then. I'm not sure what the best way to deal with that is though...

### DIFF
--- a/browser/src/UI/components/Tabs.tsx
+++ b/browser/src/UI/components/Tabs.tsx
@@ -186,7 +186,7 @@ const mapStateToProps = (state: State.IState, ownProps: ITabContainerProps): ITa
     const shouldUseVimTabs = state.configuration["tabs.showVimTabs"]
     const tabs = shouldUseVimTabs ? getTabsFromVimTabs(state) : getTabsFromBuffers(state)
 
-    const visible = state.configuration["tabs.enabled"]
+    const visible = state.configuration["tabs.enabled"] && shouldUseVimTabs !== "native"
 
     const height = state.configuration["tabs.height"]
     const maxWidth = state.configuration["tabs.maxWidth"]
--- a/browser/src/UI/components/Tabs.tsx
+++ b/browser/src/UI/components/Tabs.tsx
@@ -184,12 +184,11 @@ const getTabsFromVimTabs = createSelector(
 const mapStateToProps = (state: State.IState, ownProps: ITabContainerProps): ITabsProps => {
 
     const oniTabMode = state.configuration["tabs.mode"]
-    const oniTabsEnabled = state.configuration["tabs.enabled"]
-
     const shouldUseVimTabs = oniTabMode === "tabs"
+
     const tabs = shouldUseVimTabs ? getTabsFromVimTabs(state) : getTabsFromBuffers(state)
 
-    const visible = oniTabsEnabled && oniTabMode !== "native"
+    const visible = oniTabMode !== "native" && oniTabMode !== "hidden"
 
     const height = state.configuration["tabs.height"]
     const maxWidth = state.configuration["tabs.maxWidth"]
--- a/browser/src/UI/components/Tabs.tsx
+++ b/browser/src/UI/components/Tabs.tsx
@@ -183,10 +183,13 @@ const getTabsFromVimTabs = createSelector(
 
 const mapStateToProps = (state: State.IState, ownProps: ITabContainerProps): ITabsProps => {
 
-    const shouldUseVimTabs = state.configuration["tabs.showVimTabs"]
+    const oniTabMode = state.configuration["tabs.mode"]
+    const oniTabsEnabled = state.configuration["tabs.enabled"]
+
+    const shouldUseVimTabs = oniTabMode === "tabs"
     const tabs = shouldUseVimTabs ? getTabsFromVimTabs(state) : getTabsFromBuffers(state)
 
-    const visible = state.configuration["tabs.enabled"] && shouldUseVimTabs !== "native"
+    const visible = oniTabsEnabled && oniTabMode !== "native"
 
     const height = state.configuration["tabs.height"]
     const maxWidth = state.configuration["tabs.maxWidth"]
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -130,7 +130,7 @@ const BaseConfiguration: IConfigurationValues = {
     "statusbar.fontSize": "0.9em",
 
     "tabs.enabled": true,
-    "tabs.showVimTabs": false,
+    "tabs.mode": "buffers",
     "tabs.height": "2.5em",
     "tabs.maxWidth": "30em",
     "tabs.wrap": false,
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -129,7 +129,6 @@ const BaseConfiguration: IConfigurationValues = {
     "statusbar.enabled": true,
     "statusbar.fontSize": "0.9em",
 
-    "tabs.enabled": true,
     "tabs.mode": "buffers",
     "tabs.height": "2.5em",
     "tabs.maxWidth": "30em",
--- a/browser/src/Services/Configuration/IConfigurationValues.ts
+++ b/browser/src/Services/Configuration/IConfigurationValues.ts
@@ -151,7 +151,6 @@ export interface IConfigurationValues {
     "statusbar.enabled": boolean
     "statusbar.fontSize": string
 
-    "tabs.enabled": boolean
     "tabs.mode": string
 
     // Height of individual tabs in the tab strip
--- a/browser/src/Services/Configuration/IConfigurationValues.ts
+++ b/browser/src/Services/Configuration/IConfigurationValues.ts
@@ -152,7 +152,7 @@ export interface IConfigurationValues {
     "statusbar.fontSize": string
 
     "tabs.enabled": boolean
-    "tabs.showVimTabs": boolean
+    "tabs.showVimTabs": boolean | string
 
     // Height of individual tabs in the tab strip
     "tabs.height": string
--- a/browser/src/Services/Configuration/IConfigurationValues.ts
+++ b/browser/src/Services/Configuration/IConfigurationValues.ts
@@ -152,7 +152,7 @@ export interface IConfigurationValues {
     "statusbar.fontSize": string
 
     "tabs.enabled": boolean
-    "tabs.showVimTabs": boolean | string
+    "tabs.mode": string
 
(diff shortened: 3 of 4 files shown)


browser/src/UI/components/Tabs.tsx
```diff
@@ -186,7 +186,7 @@ const mapStateToProps = (state: State.IState, ownProps: ITabContainerProps): ITa
     const shouldUseVimTabs = state.configuration["tabs.showVimTabs"]
     const tabs = shouldUseVimTabs ? getTabsFromVimTabs(state) : getTabsFromBuffers(state)
 
-    const visible = state.configuration["tabs.enabled"]
+    const visible = state.configuration["tabs.enabled"] && shouldUseVimTabs !== "native"
```
badosu: I feel this is making configuration very complex with settings that depend on context of other settings. It's good for retrocompatibility but bad in the long run.

What about: `tabs.showVimTabs = (buffers)|native|files`

CrossR: Yep, that seems fair. Perhaps a `tabs.tabType` option instead for buffers/tabs/native vim tabs?

badosu: Well, `tabs.enabled` seems like a different configuration altogether, not sure it could be replaces.

Thoughts @bryphe ?

bryphe: Thinking about this some more since the comment on #992 , perhaps we can streamline it to not require a new setting.

Here's a table of the scenarios we're trying to support:

| Scenario | Description | `tabs.enabled` | `tabs.showVimTabs` |
| --- | --- | --- | --- |
| Default tab experience | Show buffers in Oni-style UX | `true` (default) | `false` (default) |
| Disabled | No tabs shown at all | `false`  | `false` (default) |
| Vim tabs in Oni | Show the vim tabs, in Oni-style UX | `true` (default) | `true` |
| Vim tabs, native UX | Show the vim tabs, w/o Oni UX | `false` | `true` |

The 'happy-path' IMO, is using the default experience, or turning it off - that means the 'happy-path' is completely controllable via `tabs.enabled`. I want to make sure those 'happy-path' options are very simple and streamlined to configure.

However, the more 'advanced path' (using Vim tabs with Oni's UX, or falling back to the native Vim UX entirely), also requires toggling  `tabs.showVimTabs`, with the most advanced configuration requiring changing both `tabs.enabled` to `true` and `tabs.showVimTabs` to `false` (enabling the native vim UX). I think that is pretty reasonable, as long as we document it. 

That would simplify the configuration setting and still support the ability to 'eject' and use the default VIM tab UX - does that seem reasonable?

bryphe:
Thanks for putting this together so quickly, @CrossR! 😄

badosu:
I don't understand why a fresh installation behaving differently from vim
is a happy path.

Care to share your thoughts?

Em 20/11/2017 19:12, "Bryan Phelps" <notifications@github.com> escreveu:

bryphe:
> I don't understand why a fresh installation behaving differently from vim
is a happy path.

Couple of thoughts:

A fresh installation of Oni _already behaves differently than vim today_... if you have a custom `init.vim`. You have to _opt-in_ to using your `init.vim` settings.

A goal I have with Oni is to _lower the bar_ to modal editing. Out of the box, I want the tabline to behave like Atom, VSCode, or Sublime - because for better or worse, those are the standards that a large population of users expect. Like the `oni.loadInitVim` setting, I expect that if a Vim user comes to Oni and cares enough about a tab-based workflow to enable it, they'll have enough knowledge to opt-out of Oni's functionality and bring in their own (as long as we give them the ability + settings to do this).

I think this touches a bit on your comments in #973 too - my goal for Oni is beyond just a GUI for Vim, as there are some functionalities I'd like to implement that have no parallel in Vim (like an embedded browser). Some features, like those that go beyond what Vim provides, will work best with the 'blessed' or 'happy path' settings (for example, it'd be hard to have a consistent navigation model with these new types of UIs, unless Oni provides that model).

Balancing all these aspects - having a solid out-of-box experience, lowering the bar to modal editing for new users, and catering to advanced Vim users - is certainly a challenge, and I know we don't have it right today, but we'll continue exploring until we find that balance. 

For advanced Vim users, I want to make sure they have the right knobs and settings to opt-out of Oni behavior they don't like or want, and bring in their configuration. LMK what you think.

browser/src/UI/components/Tabs.tsx
badosu: I think the idea is fine, the only problem is that to enable a specific configuration I want I'll have to check a 'truth table'.

I think it would be much more straightforward to enable or disable tabs and then (if they are enabled) choose explicitly the behaviour as a string (buffers|tabs|native).

This is just my aversion to configurations that depend on each other.

badosu:
I mean, I don't understand why buffers is easier to understand | work with for begginers, I don't understand how to work with it 😂 . If you tell me this is the default on other IDEs and more usable for beginners then great!

browser/src/UI/components/Tabs.tsx
bryphe: > the only problem is that to enable a specific configuration I want I'll have to check a 'truth table'.

Fair criticism 😄 

Perhaps this is clearer:
`tabs.enabled`: behaves as-is today
`tabs.mode`: `'buffers'|'tabs'|'native'`

It's basically your suggestion above - just renaming `tabs.showVimTabs` to `tabs.mode`. I realized the only real issue I had with the specifying the behavior directly was being confused by the case `tabs.showVimTabs`: `'buffers'` - but if we rename the setting, it alleviates my concern. Only downside is changing the setting would be a breaking change.

badosu: > Only downside is changing the setting would be a breaking change.

That's the advantage of being a 0.2.x version :-). We need to make it right now not to have to deal with this in the future

CrossR:
Shout up if the logic isn't right, there were a few different ideas in here so I could have got confused!

I do agree with the general statement that buffers are more confusing to understand for a beginner in general, but because Oni surfaces the vim buffers in individual tabs I think that makes it the same as any other editor.

Also, it stops some of the confusing stuff that you can do in vim. You can run a plugin or execute a command that you found online and it swaps your buffer, or you'll open a new file with `vim file1 file2 file3`. As a new user? That is pretty confusing, you see a single file and its awkward.

Since Oni defaults to showing all buffers in their own tabs, that confusing scenario goes away, you now have 3 tabs for those 3 files, which is what you expect. If Oni defaulted to showing vim tabs, then those scenarios could potentially still come up. (That should be true I think but when I just tried it in Oni I only got 1 tab till I cycled through, so maybe we aren't loading them correctly yet if they are passed on launch?)

browser/src/UI/components/Tabs.tsx
```diff
@@ -183,10 +183,13 @@ const getTabsFromVimTabs = createSelector(
...
 
-    const shouldUseVimTabs = state.configuration["tabs.showVimTabs"]
+    const oniTabMode = state.configuration["tabs.mode"]
+    const oniTabsEnabled = state.configuration["tabs.enabled"]
+
+    const shouldUseVimTabs = oniTabMode === "tabs"
     const tabs = shouldUseVimTabs ? getTabsFromVimTabs(state) : getTabsFromBuffers(state)
 
-    const visible = state.configuration["tabs.enabled"]
+    const visible = oniTabsEnabled && oniTabMode !== "native"
```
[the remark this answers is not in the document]
bryphe: I think that there is one case we'd still need it - if the user wants to disable the tab line completely. I want to make sure there is an easy way to 'turn off' the tab line. It looks like our current set of options support buffers, tabs, and 'native' vim rendering - but not a clear way to turn off the tabline.

Turning off / hiding the tabline is different than setting `tabs.mode` to `'native'`, because in that case, we turn off `ext_tabline`, and hand over the responsibility of rendering the tabline back to Vim - so it doesn't actually 'turn it off', just lets Vim render it. So what we're missing currently is that 'turn off' or 'hide' story.

I can see two ways to preserve this 'hide' functionality:
-  __Keep `tabs.enabled`__- it'd basically be the same as your changes today, except we'd want the `ext_tabline` option to be controlled only by the `native` setting - if `tabs.enabled` is `false`, we'd want to keep it 'externalized' (so vim doesn't try and render it, and it'd be hidden). 
- __Remove `tabs.enabled`, add a `'hidden'` or `false` value to `tabs.mode`__ - when the `hidden` value is set, we'd keep `ext_tabline` set to `true`, but not render the Oni tab UI.

I'm okay with either approach, whichever one is more intuitive - the second one means we can completely  remove the `tabs.enabled` setting, and address confusion about how `tabs.enabled` and `tabs.mode` set to `native` relate, so it seems like that might be better.

Basically I want to make sure if someone asks "How do I turn off the tabline?" we have a clear story for doing so 😄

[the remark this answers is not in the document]
CrossR: Ahhh okay,  I was forgetting the use case of not wanting a tabline of any fashion, I was only considering Oni Buffer/Oni Tab/Normal Vim Tabline.

I'll push that up in a mo.

browser/src/UI/components/Tabs.tsx
```diff
@@ -183,10 +183,12 @@ const getTabsFromVimTabs = createSelector(
...
 const mapStateToProps = (state: State.IState, ownProps: ITabContainerProps): ITabsProps => {
 
-    const shouldUseVimTabs = state.configuration["tabs.showVimTabs"]
+    const oniTabMode = state.configuration["tabs.mode"]
+    const shouldUseVimTabs = oniTabMode === "tabs"
+
     const tabs = shouldUseVimTabs ? getTabsFromVimTabs(state) : getTabsFromBuffers(state)
 
-    const visible = state.configuration["tabs.enabled"]
+    const visible = oniTabMode !== "native" && oniTabMode !== "hidden"
```
badosu: Very nice, thanks!!

bryphe:
Changes look great to me, thanks for all the thinking / iterations around this @CrossR and @badosu !

I'll bring this in when the builds have run, and I'll mention the configuration change in the next release notes

justinmk:
> because we pass the command when we start communicating with neovim.

It should be possible to change UI options at runtime via `nvim_ui_set_option()`, at least since https://github.com/neovim/neovim/commit/6e5cb0debd23693175bd05409d3f1af4015567df (unreleased)

bryphe:
Cool, thanks for the heads up @justinmk ! I split it out into #1953 . Would be neat to be able to update those properties w/o needing to reload

PR #993, opened 2017-11-20
Feature/resize buffer bar scroll

Akin909:
* created `tabs.webkit-scrollbar` class
* reduced scrollbar height to 3px
* add overflow overlay to tab bar on hover so scroll bar only appears on hover

I'm not sure whether you'd prefer for the scroll bar to always be there but smaller and overlaid or only appear on scroll, at present it only appears on scroll, but can tweak this if you'd prefer the alternative

I removed the overflow styling that was coming from the `tabs.tsx` component as it was overwriting what was set in the `less` so I've returned the wrap styles or an empty object as I didn't want to pass null or a boolean to the style prop for fear of some unexpected result.


bryphe:
Cool, looks good @Akin909 !

> I removed the overflow styling that was coming from the tabs.tsx component as it was overwriting what was set in the less so I've returned the wrap styles or an empty object as I didn't want to pass null or a boolean to the style prop for fear of some unexpected result.

Makes sense, I tried it out and seems like it works well. I was worried we might still get a scrollbar showing when we wrap, but doesn't seem to be the case.

bryphe:
Seems like there is some flakiness with the autocompletion test (the travis CI failure), it does not appear to be related to your change. Need to investigate what's happening there. I'll bring this in, though - thanks for the contribution!

Issue #956, opened 2017-11-16
Autocomplete oddities


badosu:
Suppose you have a word like `const` (on a js file), I'll denote the cursor with `_`.

In normal mode you are at `co_nst`, as soon you change it to insertion mode the menu pops up which is not something I would expect, so I type `ESC` and `ESC` to come back to normal mode. As soon as I try to get back into insertion mode the menu pops out of nowhere again, even outside of that word.

So maybe there are 2 bugs here, popping autocompletion without having inserted any character in the middle of a word, and also still thinking you are at the menu even after leaving it when coming back from normal mode.

bryphe:
Definitely both bugs - thanks for isolating a repro!

I'm able to reproduce both those issues

PR #995, opened 2017-11-21
Fix #990 - Two <ESC> key presses required to go to normal mode w/ completion menu up

### LINKED ISSUE #990: Completion: Two <ESC> presses needed to return to normal mode when completions are open
The completion menu was refactored to use a common menu code - but one impact this had is that it now takes two `<ESC>` key presses to return to normal mode when the menu is open.

This is a bit annoying and jarring, as the menu should be relatively seamless - it should not get in your way and impact your editing keystrokes (aside from providing `<enter>` or `<tab>` to quickly complete).

I believe both @hoschi and @badosu hit this

bryphe:
__Issue:__ When the completion menu is open, it takes two `<esc>` key presses to dismiss it and go to normal mode. This is problematic because you need to know to use a different set of key presses to get back to normal mode, depending on whether the menu is up - it can be especially disorienting if you press `<esc>` just as it is opening.

__Fix:__ In insert mode, we don't need an explicit mapping - switching mode will close the completion menu.

> Note: There are still some separate issues tracked by #956 (#972)

### DIFF
--- a/browser/src/Input/KeyBindings.ts
+++ b/browser/src/Input/KeyBindings.ts
@@ -14,6 +14,7 @@ export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configurati
 
     const isVisualMode = () => editors.activeEditor.mode === "visual"
     const isNormalMode = () => editors.activeEditor.mode === "normal"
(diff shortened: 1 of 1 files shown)


bryphe:
Fixes #990

PR #1005, opened 2017-11-23
Fix #980 - Use `undojoin` + `nvim_call_atomic` for atomic undo

### LINKED ISSUE #980: Rename action is not undoable with one "undo" action
When you hit `F2` to rename a string and you can't undo it by hit "undo" one time. Instead you must hit undo as much as renames occurred.

bryphe:
WIP for #980 to have rename (and other workspace edits, like formatting) undoable with a single gesture.


bryphe:
This solves the general case (at least, for a single file), but there is a problem if you `undo` and then start a new rename - we get an error that `undojoin can't be called after an undo`. We might need to find a way to clear the latest undo state, in this case?

bryphe:
Based on this thread: https://www.mail-archive.com/search?l=vim_dev@googlegroups.com&q=subject:%22undojoin+after+an+undo%22&o=newest&f=1

Seems like maybe `silent! undojoin` is a reasonable solution. It would ignore the error case, but still function correctly for the later cases.

bryphe:
Addresses #980 (at least for a single buffer)

PR #1011, opened 2017-11-23
Performance: Batch redux notifications

bryphe:
We're particularly noisy (abusive) of our redux store - in particular, a single user gesture can dispatch multiple actions (some coming from auto command notifications, key presses, etc). _Each action_ causes the redux store to dispatch an update, which then triggers a re-render.

In order to improve performance, we should batch these up - ideally under a `requestAnimationFrame`. This will give us more predictable performance.

For items that truly require an instant rendering - like the typing predictions - we should still be able to push those instantly. But those should be less frequent.

### DIFF
--- a/PLAN.md
+++ b/PLAN.md
@@ -1,0 +1,2 @@
+- How does this work with typing predictions? Seems like there may be artifacts that weren't present before
+- Any regressions?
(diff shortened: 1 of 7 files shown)


bryphe:
This introduces artifacts with predictions (seems like there may be an issue in timing with clearing predictions) - need to investigate before bringing in.

PR #1013, opened 2017-11-23
Bugfix/mouse selection buffer bar

### LINKED ISSUE #1010: Make buffer tab text not selectable by the OS:
screen shot 2017-11-23 at 3 01 37 am

By dragging the mouse just right I was able to select the text in the buffer tabs.

I think the intention was that this would be disabled.

Akin909:
* Added `user-select: none` to tabs.less to stop users highlighting buffer bar text.
* Added keys to WindowSplits - :worried: I couldn't find any id like data to use as keys hopefully the index here will be fine if not ill take them out, was just desperate to make the error go away
* Fixes #1010 (hopefully)

### DIFF
--- a/browser/src/UI/components/Tabs.less
+++ b/browser/src/UI/components/Tabs.less
@@ -54,6 +54,7 @@
 
     overflow: hidden;
     text-overflow: ellipsis;
+    user-select: none;
 
     .name {
         flex: 1 1 auto;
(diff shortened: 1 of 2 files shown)


bryphe:
Awesome, thanks for fixing this @Akin909 !

PR looks good, I'm re-running the OSX build but I don't believe the failure is related to the change.

> Added keys to WindowSplits - 😟 I couldn't find any id like data to use as keys hopefully the index here will be fine if not ill take them out, was just desperate to make the error go away

Cool, I think is reasonable for now, and it will definitely reduce noise - at the moment we only have one split. We'll need to add a proper ID later (once that is actually leveraged).

bryphe:
I tested locally and ran it, and it passed - I believe this may be failing due to the typescript version being upgraded (it's using 2.6.1 on the build machine), but the change doesn't have the #1003 fixes in it. I'll bring it and check the master build.

Thanks @Akin909 !

PR #1017, opened 2017-11-25
Fix scroll bar focus/unfocus positioning bug

Akin909:
@bryphe, whilst testing things out I noticed that following my PR for the tab bar scroll bar, If you scroll into the overflow area and then move the cursor it would lose the position of the barwhich was quite "jerky" from a UX standpoint, I managed to resolve this by specifying overflow-x: hidden on the tab bar element, which was an oversight on my part in the original PR, apologies for missing that earlier.
buffer-bar

The PR should fix this (i've tested this locally), with this change the bar now keeps the correct position but on hover the scrollbar appears and you can move back and forth without the position changing

### DIFF
--- a/browser/src/UI/components/Tabs.less
+++ b/browser/src/UI/components/Tabs.less
@@ -29,6 +29,7 @@
     width: 100%;
     color: #c8c8c8;
     background-color: rgba(0, 0, 0, 0.2);
+    overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for catching this and for the quick fix, @Akin909 ! 👍 Looks good to me.

PR #1021, opened 2017-11-26
Tidy up the remaining parts that were pointing at the old repo.

CrossR:
Just a tiny change to fix up the naming of the repo throughout, just noticed it was still redirecting when I clicked the link in the status bar.


bryphe:
Awesome, thanks @CrossR!

PR #1022, opened 2017-11-26
[WIP] Implement External Pop Up Settings.

CrossR:
Started to implement External Pop Up Settings.

I'm not 100% on what the functionality should be for hidden, or if I've done it the best way.

Currently, "oni" and "native" use the Oni and Native Vim options as expected.
Hidden doesn't show any pop up at all, but using ctrl+p or similar will still cycle through all options as normal, is that the desired effect?

I've implemented it by adding the check when it comes to render, but would it be better to just move that code into the subscription to that event? Though I guess then I would potentially need to duplicate the code, so maybe not.

Finally, I need to have a check around and see what other places could popup their own windows, but I think they all use the `NeovimPopupMenu.tsx` file.

Oh, and need to fix the lint errors in this still.


bryphe:
> I'm not 100% on what the functionality should be for hidden, or if I've done it the best way.

Ah ya, this is the most confusing case for sure. I'm thinking for 'hidden', we should not show the language service completion, but still keep the popup menu externalized.

So our settings would look like this:
- `oni` - enable both the language-server completion, as well as externalized popup menu
- `native` - disable the language-server completion, and disable externalized popup menu (fall back to native)
- `hidden` - disable the language server completion, but keep the externalized popup menu.

I think the `'hidden'` setting is reasonable behavior.. because by default Vim-completion is opt-in, unless you've done some more advanced configuration to set up an omnicompletion provider. I mainly want an easy way to turn off the type-as-you-go completion behavior we have by default in Oni for JS/TS/CSS, because I know it is annoying and distracting to some - and I think the `'hidden'` setting accomplishes that.

> I've implemented it by adding the check when it comes to render, but would it be better to just move that code into the subscription to that event? 

If we implement it as above, we actually wouldn't need to gate the render. We'd only disable externalization in the `'native'` setting, and in that case, we wouldn't get the popup-menu events at all. 

However, we would want to turn off the language-server completion - we could replace the check here:
https://github.com/onivim/oni/blob/fbea2e38ad804e797172e0a27df2e1816290c67f/browser/src/Services/Language/Completion/Completion.ts#L165

With a check for `configuration.getValue("editor.completions.mode")` against `"oni"`.

Thanks for putting together this proposal! Lots of ambiguity around these settings... helps a lot to have an implementation to discuss.

CrossR:
I think I've implemented what was discussed, but due to editing the `CompletetionProvider.ts` check to be against `configuration.getValue("editor.completions.mode")` and `"oni"`, it completely opts the user out of language server completions.

Is that what we want from `hidden`? For native that seems fair enough, I don't know if we can inject to there, but for `hidden` they may just want the auto complete to stop kicking in, but still use it when they explicitly use the keybind. Or maybe that is more an argument for splitting the config more to have one that swaps it all to vim only completion in Oni UI vs Non-auto LSP Config.

Otherwise, this should do what we want.

bryphe:
> I think I've implemented what was discussed, but due to editing the CompletetionProvider.ts check to be against configuration.getValue("editor.completions.mode") and "oni", it completely opts the user out of language server completions.

Cool, I tried out the PR and the logic looks good!

> Is that what we want from hidden?

It seems reasonable to me. We could add a keybind to explicitly expand the completion menu, which would be available in `'hidden'` mode. Sort of like the change in #1044 , where I added an explicit for opening the quickinfo/hover tool tip, even if it is disabled. We can track that separately.

> Or maybe that is more an argument for splitting the config more to have one that swaps it all to vim only completion in Oni UI vs Non-auto LSP Config.

Yes, we could potentially refactor or adjust the configuration setting later, but at least this handles the cases that are most important to me:
- `'oni'` - default completion - uses LSP completion + externalized UI
- `'hidden`' - turns off LSP completion by default, but still allows Vim completion via externalized UI
- `'native'` - allows the user to completely eject from Oni's completion UX and fall-back to the Vim experience.

And my hope is that the `'oni'` completion will be good enough that it becomes less common to change this 😄 

Thanks for all the work and thinking on this!

bryphe:
Looks like there is a lint issue blocking right now - I'll bring it in and take care of that. 

I'm planning on pushing out a release this afternoon, and this would be a great fix to include (and it will also save you from having to merge with #1049 , since I made some more changes to the completion code..) 

Thanks again for your work on this, @CrossR !

PR #1023, opened 2017-11-27
Fix #1015 - Add ability to capture logs and output for test result

### LINKED ISSUE #1015: Build / Automation: Improve logging output for integration tests
When an integration test fails, there is not very much helpful info. 

At the minimum, we should be able to get the console logs from Oni, as well as enable debug logging at the beginning of the tests - the test failures need to be diagnosable and actionable to be useful.

bryphe:
This adds a hook for automation to grab the console logs, and then output it as part of the test result.

### DIFF
--- a/test/CiTests.ts
+++ b/test/CiTests.ts
@@ -11,9 +11,9 @@ const LongTimeout = 5000
 
 const CiTests = [
     "AutoCompletionTest",
-    // "BasicEditingTest",
-    // "QuickOpenTest",
-    // "NoInstalledNeovim",
+    "BasicEditingTest",
+    "QuickOpenTest",
+    "NoInstalledNeovim",
 ]
 
 // tslint:disable:no-console
--- a/test/CiTests.ts
+++ b/test/CiTests.ts
@@ -11,9 +11,9 @@ const LongTimeout = 5000
 
 const CiTests = [
     "AutoCompletionTest",
-    "BasicEditingTest",
-    "QuickOpenTest",
-    "NoInstalledNeovim",
+    // "BasicEditingTest",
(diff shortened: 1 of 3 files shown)


test/CiTests.ts
```diff
@@ -116,8 +116,11 @@ describe("ci tests", function() { // tslint:disable-line only-arrow-functions
             })
 
             it("ci test: " + test, async () => {
+                console.log("Waiting for editor element...")
```
badosu: If I use `console.log` I can see the output when debugging oni now?

I was trying to fix something this weekend and this issue arised.

bryphe: This mainly impacts the integration tests - we'll be able to see the output of the `console.log` while the test was executing.

For debugging Oni, you should be able to use `console.log` today and see it visible in the output:
oni-screenshot-1511810392185

To confirm I just added a quick input binding with: 
```
Oni.input.bind("<enter>", () => console.log("hi from the console"))
```
And then pressed `<enter>` in the editor and saw the output.

If you're not seeing it, one thing to check is the settings up here:
image

Sometimes I forget to clear the 'filter' or have it set only on warn/error, in which case you wouldn't see the logs. LMK if you're still not seeing them; would make debugging very painful..

badosu: Thanks, gonna check it!

bryphe:
Seems like there might've been a race condition or timing issue here... to initiate the test, we wait for the `.editor` element to appear, and then we immediately query for its contents, and verify we see something there. There could potentially be a delta depending on the time it takes to load Neovim, so this condition may be too sensitive.

PR #1030, opened 2017-11-28
Bugfix/quickopen sizing

### LINKED ISSUE #1020: Tooltip continuously resizing when oni less than fullscreen
When less than fullscreen and hovering over a word with a long type signature or a lot of information the tooltip generated if oni is at less than full screen, especially if at half width continuously resizes creating a 'shuddering' effect, making the text in the tooltip hard to read.

tooltip-bug

I've had a look and I think this bug is coming from `CursorPositioner.tsx` specifically [here](https://github.com/onivim/oni/blob/6c5f054a5f028c88aa35ab275ff8110573948a88/browser/src/UI/components/CursorPositioner.tsx#L151-L156),

details-tooltip

The `this.state.isFullWidth` value is constantly flipping from true to false because the clientRect() is changing from in my specific case from 1000 to 300, as a test, on setting the left and right position to 0 the flickering stops.

A solution would be to set position right to a value and not have it vary? this seems to stop the flickering whilst allowing for space on full screen and split screen although it leaves the tooltip I'm not sure if this would impact something else or not be what you had in mind there.

EDIT: A static px value unfortunately sticks the tooltip in the right regardless of cursor position :( and a percentage value whilst working a bit better seems like a slightly messy solution

Akin909:
@bryphe, I managed to get the flickering to stop at smaller widths, granted my solution is not the best. 

I added maxWidth to the styles to prevent the fullWidth prop alternating, however I ran into an issue where the left positioning very rarely was being set to a negative value so I check its value and return it as a positive int. 

Very open to any changes you might have, I did test it out locally a fair bit and it seems to work without any strange edge cases nor do I think the general functionality has changed but not sure if there's something somewhere I might have missed. 

Also Added overflow scroll on the element so the user can scroll to end of that element as they can for one beneath it although I didn't do much with the scrollbar styling except make it smaller to keep it out of the way. This bit is likely not necessary and/or desirable so happy to remove it.

possibly fixes #1020


browser/src/UI/components/CursorPositioner.tsx
```diff
@@ -148,10 +148,15 @@ export class CursorPositionerView extends React.PureComponent<ICursorPositionerV
             visibility: this.props.hideArrow ? "hidden" : "visible",
         }
 
+        const leftSidePositioning = (x: number) =>  Math.sign(x) !== -1
```
bryphe: Looks like this could be simplified with `Math.abs`

browser/src/UI/components/CursorPositioner.tsx
```diff
@@ -148,10 +148,15 @@ export class CursorPositionerView extends React.PureComponent<ICursorPositionerV
...
+            ? x.toString() + "px"
+            : (x * -1).toString() + "px"
+
         const childStyleWithAdjustments: React.CSSProperties = this.state.isMeasured ? {
             ...childStyle,
-            left: this.state.isFullWidth ? "8px" : adjustedX.toString() + "px",
+            left: this.state.isFullWidth ? "8px" : leftSidePositioning(adjustedX),
             right: this.state.isFullWidth ? "8px" : null,
+            maxWidth: "95%",
```
bryphe: Cool, seems like this fixes the issue!

bryphe:
Change looks great @Akin909 , thank you for taking a look at this! That sizing logic is not trivial.

I left some minor feedback on the PR (we might able to use `Math.abs` to simplify the sign-change for the x-position), but overall looks good to me.

browser/src/UI/components/CursorPositioner.tsx
Akin909: Thanks for the tip @bryphe, tweaked it to use `Math.abs`, much simpler!

bryphe:
Awesome, change looks good!

The `AutoCompletion-CSS` test is a new test that isn't stable - not related to your change. I'm working on a fix in PR #1032 . I'll bring this in now - thanks for the contribution, @Akin909 !

PR #1033, opened 2017-11-29
Fix #1012 - Add 'ui.fontSmoothing' configuration

### LINKED ISSUE #1012: Add configuration for font smoothing (antialiasing)
From @benjamin-rood 's feedback on gitter - would be helpful to have a configuration setting for setting / changing: `webkit-font-smoothing`  on the body.

Perhaps `ui.fontSmoothing` with options of `none`, `antialiased` and `subpixel-antialiased`, and then just directly set the `webkit-font-smoothing` property from there.

bryphe:
Adds a `ui.fontSmoothing` configuration value, which directly sets the `-webkit-font-smoothing` property on the body.

Valid values are:
- `auto` (default)
- `none`
- `antialiased`
- `subpixel-antialiased`

More documentation about the values is available here: https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth


bryphe:
Fixes #1012

bryphe:
Also added info in the wiki: https://github.com/onivim/oni/wiki/Configuration

PR #1034, opened 2017-11-29
Fix #63: Synchronize titlebar color on OSX

### LINKED ISSUE #63: Synchronize title bar color with tabbar background color
This is probably low priority because there's a lot more important things that need to be done to make the UI smooth, but I'm just making this task because someone might already know how to do this, and might thing it's fun.

screen shot 2016-12-08 at 2 36 06 pm

I think a good default would be to synchronize the color of the title with the vim colorscheme's tabbar background color. Then of course, as we've discussed elsewhere, colorschemes/themes could specify special support for coloring the title bar.

bryphe:
Now that theming colors are integrated (#975 , addressing #950), it makes sense to synchronize the titlebar color and also have it customizable (at least on OSX). 

This change renders the browser window as 'frameless' on OSX, which means it lets us decide how to render the title bar (although the stoplight buttons are still rendered for us). 

This pushes the window title state to our UI/shell redux store, and hooks up a UI component to render it.


bryphe:
Fixes #63

Issue #1007, opened 2017-11-23
Help Explaining Scrollbars 


jordwalke:
Hello,

I've never quite understood the information that the scrollbars are attempting to convey in Oni. Could you help me understand?

**One issue**:
When inserting many newlines into a document, the scrollbar small light gray region grows smaller.

<img width="808" alt="screen shot 2017-11-23 at 12 54 47 am" src="https://user-images.githubusercontent.com/977348/33164792-3cfdea1a-cfe9-11e7-9ccf-9cec57826212.png">

That seems to make sense, but then when I hit escape and move the cursor a line up, it looks like this: (All of a sudden the light gray rectangle became small again).

<img width="814" alt="screen shot 2017-11-23 at 12 54 57 am" src="https://user-images.githubusercontent.com/977348/33164831-6244d658-cfe9-11e7-8908-fb7f127e9ba6.png">

**Another Issue**
This might just be a personal but the little white "tick" seems to represent the cursor. Is there a way to disable it? I don't really find it that useful but I know that VSCode has something similar (and it confuses me there too). I generally understand my place in the entire document by merely looking at a scroll bar "grip" size compared to the entire right border of the application. I think that a cursor "tick" is redundant information that can be derived from the size of the "grip" compared to the application window height.  Maybe I should file a separate feature request for disabling the cursor "tick"?

bryphe:
Thanks for the detailed notes, @jordwalke !

There are definitely some bugs here... The 'gray grip' should represent the _window_ - (the currently in-view area), and the 'container' should represent the entire _scrollable area_, where the _scrollable area_ is defined as (__buffer height__ + __window height__ - 1).

We add in the window height above to account for scrolling to the very end of the buffer (such that the last line is at the top of the window). The code today _doesn't account for this_ - it just uses the __buffer height__ and places the top/bottom lines within that space, which isn't correct. In addition, we don't handle the case where a line wraps (which, when we define the __buffer height__, we'd want it to account for the line wrapping).

So there are a few bugs in particular with our implementation:
- We don't take into account the bar being the _scrollable area_ - we use the entire view of the buffer. Which is why the 'gray grip' decreases in the last case - it's actually scrolling past our 'container', which isn't sized properly.
- There are some bugs in the initial population - in the first case, it looks like the scroll bar data might be stale (the 'grip' is much too small). And in the second screenshot, it's hitting the issue that the 'scrollable area' isn't correctly accounted, as above.

[...]

PR #1041, opened 2017-11-30
#1007 - Add 'editor.scrollBar.cursorTick.visible' option

bryphe:
Adds an `'editor.scrollBar.cursorTick.visible'` option, which allows disabling the cursor tick within the buffer scrollbar.


bryphe:
First round of fixes for #1007

Issue #1061, opened 2017-12-03
Unable to start go-langserver with Oni 0.2.18


josemarluedke:
Just upgraded to Oni 0.2.18 and now I'm no longer able to get `go-langserver` started. 

It was fine with the previous release. 

<img width="554" alt="screen shot 2017-12-02 at 9 26 23 pm" src="https://user-images.githubusercontent.com/230476/33522700-c00d3f9e-d7a7-11e7-888c-1d320ea9d0ba.png">

bryphe:
Strange - I wonder if there is a change in the `$PATH` behavior due to #934?

Would it be possible to check the following?
- In Oni, run `:term` and then `echo $PATH`
- In a separate bash shell, outside of Oni, run `which go-langserver`

My guess is that Oni is missing some path that is needed to find go-langserver. In the worst case, we can add it to `environment.additionalPaths` as a workaround. Sorry this regressed for you!

josemarluedke:
Inside of Oni:

```
bash-3.2$ echo $PATH
/Users/josemarluedke/.nodenv/shims:/Users/josemarluedke/.rvm/gems/ruby-2.4.2/bin:/Users/josemarluedke/.rvm/gems/ruby-2.4.2@global/bin:/Users/josem
arluedke/.rvm/rubies/ruby-2.4.2/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/josemarluedke/.rvm/bin:/Users/josemarluedke/.rvm/bin:/User
s/josemarluedke/code/go/bin :/usr/bin:/usr/local/bin::/Users/josemarluedke/.fzf/bin:/Users/josemarluedke/.rvm/bin
bash-3.2$
```
```
bash-3.2$ which go-langserver
bash-3.2$
```

Outside of Oni:


```
$ echo $PATH
/Users/josemarluedke/.nodenv/shims:/Users/josemarluedke/.rvm/gems/ruby-2.4.2/bin:/Users/josemarluedke/.rvm/gems/ruby-2.4.2@global/bin:/Users/josemarluedke/.rvm/rubies/ruby-2.4.2/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/josemarluedke/.rvm/bin:/Users/josemarluedke/.rvm/bin:/Users/josemarluedke/code/go/bin
```

```
$ which go-langserver
/Users/josemarluedke/code/go/bin/go-langserve
```

[...]

Issue #868, opened 2017-11-05
Cannot start Oni ( Failed to load python3 host )


meain:
![screenshot](https://i.imgur.com/PHSM7u9.png)

This show up every time I press a key in Oni.
Also on startup, Oni is just a blank ( black ) window and contents on screen wont appear until I enter `command` mode or `insert` mode. It seems like the issue only pops up when in `normal` mode.

> nvim starts fine without any issue

I tried fixing the python path in `config.js` by setting the python path as received from the output of `$ which python`, but still the same issue.

This is my entire config file:
```js
module.exports = {
    "oni.useDefaultConfig": true,
    "oni.loadInitVim": true,
    "editor.fontSize": "14px",
    "editor.fontFamily": "Monaco",
    "editor.completions.enabled": true,
    "neovim.global.python_host_prog": "/usr/local/bin/python3"
}
```

It throws the same even when I am not loading my neovim's `init.vim` config. I also tried setting the python path as the path of `python2`.

Also check out from this issue [issue#649](https://github.com/onivim/oni/issues/649) and tried adding python path in `Contents/Resources/app/vim/default/bundle/oni-vim-defaults/plugin/init.vim`. Still no luck.

bryphe:
Hi @meain ,

Thanks for trying out Oni! Sorry you're hitting this, and I really appreciate the troubleshooting you've done so far.

Would you mind running Neovim in the terminal and running `:CheckHealth`, and sharing the output? There might be some clues there. It does seem that Oni is having an issue finding the python provider - that may give us some clues.

Also, if you completely remove your `config.js`, are you still hitting the issue?

meain:
@bryphe Python part of `:CheckHealth`, seems fine to me. `Nvim` finds it even without seting.
```
## Python 2 provider (optional)
  - INFO: `g:python_host_prog` is not set.  Searching for python2 in the environment.
  - SUCCESS: pyenv found: "/usr/local/Cellar/pyenv/1.1.5/libexec/pyenv"
  - WARNING: pyenv is not set up optimally.
    - SUGGESTIONS:
      - Create a virtualenv specifically for Neovim using pyenv, and set `g:python_host_prog`.  This will avoid the need to install Neovim's Python module in each version/virtualenv.
  - INFO: Executable: /Library/Frameworks/Python.framework/Versions/2.7/bin/python2
  - INFO: Python2 version: 2.7.13
  - INFO: python2-neovim version: 0.1.13
[...]

PR #1066, opened 2017-12-03
Bugfix/path concatenation

Akin909:
@bryphe, whilst investigating a solution to #1061, I realised that swapping out `shell-env` for the call to `childProcess.exec` does not actually work the same way i.e. logging the path output of both processes, `shell-env` correctly picks up the path and plain process call doesn't once the app is built. I hadn't realised that that wasn't working as I still had the additional env variables set in my config so it was picking up my executables which led me to believe the process exec method was working.

With `shell-env` things appear to work correctly and the `which go-langserver` returns the correct response. Apologies thought we had cracked it earlier and wouldn't need the extra dependency, which I've re-added. I've also left out shell-env's associated dependencies and just installed shell-env.

### DIFF
--- a/browser/src/Plugins/Api/Process.ts
+++ b/browser/src/Plugins/Api/Process.ts
@@ -17,7 +17,7 @@ const mergePathEnvironmentVariable = (currentPath: string, pathsToAdd: string[])
 
     const joinedPathsToAdd = pathsToAdd.join(separator)
 
-    return currentPath + separator + joinedPathsToAdd + separator
+    return currentPath + separator + joinedPathsToAdd
 }
 
 const mergeSpawnOptions = async (originalSpawnOptions: ChildProcess.ExecOptions | ChildProcess.SpawnOptions): Promise<any> => {
@@ -31,8 +31,9 @@ const mergeSpawnOptions = async (originalSpawnOptions: ChildProcess.ExecOptions 
     let existingPath: string
 
     try {
-        const { PATH: path } = await shellEnv()
-        existingPath = path || process.env.Path || process.env.PATH
+        const shellEnvironment = await shellEnv()
+        process.env = { ...process.env, ...shellEnvironment }
+        existingPath = process.env.Path || process.env.PATH
     } catch (e) {
         existingPath = process.env.Path || process.env.PATH
     }
--- a/browser/src/Plugins/Api/Process.ts
+++ b/browser/src/Plugins/Api/Process.ts
@@ -1,11 +1,8 @@
 import * as ChildProcess from "child_process"
-// import * as util from "util"
 import * as shellEnv from "shell-env"
 
 import * as Platform from "./../../Platform"
 import { configuration } from "./../../Services/Configuration"
-
-// const exec = util.promisify(ChildProcess.exec)
 
 const getPathSeparator = () => {
     return Platform.isWindows() ? ";" : ":"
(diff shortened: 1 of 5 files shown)


Akin909:
As a side note, which has led to a bit of confusion on my end,  the `PATH` variable echo'ed by `term` in nvim does not seem to reflect the path that the neovim process is using, which is very confusing but in effect if you run `echo $PATH` (no `!`) as an `excommand` you see a path which is what I think neovim uses however the path for the terminal *in nvim in oni* seems to have ?extra or different access as the variable it outputs does not match that.

The `term` path variable in my case seems to include extra separators as I noticed in #1061 but the nvim `echo $PATH` output does not 😕

Akin909:
Apologies for what is beginning to feel like a stream of consciousness/rant, whilst adding shell env seemed to return the correct path, unfortunately in `go` files it was not picking up hover info for external libs etc. I think this is due to the fact that `go` depends on things like `GOROOT` and `GOPATH` environment variables. I've resolved this by merging the default process.env from `electron` with `shell-env`'s environment which gives the user access to all their shell variables.

This solves the problem of finding dependencies based on shell vars and also allows user to have access to all their shell variables, which at least for me is quite helpful as my vimrc references shell variables I or other programs set up.

screen shot 2017-12-03 at 20 28 00

josemarluedke:
Thanks for working on this @Akin909.

I can confirm that this solves the problem reported in #1061.

bryphe:
Thanks for catching this and fixing it, @Akin909 ! Sorry I led you down the wrong path in terms of `childProcess.exec` versus `shell-env`. 

> I think this is due to the fact that go depends on things like GOROOT and GOPATH environment variables. I've hopefully resolved this by merging the default process.env from electron with shell-env's environment.

Awesome! Yes, this makes sense.

Thanks @josemarluedke for confirming the fix, too! I suspect this could impact #868 as well, as the python provider is sensitive to not having the appropriate paths to load the python runtime.

bryphe:
Oh and this is interesting:

> The term path variable in my case seems to include extra separators as I noticed in #1061 but the nvim echo $PATH output does not 😕

I wonder if this is specific to Oni, or if it happens in `nvim` too?

Change looks good to me - the test failures on TravisCI don't seem related to your change (and when I ran locally, they passed). I'll take a look at those failures.  Thanks again for the investigation + fix here, @Akin909 !

Akin909:
@bryphe through very superficial perusing of issues re neovim's terminal I *think* it sources it's `path` differently as it's essentially another program embedded in nvim. Re the fix I tried it locally yesterday with no issues except I can't get `reason` running yet but that pre-dates this issue (I haven't set it up properly yet I think) but I made some tweaks since then and think my local build (which includes my experiments is no longer properly sourcing the tss) will the tests pick that up? if not i'd like to revert some changes rebuild and make sure there aren't any issues there.

bryphe:
@Akin909 - in theory the tests should pick it up (which may be why it is failing in the CI machine) - at least the AutoCompletion-TypeScript test should fail if we're unable to start `tss` (it looks like that is the test that is failing). The Demo-Screenshot test is also dependent on it, since it waits for the completion UX to show to generate the screenshot.

> if not i'd like to revert my build back to this commit and make sure there aren't any issues there.

If you wouldn't mind double-checking, that'd be great! I'll double-check on my machine too. I'll hold off on bringing it in until we confirm that `tsserver` is starting correctly.

Akin909:
@bryphe sorry false alarm re `tss` I've got it running, I was experiencing some crashes in the console due to some plugin tinkering, but I've built a version off of that branch i've got a functioning `tss`

screen shot 2017-12-04 at 17 23 49

bryphe:
@Akin909 Oh ok, great, thanks for the update! I built locally and it is working for me as well.

I'll bring this in, and investigate the stability of those tests.

PR #1070, opened 2017-12-04
Fix #1063: Overly aggressive definition underlining

### LINKED ISSUE #1063: Goto Definition shown in all files
After updating to the latest version of master, I now see the grey line for Go To Definition support shown in all files, including ones that I know I have no support on my machine for (ie CUDA source files on a machine with no CUDA capability), as well as strings and other things that don't have any definitions.

I can see "No language server registered" and "No grammar found for language: cuda" repeatedly in the dev console window, and if I attempt to use Go to definition I get an error of:

```
Uncaught (in promise) TypeError: Cannot match against 'undefined' or 'null'.
    at Object.<anonymous> (file:///C:/Program%20Files%20(x86)/Oni/resources/app/lib/browser/bundle.js:5:37957)
    at file:///C:/Program%20Files%20(x86)/Oni/resources/app/lib/browser/bundle.js:5:37570
    at Promise (<anonymous>)
    at o (file:///C:/Program%20Files%20(x86)/Oni/resources/app/lib/browser/bundle.js:5:37344)
    at Object.t.gotoDefinitionUnderCursor (file:///C:/Program%20Files%20(x86)/Oni/resources/app/lib/browser/bundle.js:5:37853)
    at t.CallbackCommand.y.CallbackCommand [as execute] (file:///C:/Program%20Files%20(x86)/Oni/resources/app/lib/browser/bundle.js:44:158360)
    at i.executeCommand (file:///C:/Program%20Files%20(x86)/Oni/resources/app/lib/browser/bundle.js:1:35120)
    at EventEmitter.R._neovimInstance.onOniCommand.subscribe (file:///C:/Program%20Files%20(x86)/Oni/resources/app/lib/browser/bundle.js:44:108414)
    at emitOne (events.js:115:13)
```

I'd assume there is a check that is missing or potentially invalid now after LSP changes and what not, so that it checks every single file even if there is no valid server.

bryphe:
__Issue:__ We are underlining every token, not just ones that actually have a definition provided by a language server.

__Defect:__ The result form of the definition needs two parts - a token and a location. We assumed if we got a result at all, we have a definition, but it's really gated on whether the location is available.

__Fix:__ Update logic to check for location, rather than result, and add a unit test to cover this case.

### DIFF
--- a/browser/src/Services/Language/LanguageEditorIntegration.ts
+++ b/browser/src/Services/Language/LanguageEditorIntegration.ts
@@ -115,15 +115,13 @@ export class LanguageEditorIntegration implements OniTypes.IDisposable {
     }
 
     private _onStateUpdate(newState: ILanguageState): void {
-        if (newState.definitionResult.result && !this._lastState.definitionResult.result) {
-
-            // Only show if there is actually a location specified...
-            if (newState.definitionResult.result.location) {
-                this._onShowDefinition.dispatch(newState.definitionResult.result)
-            }
+        const newLocationResult = getLocationFromState(newState)
+        const lastLocationResult = getLocationFromState(this._lastState)
+        if (newLocationResult && !lastLocationResult) {
+            this._onShowDefinition.dispatch(newState.definitionResult.result)
         }
 
-        if (!newState.definitionResult.result && this._lastState.definitionResult.result) {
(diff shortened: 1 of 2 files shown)


bryphe:
Fixes #1063

Issue #1038, opened 2017-11-29
Typing Prediction (experimental): Colors not correct in highlighted block


bryphe:
The predicted text colors always use the foreground color, instead of the highlighted block, which is jarring:
![predicted-typing-colors](https://user-images.githubusercontent.com/13532591/33398108-a64739ea-d502-11e7-81f3-1409b6934b8f.gif)
> NOTE: The input latency was artificially set to 250ms via the `debug.fakeLag.neovimInput` setting - but it's still jarring even if it shows for a split second.

As a fix, we should use the last color the cursor was on.

bryphe:
Fixed by #1071

PR #1071, opened 2017-12-04
#1038 - Fix highlight colors in typing prediction

bryphe:
__Issue:__ When typing prediction is enabled, we always use the default background / foreground colors for the predicted text. This works fine in the general case, but when predicting from a case with highlighting (which would be exacerbated once TextMate themes are enabled), you get a jarring effect. 

Easy to see with comments highlighting:
predicted-typing-colors

__Fix:__ Look back to the last non-predicted character, and use the foreground / background color for that.


bryphe:
Fixes #1038

PR #1072, opened 2017-12-04
[WIP] Externalize CommandLine and maybe wildmenu (?)

Bretley:
This is the initial work I've done so far. I kind of just took bits and pieces and threw things at the wall to see what stuck. I'm sure this is going to need a ton of revision and a ton of extra work.

- [ ] Handle recursion with the CommandLine

- [ ] Handle the cursor/position within CommandLine

- [ ] Find out what values should be configurable

- [ ] Determine what CommandLine should look like/ where it should be placed.

### DIFF
--- a/browser/src/Editor/NeovimSurface.tsx
+++ b/browser/src/Editor/NeovimSurface.tsx
@@ -11,6 +11,7 @@ import { INeovimRenderer } from "./../Renderer"
 import { NeovimScreen } from "./../Screen"
 
 import { ActiveWindowContainer } from "./../UI/components/ActiveWindow"
+import { CommandLine } from "./../UI/components/CommandLine"
 import { Cursor } from "./../UI/components/Cursor"
 import { CursorLine } from "./../UI/components/CursorLine"
 import { InstallHelp } from "./../UI/components/InstallHelp"
@@ -59,6 +60,12 @@ export class NeovimSurface extends React.PureComponent<INeovimSurfaceProps, {}> 
                 <div className="stack layer">
                     <TypingPrediction typingPrediction={this.props.typingPrediction}/>
                     <Cursor typingPrediction={this.props.typingPrediction}/>
+                    <CommandLine
+                        visible={false}
+                        content={null}
+                        position={0}
+                        firstchar={""}
+                    />
                     <CursorLine lineType={"line"} />
                     <CursorLine lineType={"column"} />
                     <ActiveWindowContainer>
--- a/browser/src/UI/State.ts
+++ b/browser/src/UI/State.ts
@@ -72,6 +72,16 @@ export interface IState {
     activeWindowDimensions: Rectangle
 
     activeMessageDialog: IMessageDialog
+    commandLine: ICommandLine | null
+}
+
+export interface ICommandLine {
+    content: Array<[any, string]>,
+    firstchar: string,
+    position: number,
+    prompt: string,
+    indent: number,
+    level: number,
 }
 
 export interface IDefinition {
@@ -219,4 +229,5 @@ export const createDefaultState = (): IState => ({
     statusBar: {},
     toolTips: {},
     activeMessageDialog: null,
+    commandLine: null,
 })
--- a/browser/src/neovim/NeovimInstance.ts
+++ b/browser/src/neovim/NeovimInstance.ts
@@ -59,6 +59,22 @@ export interface INeovimCompletionInfo {
     col: number
 }
 
+export type CommandLineContent = [any, string]
+
+export interface INeovimCommandLineShowEvent {
+    content: CommandLineContent[]
+    pos: number
+    firstc: string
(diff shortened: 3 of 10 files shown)


Bretley:
Sorry this is super bare in terms of commits, I very nearly deleted all of the changes I made... :man_shrugging:

bryphe:
Wow, excellent progress @Bretley ! Looking really cool from the gif.

And the code changes look great so far. I like the way you wired everything up - looks like you have a good handle on React + Redux 😉 Left a few minor comments.

I saw the cases for `wildmenu_show`, `wildmenu_select`, and `wildmenu_hide` - are you planning on including the wildmenu externalization in this change too? I'm okay either way.

browser/src/Editor/NeovimSurface.tsx
```diff
@@ -59,6 +60,12 @@ export class NeovimSurface extends React.PureComponent<INeovimSurfaceProps, {}>
                 <div className="stack layer">
                     <TypingPrediction typingPrediction={this.props.typingPrediction}/>
                     <Cursor typingPrediction={this.props.typingPrediction}/>
+                    <CommandLine
```
bryphe: You may not need to specify the `props` here anymore - since `CommandLine` is a connected component, it'll get these from the store via `mapStateToProps`.

browser/src/UI/State.ts
```diff
@@ -72,6 +72,16 @@ export interface IState {
     activeWindowDimensions: Rectangle
 
     activeMessageDialog: IMessageDialog
+    commandLine: ICommandLine | null
+}
+
+export interface ICommandLine {
+    content: Array<[any, string]>,
```
bryphe: Cool, thanks for thinking about how to type this. Some of the cases from Neovim are tricky!

browser/src/neovim/NeovimInstance.ts
```diff
@@ -651,6 +722,7 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                 rgb: true,
                 popupmenu_external: true,
                 ext_tabline: shouldExtTabs,
+                ext_cmdline: true,
```
bryphe: We might want a configuration for this - something like `editor.commandLine.mode` with `'oni'` or `'native'` as options (kind of like what we have for `tabs.mode`, and `editor.comlpetions.mode`).

browser/src/Editor/NeovimSurface.tsx
Bretley: This is an issue with how I did mapstatetoprops maybe? I kept getting an error when they weren't there and I'm not 100% sure why

Bretley:
@bryphe Wildmenu might or might not get done depending on how your vision of this looks. There are three ways I could see this going:

1) CommandLine becomes a Menu: Neovim sends over a cmdline event and the wildmenu functionality is instead replaced by fuzzy-style completion based on some list of keywords we have. This keeps 95% of the work in Oni and only requires that we have some precomputed list of help terms.

2) We keep neovim's version and have it in the usual spot, but instead it's a cool little textbox instead of a part of the actual neovim screen. Even if we don't do much, externalizing the 'static' parts of the neovim screen is a step we need to take in order to get more advanced scrolling/ externalize the actual text display.

3) Some hybrid of these two. It would be pretty easy to make cases for search vs ex mode vs. whatver ? means because neovim lets us know which commandline functionality is being used

Personally, I don't want the search to be over the screen because it might be covering up text,
but I do think that a popup commandline would be cool.

This is really a design choice for you.

bryphe:
Cool, nice summary @Bretley !

Initially, I think it makes sense to keep the command line separate (implementation-wise) from the quick open / command palette menu. There's two 'classes' of users I want to appeal to - Vim veterans who expect/know the command line, and new users who want to get start with modal editing - may be more comfortable in the quick open/command palette. Keeping that distinction now can satisfy both 'classes' of users, and then we can look towards ways to reconcile them later, if it makes sense.

I do like the way you've structured the UI right now - having the command line pop in an overlay is really nice! Good point about the search, though. I wonder if, in that case, we could put it in the corner - sort of like how in Chrome it pops up in the corner?

Regarding the wildmenu, if we use the UI-overlay like you've structured it, we could make the look and feel similiar to the quick open / command palette menu (even if the implementation is different) - I think it would help both classes of our users (for the vim veteran, getting adjusted to some of the new UX paradigms in Oni, and for the new user, it won't feel that different / scary from quick open or command palette).

Hope that helps and wasn't too vague... Great work so far!

badosu:
Hope we can have a configuration to disable this without the overhead of it's implementation.

Although I think this is a great feature it's very obstrusive visually for the edititing flow.

samvv:
I'll try to smoothen this out so that it can be merged. First need to see I can get it working.

bryphe:
I'll close this out since we have #1177 and #1179 in now

Issue #1076, opened 2017-12-05
Oni freezes after inserting the 1st character


muream:
Hey there, 
I just installed oni from the AUR on Arch, no configuration done whatsoever and the editor freezes as soon as I insert the 1st character.
The tabline and line numbers also appear after inserting the character and no colorscheme seems to be loaded.
As soon as the character is entered I get two warnings and an error in the console:

```
bundle.js:1 Unhandled command: mode_info_set
t.warn @ bundle.js:1
t.forEach @ bundle.js:44
_handleNotification @ bundle.js:44
_neovim.on @ bundle.js:44
emitTwo @ events.js:125
emit @ events.js:213
_decoder.on @ bundle.js:26
emitOne @ events.js:115
emit @ events.js:210
addChunk @ _stream_readable.js:252
readableAddChunk @ _stream_readable.js:239
Readable.push @ _stream_readable.js:197
Transform.push @ _stream_transform.js:147
n.push @ bundle.js:44
flush @ bundle.js:5
o._transform @ bundle.js:44
Transform._read @ _stream_transform.js:186
Transform._write @ _stream_transform.js:174
doWrite @ _stream_writable.js:371
writeOrBuffer @ _stream_writable.js:357
Writable.write @ _stream_writable.js:274
ondata @ _stream_readable.js:628
emitOne @ events.js:115
emit @ events.js:210
addChunk @ _stream_readable.js:252
readableAddChunk @ _stream_readable.js:239
Readable.push @ _stream_readable.js:197
onread @ net.js:589
```
```
bundle.js:1 Unhandled command: mode_info_set
t.warn	@	bundle.js:1
t.forEach	@	bundle.js:44
_handleNotification	@	bundle.js:44
_neovim.on	@	bundle.js:44
emitTwo	@	events.js:125
emit	@	events.js:213
_decoder.on	@	bundle.js:26
emitOne	@	events.js:115
emit	@	events.js:210
addChunk	@	_stream_readable.js:252
readableAddChunk	@	_stream_readable.js:239
Readable.push	@	_stream_readable.js:197
Transform.push	@	_stream_transform.js:147
n.push	@	bundle.js:44
flush	@	bundle.js:5
o._transform	@	bundle.js:44
Transform._read	@	_stream_transform.js:186
Transform._write	@	_stream_transform.js:174
doWrite	@	_stream_writable.js:371
writeOrBuffer	@	_stream_writable.js:357
Writable.write	@	_stream_writable.js:274
[...]

PR #1078, opened 2017-12-05
#1076 - Crash on buffer change event

bryphe:
__Issue:__ The code in `Completion.ts` makes the assumption that the order of events will always be `cursor_moved` and then `buffer_changed`. Because of this, when there is a buffer update, it assumes that `this._lastCursorPosition` will be populated. If that assumption is not true, there will be a crash.

__Fix:__ Guard against the initial case where `this._lastCursorPosition` isn't populated. This should prevent the crash, but further investigation is needed to make sure that completion works correctly (the worst case would be that the cursor movement event, in some cases, _always_ comes after the buffer update, which would give out-of-date or incorrect completions).

### DIFF
--- a/browser/src/Services/Completion/Completion.ts
+++ b/browser/src/Services/Completion/Completion.ts
@@ -146,7 +146,7 @@ export class Completion implements IDisposable {
 
         const newLine = firstChange.text
 
-        if (range.start.line === this._lastCursorPosition.line) {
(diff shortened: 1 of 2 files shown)


bryphe:
First part of a fix for #1076

bryphe:
I'll bring this in, since it is a defensive change that won't cause a worse problem - will be interesting to see how it impacts #1076.

Issue #852, opened 2017-11-02
Tab bar does not work as expected


CameronDiver:
If I leave the options as the default, I get an empty tab bar. If I set `"tabs.showVimTabs"` to true, the current buffer is shown in the single tab that is created. 

To get more than one tab to be opened, I have had to manually do `:tabnew [file]`. All I want is my open buffers to be shown in the tab bar, what am I missing?

I saw this: https://github.com/bryphe/oni/issues/835 and have neovim version 0.2.1, but I don't have the `:help ui-ext-tabline` setting, although I have installed `neovim-git` which resolved that thread.

It's probably worth mentioning I use my own vim config, with https://github.com/thaerkh/vim-workspace but I've tried without that plugin running and still no luck.

I also get the same behaviour if I don't use my config, leading me to believe that I'm completely misunderstanding the tab bar? The documentation states that the standard behaviour is for all open buffers to be displayed in the tab bar, but I'm not seeing this.

bryphe:
Hi @CameronDiver ,

Thanks for trying Oni, and for the detailed issue!

Your understanding of the tab bar is correct to me - the default & intended behavior, with `tabs.showVimTabs` set to false, should be to show buffers in the tabline. So seems like there is a problem here.

A couple questions to help clarify:
- Which version of Oni are you using? Did you build from source, or download the latest release (0.2.15)?
- I'm assuming you're on Linux - which distro are you using?

In addition, would it be possible to get a screenshot of what you're seeing? I'm curious if Oni is rendering the tab bar at all (or if it is being rendered in vim, as in #835). 

It'd be great to see the output of `:CheckHealth` as well from within Oni - that might help identify if Oni is picking up the `nvim` binary from the wrong place.

CameronDiver:
Thanks for the reply @bryphe !

Sure thing,
Distro: arch linux 
oni version (installed from `yaourt/aur`): 0.2.12.1

Apologies, github file uploading is not working for me, so I've added screenshots to imgur:
https://imgur.com/a/1CuCY

First image shows vim-tabs disabled, second image shows vim-tabs enabled, third image shows vim-tabs still enabled, and another buffer opened with `ctrl+p` (the single tab changes name).

Output of checkhealth:
```

health#deoplete#check
========================================================================
## deoplete.nvim
  - OK: has("nvim") was successful
  - OK: has("python3") was successful
  - INFO: If you're still having problems, try the following commands:
[...]

PR #1082, opened 2017-12-06
Fix buffer deletion

Akin909:
@bryphe, I'm working on resolving some of the various quirks with the buffer bar (relates to #852), largely surrounding deleting buffers and having them disappear as well as loading all buffers on `bufenter` and removing non-interactive buffers like that of `NERD_TREE` etc. from the the bar which cause crashes anyway if you try to open these.

As part of that I've been looking to correctly update the buffer bar when `bufDelete` is triggered. I tried using the `getBufferIds` function which communicates with the `msgpack api` but have found that the api returns a list of buffers which *still* includes the deleted buffer.

I've done a fair bit of googling and have not found any related issues but it seems that there's possibly a delay?? 😕 on updating the api.

Anyway a solution I've been using locally for a little while is just filtering out the buffer that triggered the buffer delete autocommand from the list of buffers and dispatching the set current buffers action.

I imagine that a solution via the `msgpack` api would be preferable in the future whether an issue might need to be raised on neovim repo? (if i'm not just mistaken about how the api is used). Another possible solution is that for the other issues re buffers, i've expanded on your `vim interop` plugin and created a function which gets (which gets whatever metadata is available) for each buffer which I plan to trigger on `bufenter` but can also trigger on `bufdelete` and use the result from that to update the state at least that way the answer to what buffers exist is coming from neovim.

What are your thought re this approach or either of the other ones?

### DIFF
--- a/browser/src/Editor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor.tsx
@@ -482,6 +482,7 @@ export class NeovimEditor extends Editor implements IEditor {
         } else if (eventName === "BufDelete") {
 
             this._neovimInstance.getBufferIds()
+                .then((ids) => ids.filter(id => evt.bufferNumber !== id))
                 .then((ids) => UI.Actions.setCurrentBuffers(ids))
         }
     }


bryphe:
>  I'm working on resolving some of the various quirks with the buffer bar (relates to #852), 

Awesome, really appreciate your help with this!

> As part of that I've been looking to correctly update the buffer bar when `bufDelete` is triggered. I tried using the `getBufferIds` function which communicates with the `msgpack api` but have found that the api returns a list of buffers which *still* includes the deleted buffer.

I just went back to the issue - I actually meant to reply to that thread, but somehow I lost my comment. Sorry about that! I did some digging around this... I learned that there is the concept of an _unlisted_ buffer. It turns out, that when a buffer is deleted (ie, via `:bdelete`), that it actually isn't deleted completely - it's removed from the buffer list and marked as _unlisted_. (`:help unlisted-buffer` and `:help bdelete` have some info).

You can see _unlisted_ buffers via `:ls!` - it seems that the `nvim_list_bufs` (which is what `getBufferIds` is using) actually shows the unlisted buffers (so it behaves more like `:ls!` than `:ls`).

My thinking is this what is causing the issue (versus a delay).

> Anyway a solution I've been using locally for a little while is just filtering out the buffer that triggered the buffer delete autocommand from the list of buffers and dispatching the set current buffers action.

I wonder how this behaves if I delete multiple buffers in sequence? For example, I have buffers [1, 2, 3], and then I `:bdelete 2`, and then later `:bdelete 3` - it seems like 2 might come back in that case, since `getBufferIds` would return [1,2,3], and then we'd filter out 3. So 2 would be like a zombie buffer... back from being deleted 😄 

One way to solve this would be to keep a list of the deleted buffers, and always ignore them in the future - that should work as long as it is not possible to 'undelete' a buffer.

> Another possible solution is that for the other issues re buffers, i've expanded on your `vim interop` plugin and created a function which gets (which gets whatever metadata is available) for each buffer which I plan to trigger on `bufenter` but can also trigger on `bufdelete` and use the result from that to update the state at least that way the answer to what buffers exist is coming from neovim.

This solution seems reasonable to me, especially if there isn't a good way to get the get this information via the `nvim_get_bufs` API right now. This would let us get the correct set of buffers deterministically - creating a function that gets us the list of buffers that we should be showing would be really useful. 
It seems like it could help us fix this buffer delete case (by using that function instead of the `nvim_list_bufs` API), and it may be able to help with some of the other cases where we are out of sync with the available buffers in Vim. (for example, `:sav` has some issues - I believe because it is creating a new buffer w/o a `BufEnter` event, and there may be more cases like that).

Also, it may be worth creating an issue in Neovim too - I imagine other UIs hit this pain point. Perhaps they could add an `nvim_get_listed_bufs` method, or an optional argument to only give listed buffers?

Thanks for your help with this, @Akin909 !

Akin909:
@bryphe  I've managed to make some progress with the functionality I described, which I've been working on another branch I'll make a PR from there and would be grateful if you could have a look it's about 90% there (I think/hope) aka it loads existing buffers when oni loads a vim session and that function can be reused for other events

bryphe:
@Akin909 sounds great!

Issue #936, opened 2017-11-13
Vendoring Python


jordwalke:
Some people who have tried to get setup with `Oni` run into issues with their python installation. Configuring python is a nightmare for people who don't use python daily. Is it possible/easy to vendor Python into Oni?

bryphe:
Thanks for inspiring people to try out Oni (and helping them get set up)! Looks like we had a record number of downloads today.

> Some people who have tried to get setup with Oni run into issues with their python installation. 

Is this the same issue as #868? It seems like there is a specific issue on OSX, where if the `python` or `python3` executable is in `/usr/local/bin`, there'd be issues loading the python host (in the worst way, Oni would be unusable).

That particular issue is fixed with #924 - I'm planning on cutting a release with this fix on Wednesday. (The root cause was that Oni wasn't getting the full set of environment variables, and this caused problems with Neovim initializing the python provider). We're also looking at solving this in a more general way by leveraging [shell-env](https://www.npmjs.com/package/shell-env)

What's interesting is, prior to the fix in #924 , Oni worked fine on OSX if there was _no python binary available_, and also if `python` or `python3` was available in `/usr/bin`... but would have problems if it was somewhere else, like `usr/local/bin`.

Alternatively - are they hitting this issue because they want to use a plugin that depends on python, like UltiSnips? And having difficulty configuring it?

> Configuring python is a nightmare for people who don't use python daily. 

This is so true! I'm trying to stay away from packaging Python, if I can help it... Especially on Windows.

jordwalke:
Sounds like you've got a handle on things then. If packaging Python is out of the question, I'll close this issue. (I mean, I can't blame you - there's no way I would ever attempt that).

bryphe:
👍 #868 should be fixed now in [0.2.17](https://github.com/onivim/oni/releases). Will be interested to see if there are still configurations that have issues.

chenglou:
Fixed for me! =D

bryphe:
Awesome, great to hear! Thanks @Akin909 for all your work on #924 & #1066 to fix these issues.

PR #1086, opened 2017-12-07
Update oni with existing buffers and deletions

Akin909:
@bryphe As discussed in #1082 here are the changes I've been working on to update the buffer bar with existing buffers on `bufenter`, and more generally to get information on all `listed` vim buffers. 
There are a couple of bugs I need to fix with this e.g how to handle deletion,  Im planning to call the `notifyWithBuffers` function on buffer delete and repopulate oni with data from neovim rather than filtering, to avoid a potential 😆 zombie buffer situation.

Any feedback or comments re tweaks would be much appreciated as `vimscript` is not my forte.

Also: I added a sort to the tabs as navigating them using `:bnext` or `:bprev` does not move across tabs in order so sorting them arranges them so that it does.


Akin909:
@bryphe just discovered this in the vim docs re auto commands

>[**BufDelete:**](http://vimdoc.sourceforge.net/htmldoc/autocmd.html#BufDelete)			Before deleting a buffer from the buffer list.

I've just tested this locally with that in mind and it seems regardless of what the api does this autocommand is called before the point at which it would actually be relevant to this use case 🤕, i.e before the before is *actually* deleted . Seems like a better approach would be to use the `BufEnter` for buffer status updates re deletion and loading vim at least. 

Which makes sense tbh since whatever is done in `BufDelete` is shortly overidden in `BufEnter`

Akin909:
Completed this functionality now and its stable locally or rather I haven't found any edge cases that have revealed any bugs.

I ended up using the `BufEnter` event as the primary method of updating buffer statuses, everytime its triggered it replaces the bufferlist with one from neovim using a `oniNotifyWithBuffers` viml function. There's definitely places that could be improved in the long run as for example I discovered python can be used in vim to execute functions, as use case for that would be importing the `vim` python module and using that to execute slower computations like getting the number of lines in each buffer.

bryphe:
> Completed this functionality now and its stable locally or rather I haven't found any edge cases that have revealed any bugs.

Cool, I'll start taking a look through this!

> I ended up using the `BufEnter` event as the primary method of updating buffer statuses, 

Sounds reasonable. There may still be events we're missing, but updating it here will certainly improve some of the cases we have.

>  I removed the `BufDelete` event because it wasn't serving any purpose given that the updating happens in bufenter

 I wonder though if it's common to delete a buffer that is not the active buffer? We might miss that case without the `BufDelete`. An example I'm thinking of is if I click the 'x' on a buffer that is not currently active - would that be handled correctly?

> There's definitely places that this could be improved in the long run as for example I discovered python can be used in `vimscript` to execute functions

Unfortunately this isn't a good option for Oni 😦  #936 briefly touches on this, but not all platforms have python available (in particular, Windows), and it's actually quite difficult to install and setup as part of Oni in the general case - so I'm trying to stay away from having to do that work if possible. Lua might be worth a try, though, so it comes standard in Neovim now.

I'll leave some more feedback on the changes!

browser/src/UI/Actions.ts
```diff
@@ -83,12 +83,15 @@ export interface ISetFont {
 export interface IBufferEnterAction {
     type: "BUFFER_ENTER",
     payload: {
-        id: number,
-        file: string,
-        fileType: string,
-        totalLines: number,
-        hidden: boolean,
-        listed: boolean,
+        currentBuffer: {
```
bryphe: Instead of distinguishing between `currentBuffer` (which is an `IBuffer`) and `existingBuffers`, perhaps it would simplify things to have just a `buffers: IBuffer[]` an an `activeBufferId`?

browser/src/UI/Reducer.ts
```diff
@@ -149,22 +149,27 @@ export const buffersReducer = (s: State.IBufferState, a: Actions.SimpleAction):
         case "BUFFER_ENTER":
             byId = {
                 ...s.byId,
-                [a.payload.id]: {
-                    id: a.payload.id,
-                    file: a.payload.file,
-                    totalLines: a.payload.totalLines,
-                    hidden: a.payload.hidden,
-                    listed: a.payload.listed,
+                [a.payload.currentBuffer.id]: {
```
bryphe: Following from the above comment in the ActionCreator - if we used a single array, as opposed to differentiating the cases, it seems like the lower `reduce` would handle both the current buffer + active buffer cases.

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -42,6 +52,72 @@ function OniNotifyEvent(eventName)
...
+  endif
+endfunction
+
+
+function OniGetEachContext(bufnum)
+  let l:context = {}
+    let l:bufpath = bufname(a:bufnum)
+
+    if strlen(l:bufpath)
+      "TODO: python can be used to retrieve the lines a buffer and len return
```
bryphe: As mentioned in the comments, unfortunately Python isn't an option for us (because there is no guarantee it's available on Windows). Lua might be an option though.

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -42,6 +52,72 @@ function OniNotifyEvent(eventName)
...
+      "TODO: python can be used to retrieve the lines a buffer and len return 
+      " the number of files this is much quicker than using viml getbufline fn
+      " https://stackoverflow.com/questions/14544618/is-there-a-way-to-get-the-number-of-lines-of-a-buffer-in-vim-script
+      " if has('python')
+      " python import vim
+      " let l:context.bufferTotalLines = pyeval('len(vim.buffers['.(a:bufnum -1).'])')
+      let l:context.bufferNumber = a:bufnum
+      let l:context.bufferFullPath = expand("#".a:bufnum.":p")
+      let l:context.bufferTotalLines = v:null
+      let l:context.line = v:null
```
bryphe: Do we need all these `v:null`s here? I'm thinking that it might be worth having a different 'type' as opposed to trying to fit this to the eventcontext

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -42,6 +52,72 @@ function OniNotifyEvent(eventName)
...
+function OniGetAllBuffers()
+  let l:buffers = []
+  let l:bufnums = User_buffers()
+  if exists("l:bufnums")
+    for l:bufnum in l:bufnums
+      try
+        let l:buffer = OniGetEachContext(l:bufnum)
+        let l:buffers += [l:buffer]
+      catch /.*/
+      "Probably dont want this outside of a debugging scenario
```
bryphe: Interesting - are there any cases we could hit this?

browser/src/UI/Actions.ts
Akin909: Sounds fair definitely crossed my mind a few times I initially thought there might be a specific need to return the current buffer as a separate item but as you point out its not really needed.

vim/core/oni-core-interop/plugin/init.vim
Akin909: Ah I had it somewhere in my head that neovim was compiled with python by default so anyone who had neovim would have setup up python but that was a fairly random assumption on my part. As I haven't been able to get to grips with it anyway ill take that comment out

bryphe: Ah ya, well it does work out of the box on OSX (and Linux too I believe) so it's an easy assumption to make! 😄

vim/core/oni-core-interop/plugin/init.vim
Akin909: Not really, nor that i've observed with the current state of the function I have it there as a last ditch attempt at catching an error if one occurs but it was essentially there for me to use to debug whilst writing it but also the try/catch would mean that even if there were an error the user wouldn't be faced with vim's not so pleasant scrolling wall of red lines

i've checked that each function gets the right kind of argument and that it doesn't recieve an invalid argument which in most cases would return in 0 being added as the value to the key. I tried a couple of different methods, and took a lot of inspiration from `buftabline` a plugin I use

Akin909:
Re. the `bufdelete` you've definitely hit on something there, as it does in fact cause the exact problem you alluded to, seems it is necessary although I was thinking of using the `bufwipeout` event which the docs indicate is triggered after `bufdelete` so hopefully after the bufferlist is updated.

bryphe:
Ah ya, I just tried out and saw the same thing. Thanks for looking at it!

We are using `bw` (`bwipeout`) in the buffer close gesture here:
https://github.com/onivim/oni/blob/9a02a9368c2fae903aef84c702e8b617c64907c7/browser/src/Editor/NeovimEditor.tsx#L409

Seems like it should hit that buffer wipeout event, though.... If it's the case that we're getting the delete event before it's actually removed, then we could still use that 'filtering' logic to filter out the deleted buffer (to hide it immediately), and then subsequent calls to get the buffer list should not return it (so we don't hit that zombie-buffer-coming-back issue)

Akin909:
I unified a the buffers as you suggested and it much more straightforward, also removed the python comments as well as re-added the `bufwipeout` event and handler and that actually works really well so you can click to delete an inactive buffer and have it disappear. 

Re. changing the `eventContext` type since other events expect different types what would your suggestion be would a union type work? by just removing those fields I dont get any type errors although your intent might have been to have the interfaces exactly match the data.

browser/src/UI/Reducer.ts
```diff
@@ -147,24 +147,21 @@ export const buffersReducer = (s: State.IBufferState, a: Actions.SimpleAction):
...
-                },
-            }
+                }
+                return byId
+            }, byId)
 
-            if (allIds.indexOf(a.payload.id) === -1) {
-                allIds = [...s.allIds, a.payload.id]
-            }
+            const bufIds = a.payload.buffers.map(b => b.id)
```
bryphe: Cool, it's nice we don't have to special case the active buffer anymore!

bryphe:
Cool, I just tried it out and it seems to be working well!

Regarding the event type - unfortunately there is no way to validate that the `IEventContext` object matches what we have in VimL. However, it does help downstream code to know what to expect.

I think it would be helpful to document this new type of return as an `IBufferEventContext`, and be able to use that strong-type for those specific buffer events. I've been slowly moving away from the `EventEmitter` event on NeovimInstance, because it doesn't give us the ability to express the type of the object. In other words, I'd like to refactor out `_onVimEvent` and subscribe to strongly-typed handlers in `NeovimInstance` instead.

I've been moving to strongly-typed events for that reason, like:
https://github.com/onivim/oni/blob/18e651becb200738e8d1061a964a5871a8746ba7/browser/src/neovim/NeovimInstance.ts#L195

In the current iteration, it's a bit tough to know what's happening in the `_onVimEvent` because there are multiple places we have to check whether `evt` is _actually_ an `EventContext`, or if it's something else - I think that might be difficult to maintain long-term.

However, if we create strongly-typed events for these operations, like:
`onBufDelete: IEvent<IBufferEventContext[]>`
`onBufEnter: IEvent<IBufferEventContext[]>`
`onBufWipeout: IEvent<IBufferEventContext[]>`
`onBufRead: IEvent<IBufferEventContext[]>`

on NeovimInstance, and then add event handlers to these (so refactor out from `_onVimEvent` in `NeovimEditor`, such that `NeovimEditor` subscribes to these events), it clearly expresses the type we expect to get, and then the type system can help us out more (and we can side-step creating a union type).

I ran into something similiar with the buffer update cases - I used to use a similiar `IEventContext` and had some logic to try and disambiguate, but it became complicated, so I factored them into their own types and used a strongly-typed event handler - I think this will help make it easier to maintain long-term.

Hope that makes sense, let me know if you have any questions or if it was unclear! Thanks for your work on this, @Akin909 !

bryphe:
Actually, I just realized that most of these are already defined on `NeovimAutoCommands`:
https://github.com/onivim/oni/blob/18e651becb200738e8d1061a964a5871a8746ba7/browser/src/neovim/NeovimAutoCommands.ts#L15

So it'd be more just switching some of those from `EventContext` to `BufferEventContext`, and then listening to those events vs. handling them on `_onVimEvent`.

Akin909:
@bryphe I've split out the events into separate handlers and created a BufferEventContext type which is an object which consists of a currentBuffer which is an `EventContext` and existing buffers which are `InactiveBufferContext` types. I reverted to using an object because I needed to be able to reference the current buffer separately as the buffer enter event triggered several other functions all of which explicitly expected an `EventContext` and for which the `InactiveBufferContext` was invalid.

Akin909:
I've tried looking into what causes the `AutoClosingPairs` tests to fail on appveyor however I don't think I've made any changes to relate to that functionality. When I run `yarn test` locally it runs 3 autoclosing pairs tests all of which pass but the `packageMetadataParser` tests instead fail, so it's hard to be sure what's going on

bryphe:
Thanks for looking into it, @Akin909 ! Unfortunately it does seem to be failing consistently on Windows 😦 

I'll troubleshoot in a bit - would it be possible to minimize the set of changes in `NeovimEditor.tsx` to the changes needed for the new events/handlers? The formatting changes add a lot of noise - minimizing the diff there to the essential pieces will make it much easier for me to debug.

Akin909:
@bryphe  I'll revert all none essential changes in `NeovimEditor.tsx` and see if that helps hopefully nothing got negatively impacted by the formatting

bryphe:
Awesome - thanks @Akin909 !

Akin909:
@bryphe  all done, branch still fails though 😭 , I'll also have a look at the last passing commit and see what I introduced after that

bryphe:
Thanks for minimizing the changes, that's very helpful!

It's strange - pressing `<CR>` when the cursor is in between seems to fail with this PR:

`window.setTimeout(|)`

I'd expect it to split to three lines:
```
window.setTimeout(
    |
)
```

But it actually ends up as:
```
window.setTimeout(
|)
```

Seems to repro on OSX for me too - it's surprising that the completion pairs test passed at all on the OSX build. 

Looking into why this behavior is different now... really strange!

bryphe:
I noticed it was also reproducing for the backspace character too.

It seems that `editorManager.activeEditor.cursor.line` is not being populated anymore. This would likely cause regressions in several other places too - just happened to get caught by the auto closing pairs test.

I'm guessing this is because we are no longer passing this value up for the current buffer?

Akin909:
@bryphe  very good catch I was really scratching my head with this 👍 , i'll have a look i think my format buffers function doesn't pass the line to reducer but we still have this value, ah which reminds that actually all the extra values that current buffer has are also not being passed over due to the same function. Ill tweak it and try the tests again

browser/src/Editor/NeovimEditor.tsx
```diff
@@ -465,40 +477,59 @@ export class NeovimEditor extends Editor implements IEditor {
...
+            currentBuffer.bufferNumber.toString(),
+            currentBuffer.windowTopLine - 1,
+            currentBuffer.windowBottomLine - 1,
+        )
+    }
 
+    private async _onBufEnter(evt: BufferEventContext): Promise<void> {
+        this._updateWindow(evt.current)
         const lastBuffer = this.activeBuffer
-        const buf = this._bufferManager.updateBufferFromEvent(evt)
```
bryphe: I believe the removal of this line is what is potentially causing the problem. Note that before this line of code was called for _all vim events_ - not just `BufEnter`. It looks like as part of the refactoring we split this out into the BufEnter/BufLeave event handlers, but unfortunately that's not functionality equivalent.

I think for now we'll have to keep the `_onVimEvent` handler for now, but just for the code above:
- the `setWindowCursor` action
- the `syntaxHighlighter.notifyViewportChanged`
- the `bufferManager.updateBufferFromEvent`

bryphe:
Cool, I think I found the issue - I just added some comments in `NeovimEditor`. Seems like there was a spot we relied on being called more frequently (for all kinds of other vim events). Let me know if you have any questions!

browser/src/Editor/NeovimEditor.tsx
Akin909: Ah in which case the line variable is needed before the reducer, ill add that in to the events I think you were likely still right about separating into handlers though as I imagine that as various use cases become know even more handlers will be needed. Im thinking of adding the line to the event handler above the switch statement so it runs before all events

Akin909:
@bryphe  I've just added the call to  initial line of the callback before the switch statement that way the function is called for all events and each handler is still free to carry out its own logic. If you'd prefer reverting to `_vimEvent` method though I can also re-add that.

browser/src/Editor/NeovimEditor.tsx
```diff
@@ -467,40 +481,57 @@ export class NeovimEditor extends Editor implements IEditor {
         }
     }
 
-    private _onVimEvent(eventName: string, evt: EventContext): void {
-        UI.Actions.setWindowCursor(evt.windowNumber, evt.line - 1, evt.column - 1)
+    private _updateWindow(currentBuffer: EventContext) {
```
bryphe: It looks like this is another case where these methods - `UI.Actions.setWindowCursor` and `this._syntaxHighlighter.notifyViewportChanged` - aren't going to be called as frequently as before. Prior to this change, they were called for every event. The `UI.Actions.setWindowCursor` is particularly sensitive because there are lots of UI pieces that depend on it being set (ie, all the quick info and context menus).

I think it's risky to change this behavior as part of this change - could we make `this._updateWindow` called for every event, as before? Similiar to what you did for `this._bufferManager.updateBufferFromEvent`.

browser/src/Editor/NeovimEditor.tsx
```diff
@@ -172,7 +172,21 @@ export class NeovimEditor extends Editor implements IEditor {
             commandManager.executeCommand(command)
         })
 
-        this._neovimInstance.on("event", (eventName: string, evt: any) => this._onVimEvent(eventName, evt))
+        this._neovimInstance.on("event", (eventName: string, evt: any) => {
+            const current = evt.current || evt
+            const buf = this._bufferManager.updateBufferFromEvent(current)
+            switch (eventName) {
+                case "BufEnter":
```
bryphe: As opposed to using the switch here, could we hook the `_onBufEnter` event by using the strongly-typed event, like:

`this._neovimInstance.autoCommands.onBufEnter.subscribe((evt) => this._onBufEnter)`? LIikewise for `BufWipeout` and `BufWritePost` - this will help ensure we can leverage the typed events for these.

This way, the only stuff happening in here is the stuff we were doing on every event anyway - the `this._bufferManager.updateBufferFromEvent`, the `UI.Actions.setWindowCursor`, and the syntax highlighter change. And later, in another change, we can refactor those out, and remove this untyped event completely.

bryphe:
Cool, thanks for updating hte PR with that change, @Akin909 !

I left a couple comments inline - it looks like the `updateBuffer` is in better shape, but there were a couple other calls that also depended on being called for _every_ event that look like they were moved as well - I think it'd be safer for this change to keep those the way they are today. We could look at continuing the refactoring in another PR.

In addition, I'd like to move to using the strongly-typed events on NeovimInstance, like:
`this._neovimInstance.autoCommands.onBufEnter.subscribe((evt) => { })` - would it be possible to move those cases in the switch statement to directly susbcribe to those events, instead? We may need to add a couple.

Let me know if you have any questions!

Akin909:
@bryphe, ✋ from the comment above I'm not sure what you mean when you say
>keep those the way they are today.
Do you mean the way they are on master aka refactor to `_onVimEnter` but the rest of the comment seems to imply the opposite so keep the the way they are in the PR,
also the missing events are those events other than those in `_updateWindow` ?

Akin909:
@bryphe I've attempted to refactor to what I *think* you had in mind but given my earlier question(s) I'm not entirely certain if this is what you intended apologies for any potential confusion or kerfuffling if I've gone off track

browser/src/Editor/NeovimEditor.tsx
Akin909: Ah of course, will do, out curiosity which events trigger this handler or is it a catch all almost for all vim events not handled in the neovim instance switch?

bryphe:
Sorry I missed your first round of questions, @Akin909 ! 

> I've attempted to refactor to what I think you had in mind but given my earlier question(s) I'm not entirely certain if this is what you intended apologies for any potential confusion or kerfuffling if I've gone off track

It's exactly what I had in mind! Thanks for being open to and making the changes 👍  Really appreciate all the work put into this.

I like that the `BufferEventContext` is strongly typed on those events - this will help make it easy to understand from the consumer side what sort of data is available. 

And the leftover stuff in `on("event")` we can refactor out later. That's called for every vim event we pass up (so every vim event that we had a notification handler for in `init.vim`). I think a good next step is to pick deterministic events to hook - like maybe cursor moved, buffer enter, etc - and get it out of the general handler. And then we can completely scrap that general handler.

This change looks great - I'll bring this in once the build is set.

Akin909:
@bryphe no worries that good to hear (that I haven't gone off track) unfortunately it seems somethings gone awry though autocompletion tests are failing and on manually recreating the test i.e. going into a `css` file typing `.test { pos` I get progress back seems to be picking up html completion though in a less file strangely enough things work correctly. i'll do some digging and re-run the tests

Akin909:
Problem solved seems to have been due to an unreliable variable, I've also added keys to a few components to quiet react, I know it implicitly uses the array indexes (at least according to the docs) if there are no explicit keys but I've on occasion had some *very* strange list behaviour because of their absence so am a bit skittish about it

browser/src/Editor/NeovimEditor.tsx
```diff
@@ -176,7 +175,9 @@ export class NeovimEditor extends Editor implements IEditor {
         this._neovimInstance.on("event", (eventName: string, evt: any) => {
             const current = evt.current || evt
             this._updateWindow(current)
-            this._buf = this._bufferManager.updateBufferFromEvent(current)
+            if (eventName !== "bufEnter") {
```
bryphe: Hmm, does this check work? I thought it needed to be `BufEnter`

Akin909: I've pushed a typo fix for that but ended up removing it having had a look at the function as the if statement there seems to be of little value as logging it, that function gets called very frequently and is not very expensive i.e. does not result in any async actions etc. on top of which it gets called by the `bufWinEnter`event then subsequently by `bufEnter`event then `bufRead` then `cursorsMoved` etc so preventing it being called one extra time seemed to have little value, ideally separating it into the essential bits for buffer updating and a function that returned the desired `OniBuffer` would be ideal as it stands i have to call it in the `bufEnter` handler to get the return  as well as in the general event handler

bryphe: Ah ok, makes sense!

bryphe:
> Problem solved seems to have been due to an unreliable variable,

Sweet, nice find!

> if there are no explicit keys but I've on occasion had some very strange list behaviour because of their absence so am a bit skittish about it here's a horrorpen showing an example

LOL I hadn't heard of a 'horrorpen' before but that is a pretty apt name 😄

Akin909:
The stuff of nightmares  😆

bryphe:
Cool, looks like the builds are green now! I'm ready to bring this in - are you all set with the changes, @Akin909 ?

Akin909:
Yep all good 👍

bryphe:
Excellent, bringing this in now. Thanks for all your work on this, @Akin909 ! 👍

PR #1087, opened 2017-12-07
Feature: Icon themes (+bundling the Seti UI theme)

bryphe:
While investigating the file explorer, I wanted to include file icons - so I split this out as a separate change. The package format is very similiar to VSCode (such that, ideally, we could also load VSCode's icon themes).

The icon theme is controllable via the `ui.iconTheme` property (and can be sent to `null` if you do not wish to have icons). The default theme right now is the popular [seti-ui](https://github.com/jesseweed/seti-ui) theme.

test

File icons are only shown when `tabs.mode` is set to `buffer` (the default) - if you are using `tab`, we do not set a file icon.


badosu:
@bryphe Chiming in a little bit late, but shouldn't this be unrelated to whether tabs mode is buffer or tabs (e.g. introducing a ui.showTabIcons = true|false option)?

bryphe:
@badosu - I wasn't sure if I could reliably get the file type for tabs, since it's possible to rename tabs AFAIK (and the icons are based off the filenames). If it's something that is desired, we could look at putting the file icon for the last active buffer in that tab, perhaps?

bryphe:
I also considered just putting a vim/neovim logo for tabs

badosu:
Did not know about the tab renaming issue, thanks for the clarification 👍

PR #1088, opened 2017-12-08
prevent text-selection of statusbar

Akin909:
Noticed shortly after closing #1010 that though it referenced only the tab bar the status bar text could still be selected, which I think isn't intentional, in which case here's a similar fix for that if needed.


bryphe:
Excellent, thanks for catching that @Akin909 ! Makes a big difference in making this feel like a 'real' app as opposed to a web app.

PR #1092, opened 2017-12-08
Swap indexes to one indexed for quick fix window.

CrossR:
Currently, this assumes all indexes that come in are 0 indexed, which may not be true.

As far as I know, all the LSPs use 0 indexed locations, so there should be no issues, but thought it was worth checking.

I may also rename `locations` to be `zeroIndexedLocations`, just so its clearer when implementing other code around this.

### DIFF
--- a/browser/src/Services/Language/FindAllReferences.ts
+++ b/browser/src/Services/Language/FindAllReferences.ts
@@ -44,14 +44,14 @@ export const findAllReferences = async () => {
 }
 
 export const showReferencesInQuickFix = async (token: string, locations: types.Location[], neovimInstance: INeovimInstance) => {
-    const convertToQuickFixItem = (location: types.Location) => ({
+    const convertToOneIndexedForQuickFix = (location: types.Location) => ({
         filename: Helpers.unwrapFileUriPath(location.uri),
-        lnum: location.range.start.line,
-        col: location.range.start.character,
+        lnum: location.range.start.line + 1,
+        col: location.range.start.character + 1,
         text: token,
     })
 
-    const quickFixItems = locations.map((item) => convertToQuickFixItem(item))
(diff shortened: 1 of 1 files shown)


bryphe:
Awesome, appreciate the fix @CrossR ! Looks good to me.

> As far as I know, all the LSPs use 0 indexed locations, so there should be no issues, but thought it was worth checking.

Yes, this is right - the LSP uses 0-indexed, and I'd like to move Oni in that direction too. Vim mostly uses one-based indices (except for some of the msgpack APIs, like `nvim_buf_get_lines`), and the TypeScript server uses one-based indices.

> I may also rename locations to be zeroIndexedLocations, just so its clearer when implementing other code around this.

Having the names reflect it definitely helps - thanks for renaming that lambda to `convertToOneIndexedForQuick` to make the intention clearer. 

We can incrementally work to rename asyou did above in future fixes - I think it definitely helps. I wonder if there are ways we can leverage TypeScript to help us enforce this, too - like create types for `type OneBasedIndex = number` and `type ZeroBasedIndex = number`, and then if you try and pass a `OneBasedIndex` to a method that expects a `ZeroBasedIndex`, it could potentially catch it at compile time... I haven't tried though.

Anyway I'll go ahead and merge - thanks for the fix!

TalAmuyal:
@bryphe another option is to have a class that holds the index with getters for 0-based and 1-based:

```Python
class Index(object):
  def __init__(self, zeroBased):
    self.zeroBased = zeroBased

  def getZeroBased(self):
    return self.zeroBased

  def getOneBased(self):
    return self.zeroBased + 1

  @staticmethod
  def fromZeroBased(self, zeroBased):
    return Index(zeroBased)

  @staticmethod
  def fromOneBased(self, oneBased):
    return Index(oneBased - 1)
```

PR #1094, opened 2017-12-09
Fix contributors link

chenglou:
Relative path leads to https://github.com/onivim/oni/blob/master/graphs/contributors which 404s


bryphe:
Ah, nice catch - thanks for the PR, @chenglou !

PR #1095, opened 2017-12-09
Add info on macOS PATH and key repeat

chenglou:
Not too sure if this is the right place to put it. Feel free to close this and put the instructions in the right place! Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,14 @@ You can also find install via a [zip archive](https://github.com/onivim/oni/rele
 - Double-click on the archive to expand
 - Drag `Oni.app` to the `Applications` folder
 
+To enable key repeat when pressing & holding a key in Oni, write the following in your terminal:
+
+```sh
+defaults write com.extropy.oni ApplePressAndHoldEnabled -bool false
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,6 +75,14 @@ You can also find install via a [zip archive](https://github.com/onivim/oni/rele
 - Double-click on the archive to expand
 - Drag `Oni.app` to the `Applications` folder
 
+To enable key repeat when pressing & holding a key in Oni, write the following in your terminal:
+
+```sh
+defaults write com.extropy.oni ApplePressAndHoldEnabled -bool false
+```
+
+> You may want to add /Applications/Oni.app/Contents/MacOS/ to your $PATH, so that from the console, you can open Oni via `Oni`
```
chenglou: Might not be ideal. The terminal tab is occupied when oni is open, until it closes

CrossR: Oni does have a "Add to Path" command, on MacOS at least, in the Ctrl+Shift+P panel I think. Should probably use that, since it is more easily removable and is done inside Oni.

chenglou: Done

bryphe:
Sweet, thanks for taking the time to add this! Having the key repeat shell command will be really helpful.

Placement is great for now - we do need to re-work / simplify our installation instructions, but that will come later.

Issue #837, opened 2017-10-28
Font size in .oni/config.js requires "px" after the number


FredChores:
![screen shot 2017-10-28 at 4 25 06 pm](https://user-images.githubusercontent.com/13815851/32139058-aaded766-bbfc-11e7-9515-9be45295de05.jpg)

Only the default font size works.  Attempting to change it, even to a value identical to default, causes the font to change to something else and borks the rendering.  The cursor element seems to show the proper rendering underneath it, I think.  This is with the current head on a current build of osx, and also with the most recent release.  I don't know if it's using the packed nvim or the one on my system.

CrossR:
Are you using "15" or "15px" as the font option?
If I miss out the "px" on the end I get something very similar to you.

I have noticed this before as well when not especially paying attention, and then with the messed up view it is awkward to swap back.

FredChores:
That did it, thanks.  I'll leave this open since it's an improvement than can be made.

CrossR:
The font config options could probably do with being improved in general.
I understand and quite like the JSON file for config changes, but I think there should be some other ways to interface with that config file as well.

Font could do with a menu option/bindable command that allows you to make it bigger and smaller, that auto updates the config option as this happens. I find myself quite often attempting to show people code and realising that I have size 12px font, and its easier to swap editor and zoom than jump into the config file and edit it there, and then go back and edit it again when I'm finished showing them.

I did have a quick look into adding that, but then I realised I don't think there is any way of saving back to the config file currently, so it seemed like something that should have a bit more thought than me just poking around and attempting to add it.

TalAmuyal:
Just my thought:
I too experienced the case of complaints of "small text" when showing code to co-workers.
Having a way of instantly scaling it would be great for this scenario.
However, I personally prefer it to be a temporary thing only.

bryphe:
Nice catch @CrossR on the `px` setting 👍 

Sounds like there are a couple issues here:
- It's easy to mess up the font-size setting. To make it more intuitive, perhaps we could add a default set of units if not specified - perhaps just using `px` if there are no units? I think that would solve the core issue. It also touches on #704, which is about making it easier to troubleshoot and understand our programmatic configuration.
- There isn't an easy way to set zoom level today - I moved this out to to #840 . It definitely sounds like a pain point! Agree with the difficulty too in saving back to the config - I added some thoughts in #840.

badosu:
Is it worthy to enable arbitrary numbers for size settings?

[...]

PR #1097, opened 2017-12-09
Add default font unit when needed

CrossR:
Tiny helper function that checks the font size a user enters and appends on `px` if no other unit is supplied, for #837.


bryphe:
Cool, thanks for adding this @CrossR ! Will make the configuration more robust and harder to mess up 👍 

I'm thinking about a way to centralize this long-term, like maybe have a way to run a 'sanitizer' or 'converter' when we load the config for certain properties, in case this ends up being a common pattern. But we can defer that until we see a pattern (like other config properties need similiar behavior).

Issue #412, opened 2017-04-28
Plugins: CSS Theming


extr0py:
Aside from just color schemes changes, I'd like to add plugin support for css styling / css skinning and theming.

This would open up the door for some very in-depth customization options, as well as slick animations. This would basically allow a plugin to include a stylesheet that is loaded - allowing the developer to customize font rendering, UI, etc.

This is a placeholder for now but will add more details as it materializes. Let me know if you have ideas / comments.

bryphe:
A good example of the scenario I'd like to support is something like https://atom.io/packages/activate-power-mode.

bryphe:
Also, add proof-of-concept by migrating the 'box-shadow' styling to a CSS file associated with onedark theme.

benjamin-rood:
I'd been interested in pitching in to this if possible? Regular vim colour styles are fugly in general, and make Oni look bad. I find dark colour schemes much harder to read than light ones.
It's clear that the functional direction for Oni is to have all the decorative customisation of an "electron" text editor but with (Neo) Vim tooling.

bryphe:
Hi @benjamin-rood ,

Would be great to have your help here! We're starting to add the ability to customize colors beyond plain vim colorschemes - there's an example here: https://github.com/onivim/oni/blob/master/extensions/theme-onedark/colors/onedark.json

And then the ability to specify them via config:
https://github.com/onivim/oni/blob/master/extensions/theme-onedark/colors/onedark.json

> It's clear that the functional direction for Oni is to have all the decorative customisation of an "electron" text editor but with (Neo) Vim tooling.

Well put, yes, this is exactly the direction we're heading 👍 

Beyond specifying colors, there are some next-level customizations we could do, by virtue of being an electron editor:
- CSS customization - this opens the door for things like animations/transitions (ie, power-mode type plugins), but is not applicable for all elements. For this, one approach is to simply specify a css file path in the theme's json file, and then we load it and add it to the DOM in our `ThemeManager`.
- Custom React components - we could add to our extensibility model ways to override React components (for example, specifying a new Cursor component). This would require a bit more thinking but allow arbitrary extension of the UI.

Let me know if you need any help getting started here - definitely appreciate your help! 

IMO, a good litmus test for this level of extensibility is if we can support a plugin like this:
https://atom.io/packages/activate-power-mode

If we can support something like that, we can likely support a wide breadth of customizations.

bryphe:
[...]

PR #1098, opened 2017-12-09
Re-activate git status bar item

Akin909:
It seemed the Git status bar fell into a state of disrepair as  it fell behind changes to the `Oni` api, I've tweaked it to work with the new api so it once again appears in the status bar.

I wasn't quite clear on what the exact status of this is as I know there was discussion in #486 regarding a general VCS api that could work with different systems, I don't know whether or not there might have been a plan regarding where to go next with the status bar plugin but as it seemed like a fairly isolated thing and could be retweaked in the future I've updated the plugin so it works again.

regarding config options I wasn't sure if that was in the plan (I haven't implemented that but if thats something I could add to config object?, i'd be happy to but I guess it leads to the bigger question of how plugins config options are going to be set in oni, e.g. a sub-object like vscode?)

### DIFF
--- a/vim/core/oni-plugin-git/index.js
+++ b/vim/core/oni-plugin-git/index.js
@@ -2,7 +2,6 @@ const fs = require("fs");
 const path = require("path");
 
 const activate = (Oni) => {
-
     const gitBranchIndicator = Oni.statusBar.createItem(0, -3);
 
     const pathIsDir = (p) => {
@@ -18,6 +17,7 @@ const activate = (Oni) => {
     };
 
     const updateBranchIndicator = (evt) => {
+        console.log('Update indicator ====================================: ', evt);
         pathIsDir(evt.bufferFullPath)
         .then((isDir) => {
             dir = null;
@@ -40,7 +40,15 @@ const activate = (Oni) => {
         });
     };
 
-    Oni.on("buffer-enter", updateBranchIndicator);
+    // Oni.on("buffer-enter", updateBranchIndicator);
+    console.log(' updateBranchIndicator: ', updateBranchIndicator);
+    console.log('Oni in [PLUGIN]!!!!!!!: ', Oni);
+    console.log(' gitBranchIndicator: ', gitBranchIndicator);
+
+    Oni.editors.allEditors.onBufferEnter.subscribe((evt) => {
+        console.log('event: ', event);
+        return updateBranchIndicator(evt);
+    });
 };
 
 module.exports = {
--- a/vim/core/oni-plugin-git/index.js
+++ b/vim/core/oni-plugin-git/index.js
@@ -1,54 +1,54 @@
-const fs = require("fs");
-const path = require("path");
+const fs = require('fs');
+const path = require('path');
+const { promisify } = require('util');
+const fsStat = promisify(fs.stat);
 
 const activate = (Oni) => {
-    const gitBranchIndicator = Oni.statusBar.createItem(0, -3);
+    const React = Oni.dependencies.React;
+    try {
+        const gitBranchIndicator = Oni.statusBar.createItem(1, -3, 'oni-plugin-git');
 
-    const pathIsDir = (p) => {
-        return new Promise((resolve, reject) => {
-            fs.stat(p, (error, stats) => {
-                if (error) {
-                    reject(error);
-                } else {
-                    resolve(stats.isDirectory());
-                }
-            });
-        })
-    };
-
-    const updateBranchIndicator = (evt) => {
-        console.log('Update indicator ====================================: ', evt);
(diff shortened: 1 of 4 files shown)


vim/core/oni-plugin-git/index.js
```diff
@@ -1,48 +1,80 @@
...
-                    reject(error);
-                } else {
-                    resolve(stats.isDirectory());
-                }
-            });
-        })
+const fs = require('fs');
+const path = require('path');
+const { promisify } = require('util');
+const fsStat = promisify(fs.stat);
```
bryphe: Cool! I hadn't tried this yet, was actually looking for an example (will be helpful for the file explorer because of all the node-callback APIs)

vim/core/oni-plugin-git/index.js
```diff
@@ -1,48 +1,80 @@
...
+  try {
+    const gitBranchIndicator = Oni.statusBar.createItem(
+      1,
+      -3,
+      'oni-plugin-git'
+    );
+
+    const pathIsDir = async p => {
+      try {
+        const stats = await fsStat(p);
```
bryphe: Nice, thanks for cleaning up and refactoring this to use `async`/`await`.

bryphe:
Awesome, this is great @Akin909 ! Thanks for investigating & bringing it back - I really like that you refactored and improved the code while fixing the issue.

> I wasn't quite clear on what the exact status of this is as I know there was discussion in #486 regarding a general VCS api that could work with different systems, 

This is perfect for now. The VCS API I imagine being further out (at least a couple months), and it will need some thinking about API design. I don't think the configuration options will be quite enough to support it, likely we will need a concept of a VCS 'provider' that gives us an API - like get modifications, create branch, notify when there is a change, etc. Still have to feel out what the surface area of that will look like, in order to support the primary version control systems (git/svn/hg)... 

But in any case, it's great to keep the status bar working for now - once we have the version control 'provider' API, I'd imagine we'd expose it on the Oni API surface, such that the statusbar plugin could ask for the branch / listen to a branch change.

I did notice one issue - if I open Oni in a directory that is not version controlled by git, I still see the 'fork' icon:

image

In that case, it'd be preferable to hide the whole thing IMO - the lone code fork could be a bit confusing.

Akin909:
@bryphe  no problem 👍 , I've removed the the lone fork symbol as well, I went back and forth on whether it communicated that the current file wasn't under version control or not, as I personally sometimes find some sort of indicator of that helpful, but I think the lone icon is probably more confusing than indicative

vim/core/oni-plugin-git/index.js
Akin909: Think node added it a couple of months ago and I've been taking every opportunity I get to promisify things, that way and avoid all the boiler plate that comes with doing it manually.

Akin909:
This is likely something for another issue/PR but I was working on trying to truncate the length of this component however there are a couple of issues I've run into. I'd want to do this inside of a `media query` and as It stands I tried creating a less file for the component and passing it a className which doesn't seem to work, I'm not sure how plugins are bundled aka if the less is read.

Another approach which did work is I passed the component a prop in the plugin's index.js and in Oni's status bar item component where it is rendered I look for the prop and if present apply a truncation class (to add  text-overflow ellipsis inside a media query which isn't possible with inline styles). I thought of adding the truncate option to the `createStatusBarItem` but this would require  changing `oni-api` to take another argument.

It's a minor point and not something I'm necessarily wanting to include here but as I often work with a split view or at least alternate frequently I noticed that the statusbar could get cluttered in which case I think plugin makers (future 😄 ) might find it difficult to apply styles to their components.

did you have any specific preferences for how that might happen given that `oni` already exposes some component creation methods but no styling methods?

bryphe:
Just tried it out, looks great!

> I'd want to do this inside of a media query and as It stands I tried creating a less file for the component and passing it a className which doesn't seem to work, I'm not sure how plugins are bundled aka if the less is read.

Ah ya, unfortunately we don't have a way to include CSS files at the moment. I was thinking about having a way to include CSS as part of themes with #412 

For this particular case, would there be any harm in adding `text-overflow: ellipsis` in general?

Also, for the status bar, I think it will be challenging if each individual status bar item has to factor this in - I believe the way some other editors like VSCode handle this is that they give a fixed width to the status bar items, and then will render them based on priority. This is nice because the individual status bar items don't have to concern themselves with the resize cases.

bryphe:
I'm good to bring this PR in - I opened #1105 to further discuss some strategies around handling overflow/sizing of status bar items.

Thanks for finding this issue and cleaning up this code @Akin909 (and thinking about the overflow scenarios) - really appreciate it!

Issue #982, opened 2017-11-18
"Find All References" produces wrong line number for jumps


hoschi:
For every entry in the list the line number is one line above the right line number. E.g. the first should be line 3 and not 2:
![2017-11-18-114042_1370x1384_scrot](https://user-images.githubusercontent.com/163128/32979420-500a47b2-cc55-11e7-9626-c85ed9b40fac.png)

Should be a zero index issue somewhere :wink:

CrossR:
https://github.com/onivim/oni/blob/105c707d156438deb85e95724a0df3365845f72b/vim/core/oni-plugin-typescript/src/FindAllReferences.ts#L22-L40

I'm not fully sure what that code does, but it looks to be the culprit.
Add 1 to line and col, do the look up and then remove 1 from both.

Getting rid of all the minus 1s seems to fix it for me though:
Left is with the fix and the cursor is on the very start of `Configuration`, right is without and its one line above and the col is off.

![image](https://user-images.githubusercontent.com/10038688/32985874-8489f598-ccbc-11e7-9795-c5ed75d7737c.png)

If I'm not missing something, I can submit the PR for it, but I assume I probably am?

bryphe:
Thanks for the detailed investigation, @CrossR!

This is confusing because there are 3 entities involved, and each has a different indexing behavior (in other words, uses either _zero-based_ line/columns, or _one-based_ line columns):
- __Language Server Protocol__ - uses _zero-based_ lines/columns
- __TypeScript Standalone Server__ - uses _one-based_ lines/columns
- __Neovim (quickfix)__ - uses _one-based_ lines/columns

So in the code above - what's happening is the request comes in via the _language server protocol_, so we have to convert it to a one-based index for the _typescript server_, and then the _typescript server_ sends us back one-based indices, so then we have to re-map it back to the protocol that the language client expects.

However, once we get those zero-based indices, we hand them off to Neovim in the quickfix window, without adjusting back to one-based indices (which Neovim expects). So it looks like the root issue is here:
https://github.com/onivim/oni/blob/6501a2b50d690ffe31e66aca71a5972c79e3e4eb/browser/src/Services/Language/FindAllReferences.ts#L49

When we're converting from `types.Location` to the quick-fix item format, we should be converting from zero-based to one-based indices.

And it looks like it is impacting other languages too (like if I 'Find all references' in a .less file): 
[...]

PR #1099, opened 2017-12-09
[WIP] Add language specific Regex tokens.

CrossR:
Still not sure on the config option name, and if we should expect the user to put [ ] around the regex.

Noticed this in #982, where in the screenshot @bryphe put up, the elements were not being correctly highlighted.

This fixes the token regex to allow for more fine grained setup, but there is still a further issue of what to do in the case of sub-classes or duplicated class names. Since we just highlight in neovim, if we had `property` and `property-longer`, both would be highlighted despite being different things. This can be seen by "Finding all references" on `Configuration` in `Configuration.ts`, and seeing how both `Configuration` and `IConfiguration` and more are highlighted, when they are independent of each other.

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -135,14 +135,17 @@ const BaseConfiguration: IConfigurationValues = {
     "language.css.languageServer.command": cssLanguageServerPath,
     "language.css.languageServer.arguments": ["--stdio"],
     "language.css.textMateGrammar": path.join(__dirname, "extensions", "css", "syntaxes", "css.tmLanguage.json"),
+    "language.css.tokenRegex": "[$_a-zA-Z0-9-]",
 
     "language.less.languageServer.command": cssLanguageServerPath,
     "language.less.languageServer.arguments": ["--stdio"],
     "language.less.textMateGrammar": path.join(__dirname, "extensions", "less", "syntaxes", "less.tmLanguage.json"),
+    "language.less.tokenRegex": "[$_a-zA-Z0-9-]",
 
     "language.scss.languageServer.command": cssLanguageServerPath,
     "language.scss.languageServer.arguments": ["--stdio"],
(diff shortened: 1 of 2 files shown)


bryphe:
Nice catch @CrossR ! Having the token regex be configurable per language definitely makes sense. Thanks for removing that hard-coding.

>  Since we just highlight in neovim, if we had property and property-longer, both would be highlighted despite being different things.

Ya, good point. I wonder if we can improve the 'search string' we use in this case:
https://github.com/onivim/oni/blob/36b144fc784511cb7ef5e84fa2c636c9c07ce904/browser/src/Services/Language/FindAllReferences.ts#L58

Perhaps we can refine that search query somehow to get better results?

CrossR:
>Perhaps we can refine that search query somehow to get better results? We could also consider manage the highlighting ourselves, if that's not feasible.

I think that would end up being pretty awkward, there is the simple case of check for word boundaries either side that would help it some what, but its also pretty common to have brackets or `:` or `=>` immediately following it, which would cause it to fail.

I see there is already some parts of Oni that use `nvim_buf_add_highlight`, could we use that? Looks like Oni has some code that groups them up and sends them for a line, would it be possible for an entire buffer? Not sure how performant that would be though. Though that raises other issues as that only highlights a single buffer whereas a search is across all buffers as far as I know.

bryphe:
> I think that would end up being pretty awkward, there is the simple case of check for word boundaries either side that would help it some what, but its also pretty common to have brackets or : or => immediately following it, which would cause it to fail.

Makes sense. Yes, perhaps managing the highlighting directly in Oni would give better results. We can directly add highlights (there's an `updateHighlights` method that the text mate highlighter uses, but we could maybe add a simpler API, like `setKeywordHighlight(...)`.

The only other thing search gives us is the ability to quickly navigate between entries with `n` and `N` - we might want an alternative to that at the minimum. I've been thinking about something like that for errors, too.

bryphe:
This PR looks great though - I'll bring this in, and we can track the search result refinement separately.

Thanks @CrossR for fixing this bug, and discovering/investigating the highlight issue too!

PR #1106, opened 2017-12-12
Performance: Remove git branch check on buffer changed

bryphe:
__Issue:__ We are checking branch status on every buffer changed event (every keypress),  which, even though kicking off the git process is async, still has non-trivial overhead:

perf-regression-git

It appears that the git spawn can take a significant amount of time - we want to ideally keep operations well under 16ms to hit 60ms, but because this exceeds that, it can cause jerkiness in rendering and responsiveness.

@Akin909 - was there a reason the `onBufferEnter` event wasn't sufficient, and we needed to hook the change event? If there is a case we need it, I'm wondering if there is a less noisy way for us to get the update?

### DIFF
--- a/vim/core/oni-plugin-git/index.js
+++ b/vim/core/oni-plugin-git/index.js
@@ -74,9 +74,6 @@ const activate = Oni => {
     Oni.editors.activeEditor.onBufferEnter.subscribe(
       async evt => await updateBranchIndicator(evt)
     );
-    Oni.editors.activeEditor.onBufferChanged.subscribe(
-      async buf => await updateBranchIndicator(buf)
-    );
   } catch (e) {
     console.warn('[Oni.plugin.git] ERROR', e);
   }


Akin909:
@bryphe I was trying to cover cases where a git branch changed say in the terminal  and a user returned to Oni, that the branch would be updated, I wasn't sure that the bufenter event would cover this. Didn't realise the impact it would have on performance 😟 . I think there are other less expensive hooks that could be used like `FocusGained` etc. but I don't believe theres a subscription hook in oni for that. I think doing simply doing without the buffchanged till its clear a replacement is needed is fair

bryphe:
> @bryphe I was trying to cover cases where a git branch changed say in the terminal  and a user returned to Oni, that the branch would be updated, I wasn't sure that the bufenter event would cover this. Didn't realise the impact it would have on performance 

Ah ok, thanks for the info! I like your suggestion to put a `FocusGained` or `FocusLost` event hook somewhere to accomplish this. I added #1109 to track adding this to our API.

> Didn't realise the impact it would have on performance 😟 .

No worries! Performance is fickle unfortunately, and I believe on OSX this would not be as apparent because there is less visible overhead (at least when I benchmark on my machine). The events that occur frequently during typing like `BufChanged` are on the critical path so we have to watch out for those, in order to maintain responsiveness. It'll be something we should document at some point, because it'd be easy for plugin authors to hit, too!

Issue #1100, opened 2017-12-11
Cannot open multiple windows


Akin909:
This is very likely to be due to a misunderstanding on my part but it seems I can't open multiple windows with Oni. Is that supposed to be the case, do I need to set a config option or is that a known bug or future feature. 

I couldn't find an issue relating to that, except #1008 which isn't quite what I mean. I'm running oni on a mac and using `cmd-n` or double clicking the `Oni` icon does not allow me to open a new window, I know that there's something about handling windows in `main.ts` with electron but I'm not sure if that currently should be permitting multiple windows or not.

bryphe:
Hmm, I just tried the same with Atom and VSCode - here's how they work:
- `cmd+n` creates a new file, not a new window. We don't have a binding for that today, but would make sense to have it. 
- None of the apps I tried supported double-clicking on the dock icon to open a new window, but some supported right-click and then 'New Window'. Oni doesn't have that today, but we could look at adding that as an enhancement. This API would be a good starting point: https://github.com/electron/electron/blob/master/docs/api/app.md#appdocksetmenumenu-macos

There's also an issue where if you close Oni, and then try and go to a menu item, you hit #484 .

Would adding a 'New Window' option to the dock menu + command palette work for you to accomplish this? Seems like that would fit well with how other apps work on OS X

Akin909:
@bryphe apologies seems I got my mac shortcuts confused. You're quite right re cmd n and the double clicking. Essentially any mechanism which leads to having multiple Oni windows (one project open in one instance another in the second) is what I was alluding too so a new window option would be excellent with the potential to bind this to a key, would a new window result in issues I noted there is an array in `main.ts` to keep track of open windows but also some code which seems to be aimed at maintaining a single instance?

Akin909:
fixed by #1111

Issue #483, opened 2017-06-08
cmd+q doesn't quit


endlist:
On MacOS, the expected behavior for all apps is that Cmd+Q will fully close an app, including open windows.  This doesn't work for Oni, I have to either right click on the icon in the dock or use the menu bar to fully quit.

bert88sta:
Keyboard.tsx has a block of code I wrote that discards all command prefixed
key chords. As of right now we're not sure if there's going to be a
configuration value or if we'll whitelist certain common non-vim bindings

Issue #67, opened 2016-12-09
Multi-monitor support


bryphe:
A scenario that frequently occurs in a multiple-monitor setup at work is that I'd like to expand VIM to encompass multiple displays, and use the goodness of the easy window splitting / window navigation across monitors. Gvim on windows, at least, doesn't seem to handle this very well - I imagine the terminal clients might be better if the terminal emulator handles resize/multiple monitors, but in my case, it doesn't fit my workflow.

With Electron, it's relatively easy to change the BrowserWindow dimensions to encompass multiple screens (or alternatively, create multiple BrowserWindows that can communicate with other).  

It's easy to get the information about all displays:
```
    require("electron').screen.getAllDisplays()
```

And it's also pretty easy to size the browser window:
```
        require('electron').remote.getCurrentWindow().setPosition(x, y)
        require("electron").remote.getCurrentWindow().setSize(.., ..)
```

What would be helpful in terms of workflow would be a command that is like "MonitorExpandLeft, MonitorExpandRight, MonitorExpandDown, MonitorExpandUp". This would size the window to expand to the boundary of the window if it hasn't already, or expand it out to the next window. This could be bound to something like <C-W> plus arrow keys by default.

This would be somewhat gated by #20 since we need a nice way to express those commands, but we could have another hacky input added until we get that fixed..

bryphe:
This is a feature I would love to have. This should allow for navigation between opened Oni windows, using `<C-w>l`, etc. The `WindowManager` can already handle and detect this case - the challenge here is bubbling that up, finding the next Oni window, and focusing it. The `main` process knows about all the active windows, so it'd be natural to send an `ipc` call to it to let it handle.

Testing with multiple Oni processes will be non-trivial here, but if we can have unit tests covering the navigation cases (figuring out which window is 'left', 'right', etc), that is sufficient.

CrossR:
The bulk of this was fixed out in #1747. I'm going to make a second issue to track the remaining items I mentioned in the end of my PR.

bryphe:
@CrossR - thanks for the fix here! It's awesome to have this in. Don't forget to claim the bounty on it! 👍 (Either bountysource or opencollective - whatever you prefer)

PR #1111, opened 2017-12-12
Add new window option

Akin909:
@bryphe following the discussion in #1100 and the link you posted I had a go at adding the functionality of opening a new window to the top menu bar (and to the dock for mac os). Tbh I was little concerned that implementing this would be more involved, i.e. managing the state of multiple windows etc. but it all seems to work fine (which I believe is largely due to your foresight in handling focusing of instances etc. in `main.ts`).

Things that it currently lacks that I imagine might become issues down the line are 
1. *mappability* 
2. an easy way to focus instances
3. cmd-Q and I imagine its alternate OS equivalents kills all windows, subjectively I would have expected each to be deleted one at a time - *looking into this*

Tbh I personally think those are both nice to haves and the menu or clicking each instance works well, and mac os, linux and i'd imagine windows all have several window manager programs, but perhaps thoughts for the future

### DIFF
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -3,7 +3,7 @@ import * as path from "path"
 import { app, BrowserWindow, ipcMain, Menu, webContents } from "electron"
 
 import * as Log from "./Log"
-import { buildMenu } from "./menu"
+import { buildDockMenu, buildMenu } from "./menu"
 import { makeSingleInstance } from "./ProcessLifecycle"
 
 global["getLogs"] = Log.getAllLogs // tslint:disable-line no-string-literal
@@ -57,7 +57,7 @@ if (!isDevelopment && !isDebug) {
     })
 }
 
-function createWindow(commandLineArguments, workingDirectory) {
+export function createWindow(commandLineArguments, workingDirectory) {
     Log.info(`Creating window with arguments: ${commandLineArguments} and working directory: ${workingDirectory}`)
 
     const webPreferences = {
@@ -129,11 +129,9 @@ function updateMenu(mainWindow, loadInit) {
     const menu = buildMenu(mainWindow, loadInit)
     if (process.platform === "darwin") {
         // all osx windows share the same menu
-        const dockMenu = Menu.buildFromTemplate([
-            { label: "New Window", click() { console.log("New Window") } }
-        ])
-        mainWindow.dock.setMenu(dockMenu)
         Menu.setApplicationMenu(menu)
+        const dockMenu = buildDockMenu(mainWindow, loadInit)
+        app.dock.setMenu(dockMenu)
     } else {
         // on windows and linux, set menu per window
         mainWindow.setMenu(menu)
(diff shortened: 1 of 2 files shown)


bryphe:
Awesome, thanks for looking at this @Akin909 ! Impressive turnaround time on the PR 😄

I just tried it out, and it's working great!

Agree with your list of items for 1-3, and that they could potentially be handled down the road. It's great that we have the 'New Window' option. 

> cmd-Q and I imagine its alternate OS equivalents kills all windows

I believe this is actually the 'right' behavior on OS X - there was some discussion around this on #483 . And when I tried on other apps (atom, vscode), it did that full-close behavior.

For 2, one item I was looking at was this: #67 

It's titled "multi-monitor support" but really it's just about quickly switching open instances of Oni. I thought it'd be cool if we could extend the window navigation commands (like `<C-w>h`) such that, when they hit a border, they would look to see if there is an Oni window present to the left/right/top/bottom, and navigate to that instance. Another option might be to have a quick-open style menu for open instances that could be navigated. But as you mentioned each platform also has their own ways to navigate windows, so it's a nice to have.

Nice work! I'll bring this in once the builds are complete.

Akin909:
@bryphe 👍 . Re. windows closing if that standard behaviour i'll leave that be.
As for window jumping that would be a really cool feature to use vim-esque key maps to move between windows, I saw somewhere in the docs that windows have x,y positions perhaps looking at those i.e comparing the current positions x to all the other windows in the array and if `otherWindow.x > currentWindow.x` then focus on that instance then map that to the appropriate key, just a thought. 
Another thing I noticed was that you can add recent files to the menu icon in mac and windows thought it isn't clear how to populate the recent files from the docs

bryphe:
> As for window jumping that would be a really cool feature to use vim-esque key maps to move between windows, I saw somewhere in the docs that windows have x,y positions perhaps looking at those i.e comparing the current positions x to all the other windows in the array and if otherWindow.x > currentWindow.x then focus on that instance then map that to the appropriate key, just a thought.

Definitely! The browser window has a `getBounds` method which gives us the positions, and the `main.ts` has that list of all the browser windows.... so it'd be a matter of figuring out how to go from one to the other.

I started on some ideas here a looong time ago - in the main process, I had this code to focus different instances:
https://github.com/onivim/oni/blob/e0c84b9b38a0902956d1859fbd9c77ee7e7f7892/main/src/main.ts#L22

And then the entry point, from the browser-side, was really simple here - just sending the `ipc` message:
https://github.com/onivim/oni/blob/e0c84b9b38a0902956d1859fbd9c77ee7e7f7892/browser/src/Services/MultiProcess.ts#L10

So I think it wouldn't be too crazy to implement the navigation - we'd just have to tell the main process which direction we want to go, and then in the main process, see if there is an Oni window that matches (like, if we move right, check for that `otherWindow.x > currentWindow.x` as you mentioned). Doesn't seem like it'd be too far out!

> Another thing I noticed was that you can add recent files to the menu icon in mac and windows thought it isn't clear how to populate the recent files from the docs

Ah interesting, I guess this might be a case where having those 'persisted' settings would be helpful - we could persist the last X opened files/folders, and show them there? I wonder if there is something special about them, or if they would be additional menu items we'd add to our dock menu?

bryphe:
Anyway I'll go ahead and bring this in as it addresses #1111 . Thanks again for your work on this, @Akin909 !

PR #1113, opened 2017-12-12
Fix #1064 - Large files crash in Oni

### LINKED ISSUE #1064: Large files open and then crash
I have a few largish text files (60+ MB), that I open, see the contents without any colourscheme etc, and then it is closed a moment later and replaced with the "Uh Oh Neovim isn't installed!" screen.
If I instead launch Oni and then open the file after Oni is open, the files load fine.

I assume the flashing with no colourscheme is deliberate to make loading faster, since I see it on an all files.
The crashing I'm assuming it is due to a timeout or perhaps Neovim not responding for a moment as the file loads?

bryphe:
__Issue:__ When trying to open a large file, the Neovim process would crash, and show our generic error message (which is confusing...).

__Defect:__ When we open a file of a supported type for our language service, we try and read the full content to send with the `textDocument/didOpen` message. This was blocked in the `NeovimInstance` on the buffer update side,  but not in other places where we may directly call `getLines`.

__Fix:__ Don't engage the language service or start `textDocument/didOpen` if the file line length goes past a configurable value (`editor.maxLinesForLanguageServices`). In addition, add a regression test, and a warning in the `getLines` method as defensive fixes to make it easier to catch this in the future.

### DIFF
--- a/browser/src/Editor/BufferManager.ts
+++ b/browser/src/Editor/BufferManager.ts
@@ -82,6 +82,10 @@ export class Buffer implements Oni.Buffer {
             end = this._lineCount
         }
 
+        if (end - start > 2500) {
+            Log.warn("getLines called with over 2500 lines, this may cause instability.")
+        }
+
         const lines = await this._neovimInstance.request<any>("nvim_buf_get_lines", [parseInt(this._id, 10), start, end, false])
         return lines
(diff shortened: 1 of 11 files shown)


bryphe:
Fix #1064

PR #1114, opened 2017-12-12
Status bar truncation

### LINKED ISSUE #1105: Status Bar: Investigate better handling of overflow scenarios
Following from @Akin909 's notes in #1098:

> This is likely something for another issue/PR but I was working on trying to truncate the length of this component however there are a couple of issues I've run into. I'd want to do this inside of a media query and as It stands I tried creating a less file for the component and passing it a className which doesn't seem to work, I'm not sure how plugins are bundled aka if the less is read.

> Another approach which did work is I passed the component a prop in the plugin's index.js and in Oni's status bar item component where it is rendered I look for the prop and if present apply a truncation class (to add text-overflow ellipsis inside a media query which isn't possible with inline styles). I thought of adding the truncate option to the createStatusBarItem but this would require changing oni-api to take another argument.

> It's a minor point and not something I'm necessarily wanting to include here but as I often work with a split view or at least alternate frequently I noticed that the statusbar could get cluttered in which case I think plugin makers (future 😄 ) might find it difficult to apply styles to their components.

We should think about ways to handle overflow in the status bar. There are two cases that are problematic:
- __Too many items__ (clutter) - if the status bar is filled up with lots of different pieces, it won't look good.
- __Too small space__ - If the window is small, even with a few status bar items, we don't gracefully handle that - and you end up with overflow problems (like the line, column here):

image

There are a few different possible solutions:
- Leave the sizing/styling up to the owner of the status bar item. This is the least work for Oni, as it defers the problem to whoever creates the status bar item. However, it means that there is potentially duplicate work across status bar items, and each plugin might have a different way of solving this.
- Let Oni manage the set of status bar items. VSCode does something similiar to this strategy - each statusbar item has a priority, and then depending on how much space is available, some status bar items may be hidden, based on priority.

Akin909:
@bryphe I've had a go at implementing a truncation system (fixes #1105), the approach i've taken is definitely debatable so let me know if you have suggestions re tweaks. 

trunc-demo
In general I use the priority to determine how each component responds to screen size changes, priority 0 don't change till things are very squished, all the others move from initially truncating to disappearing, if applicable.

I factor in count and use it to alter a components priority so a busy left or right section will reduce the priority of the components.

 Its definitely not a 100% as depending on the make up of a status bar item it'll hard be to tell what sort of css is necessary to truncate the element if it isn't all text based, an easier solution would be just disappearing unless the content is definitely known like the file name or git branch.

also since its a tiny bit jarring might be worth a small transition using the `react-transition-group`


Akin909:
@bryphe done as much as I could personally think to do with this, think it might be a situation of learning as more components appear but the general idea is 4 priority classes:

* 0 - do no truncate till very small viewport (400px)
* 1 - truncate after half screen and disappear if smaller
* 2 - disappear at just under 2/3rds
* 3 - disappear over 2/3rds

priority is assigned in creation of the component as before and a classed assigned based on that though there is one bit of processing before which basically checks the count of right and left status items and increases/aka lowers the priority depending on how busy with the aim that truncation and disappearance happens earlier. As long as there isn't a situation where everything is being set to 1 since 'in a version of the/the future theres components vying for space this works'

I abandoned the transitioning idea because I had a look at webstorm, atom and vscode and it seems thats not done in any of those on top of which I'd *personally* feel like the transitions were just visual noise

browser/src/UI/components/StatusBar.less
```diff
@@ -72,4 +73,48 @@
...
+                    display: none !important;
+                }
+            }
+
+            .item-priority-0 {
+                .status-overflow(400px);
+            }
+
+            .item-priority-1 {
+                .status-overflow(650px, 35vw);
```
bryphe: Is there a reason we specify a max for priority 1, but not for the other cases? It's not clear to me what makes this one different.

bryphe:
Thanks for the details and the notes on the levels, @Akin909 ! The priority you gave seems reasonable to me. The only one I wasn't sure about was priority 3 - do we show that over 2/3rds?

> As long as there isn't a situation where everything is being set to 1 since 'in a version of the/the future theres components vying for space this works'

Potentially we could resolve this by having an implicit priority - like the order they were registered (likely built-in status bar items would be registered first, and take precedence).

> I abandoned the transitioning idea because I had a look at webstorm, atom and vscode and it seems thats not done in any of those on top of which I'd personally feel like the transitions would just be visual noise

Seems reasonable to me! I agree, I think it would be extra noise.

The only downside I see with the approach is it makes some assumptions based on the screen size (which can be tough since people have varying monitor sizes / resolutions / orientations). At the very least, it might be worth using a non-pixel unit, like a viewport unit: https://tutorialzine.com/2015/05/simplify-your-stylesheets-with-the-magical-css-viewport-units (which could more cleanly express full viewport, 2/3 viewport, etc). It looks like you're already using them for the max width in some cases - can we use them in place of the pixel measurements too?

IMO the 'ideal' algorithm to solving this would be:
- When the status bar size changes, we'd measure the left and right container size
- While we still have space available, we'd add status bar items _by priority_, until we've reached the container size.

The nice thing about that algorithm is it's really independent of making any assumptions about screen size / coupling priorities to screen percentages, and it means we can avoid needing things like the `adjustPriority` method - that algorithm would handle those cases and potentially other corner cases by rendering as much as it can. At least conceptually it's simple - measuring the elements / dealing with widths like that is a pain, and it would need to be purely in JS as the media queries wouldn't handle that - so I think it would be more work than the proposed approach, although potentially more robust in the general case. 

But I'm okay with the approach you've taken though, as it is definitely better than what we have today!👍 The main thing I'd like to see though is if we can decouple using the `px` sizes and use some sort of viewport unit, so that it generalizes better across screen sizes. Otherwise I'm good to bring this in, and we can always iterate on it as we see fit 😄  Thanks for investigating a fix for this @Akin909 , appreciate it!

browser/src/UI/components/StatusBar.less
Akin909: The practical effect of that line is that for example the filename component a '1' truncates earlier rather than dropping to half its size. The max here gives these items more space  before they collapse, the idea I was going for was that priority could be for larger components that we wanted on the screen for as long as possible

Akin909:
@bryphe `vw` 's are more generalisable, 🤦‍♂️ , i'll give using those a go instead. Re. the solution you describe does sound more robust, I read [this](https://twitter.com/JoshWComeau/status/940029421682962432) yesterday a tweet about a react component which calculates its with as it grows or shrinks and passes it down to its child components. Also whilst working on this I had a question which is a bit tangential for `oni`,  I wonder if you've had a look at any `CSS-in-JS` solutions, like `styled-components`,  they tend to allow you a lot of flexibility with what information is available in your css, aka any react prop can be used to determine how styles can respond, just curiosity on my part.

Akin909:
@bryphe, I'd never tried before and it seemed like it should work but having tried it a bunch of different ways as far as I can see using `vh` and `vw` in a media query doesn't work or rather has not worked at all for me having gone through several posts, SO answers etc.I haven't yet seen an example using vh or post using vw or vh in a query.

It does work with em which scale based on font size, which is more fluid than a static px value also, on thinking about it this particularly approach would only really be aimed at dealing with small screen sizes which are equivalent to certain px values essentially

bryphe:
> I haven't yet seen an example using vh or post using vw or vh in a query.

Bummer, that's too bad 😦 

> It does work with em which scale based on font size, which is more fluid than a static px value also, on thinking about it this particularly approach would only really be aimed at dealing with small screen sizes which are equivalent to certain px values essentially.

That's true, that makes sense. I guess for the time being we could use the pixel values for the smaller screen sizes, and always render all items for the larger sizes.

> EDIT: I'm also looking into the more robust solution you mentioned, whereas this is more of a band-aid

Cool, I think this makes sense as a long-term solution - it's more generalized and avoids some of the caveats of the current approach.

> Re. the solution you describe does sound more robust, I read this yesterday a tweet about a react component which calculates its with as it grows or shrinks and passes it down to its child components.

Awesome! I think in our case, we wouldn't want the status bar items themselves to decide whether or not to render - it should be up to the container, since the container knows about the sizes and relative priorities of the children, whereas the children ideally shouldn't need to know that information. But we could potentially use a similiar approach.

> I wonder if you've had a look at any CSS-in-JS solutions, like styled-components, they tend to allow you a lot of flexibility with what information is available in your css, aka any react prop can be used to determine how styles can respond (this would allow for dynamic styles inside of a media query for example), just curiosity on my part.

I actually haven't used these before - happy to try it out if it can simplify things for us!

Akin909:
@bryphe following our discussion re. approaches to the bar I've started work on an alternate track i.e. the one you suggested. I'm using resize observe to keep track of the sizes of the parent sections  left and right statusbar sections as well as the children, then using the priority to remove children.

An question I have is whether or not the status bar is to be split up into three as it is right now as each container need s a width in order for the observer to track it and I can work around the usually empty center container but is it something you'd like to have in the long run as opposed to just a growing left and right which each have 50% of the screen real estate

Akin909:
@bryphe managed to get this working as we discussed and following the styled components PR I've tweaked most of the status bar components to also use styled components, I would have converted all of them however an issue I ran into is with the `.loaded` class which is set via redux globally, I think to have this work with `styled components` rather than a class a redux prop would have to be set and then passed to all styled components needing the loaded class. You could have styled components' css helper and then check if truthy render the css chunk.

I'm happy to do this or leave it to you, either way is fine, I haven't done it here already as it's i think outside the scope of what I was trying to do here. Re the discussion in #1134 I'm  also happy to take some of it on, and perhaps leave some to you so you can decide how you'd like to integrate the library etc.

browser/src/UI/components/StatusBar.tsx
```diff
@@ -8,63 +9,115 @@ import { connect } from "react-redux"
...
+interface IChildDimensions {
+    direction: string
+    width: number
+    id: string
+    priority: number
+    hide: boolean
+}
 
+const StatusBarComponent = withProps<{ maxWidth?: string }>(styled.div)`
+    white-space: nowrap;
```
bryphe: Cool, thanks for porting these over!

browser/src/UI/components/StatusResize.tsx
```diff
@@ -0,0 +1,142 @@
...
+const StatusbarSection = withProps<Section>(styled.div)`
+    flex: 1 1 auto;
+    display: ${({ count }) => count ? `none` : `flex`};
+    flex-direction: row;
+    height: 100%;
+    max-width: 50%;
+    justify-content: ${props => props.direction};
+`
+
+class StatusBarResizer extends React.Component<Props, State> {
```
bryphe: Cool, I like that you factored out this responsibility to a component

vim/core/oni-core-statusbar/index.js
```diff
@@ -4,10 +4,10 @@ const rgb = (r, g, b) => `rgb(${r}, ${g}, ${b})`
 
 const activate = (Oni) => {
     const React = Oni.dependencies.React
-
-    const workingDirectoryItem = Oni.statusBar.createItem(0, -1, "oni.status.workingDirectory")
-    const lineNumberItem = Oni.statusBar.createItem(1, -1, "oni.status.lineNumber")
-    const modeItem = Oni.statusBar.createItem(1, -2, "oni.status.mode")
+    const { workingDirectory = 0, linenumber = 0, mode = 0 } = Oni.configuration.getValue("statusbar.priority")
+    const workingDirectoryItem = Oni.statusBar.createItem(0, workingDirectory, "oni.status.workingDirectory")
```
bryphe: If we're making the priority configurable - maybe we don't even need to pass the priority in as part of the API now? If the configuration is the source of truth, maybe we could just use that. In addition, we could leverage the 'id' that we pass in here - like `statusbar.priority': { "oni.status.workingDirectory": 1}`. The nice thing about that is it makes it easy for plugins to then also be prioritized in the same system.

browser/src/UI/components/StatusResize.tsx
```diff
@@ -0,0 +1,142 @@
...
+        return (
+            <StatusbarSection
+                direction={direction}
+                count={count < 1}
+                innerRef={(elem: Element) => (this.elem = elem)}
+            >
+                {containerWidth !== undefined &&
+                    React.Children.map(children, (child: any) => {
+                        const current = this.state.children[child.props.id]
+                        return React.cloneElement(child, {
```
bryphe: I wasn't quite sure why we need to clone the element here?

browser/src/Services/Configuration/DefaultConfiguration.ts
```diff
@@ -180,6 +180,13 @@ const BaseConfiguration: IConfigurationValues = {
 
     "statusbar.enabled": true,
     "statusbar.fontSize": "0.9em",
+    "statusbar.priority": {
```
bryphe: I like you specified this in the configuration. Looks good! I left a comment below that perhaps with this, we don't need to specify the priority anymore with `createItem`? It seems like right now, there need to be two ids associated with a statusbar item - one in the configuration, and a separate one we pass to `Oni.statusbar.createItem` - I wonder if we could consolidate those?

browser/src/UI/components/StatusResize.tsx
```diff
@@ -0,0 +1,142 @@
...
+            },
+        }), this.resize)
+    }
+
+    private resize = () => {
+        const { children, containerWidth } = this.state
+        const childArray = Object.values(children)
+        const sorted = childArray.sort((prev, next) => prev.priority - next.priority)
+
+        // Loop through components sorted by priority check if component can be added without
```
bryphe: Awesome!

browser/src/UI/components/StatusResize.tsx
Akin909: @bryphe cloning the element allows me to add to its properties inside of where its rendered, so that I can clone the element add a method from its parent to the clone as a prop and reference that method in the child, for example inside of the clonedChild aka `statusBarItem` I am able to use `passWidth` a reference to the parents passWidth method or add a hide prop which toggles visibility based on the parent's state/logic. 

It's crucial to this working and as far as I know the way to attach data or functions to react.children given that here we render an unspecified number of children and they all need this behaviour and the only point at which this data is available is here.

bryphe:
> I would have converted all of them however an issue I ran into is with the .loaded class which is set via redux globally, I think to have this work with styled components rather than a class a redux prop would have to be set and then passed to all styled components needing the loaded class. You could have styled components' css helper and then check if truthy render the css chunk.

I think it's okay for now to leave the `.loaded` out - that was a quick transition-in animation, we can probably find another way to handle it later 😄 It's not critical by any means.

> Re the discussion in #1134 I'm also happy to take some of it on, and perhaps leave some to you so you can decide how you'd like to integrate the library etc.

It would be great to have your help with the remaining conversion for #1134! I think we can try and scope it - perhaps just converting the .less files over, and not trying to convert the inline styles for now, might make it easier w/ less chance of a performance regression. Glad you found that way to set inline styles though via `styled-components` - really appreciate your help!

> EDIT: Forgot to mention that I also added configuration options for the status items so users can change these. I found whilst working on this that I very much quibbled over what should or shouldn't show which I imagine will be the case for the user. If you're in agreement with this addition it'll need documentation which I can add.

The configuration looks good to me - I like that we have a configurable way to set the priority. It seems like it would be extensible for plugins too. I left some comments inline in the PR, the only consideration I had was that potentially we could consolidate the `id` we pass to create the status bar, with the id we reference in the `statusbar.priority` dictionary. In addition, we may not need the `priority` setting to be passed in to `createItem` - if they are always just going to look up the priority in the config (which they should, if they are good statusbar citizens that respect the users config), then we can just look it up for them and not need them to specify in the `createItem` call.

vim/core/oni-core-statusbar/index.js
Akin909: @bryphe do you mean deleting it from the createItem method, and the api package and then adding the priority in the statusbar component somewhere like get config value etc.? (for clarification) ^^

browser/src/UI/components/StatusResize.tsx
bryphe: Got it, I see, so this is used to augment the set of properties. Makes sense - thanks for the explanation!

bryphe:
Just a FYI - following from the `oni-api` changes: The `oni-api` [npm module](https://www.npmjs.com/package/oni-api) was published with the API changes at `0.0.10`. So you'll need to update the package.json and run` yarn install` to pick those up.

Akin909:
@bryphe just wanted to clarify whether it takes some time to kick in, as at the moment the changes to the `oni-api` don't seem to have taken effect unfortunately when I re-install the module

bryphe:
Ah sorry, there were actually some problems with the `oni-api` build - the `index.d.ts` and `index.js` weren't being overwritten because they weren't checked in, and then they weren't being included in the package because they were `gitignore`'

Only took 8 releases or so but seems like it's working now 😉 You'll need to upgrade to `oni-api@0.0.17`

Akin909:
@bryphe 👍 thanks for looking into that, seems to have resolved the typing issues.

I'm having trouble with the tests as the `"status.priority"` section I added to the config is being looked for by the status bar items and im not sure how to add it to the new test `config.js` (thanks for adding that btw) I see the env var, are you setting one locally somehow?

bryphe:
> @bryphe 👍 thanks for looking into that, seems to have resolved the typing issues.

Sure thing! I'm just looking at the branch now - the status bar looks really cool. I like that we're using the theme colors now! And the resize logic seems to be working well - much nicer experience than we had before.

> I'm having trouble with the tests as the "status.priority" section I added to the config is being looked for by the status bar items and im not sure how to add it to the new test config.js (thanks for adding that btw) I see the env var, are you setting one locally somehow?

Hmm, looking into this now.

bryphe:
I noticed when I close and re-open Oni, I see the statusbar in some weird states - not sure if this you've run into this too:
image

browser/src/Services/Language/LanguageClientStatusBar.tsx
```diff
@@ -9,7 +9,10 @@ import * as React from "react"
 
 import * as Oni from "oni-api"
 
+import styled from "styled-components"
+
 import { Icon } from "./../../UI/Icon"
+import { configuration } from "./../Configuration"
```
bryphe: Using the global configuration object is actually what is causing the test failure. We pass a `MockConfiguration` to the LanguageManager, but that doesn't get passed to the status bar object here. Instead of using the global configuration object, we should pass in a `Configuration` to the constructor here. That will help insulate this class from global state, and get the test to pass 😄

bryphe:
Also I found the issue with the unit test - the `LanguageClientStatusBar` was using the global configuration object, which makes it hard / awkward to test. For the `LanguageManager` tests, we pass in a configuration object - that should be passed down to the `LanguageClientStatusBar`.

So instead of this in `LanguageManager`:
```
    private _statusBar = new LanguageClientStatusBar()
    private _currentTrackedFile: string = null

    constructor(
        private _configuration: Oni.Configuration,
        private _editorManager: Oni.EditorManager,
    ) {
```

We should have this:
```
    private _statusBar: LanguageClientStatusBar
    private _currentTrackedFile: string = null

    constructor(
        private _configuration: Oni.Configuration,
        private _editorManager: Oni.EditorManager,
    ) {
      this._statusBar = new LanguageClientStatusBar(this._configuration)
```

Hope that helps!

Akin909:
@bryphe thanks for the heads up been stuck on that for a little while, had no idea where it was failing 👍 . 

Re the weirdness of the statusbar I haven't seen the behaviour in the screen shot though I haven't tried it with the file explorer open not that I can think of why it would be causing an issue. It's a bit of a head scratcher as you're segments are completely overlapping could you possibly post a screenshot looking at the elements, I'm not sure if theyre crossing over each other 😕 , been using it in my local build for a week now so though I'd got all/any bugs, not sure how to reproduce that.

Akin909:
@bryphe splitting the id's and not passing in priorities seems to be fairly problematic I've found a deeper problem with this or rather one which leads back to an original concern which is that the `statusb.createItem` function needs a priority and id and an alignment, if we dont pass in these as arguments at the point of creation then they need to be generated from the config, which createItem does not have access to.

```typescript
//Current implementation
class StatusBar implements Oni.StatusBar {
    private _id: number = 0

    public getItem(globalId: string): Oni.StatusBarItem {
        return new StatusBarItem(globalId)
    }

    public createItem(alignment: StatusBarAlignment, globalId?: string): Oni.StatusBarItem {

// Potential implementation
class StatusBar implements Oni.StatusBar {
    private _id: number = 0

    constructor(private _config: Oni.Configuration){}

    public getItem(globalId: string): Oni.StatusBarItem {
        return new StatusBarItem(globalId)
    }

    public createItem(alignment: StatusBarAlignment, globalId?: string): Oni.StatusBarItem {
```

This will require further changes to the `oni-api` and the statusbar which is returned from the api is already instantiated 🙇 

currently I get the id for each element this way
```
    public createItem(alignment: StatusBarAlignment, globalId?: string): Oni.StatusBarItem {
        this._id++
        const statusBarId = globalId || `${this._id.toString()}`
		//this.config is the proposed change not actual though not entirely sure how this will work
        const statusItems = this._config.getValue("statusbar.priority")
        const currentItem = statusItems[globalId]
        const itemPriority = currentItem || 0

        return new StatusBarItem(statusBarId, alignment, itemPriority)
```

bryphe:
The potential implementation you gave looks good to me - it makes sense for the `StatusBar` to have a dependency on the `configuration`, given that it needs to look up the priority.

> This will require further changes to the oni-api and the statusbar which is returned from the api is already instantiated 🙇

Not sure I understand this. As far as I can see, the only change would be to the constructor (and the internal implementation, having an instance of `configuration`) - this isn't stuff that is exposed via `oni-api`.  The only public methods for `StatusBar` from oni-api are `getItem` and `createItem`, and it doesn't look like the proposed implementation requires modifications to those signatures. Perhaps I'm misunderstanding the issue though. Although I do think it'd be good to make `globalId` non-optional though, so that could be a necessary change.

In other words, the dependency on `configuration` is an _internal implementation_ and not something that needs to be exposed to the consumer. That's actually the benefit of having this system - someone creating a new statusbar item doesn't need to worry about respecting our convention for setting the priority, and we are protected against bad actors who disregard that convention.

Perhaps the issue is how we get the `configuration` into the status bar? I can help with that piece - it's in line with some of the other refactoring I've been doing to use explicit dependencies instead of arbitrary global objects.

Akin909:
@bryphe apologies wrote that comment in a rush, without actually thinking it all the way through or trying it out actually hence the delete, turned out to be fairly straight forward in terms of the changing and didn't involve any of the exposed portions of the api. It's currently running and hasn't failed at the point it's been failing at for a while so hopefully its all resolved 🤞

bryphe:
Ah ok, great! I put together PR #1184 to help with the statusbar plumbing. Glad its working through 👍

Akin909:
@bryphe I think the CI tests have gotten the better of me here as I've incorporated your PR but still seem to be failing on CI tests due to a timeout error. The error message I get is `Error: element (".automated-test-result") still not existing after 30000ms` for which I'm unable to determine the root cause. I've tried logging and checking that everything is properly awaited, I've checked the functionality the tests cover, all of which seem to work properly. I'm still unable to run the integration tests locally due to an entirely different error due to a chrome driver timeout thought that has more or less always been the case. I've been using this PR locally and haven't personally noticed any related bugs, any help would be much appreciated

bryphe:
Ah bummer about the tests... I'll take a look at those a bit more on Tuesday when I'm back full-force 😄  Thanks for your round of investigation too! That's strange about the chrome driver timeout - do you happen to have another instance of Oni running while the tests are going? I think they are sensitive to that in some cases.

We could definitely use better logging in that case - those failures aren't actionable. So I'll look at adding some better logging there (like dumping the full console in that case).

One strategy I use when I have trouble with test failures is to look and see if I can split up the changes - a set of smaller PRs is usually safer than one larger PR. And splitting it up can help us isolate where the failure is. 

For this change, I see 4 main overarching 'changes':
- #1184 - using explicit dependencies / initialization for the status bar
- Updating the API to use the `statusbar.priority` instead of passing in the priority
- Refactoring the statusbar to use `styled-components`
- Adding `StatusResize` / implementing the actual resize logic

The act of splitting these up can help us get more info for the test, too. For example, if the tests pass for 3/4 of those changes, but fail for the last one - we have a much smaller 'surface area' of changes we need to investigate 😄  I'll bring in #1184 to start the party.

Just an idea - it's a strategy that's helped me out often when I get stuck on where to go next with a failure. In parallel, I'll work on improving those tests so we can get better logs in those cases! Logged #1205 to track this. Thanks for all your work on this!

bryphe:
Oh, one other thing I noticed when I was testing this out locally - I see there is a bit of a gap on the right side:
image

It seems like it only happens once the window gets past a certain size - the gap starts to become noticeable as the window moves past around ~1000px wide on my machine. Does that reproduce for you at all? Is there any info I can get?

Akin909:
@bryphe the gap is/was because the github Icon was not in the same section as the other components as it isn't a status Icon in the same sense as the others i.e. created via the api. I've added it to the same container now so there is no longer that gap at the beginning.

Re. breaking up the PR that's a good method to use to debug which I also tend to use but seemed a bit daunting for this given that a lot of work would need to be reproduced or reverted then recreated. Especially since the tests seem to only fail on one system (integration & unit seem to pass on windows and linux) so I think/hope its due to one odd quirk somewhere relating to `macOS` that I just haven't been able to identify.

I think in this case I can most easily split this PR into 2 with #1184 merged in I can add the changes to the api re. priority as one PR then the remaining status bar changes as another, the styled components and status resize are fairly coupled as I made a few changes to that `StatusBar.tsx` based on changes involving them and walking it back would be quite time consuming I'll start with 2 to see if that shows anything and look into a separation if not

Akin909:
Closing this PR in favour of #1114 and #1210 which are subsets of this larger PR (the remaining unincorporated changes are stylistic/relate to react keys/conversion to styled components

PR #1117, opened 2017-12-13
Fix #1009 - Persist background color

### LINKED ISSUE #1009: Eliminate White Background on Resizing Window.
There was a long standing issue with Electron apps that would result in resizing windows creating a white laggy trace behind them. I believe some people have figured out the solution - you can look at VSCode Insiders (they do not have the laggy white trace).

bryphe:
__Issue:__ When resizing the window, on Windows, you can see a white background. It's also visible during the initial load, before the document is ready.

__Defect:__ For the best experience, Electron needs to know the background color _before_ the window launches. Previously, we didn't have a way to record that.

__Fix:__ Incorporate the `electron-settings` package. This allows us to persist the background color as `_internal.lastBackgroundColor`, and use it from both the main and the renderer process.

This will also open the door for persisted configuration settings..


jordwalke:
This was definitely an improvement.

Issue #943, opened 2017-11-14
Performance: Split bundle.js into separate components (bootstrap + UI)


bryphe:
Following from @jordwalke's comments on #933 - one thing we should look at is splitting up our `bundle.js`.

![image](https://user-images.githubusercontent.com/13532591/32807715-a993fdcc-c945-11e7-940e-591b8ff82acb.png)

We have one giant `bundle.js` that handles all of our UI and Neovim interaction - but that's not really necessary, especially on first-load. 

Ideally, we could isolate the minimum dependencies needed to get up-and-running (like `NeovimInstance`, `CanvasRenderer`, etc - bootstrapping stuff), and split out from the UI elements, so that we can minimize the startup time.

bryphe:
Need to spend some time with these docs: https://webpack.js.org/guides/code-splitting/

bryphe:
This is done with #1123  - 
![image](https://user-images.githubusercontent.com/13532591/34282282-0c41e092-e679-11e7-84af-e6a2e543118f.png)

The entry point is 54.1 kb... hopefully we can continue to rev these and require less-and-less on startup. `0.bundle.js` is needed relatively shortly after, but ideally we can trim that and load more in parallel while Neovim is starting.

Issue #381, opened 2017-04-19
Performance: Evaluate V8 snapshots


extr0py:
Atom had an interesting strategy for improving startup performance in their blog post here:
http://blog.atom.io/2017/04/18/improving-startup-time.html

They are leveraging V8 snapshots to output a JS heap. This saves all the parsing and processing of JS, which is non-trivial as the bundle.js for this project gets larger.

There are some restrictions (V8 snapshots run in a bare V8 context), so it's important to have solid testing to ensure we aren't breaking anything in these cases.

extr0py:
One possible avenue to help with #355, but improving the real startup performance.

bryphe:
One suggestion that @bruchmann made on #1123 was leveraging [electron-link](https://www.npmjs.com/package/electron-link) to accomplish this. Would be interesting to see how straightforward this is to plugin to our infra/build.

bryphe:
`mksnapshot` seems to be used as well

PR #1123, opened 2017-12-13
#943 - Split bundle via async imports

bryphe:
This is the first part of an improvement in startup time. Today, we need the browser to load the entire large `bundle.js` before it can start rendering.

This change leverages the async `import` functionality, along with webpack's ability to chunk along the async boundaries, such that we can kick start some operations and parallelize the loading.

This is a first step, but helps to improve loading.


bryphe:
There may be some duplication across bundles - should look at incorporating the CommonChunksPlugin as well.

Akin909:
@bryphe may or may not be relevant to what your trying to do here or potentially a rabbit hole, though seemed worth mentioning a new `js` bundler called [`Parcel`](https://github.com/parcel-bundler/parcel) it simplifies a lot of what webpack does through painful configuration and endless plugins, it supports `TS` and `Less`

bryphe:
Ah ya, interesting! It does look easy and straightforward.

I'm wondering too if we can just avoid making bundles altogether - just point Electron to the output `index.js`. This would save us a lot of compile time, and might actually be more efficient in terms of loading. The version of Node that Electron has today unfortunately doesn't support the `import` method, but once Electron 2 is out, it might be worth profiling w/o any bundler.

bruchmann:
There’s also [electron-link](https://www.npmjs.com/package/electron-link) that might help with [improving startup time](http://blog.atom.io/2017/04/18/improving-startup-time.html).

> electron-link is a node module that takes a JavaScript file (typically the entry point of an application) and a list of modules that need to be required lazily (see Atom's build scripts for an example). Then, starting from that file, it traverses the entire require graph and replaces all the forbidden require calls in each file with a function that will be called at runtime. The output is a single script containing the code for all the modules reachable from the entry point. This file can be then supplied to mksnapshot to generate a snapshot blob.

bryphe:
Thanks for the pointer, @bruchmann ! We're tracking leveraging v8 snapshots in #381 , just added some context there - `electron-link` looks like it would be very helpful for that.

PR #1133, opened 2017-12-15
[WIP] Upgrade testing framework

samvv:
This pull request switches the testing framework to [AVA](https://github.com/avajs/ava), so that tests will run in parallel where possible and we are not constrained to e.g. instantiating an entire new application for each test. Will come in especially useful when more integration tests are added.

 - [x] Added a `upsearchSync()` to `src/Utility.ts` to enable finding a certain file in parent directories and applied it to `NeovimProcessSpawner` so that it finds the locally installed Neovim even when running from tests
 - [ ] Made integration tests run in parallel and with the locally compiled version
 - [ ] Add an option to run the integration tests again on the prebuilt binary

I also moved all test files up to the root test folder `browser/test`, which in retrospect might not have been the best idea. My reasoning was that tests can be added and removed more freely, without being forced to follow `src/` directory structure. On the other hand ... AVA runs each file in a separate process, so there doesn't really have to be a 1:1 mapping between `src/` and `test/` (what if one file in `src/` should have multiple parallel processes?). That's something you should decide, @bryphe


bryphe:
> This pull request switches the testing framework to AVA, so that tests will run in parallel where possible and we are not constrained to e.g. instantiating an entire new application for each test. Will come in especially useful when more integration tests are added.

Awesome! The parallelization of tests is really cool. This will be important as we add more unit tests. The tests seem more streamlined as well (not needing the `describe` blocks) - I like it.

> I also moved all test files up to the root test folder browser/test, which in retrospect might not have been the best idea. My reasoning was that tests can be added and removed more freely, without being forced to follow src/ directory structure

I'd prefer the unit tests (for now) to follow the directory structure. I've been experimenting with something like vim-alternates (http://www.vim.org/scripts/script.php?script_id=31), but an oni version, that lets me switch to / from the test file. My hope is we could set up something in an oni config file, like:
```
'workspace.testPaths: [{ source: 'browser/src/**/*.ts', test: `browser/test/**/${fileName}Tests.ts'}]
```
And then have a key-binding or command to switch to the 'test' file (or create one if none exists!). I think this would help solve the adding more freely aspect in a general way. I would have loved to have this sort of system in a few teams I've worked on, so I thought Oni would be a good place to test it out 😄 

One other aspect I've been thinking about is further modularizing Oni - I'd eventually like to see if we can break out some pieces into extensions:
- Auto-closing pairs
- QuickOpen
- Completion
- LanguageManager
Basically anything in the `Services` folder is up for grabs.  By mirroring the directory structure, it'd be easier to pluck out some of the relevant tests.

Let me know what you think! Thanks for making this change - the move from mocha -> AVA will be a big step forward.

bryphe:
Oh, I had one other question too - right now we use `electron-mocha` to debug the unit tests, via `npm run debug:test:unit:browser`. What's the best way to debug the AVA tests?

samvv:
> I'd prefer the unit tests (for now) to follow the directory structure. I've been experimenting with something like vim-alternates (http://www.vim.org/scripts/script.php?script_id=31), but an oni version, that lets me switch to / from the test file. My hope is we could set up something in an oni config file, like:

That's actually really cool! Yes, then it totally makes sense to keep directory structure. Will fix it ASAP.

> Oh, I had one other question too - right now we use electron-mocha to debug the unit tests, via npm run debug:test:unit:browser. What's the best way to debug the AVA tests?

I'll create and configure a new command in `package.json` for this. If you have it installed, it will be as simple as `ava`. Stack traces etc. should come nicely formatted in the terminal, and there's really no difference between a browser test and a regular test anymore thanks to [Spectron](https://github.com/electron/spectron). If you only want to run certain tests, it'll be something like `ava --match bar --match 'foo*'`

bryphe:
>That's actually really cool! Yes, then it totally makes sense to keep directory structure. Will fix it ASAP.

Sweet! I realized I don't have an issue tracking that 'alternates' work - I'll create one with more details.

> I'll create and configure a new command in package.json for this. If you have it installed, it will be as simple as ava. Stack traces etc. should come nicely formatted in the terminal, and there's really no difference between a browser test and a regular test anymore thanks to Spectron. If you only want to run certain tests, it'll be something like ava --match bar --match 'foo*'

Sounds great! 👍

bryphe:
I'll close this out since it's inactive for the time being, but I'm open to bringing it back if someone wants to look at this.

PR #1141, opened 2017-12-16
[NO MERGE] Test moving dependencies to root for #922

bryphe:
Test to see if moving dependencies to root addresses the write conflict with `rimraf` in `electron-builder`


bryphe:
Not needed now that #922 is in 👍

PR #1142, opened 2017-12-17
Add react dev tools and hook up redux dev tools

Akin909:
As per conversation on gitter I've added the react dev tools and tweaked the redux dev tools to pick up the store.

An issue I ran into relates to [permissions](https://github.com/MarshallOfSound/electron-devtools-installer/issues/63) whilst using `electron-devtools-installer`, It was reported as closed in another issue but I found that the issue continued to persist for me so I followed the advice in the link. Essentially the react devtools extension *on my mac* possibly on other platforms required me to `chmod 777` the path of the extension, I don't know how much of an issue this maybe for others if it is then it might be worth looking into the standalone app, like [`react-devtools`](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools/README.md) or simply add the advice re chmod to the dev wiki whilst electron-installer work on the issue hopefully

### DIFF
--- a/browser/src/Redux/createStore.ts
+++ b/browser/src/Redux/createStore.ts
@@ -17,7 +17,7 @@ import { RequestAnimationFrameNotifyBatcher } from "./RequestAnimationFrameNotif
 
 export const createStore = <TState>(name: string, reducer: Reducer<TState>, defaultState: TState, optionalMiddleware: Middleware[] = []): Store<TState> => {
 
-    const composeEnhancers = typeof window === 'object' &&
+    const composeEnhancers = typeof window === "object" &&
         window["__REDUX_DEVTOOLS_EXTENSION_COMPOSE__"] ? // tslint:disable-line no-string-literal
         window["__REDUX_DEVTOOLS_EXTENSION_COMPOSE__"]({}) : compose // tslint:disable-line no-string-literal
 
--- a/browser/src/Redux/createStore.ts
+++ b/browser/src/Redux/createStore.ts
@@ -17,7 +17,9 @@ import { RequestAnimationFrameNotifyBatcher } from "./RequestAnimationFrameNotif
 
 export const createStore = <TState>(name: string, reducer: Reducer<TState>, defaultState: TState, optionalMiddleware: Middleware[] = []): Store<TState> => {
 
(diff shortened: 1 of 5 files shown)


bryphe:
Awesome, thanks for getting this working again, @Akin909 ! This will make debugging so much easier 👍

> I don't know how much of an issue this maybe for others if it is then it might be worth looking into the standalone app, like react-devtools or simply add the advice re chmod to the dev wiki whilst electron-installer work on the issue hopefully

I think, since it's one-time step, adding it to the wiki is probably simpler and sufficient. I'm cool with either approach though.

Akin909:
Cool @bryphe I personally would find it really frustrating to have to switch between a standalone app and electron. I can add a section in the wiki explaining what to do if it comes up for anyone else

bryphe:
> Cool @bryphe I personally would find it really frustrating to have to switch between a standalone app and electron. 

Ah, I think I skimmed the `react-devtools` repo too fast - I thought it was an embedded UI, just outside the devtools. I agree, having it be standalone would be painful.

Akin909:
Added the following to the docs

> Note: some users *may* have issues with the `react-devtools` due to OS permissions.
to resolve this follow these steps:
> 1. Navigate to `chrome://extensions` in _*your browser*_, and find its extension ID, which is a hash string like `fmkadmapgofadopljbjfkapdkoienihi`.

> 2. Find out filesystem location used by Chrome for storing extensions:
   
 > * on Windows it is `%LOCALAPPDATA%\Google\Chrome\User Data\Default\Extensions`;
   on Linux it could be:
 > * `~/.config/google-chrome/Default/Extensions/`
 >  * `~/.config/google-chrome-beta/Default/Extensions/`
 >  * `~/.config/google-chrome-canary/Default/Extensions/`
 > * `~/.config/chromium/Default/Extensions/`
 >  * on macOS it is `~/Library/Application Support/Google/Chrome/Default/Extensions.`

 > 3. Once you have found the extension location you will need to grant the current user read and right permissions
    using `chmod 777 -R path/to/extension` (N.B. the command recursively grants the current user access to the folder and its subfolders)

bryphe:
Sweet, thanks for taking this on @Akin909 ! Looks great!

Issue #1146, opened 2017-12-18
Performance Regression: Styled components - continually adding `style` blocks to `head`


bryphe:
Since using master, I've noticed gradual slowdowns as I've been using Oni. I traced it down to to styles being continually added at the `head`:
![image](https://user-images.githubusercontent.com/13532591/34123912-f7a69fd2-e3e5-11e7-84c2-62681da818c6.png)

It seems that, whenever the `props` change, a new ID is generated and added to the head. We should look at ways to limit this - performance is critical.

bryphe:
There's multiple issues here:
- As the DOM grows, it causes layout and flow to get more and more expensive.
- The addition of styles also causes much larger repaint areas than needed.

We can't really afford to compromise on performance - for properties that are changing frequently, it seems like it's best to use the `style` property, as not only does that prevent the `head` from blowing up.

bryphe:
@samvv @Akin909 - FYI - given the performance issues I'm seeing here, I think it's fine to use `styled-components` for things where we would use CSS (like static values, or even theming, animations/transitions), but for dynamically changing values, its best for performance to continue to use inline styles.

Let me know if you have feedback or other ideas on how to handle this!

bryphe:
For now I'll revert the dynamic setting of styles in the high-frequency code paths (like `Cursor`, `CursorLine`, etc)

Akin909:
@bryphe I've been using styled components for a while and I haven't encountered this problem, not immediately sure what could be causing it, a quirk with electron or perhaps our implementation so far not sure if were referencing both the typed version of `styledComponents` in common as well as the untyped `styled-components` lib though I've not known that to be an issue, I can look at some SC repos's issues and also look at the implementation so far, seems there are issues with commas instead of the semi-colons in places which could lead to invalid css that SC may be trying to reproduce each time.

bryphe:
Hmm, seems like it is our implementation - for _static_ styles, `styled-components` is smart enough to hash it and then re-use the style based on the hash.

Unfortunately though, for styles that _change_, like this code:
https://github.com/onivim/oni/blob/9fa83a62ff35f535cef81d5e5f2d8a3f418df01d/browser/src/UI/components/Cursor.tsx#L64

```
const CursorContainer = withProps<ICursorRendererProps>(styled.div)`
    visibility: ${props => props.visible ? "visible" : "hidden"};
    position: absolute;
    left: ${props => props.x}px;
    top: ${props => props.y}px;
[...]

PR #1148, opened 2017-12-18
[WIP] #1146 - Fix perf / responsiveness regressions

bryphe:
Following from the notes in #1146 - this is bringing back the inline styling for dynamic properties (properties that change frequently).


bryphe:
For the `CursorPositioner` and `ErrorMarker` components, I reverted those changes - there were other problems besides just performance (flickering for the CursorPositioner, and for the Errors, the squiggles weren't showing up).

PR #1151, opened 2017-12-19
Fix #1149 - Handle  shift+space key

### LINKED ISSUE #1149: Shift + space is not working correctly in Oni
There's a bug in our input handling - when the user presses `Shift` + `Space`, we should be sending `<s- >` to Neovim, but we are just sending the space character ` `. Note that the control key works correctly, sending `<c- >`.

bryphe:
__Issue:__ The `shift+space` key is being sent to neovim as ` `, not `<s-space>` as would be expected.

__Defect__: The keyboard layout/mapping we leverage shows ` ` as being the result character from shift+space key, and in that case we ignore the `shift` and just send a space to neovim. This works fine for alphabetical characters - for example, if we see a `P` which is `shift+p`, we know to ignore the shift key, and just send `P` - but that doesn't make sense in the case of space.

__Fix__: Special case the ' ' key here.

### DIFF
--- a/browser/src/Input/Keyboard/Resolvers.ts
+++ b/browser/src/Input/Keyboard/Resolvers.ts
@@ -101,6 +101,10 @@ const isShiftCharacter = (keyMap: IKeyMap, evt: KeyboardEvent): boolean => {
         return false
     }
 
+    if (code === "Space") {
+        return false
+    }
+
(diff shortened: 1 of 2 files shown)


bryphe:
Fixes #1149

Issue #1081, opened 2017-12-06
set autoread doesn't seem to be working


MikaAK:
I've set `set autoread` in my `init.vim`, but my files don't actually seem to change when I've changed it from another source. This is especially annoying with package.json files when installing new packages. Oddly enough my git-gutter-plugin does update however and shows the line has changed but the content is unchanged.

bryphe:
Interesting, I'm able to reproduce this - and it seems that `nvim-qt` might have the same issue

CrossR:
There is some discussion around this in https://github.com/neovim/neovim/issues/1936 , where there is a suggested "fix".

I can confirm that if I `set autoread` and then edit a file, the file doesn't update in Oni (or nvim-qt), but If I `:checktime` it will then update.

We could potentially add the suggested change to the Oni core `init.vim`, but I'm not 100% on the consequences or if there is a performance difference. Would probably be better to add it via on Oni config option, and set it that way, so that it isn't enabled for everyone?

bryphe:
Nice find, @CrossR ! The behavior makes more sense now, after reading through the thread.

> Would probably be better to add it via on Oni config option, and set it that way, so that it isn't enabled for everyone?

Seems reasonable. I'm trying to put less and less in `init.vim` so as to minimize conflict with existing configs. It can actually already be set via a configuration in Oni: `vim.setting.autoread`: `true`.

We could check if this is set, and if so, we can call `:checktime` along with our call to focus gained in our `focus` handler (perhaps via `this._neovimInstance.command`):
https://github.com/onivim/oni/blob/57929c06a02fe6e886747bda45ac2c9412ccd2fd/browser/src/Editor/NeovimEditor.tsx#L308

CrossR:
This should hopefully be fixed now in #1156.
If you build from master it should be working, or the next release will have this in!

bryphe:
Awesome, thanks for the fix @CrossR !

MikaAK:
Nice work guys!

PR #1156, opened 2017-12-19
Wire up Autoread.

CrossR:
Following on from discussion in #1081, this allows Oni to call the check time command when the user is using `autoread`.


bryphe:
Awesome, looks great @CrossR - thank you!

PR #1159, opened 2017-12-19
[WIP] File Explorer - Theming integration

bryphe:
This hooks up the `styled-components` `ThemeProvider` in order to get theme colors for the file explorer.

Added a couple of new colors:
- `explorer.foreground`
- `explorer.background`
- `explorer.container.background`
- `explorer.cursor.foreground`
- `explorer.cursor.background`
- `explorer.highlight`

Still need to hook them all up, and validate across the colorschemes we have.

### DIFF
--- a/browser/src/Services/Explorer/ExplorerView.tsx
+++ b/browser/src/Services/Explorer/ExplorerView.tsx
@@ -24,6 +24,7 @@ require("./Explorer.less") // tslint:disable-line
 export interface IFileViewProps {
     fileName: string
     isSelected: boolean
+    theme: IThemeColors
 }
 
 export class FileView extends React.PureComponent<IFileViewProps, {}> {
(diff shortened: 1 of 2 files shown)


Akin909:
@bryphe 👍  love all the file explorer work really excited to see it :). I noticed that your integrating the theming from `themeProvider` and I wanted to point out that with regards to using the theme if the final component that you want to respond to the theme is a `styledComponent` then they all automagically aka react context have access to the theme props so if the final component you are styling is say a folderView then you can use
```js
const FolderView = styled.div
     styleA: ${props => props.theme.explorer.background}

```

This means that all child components of the theme provider no matter how far nested down have access to `this.props.theme` as long as its a styled component.

Given that its a work in progress, pardon my intruding with my suggestion just wanted to point that out to spare you potentially having to add in loads of `withTheme(comp)`

bryphe:
> I noticed that your integrating the theming from themeProvider and I wanted to point out that with regards to using the theme if the final component that you want to respond to the theme is a styledComponent then they all automagically aka react context have access to the theme props so if the final component you are styling is say a folderView then you can use

Nice, thanks for the suggestion! I'll give that a shot - seems like it will save me from a bunch of plumbing 👍

bryphe:
I'm looking at some changes to generalize the components, since it seems like some of those UI elements will likely be reused in other sidebar pane scenarios. And `styled-components` will help a lot for those! 👍

PR #1160, opened 2017-12-20
Add save window state functionality

### LINKED ISSUE #1155: Persist Window Dimensions and Location on startup
@bryphe I read [an article](https://blog.avocode.com/4-must-know-tips-for-building-cross-platform-electron-apps-f3ae9c2bffff) that illustrates how to persist window dimensions for an electron app. The end result being that oni will open in the same place at the same size as when it was closed. It's subjective whether or not users will want this but personally when I open any app I immediately have to resize it to where I want it to be which is typically split screen somewhere left or right. 

It seems you've already added the necessary infrastructure for persisting settings and the article goes through how to add this I wonder if this is something that would be desirable in your opinion if so I'd be happy to implement it possibly as a configurable option (although i'm not sure if the config object can be reached from `main.ts`

Akin909:
This PR adds saving of the previous window state so that on load of Oni the user has a window in the same position as the last time. Fixes #1155.


Akin909:
@bryphe i've tested this locally and it works great also had to explicitly set that if the window was previously maximised it re-opens maximised, all done if tests are passing but they seemed to be failing for all branches following a recent merge with master

bryphe:
Awesome, I just tested it on Windows as well - worked great for me too 👍 Nice work @Akin909 , thanks for this change! Bringing it in now

PR #1167, opened 2017-12-20
Fix #1154 - Use environment variable to specify configuration path for integration tests

### LINKED ISSUE #1154: Disappearing Config.js
@bryphe, this is an issue with development I believe as it occurred as a result of that, when running integration tests the users `config.js` is deleted and replaced then restored afterwards, if there is a break in the test process due to in my case the tests hanging and the process is killed manually there is the risk as happened to me of the config file not being restored 😭 . 

It seems like an edge case occurrence as I've run these tests several times before but is worrying that it can ever happen as Oni cannot load without a `config.js` and I didn't realise this was the cause for a while, as the app just hangs indefinitely.

I'm not sure whether or not these tests are intended to be run on CI servers exclusively 😖 but if not then its quite risky/problematic to have the actual user config interfered with whilst running tests, is it possible to have oni point to a fake config for testing purposes instead ? or default to its native config? or whether we specify that the path to search for the oni config is swapped using a `process.env.NODE_ENV=test` variable so it points to a config path for tests.

EDIT: this also raises the issue that if the `config.js` is missing for whatever reason there could/should be a fallback or a warning msg or perhaps immediately/ user is prompted to replace it with the default from oni

bryphe:
__Issue:__ The integration tests overwrite the configuration file - this is very frustrating when running locally. There is some logic for backing up / restoring the file, but it is not robust.

__Fix:__ Use an environment variable to specify the configuration path instead. This is more robust and does not rely on copying / backing up the user's configuration file.


bryphe:
Fix #1154

bryphe:
Test failure was due to `BasicEditingTest`, which is now removed - all other tests passed.

PR #1172, opened 2017-12-21
Transform the Markdown-preview plugin into a local package

TalAmuyal:
Currently, plugins are separate from Oni in the technical sense, but not completely.
After reading [this](https://medium.com/@arnaudrinquin/build-modular-application-with-npm-local-modules-dfc5ff047bcc) article, I think we should include them as a dependency using the `file:` prefix.
That **doesn't** require us to publish the plugins or separate them into a different git repository.

Inspired by that article I made the Markdown preview an independent package.

That way, it can contain its own dependencies and the `peerDependencies` section in the plugin's `package.json` should work correctly (see [unhelpful warnings](https://ci.appveyor.com/project/bryphe/oni/build/1.0.923/job/5r6jg2x7er9vktma#L59)).


TalAmuyal:
@bryphe,

On Linux Travis approves, but [on Mac it yields a timeout](https://travis-ci.org/onivim/oni/jobs/319731938#L1584).
I guess it either happens on `master` or is just inconsistent in general.

Aside from that, on Windows, it fails to install. I want to think that the reason is that my changes are non-Windows compatible , but I just can't imagine the relation between [the change](https://github.com/onivim/oni/pull/1172/files#diff-b59bd66e53ff1d95df53b911a567b60dR9) and [the error](https://ci.appveyor.com/project/bryphe/oni/build/1.0.1063/job/qleihau7n9lxad9u).

BTW, I don't know if it's a bug in `npm` or just a "feature" - it seems that if a dependency is local (`file:`-prefixed) then its dependencies are not installed.
That is the reason for [the change](https://github.com/onivim/oni/pull/1172/files#diff-b59bd66e53ff1d95df53b911a567b60dR9).
Any idea of a better solution?

bryphe:
Weird! I restarted the OSX build and it passed now. Our master build has stayed [pretty green[(https://travis-ci.org/onivim/oni/builds) - although I was wrestling with test issues most of the week 😦 

> BTW, I don't know if it's a bug in npm or just a "feature" - it seems that if a dependency is local (file:-prefixed) then its dependencies are not installed.

Unfortunately I'm not sure about this. I've always had issues trying to use the local file dependencies, so I've tried to avoid it except when debugging. We do use `yarn install` as opposed to `npm install` for the root, so that could potentially cause issues.

Besides the build issues we see here, I've also seen cases where `electron-builder` doesn't pick up separated dependencies, and you get the worst kind of bug where it works locally, but it breaks in the install package because some node_module is missing. So I'd lean towards either keeping the dependencies as-is, even if it isn't ideal.

We try and guard against these sort of breaks with the integration tests on the CI machine (that actually run against the built package), but we don't have any tests for markdown preview today, so it'd be easy for that to break that w/ a missing node module. IMO, my preference would be to focus on improving the functionality over wrestling these dependencies - because even if our node_module layout isn't perfect, it doesn't really impact the end user. Keeping them centralized at the top level is the easiest way to keep everything working for now - and then we have the flexibility to split them out to separate packages as we see fit.

One other thing is I've created an 'extensions' folder at the root:
https://github.com/onivim/oni/tree/master/extensions

I'm starting to migrate our 'plugins' there (mainly just language grammars and themes at the moment). It's great to have all the extensions 'centralized' so that we have examples, and the markdown-preview is our best example plugin. So in any case it'd be worth moving it over there!

TalAmuyal:
Gotcha, making an alternative PR.

TalAmuyal:
Replaced by https://github.com/onivim/oni/pull/1181.

Issue #1135, opened 2017-12-15
Feature: Goals / Achievements


bryphe:
(I like the format of @kkd705's feature proposal, so I borrowed it!) This is another idea for lowering the bar / discoverability, specifically for editing commands.

## PROBLEM

It's challenging to get started with modal editing. I'm sure lots of people are familiar with this graph of Vim's learning curve:

![learning curve](https://pascalprecht.github.io/public/vim-learn-curve.jpg)

A key issue is discoverability - how do I know what commands to use? How do I know what keypresses are relevant in a particular mode? How do I commit commands to memory?

One tip that I used while I was getting started with Vim was the sticky note method - I would put 3 new vim commands on a sticky note on my monitor. Once I was comfortable with those commands, I would switch to some newer ones. However, with the UI affordances that Oni has, I believe we can do better - and even gamify this.

## SOLUTION

The idea is that, we have a set of commands available in each mode. Each command will have an associated goal. For example, use the 'word' motion 50x. Every time the motion is used, we'll bump up a graph in the 'learning' pane.

Once a learning goal is met, we'll pop up a quick UI, inspired by the achievement UI in consoles.

This would accomplish two things:
- Make learning new commands/motions _discoverable_
- Help users learn commands/motions they don't use frequently

A key note is that these achievements are, in general, __modal__ - that means that, as the user changes modes, we'll automatically show them the goals or achievements that are relevant to that mode.

## EXAMPLE

The user could open the goals pane via a command-palette entry, like:
![goals-quickopen](https://user-images.githubusercontent.com/13532591/34053805-0d37b770-e17d-11e7-818c-b35721573950.png)

Which would then expose a sidebar with the list of goals/achievements:

![goals](https://user-images.githubusercontent.com/13532591/34053748-db292502-e17c-11e7-9e57-dd04f6485a94.png)

As the user progresses, the bar would increase / show some animation to make it fun. Then, we could even show console-game-like achievements:

![achievement](https://user-images.githubusercontent.com/13532591/34053845-320a4a18-e17d-11e7-86af-1c821b222323.gif)

> NOTE: A key element not shown is that, as the user switches modes, it would highlight a different set of achievements, that are specific to that mode. It'd be helpful to have a mode 'selector' that the user could use to choose which set of accomplishments to view, and it would automatically switch as the user changes modes.

## IMPLEMENTATION IDEAS

Creating the UI is relatively straightforward. The major challenges are:
[...]

PR #1179, opened 2017-12-22
Externalize wildmenu

Akin909:
@bryphe following on from #1127, I've added an external wild menu which I've also placed behind an experimental flag `experimental.wildmenu.mode: false`.

wildmenu

It's a much simpler thing to externalise as there is no cursor to deal with. I've left it pretty minimal and used theme values for styling but as to icons etc. its pretty bare bones.

Also whilst implementing it, it occurred to me that I'm able to navigate it smoothly as i have the following settings in my vimrc
```vim
"pressing Tab on the command line will show a menu to complete buffer and file names
set wildmenu
set wildchar=<Tab> "Key to use to navigate wild menu
set wildmode=full       " Shows a menu bar as opposed to an enormous list
set wildcharm=<C-Z>
```
this basically allows me to scroll the wild menu by hitting tab, not sure what the experience would be for a user new to vim i.e. arrow keys etc. I was thing we could add these settings to the `init.vim` though they might be controversial and might be better left to the user, but this I imagine wont work properly without some init.vim settings.


bryphe:
Cool, it's looking neat!

> Also whilst implementing it, it occurred to me that I'm able to navigate it smoothly as i have the following settings in my vimrc

It actually worked OK for me too (I didn't have any settings for this in my `init.vim`). I wonder if some of these are set by default?

> this basically allows me to scroll the wild menu by hitting tab, not sure what the experience would be for a user new to vim i.e. arrow keys etc. I was thing we could add these settings to the init.vim though they might be controversial and might be better left to the user, but this I imagine wont work properly without some init.vim settings.

It'd be great if we could avoid settings in `init.vim` if possible - especially for these externalized features. For this, do you think we could control the selection of the item from the Oni side? For example, add a keybinding like `command.next` - sort of like how we have `menu.next`? Or even reuse the `menu.next` binding? I'm thinking we could style it to look similiar to our other menus - like the quick open / command palette, and in that case, it would make sense for it to have the same input experience as those menus.

bryphe:
Also, I'm cool to bring this in as-is since it's behind an experimental flag -  we can continue to refine incrementally. Let me know if you're all set or if there is anything else you wanted to do as part of this PR.

Akin909:
@bryphe all done for now 🌮 
Agreed re styling it down the line, as its a bit bland perhaps a little later I can have a look at having it look more or less the same as the quickOpen menu. 
Re controlling the selection, as far as I can see that all comes from `neovim` so any oni binding would likely be a wrapper for whatever command default neovim uses to navigate the wildmenu, maybe the default `wildchar`, but as with the commandline window it isn't so much us actually sending input from the component so much as too me it feels like eavesdropping on a conversation lol

bryphe:
> @bryphe all done for now 🌮
👍 

> Re controlling the selection, as far as I can see that all comes from neovim so any oni binding would likely be a wrapper for whatever command default neovim uses to navigate the wildmenu, maybe the default wildchar, but as with the commandline window it isn't so much us actually sending input from the component so much as too me it feels like eavesdropping on a conversation lol

Makes sense. I wonder if there is a way for us to 'abuse' that command line functionality, or at least move a bit outside the box. One naive approach to get an experience that would be similiar to our existing scenarios would be, for every keystroke, we clear out the command line, and then send the new command line + `<TAB>` (or whatever the _default_ wildmenu expand is). That way, as I'm typing like `:vsp`, I'm implicitly expanding as I go. I imagine this might break more advanced functionality, but it might be good enough - and if it wasn't, the user could always 'eject' and turn off the commandline/wildmenu externalization.

bryphe:
Will bring this in! Nice work @Akin909 , thanks for all you did here!

Akin909:
@bryphe re. the replacing input of the cmdline, that might be a good approach for the `cmdline` but for the wild menu it would essentially treat each attempt at that as a new wild menu, which I think would make it very hard to use, also re. implementation each time a user hit the binding they would make a new request for a wildmenu's options which would need to be sorted and is possibly not likely to look the same i.e. the same results.

I think a user who's likely using the wild menu has probably spent a bit of time with vim and likely has a mapping or can add one. 

If they're a new user then at a certain point they're going to have to start adding mappings to vim itself, otherwise I think this could put `oni` in the situation of circumventing vim mappings by working around them. 

I think a better way to address this would be through the #1135 aka teaching people how to add them or maybe oni having some interface for that as some point.

PR #1184, opened 2017-12-23
Refactoring - Status Bar: Explicit dependencies / initialization

bryphe:
This refactors the statusbar to make the initialization timing and dependency injection explicit. A `StatusBar` now gets initialized as part of our startup flow in `browser/src/index.tsx`, and then we use that instance for the Oni API object as well as for the language client (so that it can create the language client status bar).

This also hooks in the `configuration` object, but it's not used yet - just to make it easy to bring in for #1114


bryphe:
I'll bring this in to scope down the 'surface area' of changes in #1114 - hopefully that can help us zero in on whats causing the test failures!

PR #1189, opened 2017-12-23
Automation for Reason

bryphe:
Add automated test case for completion with the `basic-reason` template


bryphe:
Looks like this is _finally_ stabilized! Going to run the test once more to make sure.

Looking forward to when I can move this to be run cross-platform (right now, it only runs on OSX).

bryphe:
2 passes in a row... looks like we're in business

build/script/travis-build.sh
```diff
@@ -27,12 +27,24 @@ npm run pack
 
 echo Using neovim path: "$ONI_NEOVIM_PATH"
 
-npm run test:integration
-
 if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+   echo Installing reason CLI tools...
+   npm install -g https://github.com/reasonml/reason-cli/archive/3.0.4-bin-darwin.tar.gz
```
jordwalke: You may want to try the newest `npm install -g reason-cli@3.1.0-darwin`
(Faster download, more updated merlin).

bryphe: Excellent! Thanks @jordwalke - testing it out now.

Issue #659, opened 2017-08-24
Flow Language Server and/or eslint?


rrichardson:
I searched but couldn't find any open issues here.  What would be the recommended approach for integrating flow and eslint for javascript support?  Maybe I could lend a hand.

extr0py:
Thanks for posting the question, @rrichardson ! This would definitely be useful giving the popularity of flow and eslint.

### ESLint

ESLint looks to be pretty similiar to TSLint (which I created a really simple plugin for here: https://github.com/extr0py/oni-plugin-tslint), in that it's a CLI tool that pushes out errors. 

I think a good starting point here would be to clone the `oni-plugin-tslint` repo, and tweak it pick up the ESLint CLI tool. The interesting bits are hooking up `buffer-saved` and `buffer-enter` events, then it runs tslint, and then it pushes the errors back to Oni uses the  `Oni.diagnostics.setErrors` API.

There aren't too many Oni plugins yet, this could potentially be the first external one ;)

The only caveat is the Oni API is in flux - but that's on my end to figure out... I'm looking at ways to have a graceful degradation when I plan on changing the API surface area.

### Flow

Looks like there is a language server implementation of Flow, which is great! I see one here: https://github.com/flowtype/flow-language-server

Seems like that would be the easiest way to hook up. I have an example implementation against a C# language server here: https://github.com/extr0py/oni-language-csharp

The place to look there is in the `Oni.createLanugageClient` API.

For language servers that have simple initialization requirements (ie, they don't care too much about finding a root path), I'd like to simplify the language server by having it just be a configuration value - that work is tracked in #504 .

The only other item here is that, Oni is hardcoded to bring in the typescript language service for JS/TS files at the moment (to give out-of-the-box completion for web projects). For cases where the user would prefer to use the Flow language service, we'd want to give them an option to disable the default typescript language service behavior.

Hope that helps - let me know if you have any other questions about those existing plugins or how they could work with Flow / ESLint

mxstbr:
Is there any update on Flow support?

Akin909:
@bryphe, I've had a quick look at implementing this though atm registering another language server for JS throws a duplicate error and the tss seems to take precedence so I've added an additional config var to use flow or tss, `useFlow: true` more or less.

bryphe:
Thanks @Akin909 for your investigation here & fixes here!

Really appreciate the documentation you added : https://github.com/onivim/oni/wiki/Language-Support#javascript-and-flow

[...]

PR #1190, opened 2017-12-23
Add flow lang server

Akin909:
Added the flow lang server (relates #659) and a toggle in the default config to swap flow for typescript as both running simultaneously causes a duplicate error. The flag `language.javascript.useFlow` is used in the typescript plugin to prevent it activating in place of flow.

FYI Process went absolutely beautifully thanks to the setup in place 😄 👍 🥇 

The only issue is that if the user sets `useFlow: false` there is still a duplicate lang server error but as far as I can tell the `tss` takes precedence and it does impact its functioning though its not ideal

### DIFF
--- a/browser/src/UI/components/BufferScrollBar.tsx
+++ b/browser/src/UI/components/BufferScrollBar.tsx
@@ -59,7 +59,7 @@ export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps, 
 
         const markers = this.props.markers || []
 
-        const markerElements = markers.map((m) => {
+        const markerElements = markers.map((m, i) => {
             const line = m.line
             const pos = (line / this.props.bufferSize) * this.props.height
             const size = "2px"
@@ -72,12 +72,12 @@ export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps, 
                 width: "100%",
             }
 
-            return <div style={markerStyle} key={m.line.toString() + m.color}/>
+            return <div style={markerStyle} key={`${m.line}-${m.color}-${i}-${top}`}/>
         })
 
         return <ScrollBarContainer>
-                <ScrollBarWindow style={windowStyle}></ScrollBarWindow>
-                {markerElements}
-            </ScrollBarContainer>
+                    <ScrollBarWindow style={windowStyle}></ScrollBarWindow>
+                    {markerElements}
+                </ScrollBarContainer>
     }
 }
--- a/browser/src/UI/components/BufferScrollBar.tsx
+++ b/browser/src/UI/components/BufferScrollBar.tsx
@@ -59,7 +59,7 @@ export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps, 
 
         const markers = this.props.markers || []
 
-        const markerElements = markers.map((m, i) => {
+        const markerElements = markers.map((m) => {
             const line = m.line
             const pos = (line / this.props.bufferSize) * this.props.height
             const size = "2px"
--- a/browser/src/UI/components/BufferScrollBar.tsx
+++ b/browser/src/UI/components/BufferScrollBar.tsx
@@ -72,7 +72,7 @@ export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps, 
                 width: "100%",
             }
 
-            return <div style={markerStyle} key={`${m.line}-${m.color}-${i}-${top}`}/>
+            return <div style={markerStyle} key={m.line.toString() + m.color}/>
         })
 
         return <ScrollBarContainer>
--- a/browser/src/UI/components/ToolTip.tsx
+++ b/browser/src/UI/components/ToolTip.tsx
@@ -19,8 +19,8 @@ export interface IToolTipsViewProps {
 export class ToolTipsView extends React.PureComponent<IToolTipsViewProps, {}> {
 
     public render(): JSX.Element {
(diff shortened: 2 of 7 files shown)


bryphe:
Sweet, thanks for looking at this @Akin909 !

The only downside to this approach is we'll include _both_ TypeScript and Flow in our install package - looks like flow+dependencies is ~10 MB.

I wonder if instead we could instruct users to use their global installation of flow, by specifying:
```
"language.javascript.languageServer.command": "flow-language-server",
"language.javascript.languageServer.arguments": ["--stdio"],
```
(of course, they'd have to run `npm install -g flow-language-server` before too)

And then just check for the presence of "language.javascript.languageServer.command" in our TypeScript plugin - if there is something specified, we don't bother registering. (We might want to do the same for `language.typescript.languageServer.command`, too!)

I think there's a couple advantages to this:
- We don't need to special case a setting for it - we can implicitly support other potential javascript strategies too (like the one from sourcegraph: https://github.com/sourcegraph/javascript-typescript-langserver)
- Keeps our install light(ish). I do plan on leveraging TypeScript more extensively (for example, for an improved experience in our configuration files - we can use our definition files to provide auto-complete, docs, etc), so I'd like to keep that. But I don't think we need two javascript strategies out-of-the-box.

Let me know what you think! I'm glad to know that `flow-language-server` basically works though, that's really helpful! Thanks for exploring this!

Akin909:
@bryphe 😱 didn't realise it was so enourmous fair enough, I'll refactor it out and we can add documentation on how to install it and the pattern for checking the language server looks much tidier so I'll swap to that thanks for the suggestion.

Just as a side note given the popularity of *flow* it think it ought to be super discoverable how to set it up as I think the majority of javascript users of oni will expecting to be able to use flow without too much difficulty

Akin909:
Added the following to the documentation
> Javascript - Flow
> If you would like to add flow support (a static type check for javascript), you will need to add the > > flow-language-server, This can be done by running yarn global add flow-language-server or npm > install -g flow-language-server, then adding the following lines to your oni config.js
 ```
	'language.javascript.languageServer.command': 'flow-language-server',
	 'language.javascript.languageServer.arguments': ['--stdio'],
```

File: browser/src/UI/components/ToolTip.tsx [hunk 1]
```diff
@@ -19,16 +19,16 @@ export interface IToolTipsViewProps {
 export class ToolTipsView extends React.PureComponent<IToolTipsViewProps, {}> {
 
     public render(): JSX.Element {
-        const toolTipElements = this.props.toolTips.map((toolTip) => {
-            return <ToolTipView {...toolTip} borderColor={this.props.borderColor} foregroundColor={this.props.foregroundColor} backgroundColor={this.props.backgroundColor}/>
+        const toolTipElements = this.props.toolTips.map((toolTip, i) => {
+            return <ToolTipView {...toolTip} key={`${i}-${toolTip.id}`} borderColor={this.props.borderColor} foregroundColor={this.props.foregroundColor} backgroundColor={this.props.backgroundColor}/>
```
bryphe: For this key - could we just use `toolTip.id` without the `i` parameter? Some of the UI in the tool tips is sensitive to repaints (like having both the _rename_ and _quickinfo_ dialog open). There's actually a bug now where if you open Rename, and the QuickInfo pops up, the Rename gets re-rendered and the focus gets messed up... If we just use `toolTip.id`, I believe it will fix that bug.

browser/src/UI/components/BufferScrollBar.tsx
```diff
@@ -72,12 +72,12 @@ export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps,
                 width: "100%",
             }
 
-            return <div style={markerStyle} key={m.line.toString() + m.color}/>
+            return <div style={markerStyle} key={`${m.line}-${m.color}-${i}-${top}`}/>
```
bryphe: I'm not sure if this is actually the correct fix for this - this seems like it could cause aggressive re-paints for the buffer scroll bar (which has caused performance issues). I think we may want to try putting a key on the scroll bar itself (or even the `ActiveWindow`). Could we track this in a separate issue / PR?

bryphe:
Sweet, thanks @Akin909 for making the changes - the TS/JS changes look great to me. And really appreciate the documentation updates!

I saw that some of the `key` parameters are updated as well in the latest PR - it might be easier to handle those in a separate change (so that we can get the flow stuff in). There are a couple of key changes that I'm concerned may regress perf, so I'd like to pull those out into a separate change if possible.

Akin909:
@bryphe that sounds good I definitely don't want to muddy this issue with key stuff especially given that the wrong keys can really make things worse I think you right that there needs to be an issue for keys though because it very hard to be sure a list of components are being rendered properly.

I'll remove the key sections and leave the changes to the ts plugin

bryphe:
> I'll remove the key sections and leave the changes to the ts plugin

Awesome, thanks!

> think you right that there needs to be an issue for keys though because it very hard to be sure a list of components are being rendered properly.

Ya, that sounds good. I wonder if there is a way we can check for this in our automation - I accidentally regressed some of the key fixes - it'd be cool if there was a way to 'protect' against this as part of our safety net in our automation / ci tests. We might be able to check and see if there are any debug errors in the console while we run the integration tests.

Akin909:
Removed most of the key changes save the tool tip id one in case that solves the bug you alluded to

bryphe:
Excellent, looks great @Akin909  - thank you! Bringing it in now.

PR #1192, opened 2017-12-23
Adding key property to tooltip

anned20:
This fixes a error/warning from React that 'key' is required

### DIFF
--- a/browser/src/UI/components/ToolTip.tsx
+++ b/browser/src/UI/components/ToolTip.tsx
@@ -17,10 +17,11 @@ export interface IToolTipsViewProps {
 }
 
 export class ToolTipsView extends React.PureComponent<IToolTipsViewProps, {}> {
-
     public render(): JSX.Element {
         const toolTipElements = this.props.toolTips.map((toolTip) => {
-            return <ToolTipView {...toolTip} borderColor={this.props.borderColor} foregroundColor={this.props.foregroundColor} backgroundColor={this.props.backgroundColor}/>
+            const key = toolTip.options.position.pixelX + toolTip.options.position.pixelY
+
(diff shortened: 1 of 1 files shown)


browser/src/UI/components/ToolTip.tsx
```diff
@@ -17,10 +17,11 @@ export interface IToolTipsViewProps {
 }
 
 export class ToolTipsView extends React.PureComponent<IToolTipsViewProps, {}> {
-
     public render(): JSX.Element {
         const toolTipElements = this.props.toolTips.map((toolTip) => {
-            return <ToolTipView {...toolTip} borderColor={this.props.borderColor} foregroundColor={this.props.foregroundColor} backgroundColor={this.props.backgroundColor}/>
+            const key = toolTip.options.position.pixelX + toolTip.options.position.pixelY
```
badosu: Are `pixelX` and `pixelY` numbers? If so this can have ambiguous keys, e.g. X=50, Y=50 | X=100, Y=0 | etc.

Maybe `${pixelX},${pixelY}` would be better.

anned20:
@badosu fixed it. Is this what you meant? I added X: and Y: for clearness and to make it even possible for the X and Y values to have a , in them

badosu:
@anned20 Yes, thanks!

bryphe:
Thanks for the PR, @anned20 and your help@badosu . Sorry for the late reply - still getting caught up after the holidays.

I just looked at this, and noticed that @Akin909 actually fixed the `key` issue too around the same time:
https://github.com/onivim/oni/commit/bf65627a45d250c90092421f10e6f751b0c556cf

So it looks like this is no longer needed, because that fix is in place in master. Apologies that there was a 'race' here... thanks again for taking the time to contribute to the project!

PR #1198, opened 2017-12-29
Feature/add focus gained and focus lost hooks

### LINKED ISSUE #1109: Plugin API: Add support for `onFocusGained` and `onFocusLost` event
From PR #1106 - it'd be nice to have a way to get focus gained / focus lost events for plugins.

Perhaps something on the `Workspace` API, like:

`oni.workspace.onFocusGained.subscribe(() => { ... })`
`oni.workspace.onFocusLost.subscribe(() => { ... })`

These would basically be proxies to the browser window focus events `blur` and `focus`: https://github.com/electron/electron/blob/master/docs/api/browser-window.md#event-blur

As part of this, we should update the git statusbar plugin to listen to these events to update the branch state.

Akin909:
This PR adds a `focusGained` and `focusLost` hooks to oni. The events themselves pass in the lastActiveBuffer on focusLost and on focusGained. 

Also subscribed the Oni git plugin to this event so in the case of changing git branch etc outside oni the plugin is correctly updated on returning to oni.

Fixes #1109

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -35,8 +35,19 @@ import { Completion } from "./../../Services/Completion"
 import { Configuration, IConfigurationValues } from "./../../Services/Configuration"
 import { IDiagnosticsDataSource } from "./../../Services/Diagnostics"
 import { Errors } from "./../../Services/Errors"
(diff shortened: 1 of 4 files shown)


bryphe:
Just tested this out - looks great, @Akin909 ! 👍 Will bring it in now.

Thanks for the contribution, and happy new year's eve! 🎉

Issue #1083, opened 2017-12-06
Visual artifacts below searched words


badosu:
See the underline below the words searched, they persist even after `:nohlsearch`:

![screenshot from 2017-12-06 13-43-04](https://user-images.githubusercontent.com/347552/33670146-85cd50bc-da8b-11e7-9672-1f54994199b6.png)

bryphe:
Interesting, looks like this might be subpixel rendering issues.

We do some work to align boundaries of rectangles to the nearest pixel here:
https://github.com/onivim/oni/blob/23822545b4aaff55c61ac58a5b7e8a4906d1f3e3/browser/src/Renderer/CanvasRenderer.ts#L259

But only for the horizontal (`x`) coordinate... potentially we need to do a similiar normalization for the `y` coordinate as well.

Akin909:
@badosu I'm currently looking at fixing this bug, but personally haven't really been able to reproduce it as I think it likely relates to the font you use and its font size, In order to test out the fix could I possibly get details about the font you are using in the screen shot and what the font size is?

CrossR:
I'm able to reproduce this with the default Oni font settings, opening any Oni .ts file, pressing `*` on `import`, and then using the `:nohlsearch` command.

![image](https://user-images.githubusercontent.com/10038688/34537861-e9497232-f0c1-11e7-9823-a35c20ff1793.png)

Akin909:
@CrossR thanks for the tip, I've just had a quick try with the default font and the instructions you suggested but nothing (unusual) happens when I do that, although looks like the cracks youre getting in your screen shot are related to the width (which is normalised) as opposed to the height which I think means that based on font and maybe font size the normalization doesn't cover enough scenarios not that I know too much  about this bug though @bryphe explained what is likely causing it in #1203.

I personally was getting a very slight visual artefact, possibly a very minimal version of @badosu's screenshot but it was occurring under words so I thought the bug only related to the lack of normalisation for the height prop 😕 , although there's also the clipping which you have at the end of words on the screen shot which I also get which I'm desperate to fix which probably also relates to width normalisation not working in some cases

badosu:
@Akin909 I am not able to reproduce this issue on the last stable version (0.2.19) anymore, I'll try checking if there's a way to reproduce them again.

CrossR:
Since moving to the new release, this has come back for me unfortunately.

Steps for me to reproduce:
- Highlight any word with `*`
[...]

PR #1203, opened 2017-12-30
bugfix/render-italics-and-bold-text

### LINKED ISSUE #878: Bold and italic font not working
Hi,

I just installed oni, and it really works good. I just noticed that both bold and italic font are not showing.
Here's screenshot from neovim:
neovim

And a screenshot from oni
oni

I'm using [Iosevka Term](https://github.com/be5invis/Iosevka) font.

Akin909:
This PR fixes #878 by adding the metadata re bold and italics from neovim to the cell data and using that when the text is rendered.

Note: I haven't included underline in the final render as that seems to be another beast but theoretically one could use this [SO](https://stackoverflow.com/questions/4627133/is-it-possible-to-draw-text-decoration-underline-etc-with-html5-canvas-text) answer which is a work around for not having a text-decoration style in canvas but wasn't sure whether the underline is wanted or whether it could/would be implemented using something similar to the error squiggles

### DIFF
--- a/browser/src/Font.ts
+++ b/browser/src/Font.ts
@@ -7,6 +7,7 @@ export interface IFontMeasurement {
 
 export function measureFont(fontFamily: string, fontSize: string, characterToTest = "H") {
     const div = document.createElement("div")
+
     div.style.position = "absolute"
     div.style.left = "10px"
     div.style.top = "10px"
@@ -27,8 +28,7 @@ export function measureFont(fontFamily: string, fontSize: string, characterToTes
     const height = rect.height
 
     document.body.removeChild(div)
-    console.log("isItalicAvailable: ",  isItalicAvailable) // tslint:disable-line
-    console.log("isBoldAvailable: ",  isBoldAvailable) // tslint:disable-line
+
     return {
         width,
         height,
@@ -38,20 +38,14 @@ export function measureFont(fontFamily: string, fontSize: string, characterToTes
 }
 
 export function addDefaultUnitIfNeeded(fontSize: string) {
-
-    const defaultUnit = "px"
-
-    if (isNaN(Number(fontSize))) {
-        return fontSize
-    } else {
-        return fontSize + defaultUnit
-    }
+    const roundFont = `${Math.round(parseFloat(fontSize))}px`
+    return roundFont
 }
 
 export function isStyleAvailable(fontName: string, style: string, fontSize = "12px") {
+    const text = "abcdefghijklmnopqrstuvwxyz0123456789"
     let canvas = document.createElement("canvas")
     const context = canvas.getContext("2d")
-    const text = "abcdefghijklmnopqrstuvwxyz0123456789"
     context.font = `${fontSize} ${fontName}`
     const baselineSize = context.measureText(text).width
     context.font = `${style} ${fontSize} ${fontName}`
--- a/browser/src/Font.ts
+++ b/browser/src/Font.ts
@@ -5,33 +5,35 @@ export interface IFontMeasurement {
     height: number
 }
 
-export function measureFont(fontFamily: string, fontSize: string, characterToTest?: string) {
-    characterToTest = characterToTest || "H"
+export function measureFont(fontFamily: string, fontSize: string, characterToTest = "H") {
     const div = document.createElement("div")
-
     div.style.position = "absolute"
     div.style.left = "10px"
     div.style.top = "10px"
     div.style.backgroundColor = "red"
     div.style.left = "-1000px"
     div.style.top = "-1000px"
-
     div.textContent = characterToTest
     div.style.fontFamily = `${fontFamily},${FallbackFonts}`
     div.style.fontSize = fontSize
 
+    const isItalicAvailable = isStyleAvailable(fontFamily, "italic", fontSize)
+    const isBoldAvailable = isStyleAvailable(fontFamily, "bold", fontSize)
+
     document.body.appendChild(div)
 
     const rect = div.getBoundingClientRect()
-
     const width = rect.width
     const height = rect.height
 
     document.body.removeChild(div)
-
+    console.log("isItalicAvailable: ",  isItalicAvailable) // tslint:disable-line
+    console.log("isBoldAvailable: ",  isBoldAvailable) // tslint:disable-line
     return {
         width,
         height,
+        isItalicAvailable,
+        isBoldAvailable,
     }
 }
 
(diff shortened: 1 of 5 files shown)


Akin909:
Another side note: I've noticed that the syntax highlighter i.e. oni's additional one seems to add bolding to certain words overwriting vim's bolding or italics which could very well be intentional though not sure if it is expected so raising it here.

bryphe:
Wow, awesome  @Akin909 ! 👍  Sorry I've been slow to go through PRs this week. Can't wait to check this out! I'll be travelling the next couple days but back in action soon. Thanks for looking at this!

> Note: I haven't included underline in the final render as that seems to be another beast but theoretically one could use this SO answer which is a work around for not having a text-decoration style in canvas but wasn't sure whether the underline is wanted or whether it could/would be implemented using something similar to the error squiggles

Seems reasonable - sounds like a lot of extra work since `text-decoration` isn't supported. (Didn't realize that, that's unfortunate). But yes, the SO answer looks promising - we could render it by drawing a line under the text. I think that'll be the better layer / more performant place to implement this.

> Another side note: I've noticed that the syntax highlighter i.e. oni's additional one seems to add bolding to certain words overwriting vim's bolding or italics which could very well be intentional though not sure if it is expected so raising it here.

Interesting, this isn't expected - but we do use some built-in syntax highlight groups (like `Identiifer`), so if those happen to have bold/italics defined, it would pick those up. Let me know if there is a specific case you're seeing this with - we can look at the highlight group being used.

Akin909:
@bryphe no worries 👍 enjoy the break, just responding here for when you're free.

*Re-underline* - __TLDR: its going to require a bit of work which I'd rather look into separately, not as part of this PR__ - the SO function works well just tried it locally though there's issues around when the underline highlight is triggered ?by neovim for example using spellbad with underline typing `sout` instead of `shout` the highlight is not consistently set to true once a typo occurs nor frequently updated so adding more characters or removing them doesnt reduce the size of the highlight as expected,

*Re-syntax highlighting* - spoke too soon as I have quite a few plugins and viml ad hoc highlighting settings in my `init.vim`, but to note it does set things like function definitions or methods to bold not sure what type these are as I dont get one specific highlight I can track, tbh its a pleasant unexpected effect imho and more related to highlight group that the textmate highlighting sets. As for user defined highlights these aren't actually overridden in my case it seems that I need to source my vimrc for it to pick up my adhoc highlighting but i'm not sure what that relates to or that its actually a part of this but rather a consequence of how oni loads in `init.vim`s

bryphe:
Regarding underline, sounds great - IMO underline is less important than bold/italic (to me at least). Happy to track that in a separate issue!

> Update: the odd syntax highlighting not corresponding to my init.vim was due to an antipattern I was using regarding applying add hoc highlights which worked (probably incorrectly) in terminal vim but not in oni which i've now resolved so don't think that issue of highlights not showing up is actually due to oni

Cool, glad you discovered the root cause! 😄 

The code looks great to me. I just tested this change out on master, and I see some clipping of text with my config:
image

I'm using these settings:
image

If I use the default, everything looks fine:
image

I believe this may be a problem with my 'Fira Code' font - perhaps it isn't able to load the bold monospace version, and it's messing up the character widths. But we might need to add a check for this in the `measureFont` method in `Font.ts` - like validate that the bold and italic version of the string also have the same width, and pass up a `isBoldAvailable` and `isItalicAvailable` - it'd be better not to render bold/italics if the font doesn't support it (and perhaps log a warning in the console).

Thanks for all your work on this, @Akin909 ! 👍

Akin909:
@bryphe I think there is a slightly larger issue with font clipping in oni in general, I initially noticed that nerd font glyphs were being clipped which is an issue in `macvim` and in some terminal emulators which i think relates to an issue raised by @CrossR though I can't find the one (in that case it was special characters not rendering) I thought this was related to `wcwidth` and how it calculated character widths for unicode8-9 symbols

Although I've also noticed the font clipping using a regular font prior to this although it was minimal especially in the case of wider characters like `w` had slight clipping at the edge (using Fira Code Nerd Font at 17px), although for this specific case as you suggest we can check that the appropriate bold and italic fonts exists as it very well might not as you point out.

screen shot 2017-12-31 at 17 53 47

Here is a screen shot of the PR in action but if you look at the `ref(0)` italic the f is clipped which I believe is a separate issue as the italic font is available in this case

Akin909:
I've added the `isBoldAvailable` and `isItalicAvailable` check although the problem persisted despite it. Following some digging it seems that the html canvas element doesn't consistently manage non-integer font sizes. In IE the decimals are preserved, but in chrome the fonts are rounded to the nearest integer. So the error you described is actually specific to decimal font sizes.

Given that the canvas in chrome rounds decimal numbers (for font size) I've rounded the values passed from the config. Essentially there's also a peculiarity re how the `canvasContext.font` works which is what is responsible for the change [see here](https://stackoverflow.com/questions/38823353/chrome-canvas-2d-context-measuretext-giving-me-weird-results). 

The font used to calculate the width when using a decimal font is essentially smaller than the rounded font used by canvas causing the smooshing. Setting a font size consistent with canvas resolves this.

The slight clipping I mentioned earlier persists it's nowhere near as noticeable and it predates this PR. Tbh I'm still not clear on what the cause might be but it seems to relate to the amount of space each word is allotted as it only slightly clips the last char of a word.

bryphe:
> Given that the canvas in chrome rounds decimal numbers (for font size) I've rounded the values passed from the config. Essentially there's also a peculiarity re how the canvasContext.font works which is what is responsible for the change see here.

Cool, thanks for the investigation @Akin909 ! 

> The fontSize used to calculate the width when using a decimal font is essentially smaller than the rounded font used by canvas causing the smooshing. Setting a font size consistent with canvas resolves this.

Great find!

> The slight clipping I mentioned earlier persists it's nowhere near as noticeable and it predates this PR. Tbh I'm still not clear on what the cause might be but it seems to relate to the amount of space each word is allotted as it only slightly clips certain characters.

Ah yes - this is related to #1083 , as well as the code here: https://github.com/onivim/oni/blob/c62d1e32f620124cf571a16aae1f3600b9e068e5/browser/src/Renderer/CanvasRenderer.ts#L259

The problem we have is with 'subpixel rendering' in the canvas - the font characters end up being fractional sizes. The fonts actually render fine in this case on the canvas, but the `clearRect` and `fillRect` actually do [antialiasing](https://www.html5rocks.com/en/tutorials/internals/antialiasing-101/).  This is also a good example: https://www.grc.com/ctwhat.htm

The problem is, on sub-pixel boundaries, when we call `fillRect`, it will actually draw _past_ the pixel boundary transparently, as this is how anti-aliasing algorithms work. Then, when we `clearRect`, it would miss part of the anti-aliased rectangle, leaving lines and gaps like in #1083 . The fix for this was to round the fillRect and clearRect call to the nearest pixel boundaries - but in some cases that catches the edge of the character. It's dependent on the font size how often this appears, but if it happens its 'only' 1px - and definitely predates this change. 

Here's a quick fiddle to demo the issue:
http://jsfiddle.net/V92Gn/5684/

I'm using:
```
        context.fillStyle = "red"
        context.fillRect(40.5, 20.5, 400.5, 200.5)
        context.clearRect(40.5, 20.5, 400.5, 200.5)
```

And I get:
image

Which you might expect from the code would clear it entirely - but due to the anti-aliasing, the filled rectangle extends slightly out of the border of clearRect. Same exact issue we have here!

bryphe:
Great work on this, @Akin909 - thank you for digging in and finding the cause for that clipping! I tried it out and it looks great with the font families/sizes I tried. Will bring this in now!

PR #1210, opened 2018-01-02
Feature: pass priority from config

Akin909:
Following discussion in #1114 this PR contains a smaller subset of proposed changes to the status bar config i.e. changes to the status item creation api. Priority and item Ids are now obtained via the users config.

PR #1211, opened 2018-01-02
Feature: Add Status bar truncation

Akin909:
As per #1184, this PR is another subset of the changes from that PR. This adds the resizing logic and components as well as converts this component to use styled components

NOTE: This PR still uses the oni-api 0.0.9 but is not dependent on changes made in #1210 I believe which is where I incorporate the new version of the oni-api


Akin909:
@bryphe weirdly enough separating this PR seems to have bypassed whatever was failing. This PR only includes the core logic for the resizing and the change to styled components for the statusbar so I imagine somewhere amidst some of the other smaller changes I made along the way I might have triggered the bug that was causing the failure which these changes do not include thankfully

bryphe:
> @bryphe weirdly enough separating this PR seems to have bypassed whatever was failing

Weird! Thank you so much for taking the time to separate it out. 

> This PR only includes the core logic for the resizing and the change to styled components for the statusbar so I imagine somewhere amidst some of the other smaller changes I made along the way I might have triggered the bug that was causing the failure which these changes do not include thankfully

Definitely possible. There is also dependency caching happening on the TravisCI server that can sometimes be out-of-date (it caches the yarn dependencies to speed up build times)... I think in some cases, when all of the CiTests fail, it can be related to that. Glad it's working now though!

bryphe:
Just tested this out, and it's working great for me @Akin909 . Makes the statusbar feel much less 'janky' - thank you for taking this on! 💯 Awesome stuff. Will bring it in now.

PR #1224, opened 2018-01-03
[WIP]Bugfix: rendering cracks

### LINKED ISSUE #1083: Visual artifacts below searched words

Akin909:
This PR normalises the height value in `CanvasRenderer.ts` to match the normalised width value, in attempt to fix #1083 I only very minimally experience this issue so I'm not sure that it will completely cover the scenario that the issue refers to as personally I can't recreate the exact bug to the same extent though @CrossR  has reproduced a similar issue which looks like it pertains to the width normalization 😦 . 

In response to that I've added rounding (`Math.ceil`) to the char width values which improves the letter clipping I mention in #878 though there are still a few edge cases. On slightly increasing the width of each char manually I'm able to resolve the edge cases (this ruins the spacing of other fonts though) so I think the font measurement function doesn't produce a value that covers certain wide space chars like nerd font/ unicode icons.

screen shot 2018-01-03 at 22 30 59

Its probably hard to make out but the line wrap arrow is supposed to be ↳  which still does not render correctly


browser/src/Font.ts
```diff
@@ -24,8 +24,8 @@ export function measureFont(fontFamily: string, fontSize: string, characterToTes
     document.body.appendChild(div)
 
     const rect = div.getBoundingClientRect()
-    const width = rect.width
-    const height = rect.height
+    const width = Math.ceil(rect.width)
```
bryphe: Unfortunately this `width` change is a bit problematic. The issue is that when we render whole tokens, the spacing between the letters isn't taken into account. You can see how the cursor is slightly off, and it gets worse letter-by-letter, until it's completely off by the end of the word:
width-distortion

bryphe: I tried a fix along these lines, and was able to address this particular issue by adding a 'kerning' value to add space between all the letters - using the difference between `Math.ceil(rect.width)` and `rect.width`. However, it actually looked really bad (unfortunately), so although it did address the cracks / subpixel rendering, it looked pretty bad.

browser/src/Renderer/CanvasRenderer.ts
```diff
@@ -278,12 +279,16 @@ export class CanvasRenderer implements INeovimRenderer {
         const delta = boundsStartX - normalizedBoundsStartX
         const normalizedBoundsWidth = Math.ceil(boundsWidth + delta)
 
+        const normalizedBoundsY = Math.floor(boundsY)
```
bryphe: The height fix seems reasonable to me, though - we add a pixel padding in between the lines anyway, so some fraction to round it up to a whole number is good. Thanks for looking at this!

bryphe:
I think the test failures are unrelated to your change - it does look like this is a bit behind master though (with all the `oni-api` version bumps 😄 ) I'm curious if bringing in the latest master works?

Akin909:
@bryphe fair enough seems like difficult bug to fix I've reverted the changes re. math.ceil and left the y axis normalisation, not sure I have any ideas re other approaches for the time being.

bryphe:
Definitely, it's not trivial! We could add a setting to use the ceiling + set a font kerning value, if that would preferable to the 1px clipping (which it may be for certain configurations).

Another thing to note is that if we re-rendered the entire canvas every time, we wouldn't have this issue - it's an artifact of the fact that `clearRect` and `fillRect` don't match up 100% in the subpixel case (and if we re-rendered the entire canvas every time, we would only need `fillRect` and not `clearRect`).

Also, one thing I've wanted to try with `CanvasRenderer` is to use `drawImage` instead of `fillRect` - It seems like its significantly [faster](https://jsperf.com/canvas-draw-image-vs-fill-rect/25), and it might exhibit different properties in terms of anti-aliasing (especially via tweaking the [`imageSmoothingEnabled`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled) context property).

It also turns out that using `drawImage` is _significantly_ faster than using `fillText` - so that is something else I wanted to look at at some point. It'd be cool if we kept a cache of common letters / tokens, and could render via `drawImage`.

bryphe:
Ran the tests locally and everything looked good - so I'll bring this in. Thanks for your help with this, @Akin909 ! 💯

Akin909:
@bryphe I like the solution of re-rendering the canvas (not sure if there'd be significant perf implications) or using drawImage (if that possible I dont know if text can be drawn as an image ?) sound quite interesting although perf benefits look huge 😱

PR #1227, opened 2018-01-04
Fix #963 - Delete word in 'lightweight' text

### LINKED ISSUE #963: Ctrl-w to delete word on the left of cursor
Another muscle memory habit that I miss while using oni. While searching for files in search popup pressing `Ctrl-W` doesn't erase the word on the left of cursor. Would be fantastic if this binding could be added.

bryphe:
__Issue:__ There are a few cases today where we use a vanilla input box instead of vim integration. This includes the rename dialog and quickopen. In these cases, the normal insert-mode muscle memory (`control+h` - delete one character, `control+w` - delete word, `control+u` - delete line) doesn't work as expected.

__Fix:__ Bake in these bindings into the `TextInputView` (lightweight text editor). This isn't perfect, and the `WordRegEx` may need some tweaking. At some point, it would be nice to replace this with a neovim-backed edit experience.

### DIFF
--- a/browser/src/Services/Menu/MenuComponent.tsx
+++ b/browser/src/Services/Menu/MenuComponent.tsx
@@ -15,6 +15,8 @@ import { IMenuOptionWithHighlights, menuStore } from "./Menu"
 import * as ActionCreators from "./MenuActionCreators"
 import * as State from "./MenuState"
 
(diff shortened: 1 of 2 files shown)


bryphe:
Fixes #963

PR #1230, opened 2018-01-04
Feature/add search icon commandline

Akin909:
@bryphe  as per discussion in #1191 I've also added a config managed option to render an icon in place of the first character. It's off by default and under the `experimental.commandline.icons`. There's currently an icon for 
`:`

screen shot 2018-01-04 at 21 28 22

 `?`

screen shot 2018-01-04 at 21 28 35

 and `/`.
  
screen shot 2018-01-04 at 21 28 43


bryphe:
Cool idea! Thanks for putting this together, @Akin909 !

Code change looks good to me - appreciate you putting it behind an experimental flag. I'm going to push out a new release next week, will be awesome to start getting feedback on these 👍 

The search next and search previous icons look great. I'm not sure about the `:` icon (the `code` icon is a bit ambiguous), but I also don't have a better suggestion at the moment 🤔

bryphe:
Just checked the tests - passing locally again, so I'll go ahead and bring this in. Thanks for thinking about this!

I'm working on a change to get the console logs a different way (it seems in some cases we don't get any logs, so hopefully this change will fix that, and let us get to the next level of investigation). Sorry about all the test issues.

PR #1236, opened 2018-01-05
Performance (Perceived): Separate loading / entrance for editor

bryphe:
Today, we show a loading 'overlay' (a blank screen) on top of the entire shell over the course of loading. 
We remove that overlay at the end of the `Oni.Start.Editors` sequence - after all the neovim instances have been spun up and fully initialized.

This means we don't show anything in the UI until we get to here:
image

This is by-design, because the first render of the canvas is rocky, since we see the colorscheme switch.

However, other parts of the UI are already rendered at this point - like the statusbar (and sidebar, later!). So we can pull the curtain up for those elements, and then have a separate overlay for the `NeovimEditor`. In effect, we're dividing the entrance into two stages:
image

In the first arrow, we'll uncover the shell, so that we can see progress there (animate in statusbar, sidebar, etc).

In the second arrow, we'll uncover the editor pane (with a slight entrance animation).

Even though the absolute time to render is no different, it _feels_ better, because you see progress in the UI state, and the subtle animations help reinforce this.

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -561,6 +561,8 @@ export class NeovimEditor extends Editor implements IEditor {
             await this._openFiles(filesToOpen, ":tabnew")
         }
 
+        this._actions.setLoadingComplete()
+
         this._hasLoaded = true
         this._isFirstRender = true
         this._scheduleRender()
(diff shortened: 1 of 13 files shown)


bryphe:
One issue is there seems to be some weirdness with the `github` icon in the statusbar - as it is rendered before the resize and shows up near the center of the window. It might be worth moving this to our plugin as opposed to special casing it.

PR #1241, opened 2018-01-06
Feature/hide app shortcut on macOS

### LINKED ISSUE #1193: OS X close Oni window (but not quit) then open new file causes error
On OS X, If you close the Oni window (but don't quit it using CMD + Q or Oni > Exit) such that Oni is still running, and then go to Oni > New File (or Open File, or Open Folder), Oni will give this error:

**A JavaScript error occurred in the main process**
~~~
Uncaught Exception:
Error: Object has been destroyed
    at executeVimCommand (/Applications/Oni.app/Contents/Resources/app/lib/main/src/menu.js:19:54)
    at click (/Applications/Oni.app/Contents/Resources/app/lib/main/src/menu.js:67:21)
    at MenuItem.click (/Applications/Oni.app/Contents/Resources/electron.asar/browser/api/menu-item.js:52:9)
    at Function.executeCommand (/Applications/Oni.app/Contents/Resources/electron.asar/browser/api/menu.js:121:15)
~~~

Tested on a Macbook Pro 2015 (12, 1)

I looked through issues for someone else mentioning this error, but I couldn't find anything, so I apologize if this is a duplicate.

### LINKED ISSUE #1196: Hide App (⌘-H) not working on macOS
No menu option available to hide the Oni application completely in macOS, and the expected shortcut <kbd>⌘</kbd>+<kbd>H</kbd> is not working.
Gonna try to fix this myself now.

Akin909:
Add `Cmd-H` shortcut to hide oni window as well as a menu option fixes #1196 
 
@bryphe  whilst working on this I observed that when hitting the close button on oni the window is closed but the app remains open in a sense (which I believe we discussed previously is default mac behaviour) however on activate when then icon is clicked for example **after** closing with the button a new oni window is created, as part of this I realised that the close button could function as a hide button so that the reference to the window still exists and it can be reused on activate. This would also theoretically fix #1193 which I imagine is occuring because though the app is running there is no window to reference

### DIFF
--- a/browser/src/Input/KeyBindings.ts
+++ b/browser/src/Input/KeyBindings.ts
@@ -29,6 +29,7 @@ export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configurati
         input.bind("<m-t>", "language.symbols.workspace", () => !menu.isMenuOpen())
         input.bind("<s-m-t>", "language.symbols.document")
         input.bind("<m-m>", "oni.editor.minimize")
+        input.bind("<m-h>", "oni.editor.hide")
 
         if (config.getValue("editor.clipboard.enabled")) {
             input.bind("<m-c>", "editor.clipboard.yank", isVisualMode)
--- a/browser/src/Services/Commands.ts
+++ b/browser/src/Services/Commands.ts
@@ -8,7 +8,7 @@ import * as fs from "fs"
 import * as os from "os"
 import * as path from "path"
 
-import { clipboard, remote } from "electron"
+import { clipboard, remote, app } from "electron"
 
 import * as Oni from "oni-api"
 
@@ -50,7 +50,7 @@ export const registerBuiltInCommands = (commandManager: CommandManager, neovimIn
 
         new CallbackCommand("oni.editor.minimize", "Minimize Window", "Minimize the current window", () => remote.getCurrentWindow().minimize()),
 
-        new CallbackCommand("oni.editor.hide", "Hide Window", "Hide the current window", () => remote.getCurrentWindow().hide()),
+        new CallbackCommand("oni.editor.hide", "Hide Window", "Hide the current window", () => app.hide()),
 
         // Language service
         // TODO: Deprecate
--- a/browser/src/Services/Commands.ts
+++ b/browser/src/Services/Commands.ts
@@ -8,7 +8,7 @@ import * as fs from "fs"
 import * as os from "os"
 import * as path from "path"
 
-import { clipboard, remote, app } from "electron"
+import { clipboard, remote } from "electron"
 
 import * as Oni from "oni-api"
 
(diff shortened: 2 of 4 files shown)


Akin909:
@benjamin-rood apologies didn't mean to trod on your toes there's it's been a tiny bug bear of mine the hide command hopefully this PR solves this issue

benjamin-rood:
Haha no problem! Sorry I have been sick with a chest flu this week, that’s why I haven’t got it done.

benjamin-rood:
Just looking at the code, it’s important to note that this does not comply with the macOS user interface guidelines nor what users expect. I get, however, that it’s a compromise to try to make the menu cross-platform friendly without creating special cases for the menu when the OS platform is `darwin`.  If you look at how the other established Electron apps do it (like VS Code or the example Electron app) they build the menu dynamically at runtime by wrapping the menu items in an abstraction. 

Would you be keen for me to clone your fork so we can add that functionality together?

There’s also some missing functionality, namely the “Hide Others” and “Show All”

The menu items need to be:

    ————
    Hide Oni
    Hide Others
    Show All
    ————
    Quit Oni

benjamin-rood:
@bryphe Could also merge this as-is and do the extra work as a separate PR, too. 
Anyway, that’s just my two cents.

Akin909:
@benjamin-rood I think my ambition with this PR was smaller than you had intended for #1196, I was primarily just trying to get the shortcut working but I've had a look at vscode's menu and you're right re the way they've implemented it. Tbh I'm fine either way i.e. if @bryphe wants to merge this as is or if you wanted to make a new PR (I haven't really made that many changes so feel free to use this branch or make another), I was also looking at the `accelerator` options for the menu items which I don't think we've used so far but seem like an easy way to get native bindings for menu commands.

benjamin-rood:
Doing it separately makes sense unless @bryphe doesn't want to merge it. Again, apologies for being sick.

bryphe:
> Doing it separately makes sense unless @bryphe doesn't want to merge it. Again, apologies for being sick.

I'm on board with bringing this in, too - the behavior with this PR is better than before.

> If you look at how the other established Electron apps do it (like VS Code or the example Electron app) they build the menu dynamically at runtime by wrapping the menu items in an abstraction.

I like this suggestion a lot as a follow-up item! The way we have the menu generation setup today is untenable in the long run. If we make it dynamic, we could also have an entry point for plugins to add options to our application menu. I could also see scenarios where, once we have per-project configuration, we might want to add menu items as part of workspace settings (for Oni, it'd be cool if we had an entry point in the menu to run unit tests or ci tests, potentially 😄 )

It seems like what we'd potentially want is, when a window is created, it would send some information / metadata from the _renderer_ (browser) process to the _main_ process to give it context to create the menu. The _renderer_ process has more flexibility, because it knows about plugins created, and what commands are available.

In terms of a model for plugins, sublime has a nice JSON-type format for specifying menu options: http://docs.sublimetext.info/en/latest/customization/menus.html

For our plugins, we could add a `menu` section to `contributes` and allow specifying this.

bryphe:
BTW, hope you are feeling better @benjamin-rood !

bryphe:
Merged - thanks @Akin909 for the contribution, and @benjamin-rood for the feedback!

PR #1242, opened 2018-01-06
Add tests for markdown preview

TalAmuyal:
So, for the past few days I worked on making plugins testable (with the `Markdown-preview` plugin as a test subject) using a mock for `Oni` itself.

@bryphe, I currently seek feedback.

The goal is to have tests for the **behavior** of the plugin itself, assuming `Oni` works correctly.

The changes in this PR contains:

- Refactoring singleton usage (received at construction time)
- Make a mock `Oni` class for plugins' tests
- Add plugin's test-runner
- Add commented-out tests I would like to implement for the Markdown-preview plugin

In addition, what do you think about:
1. The general approach taken in the PR (please code-review)
2. Taking a step forward in refactoring the singleton-usage by eliminating the singletons' `export`
3. Move more `type`-'s, `enum`-'s & `interface`-'s to the API in order to implement the behavior tests (should also be useful for implementing other plugins)

### DIFF
--- a/browser/src/Performance.ts
+++ b/browser/src/Performance.ts
@@ -1,28 +1,35 @@
 
+export interface IPerformance {
+    mark(markerName: string): void
+    startMeasure(measurementName: string): void
+    endMeasure(measurementName: string): void
+}
 /**
  * Thin wrapper around browser performance API
  */
-export function mark(markerName: string): void {
-    if (typeof window === "undefined") {
-        return
+export class Performance implements IPerformance {
+    public mark(markerName: string): void {
+        if (typeof window === "undefined") {
+            return
+        }
+
+        if (process.env.NODE_ENV === "production") {
+            return
+        }
+
+        performance.mark(markerName)
+
+        const anyConsole: any = console
+        anyConsole.timeStamp(markerName)
+
+        console.log(`[PERFORMANCE] ${markerName}: ${performance.now()}`) // tslint:disable-line no-console
     }
 
-    if (process.env.NODE_ENV === "production") {
-        return
+    public startMeasure(measurementName: string): void {
+        console.time(measurementName) // tslint:disable-line
     }
 
-    performance.mark(markerName)
-
-    const anyConsole: any = console
-    anyConsole.timeStamp(markerName)
-
-    console.log(`[PERFORMANCE] ${markerName}: ${performance.now()}`) // tslint:disable-line no-console
-}
-
-export const startMeasure = (measurementName: string): void => {
-    console.time(measurementName) // tslint:disable-line
-}
-
-export const endMeasure = (measurementName: string): void => {
-    console.timeEnd(measurementName) // tslint:disable-line
+    public endMeasure(measurementName: string): void {
+        console.timeEnd(measurementName) // tslint:disable-line
+    }
 }
--- a/browser/src/Plugins/Plugin.ts
+++ b/browser/src/Plugins/Plugin.ts
@@ -5,6 +5,7 @@ import * as Log from "./../Log"
 
 import * as Capabilities from "./Api/Capabilities"
 import { Oni } from "./Api/Oni"
+import { OniMock } from "./../../test/Mocks/oni"
 
 import * as PackageMetadataParser from "./PackageMetadataParser"
 
@@ -27,7 +28,6 @@ export class Plugin {
     }
 
     public activate(): void {
-
         if (!this._oniPluginMetadata || !this._oniPluginMetadata.main) {
             return
         }
@@ -36,8 +36,7 @@ export class Plugin {
         const vm = require("vm")
         Log.info(`[PLUGIN] Activating: ${this._oniPluginMetadata.name}`)
 
-        let moduleEntryPoint = path.normalize(path.join(this._pluginRootDirectory, this._oniPluginMetadata.main))
-        moduleEntryPoint = moduleEntryPoint.split("\\").join("/")
+        const moduleEntryPoint = this.normalizedEntryPoint(this._oniPluginMetadata.main)
 
         try {
             vm.runInNewContext(`debugger; const pluginEntryPoint = require('${moduleEntryPoint}').activate; if (!pluginEntryPoint) { console.warn('No activate method found for: ${moduleEntryPoint}'); } else { pluginEntryPoint(Oni); } `, {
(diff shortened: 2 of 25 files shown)


TalAmuyal:
Also, if suggestion 2 is accepted, I would like to make it in a preceding PR.

bryphe:
Thank you for putting this together, @TalAmuyal! It's like you read my mind - I was just thinking about how we decide when to promote an experimental feature to on-by-default, and test coverage was the first thing. Having this coverage is critical not only to validate the feature works at this point in time, but to ensure that it continues working as Oni evolves. Setting up coverage here will serve as a great example for future features.

I also like that you're refactoring the singleton usage - having the explicit arguments received at constructor time makes dependencies explicit, and makes the code testable. I've been trying to do this in other parts of the code, so appreciate your help here!

We actually have two types of tests today:
- __Unit Tests__ - Validating that a specific unit of code behaves as intended - `browser/test/`. Good example is: https://github.com/onivim/oni/blob/master/browser/test/Services/Completion/CompletionTests.ts
- __CiTests__ - These are end-to-end integration tests, that execute against a _running_ Oni instance. These live in `/test/ci`. Decent example is here: https://github.com/onivim/oni/blob/master/test/ci/AutoCompletionTest-TypeScript.ts

The unit tests today have a few mocks they use, which are here:
https://github.com/onivim/oni/blob/master/browser/test/Mocks/index.ts

In the completion example, we use a mock of `IEditor` to exercise buffer entered / changed events. By mocking out the `CompletionRequestor`, we can even exercise race conditions - like a completion request starting, and then the mode changing, and the completion request resolving: https://github.com/onivim/oni/blob/01adf733430fb1bf76062c23ebede88387a8b5e6/browser/test/Services/Completion/CompletionTests.ts#L155

My preference is always to exercise unit tests the most, as they tend to be more stable as the environment is fully controlled and there are less dependencies. However, the CiTests are necessary because they can catch things like configuration issues, etc. 

In my ideal world, for a feature, we'd have:
- A __CiTest__ validating end-to-end
- A set of __unit tests__ covering the actual code

The only concern I have is that mock the `Oni` class completely might be fragile, if we try and completely emulate everything in the context of a unit test. I put a comment in the code, but `Oni.automation.sendKeys` might be difficult to actually emulate completely, as it really involves Neovim, and we've had to re-mock that entire layer to get that right.

But I think we could do the following:
- Have a __CiTest__ that validates that editing a markdown preview document creates a new window (and does validation there) - a good template for that would be: https://github.com/onivim/oni/blob/master/test/ci/AutoCompletionTest-TypeScript.ts
Just instead of validating a completion element, we'd validate the window shows up. This would be a relatively simple test, but validate the feature is working - the nitty gritty validation would be handled by unit tests.

- Have a set of unit tests validating each of the individual pieces. And I think for these, we actually wouldn't need the full `Oni` mock. For example, looking at the code, it seems like we really just need the `Oni.EditorManager` (`Oni.editors`) and the `Oni.IWindowManager` to validate. We actually already have a mock for the editors I believe, but we could create a mock for the `WindowManager`. Looks like we may also need the `Colors` API too - but that means we don't have to mock the entire Oni API end-to-end.

If we had those two mocks, we could refactor `MarkdownPreview` and `MarkdownPreviewEditor` to take those explicit dependencies - `Oni.EditorManager`, `Oni.IWindowManager`, and `Colors`, and then use the mocks to exercise those classes. The `MockEditor` class we have lets us simulate buffer enter / update events, so there are a lot of options there. That would mean we could be decoupled from needing the `sendKeys` strategy to exercise that functionality, which is great because it minimizes the surface area of the test. We could create the `MarkdownPreview` class with the `MockEditor`, simulate buffer events, and validate is `render` gives us the right stuff.

Anyway just suggestions & thinking out loud - let me know what you think or if anything is unclear. I put a few more specific comments / questions in the PR.

But thanks again @TalAmuyal for thinking about this - having test coverage to future-proof these features is what we really need to keep the quality bar high. Really appreciate your proposal here!

browser/src/Performance.ts
```diff
@@ -1,28 +1,35 @@
...
+    endMeasure(measurementName: string): void
+}
 /**
  * Thin wrapper around browser performance API
  */
-export function mark(markerName: string): void {
-    if (typeof window === "undefined") {
-        return
-    }
+export class Performance implements IPerformance {
```
bryphe: I really like the strategy overall of making the dependencies explicit - but I wonder for this class in particular if there is any benefit, as it is essentially wrapping the `console` object (which is itself a singleton).

browser/src/Plugins/Plugin.ts
```diff
@@ -51,4 +50,31 @@ export class Plugin {
             Log.error(ex)
         }
     }
+
+    public hasTest(): boolean {
+        return (this._oniPluginMetadata && !!this._oniPluginMetadata.test)
+    }
+
+    public test(): void {
```
bryphe: This is a really creative idea! I wonder how it would scale to modules / plugins created outside the repo, though? For example - if `oni-plugin-markdown-preview` lived in a separate repo, we'd still want to test it. This also has a dependency on our settings for plugin paths, which might make it easy to miss a plugin under test. We'd probably want to be able to run `npm run test` on it, at least in that case - so I'm wondering if that would be sufficient for now.

browser/src/Plugins/PluginManager.ts
```diff
@@ -70,7 +76,7 @@ export class PluginManager extends EventEmitter {
     }
 }
 
-export const pluginManager = new PluginManager()
+export const pluginManager = new PluginManager(configuration, __dirname)
```
bryphe: Thanks for making the arguments / dependencies explicit!  👍

browser/test/Mocks/oni.ts
```diff
@@ -0,0 +1,133 @@
...
+import { ThemeManager } from "./../../src/Services/Themes/ThemeManager"
+import { MenuManager } from "./../../src/Services/Menu"
+import { Process } from "./../../src/Plugins/Api/Process"
+import { StatusBar } from "./../../src/Services/StatusBar"
+import { Workspace } from "./../../src/Services/Workspace"
+import { WindowManager } from "./../../src/Services/WindowManager"
+
+import { MockPerformance } from "./Performance"
+
+export class OniMock extends EventEmitter implements OniApi.Plugin.Api {
```
bryphe: It's really cool you're thinking about this! Per the comments - I wonder if we will actually need to mock this completely, though?

In the case of a CiTest, we have a 'real' `Oni` api object available - for validating the end-to-end.

For a unit test, I'm thinking it might be preferable to use the specific mocks that the class under test needs - like instead of passing the entire API object, we pass a mock editormanager/windowmanager/colors. This helps make the dependencies explicit and minimizes the surface area.

browser/src/Plugins/Plugin.ts
TalAmuyal: I think it should work/scale/evolve the same as the `activate` mechanism does.

browser/src/Performance.ts
TalAmuyal: It is needed only for `Oni`-s mock, as it doesn't the the `performance` module (the one that is used on the non-mock version).

TalAmuyal:
@bryphe, thanks for the warm words.

Should I go ahead and (in a separate PR) remove the singletons? (this will enforce the explicity of the dependencies and help it be maintained in the long run)

bryphe:
> Should I go ahead and (in a separate PR) remove the singletons? (this will enforce the explicity of the dependencies and help it be maintained in the long run)

Sure, that sounds like a good idea, as it is a refactoring that is somewhat independent of the tests. Thanks @TalAmuyal for the help!

TalAmuyal:
@bryphe, I feel its past time for an update, so here it is:

I invested a few days in the refactor of the singletons but I dropped it since that introduces a huge diff and you might not like the result.

Regarding the unit-tests, I made some iterations on that on a new branch but I find it difficult to make tests since.
I am currently struggling with it and I will post either a report with details of the difficulties or a new PR with the tests.

For now, I close this PR.

bryphe:
Thanks for the update, @TalAmuyal ! Appreciate it -

> Regarding the unit-tests, I made some iterations on that on a new branch but I find it difficult to make tests since.
I am currently struggling with it and I will post either a report with details of the difficulties or a new PR with the tests.

Sounds great - happy to help with the difficulties if I can.

PR #1244, opened 2018-01-06
Fixes default config json syntax.

sethwoodworth:
Before, when uncommenting the `oni.bookmarks` option, 
the next configuration line option you uncomment will cause parsing the config file to fail.


bryphe:
Looks great, @sethwoodworth ! Thank you for your investigation and for bringing the fix back in. We definitely need to improve the default configuration experience.

bryphe:
Bringing this in now - congrats on the first PR!

PR #1245, opened 2018-01-06
Feature/add file associations

### LINKED ISSUE #1213: Feature: Filetype associations
@brype whilst not really a native vim feature I think it might be nice for oni to support the ability to be selected as the default program for filetypes so that in a gui menu a user can select oni as the default program or double click a file to open it in oni.

I've found that implementing this simply requires entries for `electron-builder` in the package.json see [here](https://www.electron.build/configuration/configuration). Which i've done locally using

```json
    "fileAssociations": [
      {
        "ext": "js",
        "name": "javascript",
        "role": "Editor"
      },
      {
        "ext": "jsx",
        "name": "JSX",
        "role": "Editor"
      },
      {
        "ext": "py",
        "name": "Python",
        "role": "Editor"
      },
      {
        "ext": "go",
        "name": "Go Lang",
        "role": "Editor"
      },
      {
        "ext": "re",
        "name": "Reason",
        "role": "Editor"
      },
      {
        "ext": "vim",
        "name": "VimL",
        "role": "Editor"
      },
      {
        "ext": "ts",
        "name": "Typescript",
        "role": "Editor"
      },
      {
        "ext": "tsx",
        "name": "Typescript",
        "role": "Editor"
      }
    ]
  },
```

Whilst this allows `oni` to be set as the default program and to open files it seems that despite functionality to open files if passed arguments on launch, oni does not receive files passed to it on opening, (checked locally and the log re. `FilesToOpen` on init is empty).  I'm opening this to 
  1. Confirm this is functionality that you would like?
  2. secondly to ask for your insight re. why filepath may not be being passed to oni on init (its fairly hard to debug since a built version of the app is needed to test this functionality and I cannot see logs in main.ts)

Akin909:
This PR adds the *option* for filetype associations for macOS (linux and windows, I hope dont have either of these systems so have not been able to test this) with this mac users can set oni to be the default program for a file type (note a user will still have to manually choose oni to be the recommended application). 

As far as I can tell @bryphe if opened via the cli oni receives a filepath argument but if opened via clicking on a file *in macOS* the file is not present in the process args. The docs seem to indicate that this should be handled via the open-file event on mac which also means that if oni is open already and another gui file icon is clicked, it will also open that file (not sure if this is existing functionality on windows)

fixes #1213


bryphe:
Cool, just tried out the change - looks good to me overall! I created a `dist` package, ran it, and it showed up as a recommended application in OSX 👍 

Just had one question inline around the case where there is not an existing `BrowserWindow`.

main/src/main.ts
```diff
@@ -183,6 +182,16 @@ export function createWindow(commandLineArguments, workingDirectory) {
     windows.push(mainWindow)
 }
 
+app.on("open-file", (event, filePath) => {
+    event.preventDefault()
+    Log.info(`filePath to open: ${filePath}`) // tslint:disable-line
+    if (mainWindow) {
+        mainWindow.webContents.send("open-file", filePath)
+    } else if (process.platform.includes("darwin")) {
+        process.argv.push(filePath)
```
bryphe: In this case, where there is not an existing `mainWindow` instance, do we need to create a window? Or is some other codepath hit that we expect the window to be created?

I tested this out, and in the case where I had closed Oni (but it was still in the dock - so there is a `main` process but no `mainWindow`/renderer process active), and I tried to open a file with Oni, it never opened - makes me think perhaps we need to create a `BrowserWindow` here?

Akin909: @bryphe did you test this with the prev. pr merged #1241, I sort of created these three PRs with each other in mind (basically just sections of things ive tried in my local build) which made/makes it a tiny bit confusing but the logic there is that if you have the pr with the hide functionality then on clicking the icon the `activate` event is dispatched which should trigger a create window if none exists in which case a create window is in progress and so I push the missing file path with that in mind, they might not work as smoothly together anymore

Akin909: They're all a little inter-related unfortunately #1245  #1241  and #1246, I was trying to solve each locally and once I did I split the changes into what I hoped where separate PR's rather than one amorphous giant one, apologies didn't forsee their interdependency ☹️

Akin909: Actually nevermind all that it does cause the same issue for me locally 😆 does need a create window there (not so interdependent just a bug

bryphe: Haha cool, thanks for checking it out 😄

Akin909:
Added a create window call in the branch where one does not exist which seems to have resolved the issue you mentioned @bryphe

bryphe:
Awesome! Thanks for the fix, @Akin909 . Changes look good to me - will bring in now 👍

PR #1249, opened 2018-01-07
Remove default command line icon/Add Octicons

Akin909:
@bryphe as you suggested the code file icon doesn't quite make sense and I recently discovered an octicon react/typescript component library [`react-octicon-components`](https://github.com/rhysd/react-component-octicons) here's also a [link to the available octicons](https://octicons.github.com/). These icons seem to be more code specific than `font-awesome` and there appear to be several nice looking alternatives I've currently settled for 

*terminal*
screen shot 2018-01-07 at 19 36 45

lemme know what you think as I'm happy to change it as there are several alternatives.

### DIFF
--- a/browser/src/UI/Icon.tsx
+++ b/browser/src/UI/Icon.tsx
@@ -1,4 +1,5 @@
 import * as React from "react"
+import Octicon, { OcticonSymbol } from "react-component-octicons"
 
 export const Default = ""
 export const Large = "fa-lg"
@@ -20,6 +21,14 @@ export interface IconProps {
     name: string
     size?: IconSize
     className?: string
+}
+
+interface OcticonProps {
+    name?: OcticonSymbol
+}
+
+export function OcticonIcon(props: OcticonProps) {
+        return <Octicon name={props.name}/>
 }
 
 export class Icon extends React.PureComponent<IconProps, {}> {
--- a/browser/src/UI/components/CommandLine.tsx
+++ b/browser/src/UI/components/CommandLine.tsx
@@ -128,12 +128,6 @@ class CommandLine extends React.PureComponent<ICommandLineRendererProps, State> 
                     <CommandLineIcon name="search" key={`${character}-search`} />,
                     <CommandLineIcon arrow name="arrow-right" key={`${character}-arrow-right`} />,
                 ]
-            case ":":
-                return (
-                    <IconContainer>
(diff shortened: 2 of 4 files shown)


Akin909:
hmm seems to be linting node_modules 😦

bryphe:
Cool, that is a nice icon set! Some of them may make good sidebar icons too.

Regarding the `:` replacement - I'm thinking, at least by default, it makes sense to have `:` especially in the context of the _externalized_ command line. This way it isn't too jarring for users moving from Vim -> Oni (at least the `:` serves as a reference point to show this is the command line they are familiar with).

I think the `terminal` icon might be a bit misleading, because a new user might think they are using an embedded terminal, or think it relates somehow to `:term` in Neovim. Unfortunately I still don't have a better suggestion though!

> hmm seems to be linting node_modules 😦

Strange - it looks like its happening on `npm run lint:browser` - I wonder which specific command is causing the issue? We run these three: `    "lint:browser": "tslint --project browser/tsconfig.json --config tslint.json && tslint vim/core/oni-plugin-typescript/src/**/*.ts && tslint --project extensions/oni-plugin-markdown-preview/tsconfig.json --config tslint.json",`

My guess is that it is `tslint vim/core/oni-plugin-typescript/src/**/*.ts ` - we might need to adjust it to use the `--project` and `--config` options - most of the `tsconfig.json` explicitly exclude `node_modules`, which might not be happening in this case.

Akin909:
You're quite right got a bit excited with all the icons but I think as a starting point default is probably best it's very hard to convey with an icon what the commandline is, I'll look into tweaking the command for the typescript linter or finding the offender. I'll leave the octicons so their available if you think they'd be useful elsewhere but otherwise I'll just take out the one for the commandline and leave it as `:`

bryphe:
> You're quite right got a bit excited with all the icons but I think as a starting point default is probably best it's very hard to convey with an icon what the commandline is

Haha ya, definitely!

> I'll look into tweaking the command for the typescript linter or finding the offender.

Sounds great - thanks for the help.

> I'll leave the octicons so their available if you think they'd be useful elsewhere but otherwise I'll just take out the one for the commandline and leave it as :

Sounds good to me - these icons will be useful for the sidebar work!

bryphe:
Sweet, thanks for bringing in the icons and figuring out the lint issues! 👍 Really appreciate it!

Issue #1237, opened 2018-01-05
Hover's float window size problem


saibing:
The oni hover's window:

![oni](https://user-images.githubusercontent.com/5676390/34618702-8bfba1ee-f27a-11e7-9705-ad57c533068a.png)

The vscode hover's window:
![vscode](https://user-images.githubusercontent.com/5676390/34618719-9bb18784-f27a-11e7-9f04-01c343691599.png)

I feeling that vscode's float window is better

bryphe:
Yep, the clipping on the bottom is rough. And we're not properly rendering the `markdown` as we should (#1202 is related to this). Thanks for reporting it + the screenshot, @saibing !

Akin909:
@bryphe regarding this issue I had a look at adding some `margin-bottom` to the quick info which definitely looks nicer to me, and reducing its `max-width` so it doesn't stretch sideways but more vertically. Whilst I was at it I tried using `marked` to render the quickinfo. As far as I can tell from the LSP's typings and docs [here](https://microsoft.github.io/language-server-protocol/specification#textDocument_hover) the hover info is essentially always some form of markdown or as they call it `markup text`, however I noticed you had a todo in the code re how to handle the syntax highlighting, I was going to try and use marked to render it based on the language passed in from the lsp if one is available otherwise I was considering a fallback of the buffers 'filetype'

Akin909:
Tried using `highlight.js` in combination with `marked` which results in:

<img width="500" alt="screen shot 2018-01-07 at 21 31 08" src="https://user-images.githubusercontent.com/22454918/34654449-9f19d730-f3f3-11e7-8969-a2f26137c301.png">

<img width="472" alt="screen shot 2018-01-07 at 21 28 48" src="https://user-images.githubusercontent.com/22454918/34654451-ac505f5a-f3f3-11e7-9b4f-06c0bf6d9d5d.png">

Although think I might have gone off piste as I'm not sure what you might have intended for the highlighting

Issue #1051, opened 2017-12-01
Theming: Allow specifying of token colors in json themes


bryphe:
TextMate themes give us a finer granularity through 'scopes' in terms of colorized tokens, as opposed to the default conventional groups in Vim themes.

We should take advantage of this by allowing specification of token colors in themes, and giving real fine-grained control.

Two specific pieces of work:
- Today, TextMate themes only use existing highlight groups - we need to extend this capability so that they can specify new highlight groups
- TextMate themes already use the `editor.tokenColors` to colorize, but we need to pull that from the theme as well.

bryphe:
The end goal of this is that we'd have a function that we'd call against the active theme / thememanager, like:
`getColorForTokenKind(tokenKind: string)`

If the `tokenKind` can relate to the `scopes`, like here:
https://www.sublimetext.com/docs/3/scope_naming.html

that would be great for our TextMate theming!

bryphe:
That same 'function' could be reused in a few places:
- Highlight colors for code blocks markdown preview / quick info
- Colors in completion UI

I think it will be difficult to 'read back' these from the Vim colorscheme today, so an alternative is to _explicitly specify_ them in our themes, via a `tokenColors` setting.

bryphe:
I realized we actually can read the highlight colors, by using `nvim_command_output` with `:hi Function` or similiar. Might be applicable here - we can use those as the defaults if the token colors aren't specified by the `tokenColors` setting.

Akin909:
Ah that would be helpful 👍 , would you be able to point me to where the tokenization is happening @bryphe and I can begin to take a look at implementing this

Akin909:
Once again @bryphe need to pick your brain re how to start this out re generalising syntax highlighting so it can be used in tooltips etc., what I would like to do is get a piece of functionality which reads a line and outputs the different tokens per word, I'm not sure if the `this._tokenizeLine` function you pointed me towards would be the best place to start or importing the `PeriodicJob` class for the syntax highlighter which contains that function and using an instance of it to run the tokenize function would be better?

bryphe:
Oops, sorry I missed your response before! Thanks for reminding me.

Yes, I think there are two challenges here:
- Figuring out what the token colors should be (that's where `nvim_command_output` will help us - but I'd also like to support specifying them via themes with a `tokenColors` setting - and you can set them directly in the config like @hoschi is doing in #1379).
- Tokenizing the lines in the quickinfo, so that we can select the proper color

[...]

PR #1253, opened 2018-01-08
Feature: render markdown and fix tooltip dimensions

Akin909:
Following comments in #1237 I've had a look at implementing some changes to the tooltip:
1. Fix margin so text isn't clipped
1. Change sizing so it isn't rendered at almost full width (which is arguably a bit too big)
1. Render markdown from LSP in `quickinfo` using `marked`.
	
This is still a WIP as outstanding things include:
- [x] rendering links with a better color than standard 90s blue - used `theme["highlight.normal.bg"]`
-  [x] pass font family to code blocks using `theme` from `styled-components`?
-  [x] ~~Maybe? syntax highlighting~~ *separate pr*
-  [x] Fix tooltip rendering flicker due to width change

@bryphe  was hoping to get your feedback regarding the outstanding point especially syntax highlighting as I feel you likely would have had something in mind for this ?external lib like `highlight.js` or `prism` ? home grown highlighting to match themes

Secondly the tweaks I made re the dimensions of the tooltip, I think that as per #1237 an almost full width tooltip is *subjectively* not ideal, I tried containing this width to a max-width which it seems like `vscode` does (it seems to have a relative max-width of say 40%) but what I *personally* like about oni's hover bar is that it gives you a lot of the type info. I initially tried a slightly larger max-width than vscode possibly/maybe uses but this causes an issue with padding, so i've tried to preserve this by rendering any element that will fit the screen, adding wrapping to the `quickinfo` so text wraps vertically and also adding a `width: min-content` only for elements which are too big for the screen.

here are some screenshots:

screen shot 2018-01-08 at 21 11 52

screen shot 2018-01-08 at 21 26 14

screen shot 2018-01-08 at 21 13 40

screen shot 2018-01-08 at 21 13 50


bryphe:
Cool, thanks for looking at this @Akin909 ! The hover experience could definitely use some love 😄 

> @bryphe was hoping to get your feedback regarding the outstanding point especially syntax highlighting as I feel you likely would have had something in mind for this ?external lib like highlight.js or prism ? home grown highlighting to match themes

Regarding the syntax highlighting - this is something I'm still trying to figure out. It would be interesting to see what strategy @TalAmuyal used for the markdown-preview plugin - I wonder if we could share/reuse pieces between this?

One challenge we have with the syntax highlighting is being able to read the highlight groups. Right now, we don't really have the information on the Oni side related to the colors for each of the highlight groups. It'd be really nice to have this! One option would be to explicitly set it in the themes (like a `tokenColors` property - we have `editor.tokenColors` today, but it points to a highlight group - I'd like for it to point to an actual color eventually!) - and if it isn't specified, we just wouldn't highlight.

It might also be possible to use our existing TextMate highlighting for code blocks - we can run a TextMate grammar on each line. The code where this happens is here: https://github.com/onivim/oni/blob/c01ae065c853e50471b7e4b17f4cadbb019fe7f7/browser/src/Services/SyntaxHighlighting/SyntaxHighlightingPeriodicJob.ts#L87 , it could potentially be more streamlined in this case (this code is set up to do a few lines up to a budget, and then continue again later, so that async-job stuff causes a bit more complexity). The TextMate grammar gives us a list of tokens+scopes, which we could then match up via `editor.tokenColors`. But still need to figure out how to get actual colors there 😄 

Ideally, with whatever strategy we choose - the highlighting colors can match our theme.

One thing I noticed too between our experience and VSCode is that we render the first line in a special way:
image


versus VSCode:
image

I'm wondering if we should do away with that special handling (box-shadow/different font/different formatting) of our 'title'? 

Oh, and the width changes you made look good to me - thanks for digging in there and looking at options! 👍

Akin909:
@bryphe I very much had the same thought re the title specifically in the case of things like
```js
const obj = {
	key: 1,
	key: 2
}
```

since specifically in those case it looks odd but on applying the change I realised that tbh one of the things I've much preferred about oni to vscode's hover menu is that for certain type signatures it's the lack of a distinction makes them harder to follow, also in oni the lack of syntax highlighting makes a large block of white text hard to breakdown into its constituent parts without a visual cue e.g.

screen shot 2018-01-10 at 18 32 21

I've left in the title element but removed most of the styles to show the difference re the above. I'll defer to whatever you prefer but looking at the difference I personally found it immediately less readable, I dont know if thats because i've sort of used the visual difference of the sections to decide what information I need so either im looking for type annotation or an explanation but having them look the same I find that I cant figure it out at a glance as easily, the above is what I hope is middle ground potential  the bottom is styles removed a la vscode(which I think down the line with syntax highlighting could look much nicer, though personally still partial to oni's current separation).

screen shot 2018-01-10 at 19 27 28

bryphe:
> In the screen shots i've left in the title element but removed most of the styles to show the difference re the above (inital pr screenshots). I'll defer to whatever you prefer but looking at the difference I personally found it immediately less readable

@Akin909 , thanks for taking the time to prototype that approach and put the screenshots together! If you prefer the current approach, I'm cool with leaving the title as-is. No reason to make a change that makes it less readable IMO

> Forgot to say I like the grammar Idea especially since marked exposes a lexical parser so it identifies tokens which could then be matched to the grammar tokens so if those had colors then that could be a cool way to highlight,

Cool! That would be great. It actually relates to an issue I opened up a bit back - #1051 . This would be useful in general - it would enable highlighting here, and then I would also like to apply these token colors to other parts of the UI. As an example, I'd like to add some color to the completion UI - since we know the token type, I'd like to color either the icon / token with the highlight color. Would look really nice compared to what we have today.

>  although it's sounding like a bit of work which I'm happy to do but I think separately (I find it easier to make smaller contributions than omni-pr's, so things stay easy to debug or "reason-about")

💯 Definitely prefer smaller PRs on my side too! Makes it easier on me and is safer in general. We could use #1051 to track the highlight-color work.

Akin909:
@bryphe I've believe I completed the tweaks I intended here but would love to have your take on them, 
summary of the changes I ended up with were.
* Rendering markdown properly, using highlight mode to render links as its more distinct (I can localise this change to just the quick info but at present it affects all tooltips)
* set a max width to `55vh`
* compromised and used a `word-break: break-all` to deal with browser quirks with maxwidth which would otherwise cause unusual excessive padding without this

screenshots:

screen shot 2018-01-14 at 14 25 47

screen shot 2018-01-14 at 14 25 36

Akin909:
Also forgot to mention that I darkened the tooltips using `color` package so they standout more visually this affects all tooltips so would be good to get your take on that as well as it might not be what you have in mind.

Akin909:
**NOTE**: I've found a future/current but experimental feature bug, whilst testing the file-explorer I realised that the `cursorPositioner` is set to calculate its position based on the *entire* viewport not the editor space i.e. the file-explorer takes up the viewport space so with it open the tooltip can be pushed beyond the edge of the frame. I looked at adding the specific `.focus.editor` element's dimensions to the `setViewport` reducer except this doesn't work on initial load as the element is not in the dom. A quicker solution would be to select the element in `componentDidMount` of the positioner and use those dimension, not sure if it would be generally helpful to have the focused editors specific dimensions minus things like sidebar etc.

Either way i've essentially done some of the work for it so can add it to the main viewport reducer or use the dimensions adhoc in cursor positioner, potentially separately

bryphe:
Just testing this out now, I think the new tooltips look really nice! Good stuff.

> Also forgot to mention that I darkened the tooltips using color package so they standout more visually this affects all tooltips so would be good to get your take on that as well as it might not be what you have in mind.

The colors look good to me on the dark themes. I tested on `nord` and `onedark`. 

I did notice on the light themes, like `solarized8_light`, that it didn't quite look right:
image
This might be a pre-existing issue though, potentially related to #1260 .

And on `delek`, it looks perhaps a bit too dark:
image

I did notice one regression with the context menu:
image

Seems the icons aren't rendering for me - I wonder if somehow the font-awesome icons aren't getting the right font?

Also happened when I looked at an error message in the quick open hover:
image

> EDIT: Either way i've essentially done the work for it so can add it to the main viewport reducer and set the dimensions in cursor positioner, potentially separate (from this pr)

Good catch on the CursorPositioner bug! It would be great to have this in a separate PR.

browser/src/UI/components/QuickInfo.tsx
```diff
@@ -1,31 +1,127 @@
...
-
-        if (!this.props.text) {
+        const { html, text, padding } = this.props
+        if (!html && !text) {
             return null
         }
 
-        const lines = this.props.text.split(os.EOL)
-        const divs = lines.map((l) => <div>{l}</div>)
+        return <Title padding={padding} dangerouslySetInnerHTML={html}>{text}</Title>
```
bryphe: Are we sanitizing this `html` anywhere? In the markdown-preview plugin, we're using the `dompurify` library to sanitize the HTML. We want to avoid this type of RCE security issue that Atom hit! https://statuscode.ch/2017/11/from-markdown-to-rce-in-atom/

browser/src/UI/components/ToolTip.tsx
```diff
@@ -1,47 +1,67 @@
...
 }
 
 export class ToolTipsView extends React.PureComponent<IToolTipsViewProps, {}> {
-
     public render(): JSX.Element {
-        const toolTipElements = this.props.toolTips.map((toolTip) => {
-            return <ToolTipView {...toolTip} key={toolTip.id} borderColor={this.props.borderColor} foregroundColor={this.props.foregroundColor} backgroundColor={this.props.backgroundColor}/>
+        const toolTipElements = this.props.toolTips.map(toolTip => {
+            return (
+                <ToolTipView fontFamily={this.props.fontFamily} {...toolTip} key={toolTip.id} />
```
bryphe: Nice, thanks for moving the colors to use our `styled` themes 👍

bryphe:
So there are a couple of things to look at before bringing this in:
- The regression with the icon rendering in the completion / diagnostics UI
- Sanitizing the rendered markdown contents (if it isn't sanitized already)
- Colors in the light themes - if these can be explicitly specified via our theming json files, I'm fine with it - otherwise we might want to add additional logic.

But looking really good - thanks for all your work on this, @Akin909 ! Let me know if have questions about any of these items.

browser/src/UI/components/QuickInfo.tsx
Akin909: [`marked`](https://github.com/chjj/marked) has an option to sanitize the html which it outputs which I in turn "dangerouslySet" 
>sanitize
>Type: boolean Default: false
>
>Sanitize the output. Ignore any HTML that has been input.

It's generating the html from markdown string it gets from the `lsp` which also in the documentation state that the protocol passes sanitized strings though in the case of potentially malicious attacks if a string is passed containing a script it wouldn't be output via marked so I don't know whether another form of purification would be necessary here?

Akin909:
@bryphe thanks for the feedback 👍 , 
* Re. the sanitization I've left a comment above which in summary is that `marked` provides sanitization of the html it outputs via a config option which I'm using in this PR not sure if this is sufficient or there is a vulnerability in this process I'm overlooking,
* Re. the light themes do you mean a secondary color like maybe actually just choosing tool tip colors for each supported theme as those can be set to handpicked darker colors?
* Finally re the icon rendering I'm not able to reproduce it with the default font or with my slew of go to fonts can I ask what font you are using in the screen shot, my only change was to force the children of the tooltip to use the editor's font family in css but I can just make that specific to the code blocks which is where I specifically wanted the change

Akin909:
Having had a peek over a co-worker's shoulder at the `vscode` tooltip I also realised that for code blocks in paragraphs the backgrounds for those were being rendered differently so I've also employed the [WIP] darkening method to have the same really effect and increased line height as well it might be best to render these the way github does but this might not look great in all colorshemes as opposed to darkening an existing color in the palette.

screen shot 2018-01-18 at 11 04 04

browser/src/UI/components/QuickInfo.tsx
bryphe: Ah ok, interesting, thanks for pointing that out. Reading through the docs, it seems that should be sufficient.

bryphe:
> Re. the light themes do you mean choosing tool tip colors for each supported theme as those can be set to handpicked darker colors?
EDIT: also I concur that some of these might be due to quirks with colors especially solarized, re the lighter ones I was thinking of somehow finding out if vim bg is set to light or dark and changing the degree of darkening based on that.

We should have a way at least to specify handpicked colors for the hover info, to override the darkening behavior, since it might not work correctly in every case (or perhaps someone wants some exotic color for their scheme 😄 ). Using the strategy of seeing if the theme is light/dark would help, but there will still be situations where the user might still want to override it. 

Perhaps we could use `editor.hover.title.background`, `editor.hover.title.foreground`, `editor.hover.border`, `editor.hover.contents.background`, and `editor.hover.contents.foreground` - and if these aren't specified, we'll fallback to deriving them by darkening existing colors.

> Finally re the icon rendering I'm not able to reproduce it with the default font or with my slew of go to fonts can I ask what font you are using in the screen shot, my only change was to force the children of the tooltip to use the editor's font family in css but I can just make that specific to the code blocks which is where I specifically wanted the change

Hmm, I'll grab the latest changes and see if I can reproduce it again.

bryphe:
Here's what i saw in the devtools:
image

Seems like somehow the icon is being overrided with the `Consolas` font.

Akin909:
@bryphe I've added your recommend config options which if not set default to the darkened tooltip options, and also reverted the font family change to prevent it being forced on font awesome icons.

I haven't implemented the vim background based alteration as I realised I can determine the lightness or darkness of the original color just using the `color` package which is what I've done to moderate the darkening.

I also reverted the `font-family` change to only affect the code blocks which leaves the icom elements untouched.

bryphe:
Changes looks  good to me, @Akin909 ! I'm going to re-run the Linux job, as that should be passing. I'll bring it in after that runs.

Thanks for your work on this! 👍

Akin909:
@bryphe nice! ✨, I can add a section to the docs about the changes that users can now configure like `hover.contents.background` also did the same for the code block which show in the markdown in case ppl want that sort of granular control. Also didn't quite want to raise it as a separate issue though maybe it should be but this change raises the lol issue of how to handle opening links in oni.

I imagine your intent would be to use the internal `layer-browser` but just wanted to mention it though I know that's still a work in progress.

bryphe:
>  I can add a section to the docs about the changes that users can now configure like hover.contents.background also did the same for the code block which show in the markdown in case ppl want that sort of granular control

That would be awesome, thanks!

> Also didn't quite want to raise it as a separate issue though maybe it should be but this change raises the lol issue of how to handle opening links in oni.

> I imagine your intent would be to use the internal layer-browser but just wanted to mention it though I know that's still a work in progress.

Yep, eventually 😄 But for now I'd like to funnel them through the `browser.openUrl` command - this will do the right thing - either open in a new window, or an internal browser window, depending on if the `experimental.browser.enabled` flag is set.

bryphe:
Looks like the linux build failed, but due to the code coverage changes - I'm pulling those out in #1338 until we have a sturdier foundation for builds. I'll bring this in now. Thanks again for your work on this, @Akin909 ! 💯

PR #1257, opened 2018-01-09
TextMate Highlighting: Fix insert-mode quirkiness

bryphe:
__Issue:__ The textmate highlighting when in insert mode is not reliable, especially when typing across multiple lines.

__Defect:__ The buffer update strategy we have doesn't handle newlines well - it will overwrite an existing line when pressing enter, which can impact highlighting for subsequent lines.

Initially, I'm pulling out the insert-mode update to have a more reliable strategy. My plan is to segregate the insert-mode updates (just evaluating the current line, and not dirtying any other lines), and then do a full evaluation of the changed lines when going from insert mode -> normal mode. 

I also would like to introduce a throttle / debounce strategy, such that we are not re-evaluating syntax highlighting on every keystroke, which is very expensive when typing fast.


browser/src/Services/SyntaxHighlighting/SyntaxHighlighting.ts
```diff
@@ -101,13 +73,13 @@ export class SyntaxHighlighter implements ISyntaxHighlighter {
...
                 bufferId: evt.buffer.id,
                 lines,
+                version: evt.buffer.version,
             })
         } else {
-
-            // Incremental update
             this._store.dispatch({
-                type: "SYNTAX_UPDATE_BUFFER_LINE",
+                type: "SYTNAX_UPDATE_BUFFER_LINE",
```
bryphe: Fix typo here

PR #1267, opened 2018-01-10
Fix #1251 - TypeScript standalone server is always spawned on startup

### LINKED ISSUE #1251: Startup Performance: TypeScript Server process (`tss`) is always spawned on startup
Looking at traces from startup, it seems that the TypeScript standalone server (`tss`) is always started - even if we haven't entered a typescript or javascript file.

This should be deferred until we actually enter a TypeScript / JavaScript file.

bryphe:
__Issue:__ The TypeScript standalone server (`tss`) is always initialized on startup, even if no TypeScript file is open, and even if TypeScript isn't the strategy of choice for `javascript` files.

__Defect:__ The `TypeScriptServerHost` is created immediately in the `activate` method. Currently, we always call the `activate` method immediatley (although in the future, it'd be helpful to have an `activationEvents` concept, to defer activating plugins until some condition is met).

__Fix:__ Wait to instantiate the TypeScriptServerHost until the first `textDocument/didOpen` request comes in.


bryphe:
Fixes #1251

PR #1268, opened 2018-01-10
Fix #1265 - Flicker / movement of neovim popup menu

### LINKED ISSUE #1265: Completion flickery
The completion window (even on latest release) is still really flickery and jumpy.

completion

1. I think that the physical geometry/layout should change at most once per keystroke. There are some obvious examples where that's not the case.
2. Aside from the obvious double-overlapped/misaligned layers that are clearly visible, even when it seems it's not jumping around, a high speed replay reveals that the fade in/out is not very reliable while typing. Some times it will fade in/out twice per keystroke. In general I don't even think it should fade in/out in between key strokes. It's really hard to get the timing of that kind of interaction right, and I'd prefer a solid discrete update as opposed to a kind of unpredictable fading/animating one.

multiflicker

Here are some screenshots from the gif above zoomed in:

screen shot 2018-01-10 at 2 47 15 am

screen shot 2018-01-10 at 2 48 04 am

Also notice that while I'm typing, the little triangle moves unexpectedly:

screen shot 2018-01-10 at 2 48 57 am

screen shot 2018-01-10 at 2 49 05 am

About the animations: I have a personal philosophy about animations: That a product earns animations once its discrete transitions work perfectly without animations. Just my two cents.

In summary: Here's my suggestions:
- The dialog being open for the entire time the user is completing - never disappearing in between key strokes, and never momentarily appearing empty.
- The autocomplete dialog should update once per keystroke.
- Then once that is really solid, animations should fade the dialog window in and out, but never in the middle of interaction - only at the "bookends".

bryphe:
__Issue:__ When moving between items in the NeovimPopupMenu (or typing), there is noticeable flickering.

__Defect:__ When the state of the neovim popup menu is changed, there is status updated in the command line. This actually causes the cursor to be positioned at the command line briefly. Because the popup menu position is based on the cursor position, this actually pulls the popup menu both vertically and horizontally, causing a noticeable flicker.

__Fix:__ Ignore the cursor position changing when it is in the command line _and_ we're not in command line mode.

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -326,8 +326,19 @@ export class NeovimEditor extends Editor implements IEditor {
         })
 
         this._neovimInstance.onRedrawComplete.subscribe(() => {
-            this._actions.setCursorPosition(this._screen)
-            this._typingPredictionManager.setCursorPosition(this._screen)
+            const isCursorInCommandRow = this._screen.cursorRow === this._screen.height - 1
+            const isCommandLineMode = this.mode && this.mode.indexOf("cmdline") === 0
+
+            // In some cases, during redraw, Neovim will actually set the cursor position
+            // to the command line when rendering. This can happen when 'echo'ing or
+            // when a popumenu is enabled, and text is writing.
+            //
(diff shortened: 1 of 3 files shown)


bryphe:
Fixes part of #1265 , and #851

PR #1271, opened 2018-01-11
Fix #1264 - 'showmatch' causing tab to lose focus

### LINKED ISSUE #1264: set sm! causes tab to loose focus.
Vim's showmatch feature causes the current "tab" to loose focus while the matching paren is being highlighted.

bryphe:
This change addresses the 'showmatch' option. The first attempt at a fix was to just treat the 'showmatch' mode like regular insert mode, but the highlight of the matched character doesn't work correctly in that case.

Note that the change is bigger than strictly required, as I did some timeboxed refactoring to help enable better unit testing of this scenario (need to continue to refine / whittle away at `NeovimEditor` and the implicit dependencies).


bryphe:
Kept the cursor in 'block' mode, and added a small bump like we have for motion. The tab will still show the insert mode highlight in this case. Fixes #1264

PR #1274, opened 2018-01-11
Fix #1263 - Rename also renames strings

### LINKED ISSUE #1263: Rename action changes not only symbols
```js
let put = 'fooo'
// put it down
let bar = {
    myString:'put'
}
```

When you try to use the rename feature on `put` the content of `myString` and the comment also gets changed, which I thought would not be the case. Shouldn't it only change it to

```js
let post = 'fooo'
// put it down
let bar = {
    myString:'put'
}
```

When I rename it to `post`?

bryphe:
Adjusts the `findInStrings` and `findInComments` to more conservative defaults, so that the rename operation (for TypeScript) does not modify strings / comments.


bryphe:
Fixes #1263

Issue #545, opened 2017-07-20
Is it possible to individualize vim tabs and splits?


bert88sta:
I feel as though putting each tab in its own container, which then in turn contains multiple splits, could be beneficial for rendering. If we're forced to track where the edits are taking place or where the scrolling is, then we don't have to redraw other splits.

Also, one of the things I've always wanted (that seems impossible in vim) are split borders. I've abandoned airline in favor of the builtin colorscheme and Oni's statusline, but this annoys me:

<img width="1680" alt="screen shot 2017-07-20 at 9 45 09 am" src="https://user-images.githubusercontent.com/8421641/28420440-28cb0f30-6d30-11e7-829b-9f09bfb081cc.png">

(these are 2 separate splits, but if they were the same language it would be hard to tell where one begins and one ends)

extr0py:
Yes, great suggestion. I'm hitting this too - sometimes it's difficult to know where the split boundaries are in the current implementation!

I'd like to provide more control over the actual split window UI - borders, drop shadow, etc. And explicitly controlling the rendering here will allow us to do fancier things and more in-depth performance optimizations in the buffer rendering and responsive. Some of the thinking around this is captured in #362.

bryphe:
For this, I'm thinking about adding an `editor.splitMode` configuration, with a value of `oni` or `native` - having a value of `oni` would result in using containerized splits as you describe (along with multiplexed Neovim processes).

badosu:
@bryphe Seems like this was addressed now that you can have multiple Oni windows, right?

PR #1276, opened 2018-01-11
Bugfix/update colors when config changes

### LINKED ISSUE #1275: Regression - Theming: Theme colors used by 'styled-components' are not updated when changing colorscheme
__Repro:__
- Open Oni
- Edit `config.,js`
- Change colorscheme

Expected: The editor + all shell / chrome should be updated. (You can check 'QuickOpen' to validate)
Actual: Only the editor colorscheme is updated.

It seems that the `ThemeManager` we use with our `styled-components` isn't getting the updated colors.

Akin909:
Adds a check in the config changes handler to update the theme if key that was changed includes `ui.colorscheme`

Fixes #1275

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -520,7 +520,7 @@ export class NeovimEditor extends Editor implements IEditor {
     }
 
     public async newFile(filePath: string): Promise<Oni.Buffer> {
-        await this._neovimInstance.command(":vsp " + filePath)
+        await this._neovimInstance.command(":new " + filePath)
         const context = await this._neovimInstance.getContext()
         const buffer = this._bufferManager.updateBufferFromEvent(context)
         return buffer
@@ -707,7 +707,7 @@ export class NeovimEditor extends Editor implements IEditor {
         .then(ids => this._actions.setCurrentBuffers(ids))
     }
 
-    private _onConfigChanged(newValues: Partial<IConfigurationValues>): void {
+    private async _onConfigChanged(newValues: Partial<IConfigurationValues>): Promise<void> {
         const fontFamily = this._configuration.getValue("editor.fontFamily")
         const fontSize = addDefaultUnitIfNeeded(this._configuration.getValue("editor.fontSize"))
         const linePadding = this._configuration.getValue("editor.linePadding")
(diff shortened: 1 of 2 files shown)


browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -707,17 +707,21 @@ export class NeovimEditor extends Editor implements IEditor {
...
 
         this._actions.setFont(fontFamily, fontSize)
         this._neovimInstance.setFont(fontFamily, fontSize, linePadding)
 
-        Object.keys(newValues).forEach((key) => {
+        Object.keys(newValues).forEach(async (key) => {
             const value = newValues[key]
             this._actions.setConfigValue(key, value)
+            if (key && key.includes("ui.colorscheme")) {
+                await this._themeManager.setTheme(value)
```
bryphe: Thanks for the change, @Akin909 ! 🎉 

I believe we might be setting the theme already here, so this line might not be needed:
https://github.com/onivim/oni/blob/d60fa91374b4d3f30730d96d6b635cbf56c3c3ed/browser/src/Services/Themes/index.ts#L12

The `ThemeManager` actually has an event `onThemeChanged` - maybe we could listen to that either in `Shell.tsx` or in `browser/src/index.tsx`? The reason I'm thinking about those places as opposed to `NeovimEditor` is that, eventually, we might have more than one way `NeovimEditor` instance - like for #545 . So in that case, this might get called multiple times. There's only ever one 'Shell' though at the moment.

I was actually trying to remove all the `Shell` references from `NeovimEditor`, but the title one is stuck there at the moment. But I'm hoping to get those out to fully support that 'multiplexing' scenario 😄 

Let me know what you think!

Akin909:
Thanks for pointing that out @bryphe 👍 that simplified things quite a bit

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
Akin909: @bryphe hopefully my changes are ok? I didn't use the listener you mentioned as looking at the activate function it seemed to be subscribed to config changes to colorsheme already just not updating the store with this info so I added an action there though I might have entirely missed the point of your suggestion 😟

bryphe:
Awesome! Fix looks great, will bring it in when the tests are done.  Thanks @Akin909 ! :+!

PR #1282, opened 2018-01-12
#68 - Part 2: Implement loading configuration for workspace

bryphe:
This change adds a `WorkspaceConfiguration` object + unit tests.

The `WorkspaceConfiguration` listens for workspace changes, and checks to see if a workspace configuration is available. If so, it adds it as a configuration provider.


bryphe:
I also added a stub oni config for now - but I'd like to use this to prototype some new functionality (alternate paths, watch/preview capabilities specific to Oni)

Issue #1226, opened 2018-01-04
Feature: Start / Welcome Screen


bryphe:
Today, Oni is not very welcoming to the new user, especially a new user who is not familiar with Vim.

As part of our _lowering the bar_ initiative, we should provide a better experience on first launch. Here's a brain dump of some ideas I'm thinking of including on the main page:

- Header + tagline (Oni: Modern Modal Editing?)

- Get started
  - New File
  - Open Folder
  - Command Palette (`<Control+Shift_P>`)
- Recent (list files / workspaces)
- Learn
   - (TBD) Modal editing 101 (entry point to our gamified vim tutor)
   - (TBD) Entry points for achievements / goal feature
- Customize
   - Edit `config.js`

Making this keyboard accessible via standard vim commands is important, of course :)

Let me know if you have ideas / feedback!

TalAmuyal:
While reading your comment, the sentence "Making this keyboard accessible via standard vim commands is important" kept going through my head, so I'd like to `like` that note twice :P

Akin909:
😆  A me too is a little unhelpful here but absolutely I think re. keyboard navigation of the screen but I love the idea. As a tiny suggestion re copy of all things, given its the start screen I think the line about oni should include vim in some way so if a user is completely unfamiliar with oni and vim its clear that there's a relationship  🤷‍♂️

athaeryn:
Could oni include [startify](https://github.com/mhinz/vim-startify) preconfigured with the things you listed? Building this with startify gets you mouse + keyboard control for free. :) Users could also edit the config to customize their start screen as they learn more (disable the Learn links, add things like recent sessions, etc.).

bryphe:
> Could oni include startify preconfigured with the things you listed? Building this with startify gets you mouse + keyboard control for free. :)

Good question 😄  It is true that we'd get the mouse+keyboard control for free using a Vim solution. I'd prefer not too, though, primarily because one of the implicit goals I had for the start screen was to highlight Oni's capabilities. In other words, if it is designed correctly, it should be obvious what _breaking free of terminal limitations_ means. It should convey, right from opening the app, that Oni is providing something new and different than terminal vim, but that there is crossover - you don't need to learn a new interaction model.

It's sort of accomplishing two things - letting the user know how to get started, and showing vim users how their knowledge maps into some of these new experiences we're providing & working on. `vim-startify` helps with the former but not the latter.

[...]

PR #1285, opened 2018-01-12
#1226 - Part 1 - Initial layer UX for welcome screen

bryphe:
This is just putting together an initial 'layer' for the welcome screen. It looks like ugly right now , and has no functionality, but it at least is a starting point.

A few next steps for 'Part 1':
- Make it look not so ugly
- Add an entrance animation
- Add placeholders UX for each of the section
- Add an experimental configuration `experimental.welcome.enabled`

Following that:
- Add full functionality for each section
- Implement integration with Neovim for proper keyhandling


bryphe:
WIP for #1226

PR #1286, opened 2018-01-13
Fix #978 - Add 'tabs.showIndex' option

### LINKED ISSUE #978: Add tab index for absolute navigation
At the moment it is hard to navigation between tabs with `{count}gt` when you have more than ~5 tabs. Even more harder when tabs are wrapped. The index from traditional vim should be added again to make this as easy as before.

2017-11-18-111401_2058x1409_scrot

bryphe:
This adds a `tabs.showIndex` configuration option, which, when enabled, adds a number to the tab for easy switching via `{n}gt`:
image


bryphe:
Fixes #978

bryphe:
Added documentation here: https://github.com/onivim/oni/wiki/Configuration#tabs

PR #1287, opened 2018-01-13
Tabs: Show file icon by default

bryphe:
- Add a setting `tabs.showFileIcon`
- Default to true, so we always show file icons in both buffer / tabs mode


bryphe:
I'm not sure where this issue was - but I know when I originally implemented the file icons, I didn't implement them for tabs. I was under the impression that customization of the filename was more common, but the default is to use the filename, and in that case, we should be able to show an icon.

I added file icons by default to tabs here, but also a setting to disable them if it is not desirable

bryphe:
Added documentation here: https://github.com/onivim/oni/wiki/Configuration#tabs

PR #1288, opened 2018-01-13
Fix #1220 - Add 'editor.insertSpaces' and 'editor.tabSize' options

### LINKED ISSUE #1220: Please provide an option to set tabSize
I want to set `tabSize` directly without having to disable `oni.useDefaultConfig`.
thanks.

bryphe:
This change adds `editor.insertSpaces` and `editor.tabSize` options, so that the space/tab settings can be adjusted directly from the `config.js`.

If `editor.insertSpaces` is set to 'null', we will not set `expandtab`/`noexpandtab`.
If `editor.tabSize` is set to 'null', we will not set `tabstop`, `shiftwidth`, `softtabstop`.

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -103,6 +103,9 @@ const BaseConfiguration: IConfigurationValues = {
     "editor.cursorColumn": false,
     "editor.cursorColumnOpacity": 0.1,
 
+    "editor.tabSize": 4,
+    "editor.insertSpaces": true,
+
     "editor.tokenColors": [{
         scope: "variable.object",
         settings: "Identifier",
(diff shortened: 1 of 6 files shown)


badosu:
@bryphe Does this mean that I have to add these two configurations to have my `.vimrc` configuration respected?

bryphe:
Ya, good point @badosu - at this time, you would need to explicitly set those configurations to null. 

What I'd really like longer term is to have automation indentation detection - we should only care about these settings for new files - existing files we should just match indentation behavior.

badosu:
I think this may be an issue for vim users that expect their configuration to 'just work' when migrating to Oni, it's not my case but it's typical for people working on different languages to have different indentations defined via filetype detection, this PR will override this.

Until we have a configuration like `editor.<ftype>.tabSize` I think it's better to leave it as null by default.

bryphe:
Been a looong time but finally got back to this - good points @badosu . I think the latest iteration will improve this behavior. If nothing is set (the default), it'll pick up the Neovim defaults - and we won't aggressively override them anymore. Otherwise,  the user can use `editor.detectIndentation` to automatically set space/tabs based on the file contents - or explicitly override it with `editor.tabSize`/`editor.insertSpaces` or per-filetype `editor.<language>.tabSize`/`editor.<language>.insertSpaces`.

I'm also adding an API to the `Editor` here: https://github.com/onivim/oni-api/pull/42 so that we don't have to dive to the Neovim level (will make it more testable and less coupled to the Neovim strategy).

PR #1293, opened 2018-01-14
remove icon for default command line

Akin909:
@bryphe apologies re the bugs with #1249 🙇 , this PR just removes the case for the icon for `:` so there is no default icon as discussed in #1249 without any octicon baggage. Should be bug free just a small deletion.


bryphe:
> @bryphe apologies re the bugs with #1249 🙇 

Haha, no worries! It's unfortunately easy to break, especially since there isn't a test covering this developer case. I've broken this too before. 

Appreciate your understanding - thanks for bringing back the icon fix!

bryphe:
Looks good to me! Bringing this in now 👍

PR #1301, opened 2018-01-16
Fix #1243: Regression in 'editor.quickInfo.show' command

### LINKED ISSUE #1243: Add command to explicitly show hover
When `editor.quickInfo.enabled` is false, there is now way to show the quickInfo UI.

We should have a command that is available for explicitly showing it, like `editor.quickInfo.show`.

bryphe:
__Issue:__ When `editor.quickInfo.enabled` is `false`, we should still allow the user to explicit pop up the hover information.

__Defect:__ The `editor.quickInfo.show` command is available, but after the refactoring it didn't have access to the `LanguageIntegration` object.

__Fix:__ Fix the initialization order in `NeovimEditor`'s constructor


bryphe:
Fixes #1243

PR #1309, opened 2018-01-16
Fix #1295: Crash when opening a window w/o navigating to it

### LINKED ISSUE #1295: Opening Undotree window causes pixelToScreen crash
Apologies for the very vague issue description, but I've noticed since I believe the layers/window changes which I haven't looked at were integrated whenever I try to open the plugin `undotree` or use a custom fn I have to toggle `quickfix` or `loclist` I get the following error

screen shot 2018-01-08 at 20 52 58

the `e.screenToPixel` function seems to fail I can dig a little deeper and report back but I can consistenly reproduce it using `undotree` or my custom fn not sure what exactly screen to pixel does or whether or not either of the triggers opens a buffer in a special way that causes this error.

A thought the error message also raises is whether it might be worth adding a `react Error Boundary` at a higher level component and take advantage of `componentDidCatch` so the app isn't taken down potentially not sure this is a 100% catch all or that recovery from certain errors is possible.

bryphe:
__Issue:__ Use of any plugin or VimL which creates a window, but does not navigate to it, or immediately navigates away, causes Oni's UI to crash.

__Defect:__ We always assume that the windows `screenToPixel` function is available, but in the case where we've never navigated to the window, and gotten the full context for it, it's not available. This is used to determine the size of the window.

__Fix:__ In the case where we don't have a `screenToPixel` function, we won't render the layers (they'll be size 0). This is fine, because in all the scenarios we need the layers today, the user explicitly navigates.

Also added a regression test - a quick and easy way to repro this is `:vsp` followed by `:wincmd p`. This regression test simply executes that and validates our UI is still around.


bryphe:
Fixes #1295

bryphe:
Given the TravisCI issues / delays, I ran the tests locally, and they were green.

PR #1310, opened 2018-01-16
Feature/initial Git VCS provider

### LINKED ISSUE #1308: Add git changes indicator to git plugin
## Problem/Desired Feature

@bryphe I've noted in apps like `vscode` or `hyperterm`, or even vim statusline plugins like `airline` with `fugitive` the git plugins include information regarding changes to branch such as deletions and additions, which is quite helpful at a glance to know where you stand

### Example (hyper term)
screen shot 2018-01-16 at 17 40 22

I was thinking something simple like a plus for additions, minus for deletions and maybe a similar symbol to the above for changes. 

I realise this will require some extra methods on the git service which theres some future plans to refactor or re-implement but would adding those methods and showing that in the git service/plugin as it stands be ok?.

### Implementation Ideas

I was thinking this could be approached using an `npm` package (which is one of the many values of having Oni as a frontend aka access to node libraries)

The package which comes to mind is [`simpleGit`](https://github.com/steveukx/git-js) which is a lightweight wrapper around git and a very good jumping off point for loads of git functionality like `cloning repos`, `renaming branches`, `merges` etc.

### Potential Issues
The way that vim handles changes in buffers is an area of concern as I've looked at the implementation of plugins like `fugitive` and `gitgutter` and one particular area of concern is that whilst editing a buffer if unsaved edits are actually not saved to the same location on disk so these plugins use a series of work arounds for that. 

For the scope of what I intend a naive approach would be only reflect changes on save, however the limitation there would be that by going down this route rather than using something like `vim-fugitive` is that this is a later issue which might be faced although potentially a similar workaround can be implemented at the time if it comes to it

Akin909:
Having had a bit of spare time I've refactored this PR to move the `git` specific functionality into the `oni-plugin-git` and use instead an abstracted `VersionControlManager` to render ui elements as per #2110, just wanted to point out that this is still an initial implementation so doesn't have all the bells and whistles but hopefully can be view a further step towards achieving #2110 

### Sidebar plugin
vcs

### Status bar changes
screen shot 2018-05-12 at 14 37 36

But what I have added so far:
 * Git VCS is now derived from `oni-plugin-git`
 * VCS service in oni now handles rendering `statusbar`, `sidebar`, `menu`
 * functionality changes deletions and additions show in status bar
 * Sidebar pane which shows modified files, selecting one stages the file.
 * Menu triggered via `oni.{vcs}.branches` show local branches, selection checkout the branch

Additional Changes (development-related):
 * I've switched some oni `package.json` commands to `npm-run-all` (which uses yarn or npm depending on the command) has a much tidier syntax which could potentially make the `package.json` readable again.

```js
example:  {
	build:js: "yarn do stuff",
	build:html: "yarn do stuff",
	build:css: "yarn do stuff",
	// with npm-run-all: accepts globs * and ** to run matching commands
	// run-p runs in parallel and run-s in series
	build: "run-p build:*"
}
```
### Todo

- [x] Fix buffer layer test - this fails due to issues building this branch which relates to onivim/oni-types/pull/1 i.e. issues importing subject from `rxjs`
- [x] Add enzyme tests for components 
~~- [ ] Render lists with react-virtualized~~ Required changes in #2182 - so will hold off adding this till that is merged in
- [x] Make sections collapsible
- [x] Fix filename overflow issue
- [x] Gate sidebar under experimental flag

wil fix #1308


Akin909:
Potentially need to switch `lib` to `nodegit` which explicitly supports `electron` but has a slightly more involved API as a packaged version of this PR doesn't show the changes (granted that this is due to the package and not an undiagnosed bug).

vim/core/oni-plugin-git/index.js
```diff
@@ -31,8 +30,16 @@ const activate = Oni => {
...
         let branchName;
+        let summary;
         try {
           branchName = await Oni.services.git.getBranch(dir);
+          try {
+            summary = await Oni.services.git.getGitSummary(
+              Oni.workspace.activeWorkspace
+            );
+          } catch (e) {
+            console.warn('[Oni.Git.Plugin]: Could not get Summary', e);
```
bryphe: Is there a case where we could not the summary, but we could get the branch? If not, it might make sense to simplify and consolidate these try/catches

bryphe:
> What I've implemented here is using simple-git which thankfully is typed and promise compatible. It provides very handy wrappers around git commands and added the number of additions and deletions to the git statusbar. Given that I rushed ahead I'm more than open to feed back re choice of library, changes to the Git service etc.

Excellent. I think `simple-git` is fine - it seems like a pretty thin wrapper around the CLI git interface.

Performance is important to consider - but it seems like we're just hooking this to the existing `FocusGained` hook, as well as on buffer saved, which I think this is fine.

I tried this out locally though and didn't see the status indicator - is that related to needing to switch to `nodegit`?

vim/core/oni-plugin-git/index.js
Akin909: That's essentially there for development purposes I'll remove that once the PR is finalised

Akin909:
@bryphe  are the changes not working in `development` for you or when you run the packaged app? I thought there might be a need to switch the library because it wasnt working *once packaged* and `nodegit` advertise their compatibility with electron so I thought electron might have a unique issue with this but having looked at the code base for `simple-git` its essentially a wrapper around child process git command execution which we do to get the branch so it's likely there's a bug in my implementation somwhere

Akin909:
@bryphe I think I've gotten to the bottom of the error here which is that the packaged app has no access to git, I'm not sure how the `Child process exec` method used for the branch name works but I think that may have to do with using `shellEnv` which means that it has access to the system git

A solution to this would be just to replicate a similar method to manually determine the changes but something that occured to me here is that using `nodegit` or `simple-git` could make for a very straight forward jumping of point for way more git functionality in oni, like `merging, cloning, branching etc.` 

Node git seems to come with git bundled but there's an issue on the repo about current [issues](https://github.com/nodegit/nodegit/issues/1347) installing it which I'm facing on top of which its api is extensively but arguably poorly documented. I'm looking into seeing if the packaged app can be bundled with git and am wondering if you know about this as I can't find any information regarding this? (that way I can stick with simple git and a use a bundled git executable)

Akin909:
I've settled on creating a `diffSummary` function manually having had a look at how `simple-git` does the same. I don't think though that the `Git` service if based on hand written commands like this will scale well if adding more functionality like this. 
An example of which is that by doing it manually if git gets updated if the output of the command I'm using changes the parsing I'm doing could entirely fall apart so its somewhat brittle.

I'm going to log an issue with `simple-git`, maybe they can provide an explanation or point me in the  right direction of whatever the cause maybe and if no joy there I can always check back in with `nodegit`.

bryphe:
Yes, `simple-git` looks like a good choice! Thanks for logging the issue there.

> EDIT: I'm considering removing the perFile changes as I'm not sure they aren't just confusing and currently I'm not entirely certain they actually work correctly

Seems reasonable - we can continue to build on it!

browser/src/Services/Git.ts
```diff
@@ -5,28 +5,109 @@
...
+        if (isRepo) {
+            // status = await project.diffSummary()
+            // if (status.files || status.deletions || status.insertions) {
+            //     return status
+            // }
+            const options: IExecOptions = {
+                cwd: currentDir,
+            }
+            const cmd = `git diff --stat=4096`
+            try {
```
steveukx: Following up from your issue https://github.com/steveukx/git-js/issues/241 - this block should be equivalent of using:

```
status = await project.diffSummary();
```

Do you find that the `status` is an instance of the `simple-git` object rather than an instance of `DiffSummary` when run this way?

Akin909: @steveukx the bug I'm currently facing is pretty odd and I'm not sure it isn't electron related. Basically when I run `await project.diffSummary()` **in development** it works perfectly for me though when I build and package the electron app the same variable `status` now returns the `simple-git` object so my assumption is somehow something changes once packaged.

Except when I run the raw command via `childProcess` that outputs the git command 😕, haven't found reference to a similar issue anywhere so i'm stumped

steveukx: That is strange - is it certain that the library is included as `require('simple-git/promise')` ? if it's included as `require('simple-git')` you would get the `simple-git` object back to chain it on to additional commands.

Akin909: Yes I've `imported` it as `git/promise` tried `simple-git` but switched over once I realised we would be able to use promises

Akin909:
@bryphe @CrossR I've slowly started transitioning this PR into a small initial implementation of a `GitVCSProvider` with a git branch menu for switching branches as well the status bar icons to show changes. I'm at my wits end with `simple-git` though as it does *exactly* what we need in a v. nice way but in *development* but the moment I run `yarn build && yarn run pack` the package stops functioning properly.

What happens is that each function call e.g. `simpleGit(directory).status()` which in developement returns the status in production just return the class which simple git provides, can either of you think of any reason that the behaviour would change going from dev to prod i'm wondering if it has to do with our minification or chunking ?? as there is a line in simple git [here](https://github.com/steveukx/git-js/blob/0770825ba9c111ce0affc353c7b3514aee01a7ef/promise.js#L59) which I think might be doing something to change the way it behaves

I've raised an issue on the repo and had some feedback here but I think I might need another eye that isn't too close to it to see if theres something I'm missing? I'd hugely appreciate any input it's soo nice in development and I can see being able to build on it quite straightforwardly if we can just get past this hump.

CrossR:
I'll try and have a bit more of a detailed look in a bit, but I can at least say that I see the exact same behaviour when I tested it last time (maybe a week back?), so it is repeatable and nothing specific to your machine etc.

bryphe:
> What happens is that each function call e.g. simpleGit(directory).status() which in developement returns the status in production just return the class which simple git provides, can either of you think of any reason that the behaviour would change going from dev to prod i'm wondering if it has to do with our minification or chunking ?? 

Hmm, this is definitely possible! What I'd recommend is setting up our `webpack` config so that we don't bundle `simple-git` - by default, webpack pulls everything in. We have a few libraries that have problems like this - so we set them up as `externals` in webpack - happens here:
https://github.com/onivim/oni/blob/ed3f49c04404d2bb5a50c06c3e5f365c962905ae/browser/webpack.development.config.js#L8

So you might want to try adding a line like: `"simple-git": "require('simple-git')"` and rebuilding.

At the very least, that library won't be minified anymore - so it means we'll hopefully be able to troubleshoot and figure out whats going on! 👍  Let me know how that goes.

Akin909:
@bryphe thanks for the help re. the externals 🎉 haven't tried it out but that could be the end of 2months of stalled progress 👍 , v. much appreciated.

Re. the VCSProvider, I think I perhaps overstated what I was attempting as whilst working on this or rather fiddling since it was only occasionally working I noted the issue you raised #2110 and thought it seemed like several PRs (ideally, rather than a mega one) worth of work to move out all the git functionality into a plugin and have it implement oni's interface etc.

I was a aiming for what I was thinking of as a *small step* in that direction i.e. using a class there and having it implement an interface etc. so that going forward it could be progressed, definitely not intended to be a fix or anything close to that for #2110 more of an acknowledgement of the desired direction 😄 hopefully as it is the git provider can just be pasted into the plugin and the plugin logic added to core.

Akin909:
Hurray!!!!!!! it does work once the webpack config is updated again much appreciated re the input @bryphe. I only recently added `fsevents` to that for `chokidar` due to a similar issue but had zero perspective after months on how to fix the issue (although now I wonder if I should have just required chokidar in the externals although the errors I was getting related only to fsevents and settled once I required it in externals).

Given that it works currently I just wanted to outline what it entails (which I imagine you've seen by now) aka just a change to the status bar item and a local branch menu which checkouts the branch on selection.

screen shot 2018-05-08 at 22 16 22


I think I might have stepped in it a little by alluding to #2110 but I wasn't envisioning this PR as a medium-big step in the direction of that but rather as a v. small one, I'd very much like to remain close to the original goal of this PR which was to add a couple of features to oni's current git functionality.

Moving forward now that it works once I get the time to put in I can look a bit at changing the services to match with what #2110 intends but for now granted the functionality I've added is stable I'd rather wind this PR down.

Akin909:
Having had a bit of spare time I've refactored this PR to move the `git` specific functionality into the `oni-plugin-git` and use instead an abstracted `VersionControlManager` to render ui elements as per #2110, just wanted to point out that this is still an initial implementation so doesn't have all the bells and whistles but hopefully can be view a further step towards achieving #2110

Akin909:
@bryphe this is currently blocked by an error trying to build the plugin using `oni-types` I noted onivim/oni/pull/1492 referenced this bug and that it relates to how `Subject` is imported from `rxjs` I will try and push a PR for that though unless another another is cause? (I use the types to get the `Event`)

Note: RxJS have recently moved to v6 which now recommends importing as `import { Subject, etc,} from "rxjs"` which might resolve the issue

browser/src/Plugins/PluginManager.ts
```diff
@@ -23,6 +24,11 @@ export class PluginManager implements Oni.IPluginManager {
     private _anonymousPlugin: AnonymousPlugin
     private _pluginsActivated: boolean = false
     private _installer: IPluginInstaller = new YarnPluginInstaller()
+    private _pluginsLoaded = new Event<void>()
```
Akin909: Added this when initially trying to figure out how to get access to the `VCSProvider` however I've left it in as there didn't seem to be away previously to await all plugins being loaded then execute a function

Akin909:
@bryphe it'd be great to get your feedback on this PR as I'm sure if I spend any more time on it I could get v. carried away. There's still plenty that can be done down the line but as far as an _mvp_ that moves the vcs integration hope this covers a lot of what was intended

CrossR:
Just had a quick test of the functionality, I'll check the code in a mo as well.

The status bar showing changes works great! The side bar modified and untracked works fine, but the staged section seems a bit confusing to me. It seems to not show new files. The showing of existing files worked fine, though I couldn't work out how to unstage (though I've not read the code yet, so that might be cleared up in a second).

image

I added two test files, and they showed up as untracked fine, but did not show up as staged once selected.

Only other thing I could see was that the status bar correctly updates on save, whereas the sidebar I found myself closing and reopening it to get it to update.

vim/core/oni-plugin-git/src/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+            throw new Error(error)
+        }
+    }
+
+    private _getModifiedAndStaged(files: FileSummary[]): { modified: string[]; staged: string[] } {
+        return files.reduce(
+            (acc, file) => {
+                if (file.working_dir === "M") {
+                    acc.modified.push(file.path)
+                } else if (file.index === "M") {
```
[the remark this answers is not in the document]
Akin909: Thats almost certainly where its going wrong, `simple-git` is fairly well documented although at some points re. the result the docs just point you to the code so it isn't always immediately clear what certain things return so will definitely take a look there to see what the symbol is for new untracked files to add to that section

browser/src/Services/VersionControl/VersionControlView.tsx
```diff
@@ -0,0 +1,224 @@
...
+                            titleId="staged"
+                            files={staged}
+                            selectedId={selectedId}
+                            visibility={this.state.staged}
+                            onClick={this.props.handleSelection}
+                            toggleVisibility={() => this.toggleVisibility("staged")}
+                        />
+                        <GitStatus
+                            files={untracked}
+                            symbol="?"
```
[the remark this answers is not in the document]
Akin909: @CrossR thanks for the feedback, i've run it locally intermittently for a while so definitely don't have fresh eyes anymore, I could do a quick replacement with another FA icon and see how that looks, re staged and unstaged i'd be up for changing that but I think a little later only because it'll require a bit of work to remove that since I've structured it with that particular separation in mind

Akin909:
@CrossR thanks for the feedback those are definitely the sort of issues I need feedback on since I made a lot of assumptions, I'll have a look at ensuring the sidebar correctly updates on save and we actually pick up modified prev. untracked files 👍

Akin909:
@CrossR I think I've addressed the issues you raised re. saving I've added an event handler to pick that up, re. the icons I had a fiddle but tbh think I'd rather save all that for another PR since tbh I cant settle on ones that I think work well, more because FA I think don't really have ideal symbols for git stuff, there are `octicons` which are specifically git icons and symbols I was thinking of for those bits so will leave it till i've explored those.

re. unstaging tbh I haven't added that yet I sort of stopped adding functionality to it when I realised how big the PR had grown so I still plan on adding (in another PR) unstaging, commiting then, uncommiting, but commiting will require a bit more work as I'd need to provide and input/textarea for users to enter a message

CrossR:
Yeah, the points I made about the icons and looks were just comments for future PRs, they can wait!

The staging/unstaging works great for me now, for both new and existing files, but I now get an error and no status bar Item. Though I admit, I may have jumped the gun and have tested it a bit too early...

image

Is the error I get if useful, seems to happen on Save.

Akin909:
@CrossR thanks for checking back again, yeah theres this strange error which seems to crop up with the types getting mismatched its not actually an error in the code as far as I can see, pretty sure its related to a change I made in the package.json, I'll work on fixing that bug and improving test coverage then will poke you again when thats done so you dont have hop back and forth testing interim versions

Akin909:
@bryphe I've added tests for all the new vcs stuff using `jest` but the codecov and diff don't seem to reflect this no matter what I add, is this likely due codecov not picking up jest test?

@CrossR think I've addressed the issues in your review except for the icons which I'm going to leave for now if you wanted to have another look over this if you get a chance (side note I put the new sidebar item under an experimental flag till the functionality is fleshed out)

Akin909:
Actually seems like jest is not collecting coverage at all re. setting up the upload all examples I've seen indicate that this is done with the current settings we are using aka it is set to generate a coverage report and out put it into the the coverage directory not sure if anything else is required remotely

CrossR:
Changes look good to me, and the bits that weren't working for me before (new files + reload on save) work great now! Very excited for this!

Akin909:
@CrossR thanks for the review 👍 bringing this in now them 🎉

bryphe:
Woohoo! Congrats on getting this in @Akin909  - really exciting! Lays the foundation for lots cool possibilities with VCS

Thanks for all your hard work on this @Akin909 , and thanks @CrossR for reviewing it! 💯

PR #1312, opened 2018-01-17
Fix #571 - Persist Opened Folder

### LINKED ISSUE #571: Open Folder: Persist opened folder on launch
If the user has opened a folder, and then launches Oni again from an install directory, we should persist that and default to opening from their previous folder.

This fits in with persisting other editor settings - like window size (#452), open files, etc.

bryphe:
This change adds the ability to persist configuration values, by calling `Oni.configuration.setValues({ ...values }, true /* persist */)`, plus some unit tests.

It also includes persisting the current workspace path, in the `workspace.defaultWorkspace` setting. It's meant really to be persisted, but you can override the persisted setting by include it in your `config.js`.


bryphe:
Fixes #571

PR #1313, opened 2018-01-17
Fix #1235 - Use 'ONI_WEBPACK_LOAD' environment variable instead of 'NODE_ENV'

### LINKED ISSUE #1235: Blank black screen
I am on OS X 10.13.2, running latest head Neovim, and build Oni from master. When I launch Oni I get a blank black screen.

bryphe:
It's still not clear all the cases that cause #1235 , but there is definitely a conflict with the `NODE_ENV` variable - if `NODE_ENV` is set, we'll try and load from webpack-dev-server instead of the local file. This is problematic for anyone who uses `NODE_ENV` in their local workflow, so we should use a different environment variable to gate this.

### DIFF
--- a/index.html
+++ b/index.html
@@ -28,7 +28,7 @@
       <script>
         console.timeStamp("browser.domloaded")
         var path = "lib/browser/bundle.js"
-        if(process.env.NODE_ENV === "development") {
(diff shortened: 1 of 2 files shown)


bryphe:
Related to #1235

PR #1318, opened 2018-01-18
Bugfix/use editor dimensions

Akin909:
Based off of #1253 (so this PR must follow it and update with it and master). This PR changes the viewport reducer to add dimensions for the `.focused.editor` which fixes the situation where the sidebar is open which reduces the viewport which is the current basis for positioning the hover and autocomplete menus.

Initially the `focused.editor` DOM element does not exist so on `componentDidMount` of the cursor positioner component `updateViewport` is called to update the store with new dimensions

### Issues

Currently there is a quirk on initial load where the position initially used is the default which suddenly gets updated in `cdm` causing a visible very shortlived and once of jerk to the opening, this can probably be fixed by moving the `updateViewport` call to the mounting of the editor element itself or another react component the renders before the `cursorPositioner`


Akin909:
@bryphe completed this now I refactored away from depending on the classNames `focus editor` and instead added a resize observer to the `NeovimSurface` couldn't be placed higher as the store is initialised for the editor only a component above.

browser/src/Editor/NeovimEditor/NeovimEditorActions.ts
```diff
@@ -52,8 +52,12 @@ export interface IAddBufferLayerAction {
 export interface ISetViewportAction {
     type: "SET_VIEWPORT"
     payload: {
-        width: number
-        height: number
+        width: number,
+        height: number,
+        focusedEditor?: {
```
bryphe: I'm wondering why we need separate `width` and `height` here - is there a way we can simplify/consolidate these? Ideally, `NeovimEditor` should only need to be concerned with its own width and height.

bryphe:
Thanks @Akin909 ! How do I reproduce this? When I test quick info / completion, it seems to show up in the proper spot:
image

Just want to make sure I understand the repro (and the fix).

browser/src/Editor/NeovimEditor/NeovimEditorActions.ts
bryphe: (So in other words - it seems like we should only care about the width and height of the editor)

Akin909:
@bryphe you make a good point re consolidating the height and width I was under the impression that we needed the full viewport dimension somewhere for some reason but if that data is primarily used in an editor context consolidation is much better.

Re. reproducing the error the easiest way to do this on `master` go to split screen, find a long line and wait for the tool tip to appear depending on the size of the tooltip aka if it is wide it will run off the screen without the changes in this PR. another way to reproduce it is to get the signature helper to open near the end of a long line and it will begin to extend of the screen.

The crux of the issue is that the width of the `document.body` is on my machine for example approx 1200px and the actual editor not including the sidebar is about 900px so the tooltip cursor positioner thinks there is more space available than there actually is also, once the multiplexing is merged this will become more apparent as the tool tip will bleed across instances

bryphe:
> @bryphe you make a good point re consolidating the height and width I was under the impression that we needed the full viewport dimension somewhere for some reason but if that data is primarily used in an editor context consolidation is much better.

Ah ya, sorry about this - it's actually vestigial from refactoring. I moved a bunch of redux state from the original store into a per-editor store, and that was left. But I think it can be consolidated now.

And thanks for the details about the repro - that makes sense! I did notice the tooltips going offscreen but didn't realize it was due to this. Makes sense that we need to adjust this.

Akin909:
Cool I've tweaked this to consolidate the height and width or rather just have one which references the individual editor

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -245,15 +245,6 @@ export class NeovimEditor extends Editor implements IEditor {
             this._symbols,
         )
 
-        const updateViewport = () => {
```
bryphe: Nice, thanks for cleaning this up! Doesn't make sense anymore, now that the editor is more self-contained.

browser/src/Editor/NeovimEditor/NeovimEditorReducer.ts
```diff
@@ -193,11 +193,13 @@ export const definitionReducer = (s: State.IDefinition, a: Actions.SimpleAction)
 }
 
 export const viewportReducer = (s: State.IViewport, a: Actions.ISetViewportAction) => {
+    const { width, height } = a.payload
     switch (a.type) {
         case "SET_VIEWPORT":
             return {
-                width: a.payload.width,
-                height: a.payload.height,
+                ...s,
```
bryphe: Glad this is simplified now!

bryphe:
Just tried out the change, looks good to me - I like that the logic is a bit simplified now and the resize logic is removed, and we just have the `ResizeObserver` now - fits in line with how the `NeovimEditor` is self-contained inside the window manager. Looks like there is just a merge conflict to resolve, but otherwise I'm set to bring this in. Thanks for all your work on this, @Akin909 !

Akin909:
@bryphe fixed the conflict this is good to go now

bryphe:
Thanks @Akin909 for your work on this! Bringing it in now.

PR #1325, opened 2018-01-19
[WIP] Feature/prepush hook

### LINKED ISSUE #1317: Pre-Push +/- Pre-Commit hooks
@bryphe  this issue is aimed primarily at development, I'm not sure what your feeling might be regarding setting up `prepush` +/- `pre-commit` hooks for oni which at the very least run `tsc` and `yarn lint` commands to avoid situations where a developer might push code which was inevitably going to fail in CI as they didn't run linting or tsc before hand I personally find that I forget all the time and locally and at work use [`husky`](https://github.com/typicode/husky) which is an npm package that sets up a dev's `git` folder with a pre-push hook.

As for pre-commit hooks I personally them a little hit and miss as they can interrupt the flow of developing but its also possible to run things without verification.

Also related but definitely more controversial is the idea of using the spectacular [`Prettier`](https://github.com/prettier/prettier) auto formatter (I'd suggest having a look at its repo if you're unfamiliar with it as its very much unlike other auto-formatters in the way it works) to run pre-push or pre-commit to format all repository code. Whilst I appreciate that people tend to be quite invested in how they format code using an opinionated formatter like prettier can ensure that the formatting of the code base is always consistent with a few preset rules, and that there aren't disparate areas of unique formatting which don't match other areas, whilst tslint provides some guards which are more about say preferring arrow functions etc. 

`Prettier` is just very robust means of formatting code which is I would say far more advanced at actually *re-formatting code* . This would be more about consistency as hopefully `oni` grows and hopefully gets more contributors and differences become more apparent. Also it would require no effort on the contributors part to ensure their code matched with the rest of the code base, as prettier would just run and only fails if the code has parsing errors in it in which case it just fails and the hook itself should fail.

Either way depending on your feeling I am happy implement the simple hooks pre-push for linting and tsc or if you're interested in `prettier` I can set that up as well.

Akin909:
This PR will fix #1317 by adding a `prepush` hook for `yarn lint and tsc` and [WIP] uses `prettier` (with a `prettierrc` config to match the project) as well as `pretty-quick` to run reformatting.

### To Do
~~Prettier hook is currently `precommit` which I am going to try to change to `prepush` however the issue with this is that `prepush` all code has already been committed so the hook will have to autoformat commit and re-push. FYI using `pretty-quick` gives a clear concise update on the result of the formatting as well as what was formatted and on average takes about a second on pre-commit however I will try to move this action away from `pre-commit` as discussed with @bryphe~~

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,6 @@
   "printWidth": 100,
   "semi": false,
   "singleQuote": false,
-  "trailingComma": "es5"
+  "tabWidth": 4,
+  "trailingCommas": "all"
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -3,5 +3,5 @@
   "semi": false,
   "singleQuote": false,
   "tabWidth": 4,
-  "trailingCommas": "all"
+  "trailingComma": "all"
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "printWidth": 100,
+  "semi": false,
+  "singleQuote": false,
+  "trailingComma": "es5"
+}
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "build:test:unit": "cd browser && tsc -p tsconfig.test.json",
     "copy-icons": "node build/CopyIcons.js",
     "debug:test:unit:browser":
-      "cd browser && tsc -p tsconfig.test.json && electron-mocha --interactive --debug --renderer --require testHelpers.js -R testCoverageReporter --recursive ../lib_test/browser/test",
+      "cd browser && tsc -p tsconfig.test.json && electron-mocha --interactive --debug --renderer --require testHelpers.js --recursive ../lib_test/browser/test",
     "demo": "npm run build:test && mocha -t 30000 lib_test/test/Demo.js",
     "dist:win":
       "cross-env ELECTRON_BUILDER_ALLOW_UNRESOLVED_DEPENDENCIES=1 build --arch ia32 --publish never",
@@ -290,6 +290,7 @@
     "sudo-prompt": "7.1.1",
     "ts-loader": "2.3.2",
     "tslint": "5.7.0",
+    "vscode-snippet-parser": "0.0.5",
     "wcwidth": "1.0.1",
     "webdriverio": "4.8.0",
     "webpack": "3.5.3",
--- a/package.json
+++ b/package.json
@@ -134,6 +134,10 @@
     ]
   },
   "scripts": {
+    "prepush": "lint-staged",
+    "lint-staged": {
+      "*.{ts,tsx,less,md}": ["prettier --write", "git add"]
+    },
     "build": "npm run build:browser && npm run build:main && npm run build:plugins",
     "build-debug": "npm run build:browser-debug && npm run build:plugins",
     "build:browser": "webpack --config browser/webpack.production.config.js",
@@ -256,6 +260,7 @@
     "fs-extra": "4.0.2",
     "fuse.js": "2.6.2",
     "github-releases": "^0.4.1",
+    "husky": "^0.14.3",
     "innosetup-compiler": "5.5.9",
     "jsdom": "11.0.0",
     "less": "2.7.1",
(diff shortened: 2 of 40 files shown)


bryphe:
Just tried this out - I think the `precommit` action is reasonable here. Great find on using `pretty-quick` on the staged files - seems like a reasonable compromise. I tested a change and it indeed is fast, and doesn't seem like it gets in the way of the 'flow'. Open to trying it out this way!

package.json
```diff
@@ -134,55 +114,76 @@
     ]
   },
   "scripts": {
+    "precommit": "pretty-quick --staged",
+    "prepush": "tsc -p browser/tsconfig.json && yarn lint",
```
bryphe: Seems like this is only build the browser - would it make sense to run `npm run build && npm run lint` to validate all the projects we build?

package.json
```diff
@@ -134,55 +114,76 @@
...
     "build:plugin:oni-plugin-typescript": "cd vim/core/oni-plugin-typescript && npm run build",
-    "build:plugin:oni-plugin-markdown-preview": "cd extensions/oni-plugin-markdown-preview && npm run build",
+    "build:plugin:oni-plugin-markdown-preview":
+      "cd extensions/oni-plugin-markdown-preview && npm run build",
     "build:test": "npm run build:test:unit && npm run build:test:integration",
     "build:test:integration": "cd test && tsc -p tsconfig.json",
     "build:test:unit": "cd browser && tsc -p tsconfig.test.json",
     "copy-icons": "node build/CopyIcons.js",
-    "debug:test:unit:browser": "cd browser && tsc -p tsconfig.test.json && electron-mocha --interactive --debug --renderer --require testHelpers.js -R testCoverageReporter --recursive ../lib_test/browser/test",
+    "debug:test:unit:browser":
```
bryphe: Does prettier make these revisions to the `json` file? I notice that there is some tool (I think its `yarn`) that un-does these edits.

bryphe:
I'm thinking it'd be good, as a separate change, to just run prettier over the whole code base and get a good default - that way PRs won't have a ton of formatting changes. 

We have a flood of PRs on deck though, since TravisCI's mac builds have been down... but once those are in, and things are quieter (less merge conflicts), it may make sense to minimize the surface area of incoming changes.

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "printWidth": 100,
```
bryphe: Thanks for setting up these defaults! 💯

package.json
Akin909: @bryphe yes that was `prettier's` doing, it works on `md` and `json` as well an they are currently expanding its support for other languages and filetypes

package.json
Akin909: @bryphe great suggestion I was just testing out commands at that point but those would be good commands to run there wasn't sure which to use

bryphe:
Sounds great, @Akin909 ! Really appreciate the work you did here.

Once the queue dies down:
image
And we can clear _our_ PR queue, we can bring this in + run `prettier` on the whole code base.

I like having an enforced style a lot, as it then keeps discussions focused on higher-level things like design/architecture/features as opposed to code style/whitespace/etc. 👍

bryphe:
Cool, the change looks good @Akin909 ! Can we scope this PR to just the hook + prettier updates? And then have a second PR that runs `prettier` over everything? I think that will help us stay ahead of the merge conflicts.

But I'm ready to bring this in when you are-

Akin909:
@bryphe I've tried to minimise changes here however to keep up to date with master when I merge the upstream the hook runs although in hindsight I could have run them no verify it might be worth creating a new branch hopefully which we could merge with haste to avoid the commit hook kicking in at some point due to upstream changes

bryphe:
> @bryphe I've tried to minimise changes here however to keep up to date with master when I merge the upstream the hook runs although in hindsight I could have run them no verify it might be worth creating a new branch hopefully which we could merge quickly to avoid the commit hook kicking in at some point due to upstream changes

That sounds like a good plan! Then, all future changes will be 'prettified' so it should make the merge simpler.

Akin909:
Closed in favor of #1361

Issue #1300, opened 2018-01-16
Input: '<space>' should be used for bindings, as opposed to the literal space character


bryphe:
Repro:

- Add  `oni.input.bind("<space>", () => { alert("hi") })` to your config, save.

The `alert` should show, but does not.

Note that if instead you use an explicit space character via `oni.input.bind(" ", () => { alert("hi") })`, it works.

This is not ideal, because this is sensitive to whitespace trimming - we should be using the explicit `<space>` instead.

someguynamedmatt:
@bryphe Been a while since I've expolored through the codebase. Would you be able to point me towards the file that reads from the users config file? In other words, whichever file parses their settings?

Thanks!

bryphe:
Hi @someguynamedmatt , good to see you here again 😄 

> Would you be able to point me towards the file that reads from the users config file? In other words, whichever file parses their settings?

For sure - this is the file that loads the user's configuration file:
https://github.com/onivim/oni/blob/a5cbb53fce654c0047cd60f46a48028e84e53ff4/browser/src/Services/Configuration/FileConfigurationProvider.ts#L115

And the input stuff - the implementation of `oni.input.bind` - is here:
https://github.com/onivim/oni/blob/a5cbb53fce654c0047cd60f46a48028e84e53ff4/browser/src/Services/InputManager.ts#L28

someguynamedmatt:
@bryphe Fantastic! Thanks! I'll start messin' around with it when I get a free moment.

bryphe:
Fixed on master w/ #1333 - thanks @someguynamedmatt !

PR #1333, opened 2018-01-21
[ISSUE #1300]: <space> rather than literal

someguynamedmatt:
#1300 @bryphe I think this is the solution to the issue, assuming I understood the issue _correctly_ 😄

### DIFF
--- a/browser/src/Input/Keyboard/Resolvers.ts
+++ b/browser/src/Input/Keyboard/Resolvers.ts
@@ -40,6 +40,7 @@ const keysToRemap: { [key: string]: string } = {
     "ArrowRight": "right",   // ArrowRight
     "ArrowDown": "down",    // ArrowDown
     "Insert": "insert",
+    " ": "space",
 }
 
 export const remapResolver = (evt: KeyboardEvent, previousResolution: string | null): string | null => {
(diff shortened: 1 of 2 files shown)


bryphe:
Changes look great to me, @someguynamedmatt ! Looks like the right spot to me 😄 

Could you please add a test to cover this case as well? We have some cases for this started here:
https://github.com/onivim/oni/blob/9ba6f55b6ee5664e9b7cea8651e2e2368fca3309/browser/test/Input/Keyboard/ResolverTests.ts#L19

It should be pretty similiar to the existing case we have there, except for space instead of backspace.

You can test it by running `npm run test:unit:browser`, or debug it via `npm run debug:test:unit:browser`.

Thanks for the PR, appreciate it!

someguynamedmatt:
@bryphe I’m always forgetting the tests. Lol. I’ll get that up soon.

someguynamedmatt:
@bryphe Good to go.

bryphe:
Excellent, looks great! 👍

bryphe:
Linux failure isn't related to your change, so I'll bring this in now. Thanks for the contribution, @someguynamedmatt ! 👍

PR #1334, opened 2018-01-21
Feature/buffers menu

### LINKED ISSUE #1173: Open Buffers/Files QuickOpen Window
@bryphe now that oni has access to a list of buffers I was thinking that It would be really great to use this list as a quick open menu you could use to navigate different buffers.

In the same vein as `fzf`'s `:Buffers` command which is what I currently use then you could use this menu to switch buffers. The difference between this and the regular find file command would be that it would only show you the files you have open rather than having to constantly filter through everything.

I realise that the file explorer will potentially solve this problem however my thinking was that generally I personally wouldn't open something like `NERDtree` or have a split in my editor to find my open buffers but would rather be shown a short list I could quickly select one from.

I can implement this (I think) but wanted to hear your thoughts before I went ahead in case you have other plans regarding something like this. My general thinking re implementation would be to add another function to the `MenuManager` like `listOpenBuffers` somehow getting that from the either redux or a vim event, probably redux

Akin909:
This PR adds a open buffers menu, which fixes #1173,

### Changes: 
* Added a oni-plugin-buffers plugin and registered associated commands
* Created an `inactiveBuffer` class to ensure active and inactive buffers are passed in (bufferManager only had access to the visited buffers)
* Changed menu styling to prevent text wrapping if long item name

### Todo

- [x] Fix `openFile` typings (`oni-api`)
- [x]  Add commands to open splits, tabs and delete buffers
- [x]  Add `InactiveBuffer` to `Oni` api and reference `Oni.InactiveBuffer`
- [x] Add `metadata` field to `menuItem` to store original value of absolute file path

- [x] Fix buffer population - `buffersById` in `bufferManager` does not reflect buffers in vim.
- [x] Truncate filenames


Akin909:
@bryphe Ideally I'd prefer to add the change were it'd be most useful long term although an issue I faced having just tried this out is that it doesn't actually read the non-implemented function understandably, can you  point me towards how to get it to look at the correct function? I'm not clear on how it calls the function in neovim,

From #1316 seems there are a few changes that will be happening do I need some of the larger changes to be incorporated for this to work or is there a smaller change I can make to get this working. its all in all a small change that can easily be copy/pasted over later either way

bryphe:
> Ideally I'd prefer to add the change were it'd be most useful long term although an issue I faced having just tried this out is that it doesn't actually read the non-implemented function understandably, can you  point me towards how to get it to look at the correct function? I'm not clear on how it calls the function in neovim,

Hmm, is this regarding `openFile`? 

> From #1316 seems there are a few changes that will be happening do I need some of the larger changes to be incorporated for this to work or is there a smaller change I can make to get this working. its all in all a small change that can easily be copy/pasted over later either way

Ah I was just trying to give you some visibility in terms of how I pictured things changing (since right now, most commands are 'global'). But for this, feel free to create a global command or whatever is easier. As you mentioned, it's a pretty small tweak to move it around.

Akin909:
@bryphe is it possible currently in Oni to create a binding that only applies when a menu is open?. I realise that I've now added several commands all of which can eat up bindings and are only valid when the menu is open

bryphe:
> @bryphe is it possible currently in Oni to create a binding that only applies when a menu is open?

Yep! That's the third argument to `bind`: 

```
        input.bind("<m-t>", "language.symbols.workspace", () => !menu.isMenuOpen())
```

The last argument is a `filter` - so in this case, the binding is only active when a menu is not open.

We have other ones, like to only make the binding available for normal mode:
```
           input.bind("<S-C-P>", "commands.show", isNormalMode)
```

Hope that helps!

And I saw you were waiting on the change for `oni-api`, I just brought that in/bumped the version/will release. Can't wait for this feature! 👏

Akin909:
@bryphe thanks for explaining this PR I think is almost done although I'd like to hide the implementation details about passing in a third argument from the user is it possible to create a command has already taken care of these details for the user from the plugin makers perspective

I know that there is commands file which has done that for certain command so would I need to add a command there or is there a way using the plugin API to accomplish this

bryphe:
> @bryphe thanks for explaining this PR I think is almost done although I'd like to hide the implementation details about passing in a third argument for the menu bindings from the user is it possible to create a command that has already taken care of these details for the user from the plugin makers perspective

Awesome! Yes, I agree, it's simpler to just have the binding be enabled when it makes sense, and not other times.

There's a couple different ways to accomplish this:
- When you register your command, you can pass in an `enabled` callback - that makes the command only enabled in specific cases. Even if the user has bound to it, it will only get executed if it's actually enabled. That's how the context menu bindings work, for example.
- You can also optionally return a result from the command itself - if you return `false` that means it was unhandled, and it should let the input manager try other bindings.

I like the explicit `enabled` callback for the command.

vim/core/oni-plugin-buffers/index.js
```diff
@@ -0,0 +1,109 @@
...
+  Oni.commands.registerCommand({
+    command: 'bufferlist.split',
+    name: 'Split Selected Buffer',
+    execute: () => openBuffer(menu, 'horizontal'),
+  });
+
+  Oni.commands.registerCommand({
+    command: 'bufferlist.vsplit',
+    name: 'Vertical Split Selected Buffer',
+    execute: () => openBuffer(menu, 'vertical'),
```
bryphe: This is where I was mentioning you can add an `enabled` filter - like - so you could bake in the filter to the command, instead of having the user needing to know to apply the filter. It definitely is more convenient to do it here, so that we always do the 'right' thing - and the user doesn't need to think hard about it.

bryphe:
Just tried this out, looks awesome! 💯 I'll definitely be using this.

image

Great work, @Akin909 !

Akin909:
Thanks @bryphe it's almost there just tried the filter method you recognised but ran into a type error with `openFile` so I've made a PR to the `oni-api`

bryphe:
Just pulled in the new `oni-api` typings - thanks for fixing those!

Akin909:
@bryphe all done with this now thanks for merging the `oni-api` PR, I just wanted to draw your attention to 2 changes I made (aka to oni autocommand hooks).

Whilst trying to hook up the buffer delete command I realised that currently in order to properly respond to a buffer being deleted oni is entirely dependent on a subsequent `bufEnter` action. I tried to fix this by adding 2 events to the `neovimEditor` - `BufDelete` which we had previously but I removed at some juncture as it actually occurs *before* the buffer is deleted and `BufUnload`. I realised that `BufWipeout` (which is being used currently) and `BufUnload` are called not entirely consistently or rather seemingly irregularly based on some vim-logic im unaware of aka sometimes `BufWipeout` which should be called *after* `BufDelete` is occasionally not called, and `BufUnload` I think is more consistently called, though __None__ of these commands are actually an accurate hook for detecting __After__ a buffer has been deleted.

I currently have no idea how to do this without forcing a bufEnter event which is what I do to ensure the mapping works, (or manually filtering out the deleted buffer which we discussed previously). I've left them in as I believe they can be useful hooks to have for some other `Oni` purpose rather than deleting them just to have to write them in again at some future date

browser/src/Services/EditorManager.ts
```diff
@@ -93,9 +93,22 @@ class AllEditors implements Oni.Editor {
         return this._activeEditor.neovim
     }
 
-    public openFile(file: string): Promise<Oni.Buffer> {
-        Log.warn("Not implemented")
-        return Promise.resolve(null)
+    public async openFile(file: string, method = "edit"): Promise<Oni.Buffer> {
+        const cmd = new Proxy(
```
bryphe: Could we just have this call into the `activeEditor`'s `openFile` method? I'd like for the `EditorManager` to be agnostic to `Neovim`, if possible - ideally, the `EditorManager` could support all sorts of `IEditor` implementations - even ones that didn't depend on Neovim. Looks like this is duplicated with `NeovimEditor`'s implementation anyway.

vim/core/oni-plugin-buffers/index.js
```diff
@@ -0,0 +1,111 @@
...
+        !menu.isOpen() ? createBufferList() : menu.hide()
+    }
+
+    const deleteBuffer = async menu => {
+        if (menu.selectedItem) {
+            await Oni.editors.activeEditor.bufferDelete(menu.selectedItem.metadata.id)
+            // This line forces vim to navigate to the next buffer after deleting
+            // NOTE: This is essential as otherwise Oni does not properly register the buffer
+            // delete event
+            await Oni.editors.activeEditor.neovim.command("bn")
```
bryphe: If this will always need to be done along with the `bufferDelete`, perhaps we should just push this inside the `bufferDelete` method?

bryphe: In addition - can we check if the current buffer's id (`activeEditor.activeBuffer.id`) is equal to the one we're deleting? If I understand correctly, we only need to do the `bn` if the buffer is the same, so if `activeEditor.activeBuffer.id === menu.selectedItem.metadata.id`.

Akin909:
Actually found an unusual bug, it seems the on using the filterFunction messes with the meta dat fields I've added to the menu Items which the plugin depends on for its functionality 🙁 , I'll look into that report once it's sorted out

vim/core/oni-plugin-buffers/index.js
```diff
@@ -0,0 +1,111 @@
...
+            } finally {
+                menu.hide()
+            }
+        }
+        return
+    }
+
+    Oni.commands.registerCommand({
+        command: "bufferlist.delete",
+        name: "Delete Selected Buffer",
```
bryphe: Could we add a common 'prefix' for these, like:
````
name: "Buffer: Delete",
detail: "Delete selected buffer",
```

Trying to organize the commands a bit 😄

bryphe:
Thanks for the details, @Akin909 ! I left a few comments on the PR. I'm not exactly sure how the buffer lifetime works, or why'd we see inconsistent results.

The buffer navigation seems like a reasonable workaround, but I'm wondering - can we check the case where the deleted buffer is equal to the current buffer, and conditionally execute `bn`? Or do we always need to execute `bn`? If it does always need to be executed to get the latest list - could we just do `bn`/`bp`/ in the case where we're not on the buffer we're deleting?

vim/core/oni-plugin-buffers/index.js
Akin909: 👍

Akin909:
Tbh I'm sure there's some reasoning there somewhere or a way around this issue but from reading the vim docs and autocommand hooks I'm not clear on where is the earliest *correct* place to hook into following a buffer being delete aka what action is unconditionally always run following a `BufDelete` when the buffer has definitely already been unloaded, but currently given up (at least for this PR)

As for the workaround I could pass in `%` as the buffer to be deleted and if it isn't the current buffer execute a `bn/bp` otherwise just a `bn`, although as what I hope is a temporary solution.

Also looking at the comments and will make those changes

vim/core/oni-plugin-buffers/index.js
Akin909: @bryphe did you intend for the commands to be in caps with the space there, I've changed the command to be `buffer:action` as from looking at the commands none start with caps and none have spaces in them ❓

bryphe: Yeah, no set standard yet, but here's what I'm thinking:

- command: camel case / lowercase - ie, `workspace.openFolder`
- name: noun + verb, capitalized - `Workspace: Open Folder`
- detail: sentence case - `Set a folder as the current workspace` 
(not a great example, but hopefully that helps).

I think the commands you have look good - I was thinking for the `name`/`detail` fields.

Like looking at the command palette we have today:
image

The 'Workspace' ones look good to me, but the 'Show Errors' looks like it's just kind of floating. It'd be nice to group it with some other error/diagnostic related commands (maybe like `Errors: Show`, `Errors: Go to next`, `Errors: Go to previous`, etc..)

Akin909:
All done 👍 I've changed the command names, moved the post deletion navigation to the command itself, and removed the `openFile` implementation to the neovimEditor, as well as fixed the metadata bug

bryphe:
Looks great, @Akin909 ! Bringing this in now. Really excited to start using this and have this part of the next release. Thanks for the contribution!

PR #1335, opened 2018-01-22
Automation: Fix instability with 'NoAdditionalProcesses' test

bryphe:
The 'NoAdditionalProcesses' test expects there to be two processes on startup, but it's actually only waiting for the editor neovim instance to be started - so in some cases, it's only seeing one process start.

In addition, I noticed that occasionally the very first test is failing - I believe this may be because it is slow to warm up on first boot & timing out. Trying to add more logging to catch and fix that.


bryphe:
Because of the TravisCI Mac delays, I've run this locally - all the tests passed.

PR #1338, opened 2018-01-22
Automation: Enable CiTests for linux builds

bryphe:
I want to have the tests run on at least two platforms build-over-build - since we are disabling OSX due to issues on Travis, this is an attempt to get the CiTests to run on the Linux machines.

### DIFF
--- a/build/script/travis-build.sh
+++ b/build/script/travis-build.sh
@@ -10,6 +10,11 @@ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
   export DISPLAY=:99.0
   sh -e /etc/init.d/xvfb start
   sleep 3
+
+  # Install Neovim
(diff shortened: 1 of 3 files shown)


build/script/travis-build.sh
```diff
@@ -10,15 +10,24 @@ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
   export DISPLAY=:99.0
   sh -e /etc/init.d/xvfb start
   sleep 3
+
+  # Install Neovim
+  curl -LO https://github.com/neovim/neovim/releases/download/v0.2.2/nvim.appimage
```
someguynamedmatt: Was this why the Linux builds were failing? (No Neovim instance?)

bryphe: Yes, exactly! Most of the ci tests rely on having Neovim around - since there wasn't an `nvim` to pick up, those would all fail.

PR #1339, opened 2018-01-22
Bugfix/broken tooltip conditional

Akin909:
@bryphe seems I missed a bug in a conditional I'm using to add padding to the documentation for the tooltip I noted the errors in my console whilst using a `vue` file  (I failed to check an iterim value existed which causes this error).


bryphe:
Thanks for catching this, @Akin909 !

PR #1358, opened 2018-01-25
Fix #98 - Create basic notification UX

### LINKED ISSUE #98: Implement notification UX

bryphe:


bryphe:
TODO:
- Hook up onClick event
- Hook up close event

PR #1361, opened 2018-01-25
Feature/prepush hook v2

Akin909:
@bryphe as discussed in #1325 this PR mirrors #1325 without any other changes. I'd recommend we merged this quickly if that's okay with you since given all the hooks any changes in master that get pulled in could accidentally cause the hooks to run (granted I forget to run `--no-verify` flag, which is inevitable)

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "printWidth": 100,
+  "semi": false,
+  "singleQuote": false,
+  "tabWidth": 4,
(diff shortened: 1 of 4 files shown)


bryphe:
Sweet, thanks for splitting this out @Akin909 ! Bringing this in now-

bryphe:
Since everything will be 'prettified' now, hopefully that will minimize merge conflicts now.

PR #1368, opened 2018-01-26
Fix #1234: Initialize neovim instances in sequence instead of parallel

### LINKED ISSUE #1234: Oni crashes on Ubuntu when opening from terminal
when I start oni from the terminal it crashes

terminal_001

oni_002

platform: Ubuntu 16.04
oni version: v0.2.19
neovim version: v0.2.2

bryphe:
Not sure why this is the case, but initializing the Neovim instances in parallel, on linux, seems to get the process stopped. Sequentially initializing them addresses the issue.


bryphe:
Fixes #1234 and the similiar issue with `npm run start`

someguynamedmatt:
👍 So awesome

PR #1375, opened 2018-01-26
[WIP] Fix #1258 - Fix quotes auto-insertion

### LINKED ISSUE #1258: auto closing paires not working with quotes
config part:
```js
{
    "autoClosingPairs.enabled": true,
    "autoClosingPairs.default": [
        { "open": "{", "close": "}" },
        { "open": "[", "close": "]" },
        { "open": "(", "close": ")" },
        { "open": "'", "close": "'" },
        { "open": "`", "close": "`" },
        { "open": '"', "close": '"' },
    ],
}
```

When I write a `(` I get `(|)` and I can type a `)` which leads me to `(foo)|` as expected. With quotes it doesn't work and I get `"foo"|""` instead.

CrossR:
Currently seems to work, but need to do a little more testing, as well as add a test so we know if this functionality breaks in the future.

My only real concern is that there is already a `TODO`  there about collapsing the code down due to the multiple async calls and I have added more  async calls now... Should I look about fixing that too? Though not 100% where to start with that.


bryphe:
Awesome! Thanks for the fix, @CrossR . Looks good to me.

> Currently seems to work, but need to do a little more testing, as well as add a test so we know if this functionality breaks in the future.

We might be able to add this to the "AutoClosingPairsTest" - https://github.com/onivim/oni/blob/8d796c4044679a9f78d91cb201a0bc6400c910a5/test/ci/AutoClosingPairsTest.ts#L27

We'd have to specify a custom configuration for it though to include the quote - some other CiTests do this here (include a custom `config.js` and specify the path as `configPath`):
https://github.com/onivim/oni/blob/8d796c4044679a9f78d91cb201a0bc6400c910a5/test/ci/Editor.ExternalCommandLineTest.ts#L21

It'd be nice to get this hooked up unit tests, too. I think we could at least get the `handleOpenCharacter` unit testable, by doing the following:
- Replace the `neovim.callFunction("getpos")` with the Buffer's `getCursorPosition` method.
- Refactor the code inside the `nvim.blockInput` block of `handleOpenCharacter` to a new function that we export - this means we can unit test the 'guts' which only depend on the buffer and the auto closing pair, and that's where the interesting stuff happens anyway
- We can use the `MockBuffer` to exercise cases - toss in some data and then validate that the output cursor position / line data is as we expect.

The snippet tests actually do some kind of similiar stuff - validate that the cursor moves to the right spot, or that the data is correct - might be useful if you want to pursue this:
https://github.com/onivim/oni/blob/8d796c4044679a9f78d91cb201a0bc6400c910a5/browser/test/Services/Snippets/SnippetSessionTests.ts#L41

Thanks for thinking about the tests!

> My only real concern is that there is already a TODO there about collapsing the code down due to the multiple async calls and I have added more async calls now... Should I look about fixing that too? Though not 100% where to start with that.

Nah, this isn't critical - I think it's just a premature concern on my part. And we can split it out if anything - but it's okay to defer this until we see a problem. The good thing is the `blockInput` call should ensure that the input sequence is deterministic.

CrossR:
I think I've got the CI test working now, at least it passes on my machine so should be fine.

The Unit tests I'm less sure about, currently the last one I added is failing and I'm not fully sure why, so I need to have another look. I could be doing something stupid though...

Oh and I fixed a few of the comments in the CI tests, since they looked out of date / copied + pasted.

CrossR:
The tests seemed to periodically work for me (CI ones, the last  Unit one has failed but I haven't looked at it yet). I'll check it out tomorrow and see if its obvious then.

browser/test/Services/AutoClosingPairsTests.ts
```diff
@@ -1,22 +1,97 @@
...
+        let mockEditor: Mocks.MockEditor
+        let mockBuffer: Mocks.MockBuffer
+
+        beforeEach(() => {
+            mockEditor = new Mocks.MockEditor()
+            mockBuffer = new Mocks.MockBuffer("typescript", "test.ts", [""])
+            mockEditor.simulateBufferEnter(mockBuffer)
+        })
+
+        it("inserts both characters of the pair", async () => {
```
bryphe: Great set of test cases! Thanks for going the extra mile and implementing these 💯 Excellent foundation for us as we fix additional bugs or add more functionality here.

browser/test/Services/AutoClosingPairsTests.ts
```diff
@@ -1,22 +1,97 @@
...
+                    mockBuffer.setLinesSync([str])
+                },
+                pair,
+                mockEditor as any,
+                false,
+            )
+
+            const [firstLine] = await mockBuffer.getLines(0, 1)
+
+            assert.strictEqual(firstLine, "()")
```
bryphe: I'm glad we have this core functionality under test now - in addition to the tests exercising the new functionality below 👍

test/ci/Api.Buffer.AddLayer.tsx
```diff
@@ -1,7 +1,5 @@
 /**
- * Test script to verify the scenario where no neovim is installed
- *
- * We should be showing a descriptive error message...
+ * Test script for the Oni Layers API, tests the adding of new files and splits.
```
bryphe: Thanks for cleaning these up!

test/ci/AutoClosingPairsTest.ts
```diff
@@ -46,10 +49,15 @@ export const test = async (oni: Oni.Plugin.Api) => {
     const expectedResult = [
         "const test = {",
         "    window.setTimeout(() => {",
-        "        ",
+        '        let testString = "Oni"',
```
bryphe: Awesome 😄

bryphe:
That's awesome you implemented both the CiTest validation and unit tests, @CrossR ! Really appreciate it. Gives me a lot of confidence in the functionality and that it will work build-over-build.

browser/src/Services/AutoClosingPairs.ts
```diff
@@ -213,6 +223,40 @@ export const getWhiteSpacePrefix = (str: string): string => {
...
+    editor: Oni.Editor,
+    openCharacterSameAsClosed: boolean,
+): Promise<void> => {
+    // TODO: PERFORMANCE: Look at how to collapse this instead of needed multiple asynchronous calls.
+
+    const activeBuffer = editor.activeBuffer as IBuffer
+
+    if (openCharacterSameAsClosed) {
+        const lines = await (activeBuffer as any).getLines(
+            activeBuffer.cursor.line,
```
bryphe: For the tests, it looks like `MockBuffer` doesn't have a `cursor` property yet - we may need to add it to get the tests working here / exercise this completely.

CrossR:
I've added a cursor property now, and removed the unneeded `_cursorPosition` var, as well as changing `getCursorPosition()` and `setCursorPosition()` to work with the new property and work the same as the non-mocked version. Had to update the SnippetTests to work with that change, but the tests pass, so seems fine!

Still need to check the CI tests, but a lot of the CI tests have been flaky in my dev env. It might be time for me to delete my current dev VM and start a new one, just to make sure I've got the right dependencies and such, since it only takes a few mins to re-setup.

bryphe:
> I've added a cursor property now, and removed the unneeded _cursorPosition var, as well as changing getCursorPosition() and setCursorPosition() to work with the new property and work the same as the non-mocked version. Had to update the SnippetTests to work with that change, but the tests pass, so seems fine!

Thanks for making these changes!

> Still need to check the CI tests, but a lot of the CI tests have been flaky in my dev env. It might be time for me to delete my current dev VM and start a new one, just to make sure I've got the right dependencies and such, since it only takes a few mins to re-setup.

Worth a shot, just to see. There are some tests that fail intermittently on the build environment too, but it's looking a bit more stable.

It seems like we're still hitting that `tslint` issue - there's a change I've been wanting to do to not require those extra `.js` files, and line the configuration for the tests - so I'm hoping that will avoid this `tslint`/`prettier` conflict.

CrossR:
Hm....looks like it passed some and failed other which is more confusing...

browser/src/Services/AutoClosingPairs.ts
```diff
@@ -31,15 +32,14 @@ export const activate = (
...
+    ) => () => {
         const neovim: NeovimInstance = editor.neovim as any
         neovim.blockInput(async (inputFunc: any) => {
-            // TODO: PERFORMANCE: Look at how to collapse this instead of needed multiple asynchronous calls.
-            await inputFunc(pair.open + pair.close)
-
-            const pos = await neovim.callFunction("getpos", ["."])
-            const [, oneBasedLine, oneBasedColumn] = pos
-            await editor.activeBuffer.setCursorPosition(oneBasedLine - 1, oneBasedColumn - 2)
+            checkOpenCharacter(inputFunc, pair, editor, openCharacterSameAsClosed)
```
bryphe: Looking at the logs from the automation, I do see one bug here - we should either be `return`ing this or `await`'ing it, so that we're actually blocking.

What's happening here is that we're not really blocking the input - so we're allowing the editor to accept further input, which might cause unpredictable races.

In the log I saw from automation, with the resulting string of `window.setTimeout()()`, this could actually explain why we're seeing it. When the first `(` is pressed, we start the logic for `checkOpenCharacter`, but we don't wait for it to complete. In the meantime, a second `(` comes in, and if in the first `checkOpenCharacter`, we haven't placed the cursor, we're going to get unpredictable results. So it seems like there is an actual race condition here that may be impacting the CiTests.

CrossR:
Thanks! Completely missed that race condition.

The issue now seems to be that for some reason the characters `"Oni"` aren't being typed in some of the tests. I'll see if I'm missing something obvious again.

`"expecd":["const test = {"," window.setTimeout(() => {"," let testString = \"Oni\""," })","}"]`
`"actual":["const test = {"," window.setTimeout(() => {"," let testString = "," })","}"],`

bryphe:
Ah ya, I just looked at the logs too. It's strange that, in some of the platforms, it seems to have passed (it looks like 'x86' passed, but 'x64' failed).

CrossR:
I think the errors were more race conditions since I hadn't changed the waitfor, so it was terminating early in some cases when 5 lines were hit.

There is a single fail, but I don't think its me...Are you able to restart that test?

bryphe:
> I think the errors were more race conditions since I hadn't changed the waitfor, so it was terminating early in some cases when 5 lines were hit.

Ah ok, got it! Yes, the error doesn't look like yours - looks like a transient failure with yarn instead. Just restarted the build 🤞

bryphe:
Looks like this is good to go - green builds now. Bringing this in now! Thanks @CrossR for all your work troubleshooting the tests.

PR #1380, opened 2018-01-27
bugfix: Ocaml-lang-server path

Akin909:
@bryphe I've changed the path to point towards `node_modules/ocaml-language-server/bin/server/index.js` as by looking at the `.app` directory in the packaged app this was the only initialisation point I found, no executable. This works on my mac when built although I have not been able to test it works on `windows` (as I dont have a windows machine) but as there is no executable I use the same entry point for both windows and mac

### DIFF
--- a/browser/src/Services/Configuration/ReasonConfiguration.ts
+++ b/browser/src/Services/Configuration/ReasonConfiguration.ts
@@ -8,9 +8,14 @@ import * as path from "path"
 
 import * as Platform from "./../../Platform"
 
-export const ocamlLanguageServerPath = Platform.isWindows()
-    ? path.join(__dirname, "node_modules", ".bin", "ocaml-language-server.cmd")
-    : path.join(__dirname, "node_modules", ".bin", "ocaml-language-server")
+export const ocamlLanguageServerPath = path.join(
+    __dirname,
+    "node_modules",
+    "ocaml-language-server",
+    "bin",
+    "server",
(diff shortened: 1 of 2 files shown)


jordwalke:
@Akin909 Thanks for paying attention to this workflow.

Akin909:
@jordwalke, no worries it's been a blocker for me too for a little while 👍  hopefully this resolves that. @bryphe I've also add a small style change to the tool tip as I noticed certain popups in reason files are quite large so I've added a max width to the content (an example is hovering over `ReasonReact` outputs a massive chunk of text) although it includes some strange characters which I think are coming from the `lsp` rather than our parsing of the markdown string as I've observed this prior to the tooltip changes

Akin909:
Here's a screen shot of the characters I mentioned above:

screen shot 2018-01-27 at 12 00 52

bryphe:
Awesome, thanks for fixing this @Akin909 !

The characters in the quick info do look strange... Not sure where the bug is, but it seems like the `ocaml-language-server` is basically passing the call through to `merlin`: https://github.com/freebroccolo/ocaml-language-server/blob/8f69e63bd26d26505f77498c6dc2dea037527c50/src/bin/server/command/getDocumentation.ts#L12 - so it might be that merlin is sending us those.

Bringing this in now. Looking forward to trying to run our Reason test again - this might fix it 👍

PR #1382, opened 2018-01-28
Bugfix/check path exists workspace

### LINKED ISSUE #1381: Oni fails to start if persisted folder no longer exists
When oni initialises it now looks for a persisted folder and executes a `changeDirectory` however this command does not check to see if the directory it is passed exists, as such in my case where I opened a folder in oni, later moved the folder outside of oni and closed oni, my next attempt to open Oni results in a blank screen which hangs indefinitely.

Running `oni --debug` the console shows an error as the folder that oni is trying to navigate to does not exist which halts the whole startup process, in order to resolve this as the persisted setting holds every time I opened oni manually called `Oni.workspace.changeDirectory('valid/path')` and restarted.

A solution to this is that the `changeDir` function should always check that the directory it is passed is valid before attempting to switch to it. A more general issue is that I think every function or rather main startup process should have a `try/catch` or some form of error handling so that oni doesn't end up hanging without any feedback given back to the user

Akin909:
This PR hopefully fixes #1381, adds a check to the `changeDirectory` function in the workspace service, to check if the directory is valid, if not it changes the directory to null and passes null to the event it does not however call the `process.cwd` function.

I was unclear on what the desired functionality should be if an invalid directory was passed my thinking was to not dispatch an event or change the active workspace but rather ignore the call entirely i.e. return early except I notice that to close the folder you explicitly pass `null` to the function so not sure if its best to emulate that (which is what I've currently done).

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -397,8 +397,8 @@ export class NeovimEditor extends Editor implements IEditor {
             this._actions.setNeovimError(true)
         })
 
-        this._neovimInstance.onDirectoryChanged.subscribe(newDirectory => {
-            this._workspace.changeDirectory(newDirectory)
+        this._neovimInstance.onDirectoryChanged.subscribe(async newDirectory => {
+            await this._workspace.changeDirectory(newDirectory)
         })
 
         this._neovimInstance.on("action", (action: any) => {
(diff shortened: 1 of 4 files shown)


bryphe:
Change looks good to me, @Akin909 ! Thanks again for finding / fixing this. Bringing it in now.

bryphe:
> return early except I notice that to close the folder you explicitly pass null to the function so not sure if its best to emulate that (which is what I've currently done).

This behavior seems reasonable to me. I'm working on some "empty experience" UX (like, what to show in the explorer sidebar if there is no active folder) - having it explicitly set to null when there is no workspace available will help with those features.

PR #1400, opened 2018-01-30
feature(tiny): add margin right to explorer icons

Akin909:
With the explorer now open as default my eye was immediately caught by the icons sitting very close/right beside the node name which might be the intention but if not I've added a `margin-right: 7px` to allow a little more space between these, added this locally which hasn't broken the aesthetic I think.


bryphe:
Cool, just tried it out and it looks great - thanks for thinking about the aesthetics! 👍 Bringing it in now.

PR #1402, opened 2018-01-30
Feature/navigate to project root

### LINKED ISSUE #1385: Oni Directory Navigate to Project Root Option
When working on a project, I usually have this under some form of version control (90+ % of the time) so I use [`vim-rooter`](https://github.com/airblade/vim-rooter), which is a `viml` plugin that automatically sets vim's directory to that of the *current buffer's __project__ root* which it identifies by looking for a `.git, node_modules` and various other config files (the preceding were just examples).

I think as a configurable option It would be very handy to have `oni` be able to replicate this functionality as for example any project wide command a user might run is likely intended for the project or if you do a search in project the starting dir I think as a good default should be the project root *for the file being edited*. Also I think unless specifically sought by the user, *opinion alert*  whilst working on a buffer Its most useful to have oni pointing towards that buffer's project root, which could be very handy with the file explorer as the explorer's contents could reflect the current files project root from buffer to buffer

Akin909:
This PR will fix #1385 by adding a `workspace.autoDetectProjectRoot` setting with three options as recommended by @bryphe `never`, `noworkspace` and `always`, it mirrors the functionality of `vim-rooter` but for oni's workspaces.

The `noworkspace` option (default) will navigate to the current buffer's project root is no workspace has been set.

`always` will change the workspace to match the currently open buffers project root.

The never options leaves this as they are with no dir changes.

It also a a `workspace.autoDetectRootFiles` an array of files to be used to mark a project root which the user can configure.


bryphe:
Wow, that was fast! 👍 Very nice 😄

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -714,6 +714,7 @@ export class NeovimEditor extends Editor implements IEditor {
             )
         }
 
+        this._workspace.autoDetectWorkspace()
```
bryphe: Would it make sense for us to check this on buffer enter? I'm thinking of the case where I open Oni, with no workspace open, and then open a file - do you think we should try and `auto-detect` the workspace at that time?

browser/src/Services/Configuration/DefaultConfiguration.ts
```diff
@@ -282,6 +282,16 @@ const BaseConfiguration: IConfigurationValues = {
     "ui.fontSmoothing": "auto",
 
     "workspace.defaultWorkspace": null,
+    "workspace.autoDetectWorkspace": "noworkspace",
```
bryphe: Awesome, looks like a great set of defaults!

Akin909:
Updated as requested @bryphe

bryphe:
Awesome! Works great. Thanks @Akin909 - this will make it much easier to get started with the explorer / search features.

PR #1415, opened 2018-02-01
Add tests for Markdown preview plugin

TalAmuyal:
This PR is not complete, it requires a change in Oni's core and the API so it is also a proposal and a request for comments.

Specifically, I added `IPlugin ` & `IPluginManager` to the API for the tests (`IOniWithPluginApi`s content should be an addition to `Oni.Plugin.Api` and not an interface of its own).


test/ci/MarkdownPreviewTest.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const typelessOni = typedOni as any
+    const oni = typelessOni as IOniWithPluginApi
+
+    const markdownPlugin = await oni.plugins.getPlugin("oni-plugin-markdown-preview")
+    assert.notEqual(markdownPlugin, null)
+    if (markdownPlugin === null) {
+        return
+    }
+
+    describe("preview pane", async () => {
```
bryphe: Interesting - did you need to do anything special to get `describe` available here? I think this depends on `mocha` and I'm not sure if we have it in the context of these tests, which are really running inside oni

test/ci/MarkdownPreviewTest.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+interface IOniWithPluginApi {
+    plugins: IPluginManager
+}
+
+export async function test(typedOni: Oni.Plugin.Api) {
+    const typelessOni = typedOni as any
+    const oni = typelessOni as IOniWithPluginApi
+
+    const markdownPlugin = await oni.plugins.getPlugin("oni-plugin-markdown-preview")
```
bryphe: Is this the API you were asking about? If so, I'm totally on board. 

This is something I've wanted for a while - the long-term vision for the plugins is to have a way for them to export functions that can be used by other plugins, or in a user's configuration, etc. Having a `plugins.getPlugin("<plugin-id>")` seems like a reasonable way to accomplish this. I'd even like to move some of Oni's core functionality to plugins like this - for example, it'd be cool to move our configuration story over to a module/plugin, and have `Oni.configuration` just be a convenience method for `Oni.plugins.getPlugin("oni-configuration")`. Not something I'd expect to be implemented as part of this - just wanted to share some related thoughts 👍

bryphe:
Thanks for the thinking around this, @TalAmuyal ! I mentioned this in a review comment but the API changes you proposed are definitely in-line with what I'm thinking for the direction of plugins/extensibility. Would be great to have this.

Today, the only export that matters for plugins is the `activate` method, but I could see this extending it - we could set it up such that any function exported could be called via `plugins.getPlugin("my-plugin").myExportedFunction(...)`. Would be helpful for tests like these, and our extensibility scenario in general.

test/ci/MarkdownPreviewTest.tsx
TalAmuyal: As you can see in the PR's diff, there is nothing special - I just assumed it will work.
On the other hand, I didn't really tested it since there in no implementation for `getPlugin` & `getStatus` in Oni yet.

TalAmuyal:
@bryphe, what do you think of making the plugin's `activate` to return the object's instance (or null or some null object for non-implementing plugins)?
That will also support a future `deactivate` function or multiple calls to `activate`..

If that doesn't seem right to you, what do you prefer? (will try to implement it)

TalAmuyal:
In the mean time, I implemented my suggestion so I could test the Markdown-preview tests.

The tests fails, but with no indication of the reason: https://travis-ci.org/onivim/oni/jobs/336904109#L4093

So the real important issues are:
1. The test fails with no indication for the reason
2. I can't (or just don't know how to) run the CI tests locally
3. It takes forever for the tests to run

What can we do with these issues? (for me, 1 and 3 are blockers and 2 is a real pain)

CrossR:
>I can't (or just don't know how to) run the CI tests locally

For me at least, I run the CI tests locally using `npm run test` to do both Unit and CI tests, or `npm run test:integration` for just the CI tests. When I'm running these on Windows, they seem dependant on having some form of built executable sat around which I generate with `npm run dist:win`. You can also comment tests out of `test/CiTests.js` to skip the other tests whilst you are developing to speed it up a bit.

As for debugging, I'm stuck on how best to debug as well with my CI tests. It works fine when using `npm run start` so I just need to work out the differences in the environment that cause the tests to fail.

TalAmuyal:
@CrossR, thanks for the swift and detailed reply!
I'm on Ubuntu, so I don't think `npm run dist:win` will do the trick for me.
Also, I didn't find any other `npm run dist:`.
When I run `npm run test:integration` I get thousands of errors like:
```
src/CSS.ts(8,5): error TS2304: Cannot find name 'require'.

8     require("./UI/components/common.less") // tslint:disable-line no-var-requires
      ~~~~~~~
```
and
```
test/UtilityTests.ts(78,41): error TS7006: Parameter 'val' implicitly has an 'any' type.

78             outputObservable$.subscribe(val => {
```

So I think I'm missing some fundamental step here...

CrossR:
Hmmm... Maybe this is useful?

https://github.com/onivim/oni/blob/master/build/script/travis-build.sh

Looks like that is what Travis uses to build the linux build.

So based on that, it looks like `npm run pack` is needed for Linux builds, it seems to be the same command as `npm run dist:win` just without a `--arch ia32` flag being passed.

bryphe:
Hey guys! Sorry about the issues with the CiTests (and thanks @CrossR for the help troubleshooting!)

I started a cursory section here on the wiki: https://github.com/onivim/oni/wiki/Development#integration-tests

But it would be great to tune that up a bit with better examples / bullet lists of the workflow, etc. (Helpful to add some of the blockers too, like adding `rpmbuild`)

Also, I know there are a definite couple of pain points:
- The CiTests are very sensitive to the configuration. For example, the LargeFilesTest will fail if you have the experimental textmate highlighting enabled (this is actually a real bug, with the textmate highlighting feature - which is good because it means the CiTest is doing its job!). Some CiTests specify their own configuration, so they are insulated against the user's configuration, but if a configuration is not specified for a CiTest, it will use whatever the user's configuration is. Opened #1436 to track fixing this pain point.
- You need a packaged build to run the CiTests. This is annoying, too. #1437 tracks addressing this.
- Debugging isn't straightforward (#1438). The way I usually do this today is open Oni, and then run `Oni.automation.runTest("/path/to/built/test/js")`, but it's not perfect, due to config differences.

Please feel free to update the documentation and leave feedback for these items. Definitely room for improvement, I want the development experience to be smooth - your time is valuable!

TalAmuyal:
It works! (thanks!)
Though the logging is useless:
```
  4) ci tests MarkdownPreviewTest ci test: MarkdownPreviewTest:

      AssertionError [ERR_ASSERTION]: false == true
      + expected - actual

      -false
      +true
```

Shouldn't it include at least the line number?

bryphe:
Oh and I almost missed this from above:

> @bryphe, what do you think of making the plugin's activate to return the object's instance (or null or some null object for non-implementing plugins)?
That will also support a future deactivate function or multiple calls to activate..

> If that doesn't seem right to you, what do you prefer? (will try to implement it)

I think that sounds great! It seems clean & intuitive to me - if the plugin wants to expose functions / objects externally, it can return it from `activate`.

bryphe:
Ah ya, sorry, the logging isn't great... If you output it to a log file, and you scroll up to `---RESULT: <your test name>`, you can see a more detailed error. Would be great to show this at the bottom instead.

badosu:
@bryphe @TalAmuyal A little sidetracking on this issue, but how can I test the markdown preview today?

TalAmuyal:
Add `"experimental.markdownPreview.enabled": true` to your settings to enable the plugin.
The preview-pane opens automatically when you open a Markdown file, but there is no logic to auto-close it yet so you can add a binding to toggle the preview-pane: `oni.input.bind("<f8>", "markdown.togglePreview")`

badosu:
Works great, thanks! Let me know if I can put this in the wiki.

TalAmuyal:
__Update__
I found where my code fails and am working on it.
I will later post additional notes regarding problems I'm facing.

TalAmuyal:
Well, it seems that `describe` and `it` doesn't work.
It's a shame that there is no compilation error...

TalAmuyal:
At last, after a full day of working on it, the tests pass!
@bryphe, please note that I've added 2 modules which greatly helped my with debugging: `Assert` which wraps `assert` and `Debug` which provides an inspection function (I hope it is not redundant).

I'm waiting for your feedback.
Afterwards I will submit a change to the `oni-api` package that includes the `plugins` addition.

extensions/oni-plugin-markdown-preview/src/index.tsx
```diff
@@ -168,11 +171,30 @@ class MarkdownPreview extends React.PureComponent<IMarkdownPreviewProps, IMarkdo
...
 
     constructor(private _oni: Oni.Plugin.Api) {
         this._oni.editors.activeEditor.onBufferEnter.subscribe(args => this.onBufferEnter(args))
     }
 
+    public isPaneOpen(): boolean {
+        return this._open
+    }
+
+    public getUnrenderContent(): string {
```
bryphe: Minor nitpick - should this be `unrenderedContent`?

browser/src/Plugins/Plugin.ts
```diff
@@ -51,8 +61,9 @@ export class Plugin {
 
         try {
             vm.runInNewContext(
-                `debugger; const pluginEntryPoint = require('${moduleEntryPoint}').activate; if (!pluginEntryPoint) { console.warn('No activate method found for: ${moduleEntryPoint}'); } else { pluginEntryPoint(Oni); } `,
+                `debugger; const pluginEntryPoint = require('${moduleEntryPoint}').activate; if (!pluginEntryPoint) { console.warn('No activate method found for: ${moduleEntryPoint}'); } else { pluginContainer._instance = pluginEntryPoint(Oni); } `,
```
bryphe: Cool, I like this a lot! Some interesting possibilities this opens up... like library plugins that provide common functionality, or plugins that supply functions that you can use with `input.bind`, etc. Great way to get access to the plugin to test, too.

test/ci/MarkdownPreviewTest.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+async function awaitEditorMode(oni: Oni.Plugin.Api, mode: string): Promise<void> {
+    function condition(): boolean {
+        return oni.editors.activeEditor.mode === mode
+    }
+    await oni.automation.waitFor(condition)
+}
+
+async function openFile(oni: Oni.Plugin.Api, path: string): Promise<void> {
```
bryphe: There's actually a similiar utility in `test/ci/Common.ts` called `navigateToFile` - we might want to see if we can reconcile that with this?

bryphe:
> At last, after a full day of working on it, the tests pass!

Woohoo! 👍 Thanks for all your hard work on this!

> @bryphe, please note that I've added 2 modules which greatly helped my with debugging: Assert which wraps assert and Debug which provides an inspection function (I hope it is not redundant).

Seems reasonable to me. Is `Debug` used now, or transiently? There is also the `console.dir` utility which I use quite a bit (it's similiar to the `inspect` method there, it prints out the objects properties). If we're not using the `Debug` module currently, I'd lean towards leaving it out. I'm hoping that we could get a way to pop open the debugger + break for #1438 

Just had some very minor feedback, but otherwise looks good to me. Can't wait to enable this. 😄

bryphe:
Also the OSX changes don't look related - seems like it might be an issue with the cached modules. I'll try clearing the cache and restarting the build.

extensions/oni-plugin-markdown-preview/src/index.tsx
TalAmuyal: Oops...

TalAmuyal:
Seems that for MacOS all of the CiTests fails on `waitFor timeout expired for condition: ()=>i.editorManager.activeEditor.neovim&&i.editorManager.activeEditor.neovim.isInitialized"`

Have you already cleared the cache?

BTW, refactored according to the CR.

bryphe:
I cleared the cache and re-ran the tests - seems like they get farther, but I'm seeing this on the OSX tests:

```
[SEVERE] file:///Users/travis/build/onivim/oni/dist/mac/Oni.app/Contents/Resources/app/lib/browser/bundle.js 0:24588 Error: spawn /Users/travis/build/onivim/oni/dist/mac/Oni.app/Contents/Resources/app/node_modules/oni-neovim-binaries/bin/nvim-osx64/bin/nvim ENOENT
    at exports._errnoException (util.js:1024:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:192:19)
    at onErrorNT (internal/child_process.js:374:16)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

It means somehow there isn't a bundled neovim process, which is strange (usually indicates something went wrong with packaging?) There isn't anything in your change that would cause this. Potentially still a cache issue.

Could you please try merging `master` and we'll try again? I'll rebuild in the meantime too 😄

TalAmuyal:
Done.
Travis keeps not liking me, and to top it off AppVeyor randomly chooses if to fail the x86 or x64 build...

bryphe:
Looks like AppVeyor tests are good now! I'm still not sure what's happening on OSX, but I'm running the tests locally and I built the local package - everything seems fine. I'll bring this once I finish running locally.

bryphe:
Tests passed locally for me - thanks for all your hard work getting the test infrastructure around this, @TalAmuyal ! Will be helpful for future features as well 👍

bryphe:
Now that we have test coverage, I think we're good to turn this on by default, when you're ready!

PR #1419, opened 2018-02-01
Bugfix/window number error removing bufferbar highlight

Akin909:
Noted recently that an error with the `windowNumber` subscription which happens consistently on startup which seems to take down the buffer bar selection, so it is no longer clear which buffer is selected. 

screen shot 2018-02-01 at 19 51 10

I couldn't figure why the error might be happening but I added if block to the actions in the handler to prevent the error being thrown

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -305,7 +305,8 @@ export class NeovimEditor extends Editor implements IEditor {
         })
 
         this._windowManager.onWindowStateChanged.subscribe(tabPageState => {
-            const inactiveIds = tabPageState.inactiveWindows.map(w => w.windowNumber)
+            const filteredTabState = tabPageState.inactiveWindows.filter(w => !!w)
+            const inactiveIds = filteredTabState.map(w => w.windowNumber)
 
             this._actions.setActiveVimTabPage(tabPageState.tabId, [
                 tabPageState.activeWindow.windowNumber,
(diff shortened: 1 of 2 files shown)


Akin909:
On further investigation it seems [here](https://github.com/Akin909/oni/blob/e9f9311ee245ad49b2825c5720910e18464cfe15/browser/src/neovim/NeovimWindowManager.ts#L92) `TabState` can be passed in as null, which seems expected to me as If only one window exists there wouldn't be any inactive window (at a glance dont actually know what's going on there), but I found that this error could be specifically triggered by running `bd!`, so in the subscription handler I check that there are no falsy values in the inactive windows array before the subsequent operations

Akin909:
The component did catch is essentially a somewhat unrelated suggestion of a point where some fallback UI can be set to render if the editor crashes, but even unimplemented as it is any errors in the `NeovimSurface` will get caught there and Oni at least will remain open

bryphe:
Cool, thanks for the investigation here, @Akin909 ! The fix looks safe to me, and does seem like it would make it more robust.

> This bug also may also relate to opening the welcome screen although I currently cannot test this as using the screen swallows up input like <space>, <enter> etc. so hard to navigate or run commands once oni is opened with the welcome menu

Ah ya, for sure, this needs to be fixed before I can turn it on by default!

> The component did catch is essentially a somewhat unrelated suggestion of a point where some fallback UI can be set to render if the editor crashes, but even unimplemented as it is any errors in the NeovimSurface will get caught there and Oni at least will remain open

This is a good idea - I was thinking it might make sense to add it to `<WindowSplitHost />` which is the container for all the splits (including editor splits, explorer, sidebar, etc - so it would sort of catch everything).

bryphe:
Looks good to me! I'll bring this in. Really appreciate your help catching these issues, @Akin909 👍

Issue #1408, opened 2018-01-31
open folder doesn't work


rockneverdies55:
Just upgraded to `0.2.21` and it seems like nothing is happening when I use `Open Folder` in the file menu or click to the very left bottom corner link. Open File seems to be working and launching the system file explorer as expected. The os is windows 7.

Akin909:
@rockneverdies55 this was a regression in the most recent release but has since been fixed (#1393) in master and will be in the next release or if you wanted to you could also follow the `README.md` and build `Oni` locally.

CrossR:
Actually, it looks like the fix in #1393 fixed the menu, but I'm less sure if it fixed the status bar + quick open. At least both those cases seem to be using `oni.openFolder` from a quick search in the code base.

May be worth leaving this open to track those two.

Akin909:
@CrossR thanks for pointing that out 🤦‍♂️  just pushed # to fix this checked to `quickOpen` menu command is working think it was just the statusbar that was still down #1421 to fix that

bryphe:
Thanks for your help fixing this, @Akin909 ! Looks like we're in good shape now in master - I'll close this out.

PR #1421, opened 2018-02-01
bugfix/fix open folder in statusbar

Akin909:
The open folder command was renamed to `workspace.openfolder` this PR updates the statusbar file item command so it functions properly. Relates #1408


bryphe:
Awesome, thanks for the help here @Akin909 ! Looks good to me 👍

Issue #1412, opened 2018-02-01
Resize Oni splits


Akin909:
Following on from #1406 this issue relates to creating functionality which handles resizing Oni splits and sidebars.

From @bryphe 
> Yes, I think we'd want to split this up into two tasks:

>API - Add an API on the WindowManager for resizing splits. I'm thinking that the split method could 

> remove some sort of WindowHandle that lets you specify a size, like:
> const splitHandle = windowManager.split("horizontal", newEditor)
> splitHandle.setSize(250)
> (It would also be a natural place to show and hide, which would be great for toggling sidebars / terminals / etc).

> Drag-to-resize UI gesture - once we have the API set, we could hook up a drag handle somewhere. #1405 adds a <div className="split-spacer"> next to each window, so that would be a natural place to add it - we could turn that into something like a ResizeHandle. It's close to the WindowManager so it seems like we could plumb through the API call from above in a fairly straightforward way.

bryphe:
Thanks for splitting this out, @Akin909 ! Excellent - I can take a look at this piece (it's sort of related to the toggling functionality), and then once we have the API, hopefully it's a bit more straightforward to implement a UI gesture like a drag handle.

bryphe:
I think with this, #1414 wouldn't be too bad either - we'd simply add some commands to talk to this API that can control the split / window size.

PR #1424, opened 2018-02-02
Feature/toggle sidebar

Akin909:
Whilst I absolutely love the sidebar I couldn't bear having it open all the time so this PR aims to add toggleability its very simply implemented toggling a boolean in redux and adding an Oni command, wasn't sure if you had a plan for this as part of a larger split handling bit of functionality but I implemented this locally, if it's not quite not what you intended no worries 👍 , as I know there #1363  etc and #1412 and you'd discussed different means of handling splits

### DIFF
--- a/browser/src/UI/Shell/ShellActionCreators.ts
+++ b/browser/src/UI/Shell/ShellActionCreators.ts
@@ -59,6 +59,10 @@ export const setViewport = (width: number, height: number) => ({
     },
 })
 
+export const toggleSplit = () => ({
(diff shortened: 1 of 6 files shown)


TalAmuyal:
So is there a way to bind a toggle command for the sidebar in `master`?

bryphe:
There unfortunately isn't a good way right now - but you can set `sidebar.enabled` to `false` for now, if it's in the way. Toggling is high on the list of priorities!

Issue #1413, opened 2018-02-01
Welcome Menu - Input and loading


Akin909:
I've again bundled two bugs together as they both pertain to to the welcome menu.
The first of which is that the welcome screen now swallows input aimed for vim relates to #1398 which means that on opening vim one of my first actions which is to open the `commandline` and open a file or a session is now blocked.

Secondly the welcome screen ~~only appears if I have oni open at full screen if oni starts at half screen for some inscrutable reason the welcome menu does not appear instead I get an empty buffer titled welcome~~ only appears with my `init.vim` disabled going to try commenting out sections to find the cause, although I think if a user's init.vim might somehow break the welcome menu due to maybe a competing `BufEnter` command etc there might need to be logic to give the menu priority

EDIT: Apologies, I appreciate this issue is ambiguous, what I'm trying to convey is that with the welcome menu enabled my keyboard input seems to disappear in a confusing manner hitting certain keys does nothing 😕, and regarding its appearance it's likely an interaction with another plugin but potential plugin interaction is the concern as has happened in my case

Akin909:
Looked into both of these and it seems the issue with when welcome screen appears relates to my `vimrc` as not loading it seems to allow the screen to appear as expected. I noticed that vim actually seems to be creating a real buffer for the welcome screen.

I dont know much about how the buffer layer works but I was thinking if it makes actual vim buffers it might be good to have them be `scratch` buffer and setting them to `nofile` as I've noticed thats its possible to actually run the save command and create a file in the current workspace called `WELCOME`

As to the specific cause of the error that will take a bit more digging as my config is enormous

bryphe:
Thanks for testing out the welcome screen @Akin909 - really appreciate it and it will help us iron out all the issues before the next release 💯 

> which means that on opening vim one of my first actions which is to open the commandline and open a file or a session is now blocked.

This is definitely a blocker to turning on the feature. The welcome screen even advertises that this should be available 😄 

> I dont know much about how the buffer layer works but I was thinking if it makes actual vim buffers it might be good to have them be scratch buffer and setting them to nofile as I've noticed thats its possible to actually run the save command and create a file in the current workspace called WELCOME

[...]

PR #1425, opened 2018-02-02
Fix #1396 - Navigation Quirks

### LINKED ISSUE #1396: Navigating file explorer Quirks
@bryphe given you're looking to get the file explorer into the next release wanted to give some feed back on issues I've noted (following on from the gitter conversation).

As I mentioned with my own `init.vim` enabled I find that pressing `j` or `k` causes the cursor to occasionally jump over nodes in the tree. I've tried deactivating my config which actually results in an error which crashes oni.

screen shot 2018-01-30 at 08 15 44

Will do some more digging and add a screen cast to illustrate the issue shortly.

bryphe:
This change hopefully addresses #1396 (and the navigation part of #1413 ), by disabling the `init.vim` loading of our `SharedNeovimInstance`. For the UI that depends on it, it's mostly canonical vim navigation (`h`, `j`, `k`, `l`, `gg`, `G`, etc), so hopefully that will be sufficient. Worst case, we could specify an alternate config later to use for those specific pieces of UI.

Instead of the `NeovimProcessSpawner` needing to check the configuration for the options, we've pushed that upwards. This lets the `SharedNeovimInstance` override the `loadInitVim` and `useDefaultConfig` options, and force them to `false`. 

This should minimize conflicts in those pieces of UI, but we may need more work down the road in terms of extending or customizing the mappings for these components.

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -687,6 +687,9 @@ export class NeovimEditor extends Editor implements IEditor {
         const startOptions: INeovimStartOptions = {
             runtimePaths: this._pluginManager.getAllRuntimePaths(),
             transport: this._configuration.getValue("experimental.neovim.transport"),
+            neovimPath: this._configuration.getValue("debug.neovimPath"),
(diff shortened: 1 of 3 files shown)


bryphe:
Fixes #1396 , by making the explorer navigation not dependent on a user's `init.vim`.

PR #1429, opened 2018-02-02
Update default config.js for Nord colourscheme. 

CrossR:
Realised with a new Oni install, `onedark` is still in the `config.js` that is used as the template for new config files, potentially causing new users to view the config file and change their colour settings without realising.

Only a tiny fix, but at least streamlines it a tiny bit more.


bryphe:
Ah thanks for catching this, @CrossR 👍

PR #1430, opened 2018-02-02
Sidebar: Initial vim marks integration

bryphe:
The goal of this functionality is to leverage vim marks as a bookmarks provider, which would allow:
- Showing local and global marks in the bookmarks pane
- Updating the bookmarks pane when a mark is set
- Easy deletion of marks from the bookmarks pane


TalAmuyal:
When I saw the "Bookmarks pane" feature I told myself "meh, I've got Vim marks for that", but integrating Vim marks into the bookmarks - now that's the way to do it!

**Disclaimer:** All that is said above is meant in good spirits and no harm was intended.

bryphe:
> When I saw the "Bookmarks pane" feature I told myself "meh, I've got Vim marks for that", but integrating Vim marks into the bookmarks - now that's the way to do it!

Haha, ya, I was considering adding some separate bookmarks functionality to Oni... but then I was like, wait a minute, it's already there...

I don't use Vim marks much because I forget about them (or forget which marks I have set), so I'm hoping this will help me incorporate them into my workflow.

PR #1435, opened 2018-02-03
Feature/get, unmap and restore user bindings

Akin909:
@bryphe whilst looking into #1396 I mentioned I wrote some functionality to get a user's mappings, unmap them and restore them. Whilst not tied into any specific issue it occurred to me that future use cases for this functionality could be 
* showing a user their vim bindings in oni,
* giving oni the ability to potentially add some ux to change user's vim binding in oni.
* the original inspiration for this functionality temporarily removing user's bindings in oni's buffer layer

screen shot 2018-02-03 at 10 31 12

A above is a screenshot of the output of the command, which seems like a potential gold mine to me hence the PR, though if there isn't as much utility in this as I'm imagining no worries 👍 , just thought I'd run it past you rather than have it locally and it never see the light of day


Akin909:
Cool 👍  I'll tweak this to move it into the instance although so far to add a command for `neovimEditor` we've put methods in there (`NeovimEditor`) so I'm not sure how you'd be able to call this function from the editor aka `Oni.activeEditor.getUserMapping()`?

bryphe:
> Cool 👍 I'll tweak this to move it into the instance

Sweet, thanks!

> although so far to add a command for neovimEditor we've put methods in there (NeovimEditor) so I'm not sure how you'd be able to call this function from the editor aka Oni.activeEditor.getUserMapping()?

At least for me, it's not clear at the moment how we'd actually use this functionality or what the API surface area should look like (it'd be guesses). It's hard to know the right abstraction depending on the sort of functionality we decide to build on top of it - it might be that, if we decide we need to save/restore key bindings for a scenario, we actually don't need an `Oni.editors.activeEditor.getUserMapping()` exposed publicly, but just a `saveBinding`/`restoreBinding`, or something. In other words, someone might not need to directly call `getUserMapping()`, but API methods on `Oni.activeEditor` might call into it on the `_neovimInstance` it has. Hard to predict since we don't know the user scenario!

There also may be some potential for confusion with `InputManager` - how does `bind`/`unbind` relate to `getUserMapping`, etc? Should we start integrating the `InputManager` with this functionality, so that it is aware of Neovim's bindings? It's sort of the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle. 

Since we don't have a scenario yet, it's better to keep the `Editor` interface as-is, and leave it as functionality on `NeovimInstance` (which is a safe place for any lower level interactions with Neovim). Once we have a use case, we can start to figure out the right level of abstraction for it in terms of Oni's interfaces.

Also, we do expose the `neovim` instance on the editor today (although this is really a crutch), via `Oni.editors.activeEditor.neovim` - so from configuration/plugins, they could still access this method.

Akin909:
@bryphe happy to move this over to the `NeovimInstance` if youre still amenable to having this functionality somewhere for a future use case

bryphe:
>@bryphe happy to move this over to the NeovimInstance if youre still amenable to having this functionality somewhere for a future use case

Thanks for checking in! 👍 

I'm thinking about this more, and I'd prefer to not bring in the functionality at the moment - given that there isn't a use case for it.

It will be a good point of reference later, though, if we do need to consult the user mappings!

PR #1443, opened 2018-02-06
Feature/syntax highlight hover

Akin909:
This PR adds syntax highlighting to the hover tool tip other changes include:
*	Adding deep merges to the config so it supports sub object fields
* Adding an adhoc highlighter to generate tokens for given lines as well as fn to generate vim highlights for a line
* Added a function to get vim highlights actual color values
* Added functionality to pull token colors from the `json` theme, user config and set defaults based on available vim highlights.
* I've removed splitting up the hover info on every new line and now only render the docs and the titles differently.
### Todo

- [x] Set highlight colors for various tokens or so tokens such as `variable.other.object` render with a color

- [x] Fix failing test due to config changes

- [x] Fix Regex quirks re .`|` characters

- [x] Add unit tests

- [x] resize hover container

Screenshots

screen shot 2018-02-07 at 22 05 33
screen shot 2018-02-04 at 16 06 44

*Discussed below:*
Fix formatting for code block and add highlighting to code blocks in tooltip


browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -727,6 +741,31 @@ export class NeovimEditor extends Editor implements IEditor {
...
     }
 
+    public async getVimHighlights(): Promise<IVimHighlight[]> {
+        const configurationColors: Array<{
+            settings: string
+            scope: string
+        }> = this._configuration.getValue("editor.tokenColors")
+
+        // FIXME: Specify Highlights to check by differentiating vim highlights as no having "."
+        const tokens = Object.keys(configurationColors).filter(c => !c.includes("."))
```
bryphe: Could we move this logic to `NeovimInstance` as `getHighlights`? This way, we don't need a public `getVimHighlights` method, as I'm hoping to keep the `Editor` interface as agnostic of Vim as possible.

Once we have that, we could change below to:
```
const vimHighlights = await this._neovimInstance.getHighlights()
```

browser/src/Services/Themes/ThemeManager.ts
```diff
@@ -319,6 +484,19 @@ export class ThemeManager {
         return this._themeLoader.getAllThemes()
     }
 
+    public async getVimHighlightColors(tokenColors: IVimHighlight[]) {
```
bryphe: It seems like we're actually setting the token colors here, so we might want to rename this to `setTokenColors` instead.

browser/src/Services/Themes/ThemeManager.ts
```diff
@@ -269,11 +368,78 @@ export const DefaultThemeColors: IThemeColors = {
...
+        function: {
+            color: null,
+        },
+        constant: {
+            color: null,
+        },
+    },
+}
+
+export interface IVimHighlight {
```
bryphe: Ideally, we could make the ThemeManager agnostic of Vim. The `highlight group` is really an implementation detail - it'd be great if we could make this care about color/bold/italic/background-color, and then leave it up to the `setHighlights` in `Buffer` to rationalize highlight groups from that. It sounded like there might be challenges here though - let me know what sort of hurdles you ran into.

browser/src/UI/components/QuickInfo.tsx
```diff
@@ -3,6 +3,79 @@ import * as os from "os"
 import * as React from "react"
 import styled, { boxShadowInset, css, fontSizeSmall, withProps } from "./common"
 
+const markedCss = css`
```
bryphe: The changes for the rendering look great to me! Thanks for all your work here. I wonder if we could split this up into two changes? It seems like the synchronizing highlight groups has some challenges, but we could at least bring in the rendering with the `editor.tokenColors` we have today

browser/src/UI/components/QuickInfoContainer.tsx
```diff
@@ -0,0 +1,166 @@
...
+    description: {
+        __html: string
+    }
+}
+
+const quickInfoTokens = {
+    "editor.tokenColors": {
+        "meta.import": {
+            scope: "meta-import",
+            settings: "Define",
```
bryphe: Ah ya, this is where I was hoping we could specify color/background color/bold/italic etc (and in the user's config / theme as well) instead of the vim highlight group. The vim highlight group was really just a crutch to implement.

extensions/theme-onedark/colors/onedark.json
```diff
@@ -45,6 +45,11 @@
         "highlight.mode.operator.foreground": "#282c34",
 
         "highlight.mode.visual.background": "#56b6c2",
-        "highlight.mode.visual.foreground": "#282c34"
+        "highlight.mode.visual.foreground": "#282c34",
+        "editor.tokenColors": {
+            "variable.parameter": {
+                "color": "#d19a66"
```
bryphe: Nice! 😄

browser/src/Editor/BufferManager.ts
```diff
@@ -295,6 +295,9 @@ export class Buffer implements IBuffer {
         }
 
         const getToken = (lineContents: string, character: number): Oni.IToken => {
+            if (!lineContents || !character) {
```
bryphe: Interesting, I wonder if this is related to an issue in my PR #1298... perhaps I'm hitting this too!

browser/src/Editor/NeovimEditor/markdown.ts
```diff
@@ -1,11 +1,135 @@
...
+const renderWithClasses = ({
+    tokens,
+    colors,
+    text,
+    element = "span",
+    container = "paragraph",
+}: IRendererArgs) => {
+    // This is critical because marked's renderer refuses to leave html untouched so it converts
+    // special chars to html entities which are rendered correctly in react
+    const unescapedText = unescape(text)
```
bryphe: Thanks for all your work here to get this set up! 👍

browser/src/Services/Configuration/Configuration.ts
```diff
@@ -172,16 +175,16 @@ export class Configuration implements Oni.Configuration {
...
+        let currentConfig = merge(
+            this._defaultConfiguration,
+            this._persistedConfiguration.getPersistedValues(),
+            this._setValues,
+        )
 
         this._configurationProviders.forEach(configProvider => {
             const configurationValues = configProvider.getValues()
-            currentConfig = { ...currentConfig, ...configurationValues }
+            currentConfig = merge(currentConfig, configurationValues)
```
bryphe: Cool, this might be helpful for other config settings as well.

Akin909:
@bryphe 👍 thanks for the feedback its good to get some input re the implementation.

I've made changes as per your request so
* removed the `vimhighlight` fn from the `editor` to the instance.
* renamed the vimhighlight fn in the `themeManager`
* specified a list of highlights to use.
Re. being able to read in themes like `vscode's` I've made the scopes an array and am working on converting theme scopes specified by an array internally to a dictionary.

Summary of other changes
* On load the theme config is populated with vim highlight colors based on a fallback vim id specified in the config *this part is essential for highlighting if not specified in theme or user config*
* Merge user, theme and vim colors respectively based on priority
* created a separate *enhanced* theme for the hover tool tip which uses a wider array of tokens than the syntax highlighting

browser/src/neovim/NeovimInstance.ts
```diff
@@ -534,6 +543,45 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
         return path.normalize(currentWorkingDirectory)
     }
 
+    public async getVimHighlights(): Promise<ITokenColorsSetting[]> {
```
bryphe: Thanks for moving this!

browser/src/neovim/NeovimInstance.ts
```diff
@@ -534,6 +543,45 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
...
+                    token,
+                    true,
+                ])) as IHighlight
+                const highlightOrDefault = (color: number) => (color ? Color(color).hex() : null)
+                const { bold, foreground, background, italic } = highlight
+
+                return {
+                    scope: [token],
+                    settings: {
+                        fallback: token,
```
bryphe: The fact that the system end-to-end still needs to know about the Vim highlight group means we haven't fully decoupled it yet. It'd be great if we could break our dependency on this, because it adds complexity along the entire stack. If we can just speak in `tokenColors` without having to worry about resolving a fallback, that would make life easier everywhere we use this.

One idea we could try is to _invert_ that responsibility. Right now, we create a bunch of `tokenColors` and specify a `fallback`. What if, instead, we create a set of default `editor.tokenColors` from this method?

For example, in `QuickInfoContainer`, we have this list of scopes:
```
...
       "variable.language": {
            scope: [],
            settings: {
                fallback: "Identifier",
            },
        },
        "variable.parameter": {
            scope: [],
            settings: {
                fallback: "Identifier",
            },
        },
```

Could we instead, here, specify a mapping like:
```
const VimHighlightGroupsToScope = {
"Identifier": ["variable.language", "variable.parameter"]
}

And then, from this method, return essentially that full set?
```

We're basically inverting the dependency, and letting `NeovimInstance` do the job of converting completely to the `ITokenColor` world. Then, if this is the default set, they can naturally be overridden by the theme and user configuration. 

This gives us a few benefits:
- Consumers don't need to worry about resolving `fallback` or what `fallback` means
- We don't have to reimplement the logic in `QuickInfoContainer` in different places

I would prefer to not have the vim highlight group as part of the `ITokenColor` model (even as `fallback`), because of the coupling it creates for the consumer - the consumer has to know about vim highlight groups to work. It seems like we can push that logic into `getVimHighlights` by inverting how we create our token colors - letting that function give us a full set, instead of the consumer needing to reverse map.

browser/src/Services/Themes/ThemeManager.ts
```diff
@@ -141,11 +153,78 @@ export const getHoverColors = (
...
+    defaultTokens,
+    themeColors,
+    vimColors,
+}: {
+    config: Configuration
+    defaultTokens: IEditorTokens
+    themeColors: IEditorTokens
+    vimColors: { [token: string]: ITokenColorsSetting }
+}) => {
+    const userTokens = config.getValue("editor.tokenColors")
```
bryphe: One thing that is a little bit confusing here is that the `ThemeManager` is operating at a different level of abstraction for `tokenColors` vs `colors`.

For `colors`, the `ThemeManager` is mainly responsible for loading the colors and getting the defaults. The `Colors` calls in `Services/Colors` brings together the theme colors with the user's configuration, and handles the merging (it's basically a layer on top, that coalesces it with the configuration). It's main responsibility is listening for both the `ThemeManager.onThemeChanged` and `configuration.onConfigurationChanged` to pick up color changes in either place.

For `tokenColors`, the `ThemeManager` is doing the merging of the user configuration, which is more responsibility than it has with `colors`. I could see us later wanting to have some similiar logic as for `colors` - where we listen to configuration changes and update them. It might be worth decoupling that / abstracting that like we have for `colors`, at some point.

bryphe:
Thanks for all the work on this @Akin909 !

> On load the theme config is populated with vim highlight colors based on a fallback vim id specified in the config this part is essential for highlighting if not specified in theme or user config

I had a comment on this in the file too - I'm wondering, is it not possible to have a default map like this:
```
const VimHighlightGroupToDefaultScopes = {
"Identifier": ["variable.language", "variable.parameter", ....],
...
}
```

And then we use that to map the vim highlight groups to a set of token colors that we return from `getVimHighlights` w/o a `fallback` property? That means that, the only part of the code that has to worry about vim highlights is the `getVimHighlights` call - the whole rest of the pipeline can be agnostic of it and just deal with `ITokenColors`.

> created a separate enhanced theme for the hover tool tip which uses a wider array of tokens than the syntax highlighting

Just curious - why does there need to be a different theme? Can we use the wider array of tokens for syntax highlighting, too, and get more fidelity there?

Akin909:
@bryphe re.

> created a separate enhanced theme for the hover tool tip which uses a wider array of tokens than the syntax highlighting

> Just curious - why does there need to be a different theme? Can we use the wider array of tokens for syntax highlighting, too, and get more fidelity there?

I found that having *all* highlighting for tokens centralised meant that any change or restructure that I wanted to make to highlight the tooltip necessarily impacted the `syntaxHighlighting` and I soon discovered that what I wanted re highlighting the hover didn't always correspond to syntax highlighting and meant compromising on both.  For example if based on the original theme I wanted to restructure colors or the layout of the theme It would have the side-effect of breaking syntax highlighting in some unexpected way. 

Also I found trying to co-ordinate syntax highlighting the hover with sensible colors and having to also deal with the way they impacted the syntax highlighting overwhelming as a choice to color a specific type one way in the hover could end up cascading into a bunch of often unexpected/undesired color changes all over the place in the syntax highlighting for vim

Also I *very* much wanted to keep this PR and in general any highlighting of ui based on tokens flexible so that that in other components like the hover or later maybe `codelens` windows wouldn't be dependent on the structure of the central theme. I plan/planned on generalising the quick info container's functionality into a `Render Prop` to be used in other components so they can all inherit the theme but restructure it or use it as needed without influencing the central theme object.

Re. consolidating/ removing the fallbacks i've had some success implementing this although in this case it creates the following issues:
* invalidates the above plan or rather is a spanner in the works as without access to what highlights are available in vim I'd not be able to create a component that has fallbacks as the original theme is the only one that knew anything about vim's highlights as they aren't saved anywhere, so in order to color anything else it would have to come through the theme.

This also means that if we wanted to say specify a different initial color for the hover tool tip say highlight this kind of operator as `red` since it occurs rarely in the hover or makes it look better then we would also have to have the fallback be the same color in the editor (not sure i've elaborated this concern well)

* A bigger issue is that with a centralised theme with no reference to vim , so far oni has been dependent on vim colors in the config to syntax highlight, trying to color the hover ui is now linked to `syntaxHighlighting` which increases complexity a bit now and tbh I was hoping not have to tweak the already present colors as figuring out how to color the hover box has been proving challenging on its own

Apologies for the enormous wall of text this PR has proven far more time/labour intensive than I intended/hoped 🙍‍♂️

bryphe:
> An issue is that with a centralised theme with no reference to vim , so far oni has been dependent on vim colors in the config to syntax highlight, trying to color the hover ui is now linked to syntaxHighlighting which increases complexity a bit now and tbh I was hoping not have to tweak the already present colors as figuring out how to color the hover box has been proving challenging on its own

Ah ya, this is a deficiency in the existing `syntaxHighlighting` implementation. Ideally, it shouldn't reference the highlight groups by name either, for the same reasons that `fallback`. I did it as a shortcut to get the highlighting working, but I think t hat we should fix it before building on top of it, before the coupling becomes intractable. Once `syntaxHighlighting` no longer needs the highlight group, I think we'll be able to get rid of `fallback` on the whole stack.

I can see breaking it up in the following way.
1) Add the `getHighlights` method and make the ThemeManager of the initial set of highlights, come from Vim
2) Consolidate the `tokenColors` from configuration + `themeManager` and make a unified place (like the `Colors` class today).
3) Update the syntax highlighting to use the `ITokenColor` and not the vim highlight group. This means tweaking the `Buffer` implementation so that it creates ad-hoc highlight groups. This means that we break our dependency on the vim highlight group and no longer need a `fallback`.
4) Add in the hover implementation + the enhanced tokens specific to it. There's a lot of work you did here in terms of parsing, hooking up the textmate highlighting, etc!

> Also I found trying to co-ordinate syntax highlighting the hover with sensible colors and having to also deal with the way they impacted the syntax highlighting overwhelming as a choice to color a specific type one way in the hover could end up cascading into a bunch of often unexpected/undesired color changes all over the place in the syntax highlighting for vim

Makes sense, I can see where you may want to override them specifically for hover.

I'll put together some PRs for 1-3 to break our dependency on the highlight groups for syntax highlighting (building on the work you did!)... And then, hopefully we can apply this change, just without needing `fallback` anymore!

Akin909:
@bryphe just a heads up I've just pushed some changes almost exactly as you posted. I've removed the `fallback` and have added the functionality to generate tokens which are then feed into the theming. The dependency on vim highlights still exists and as work around I use the map of highlight to color to resolve that in `SyntaxhighlightReconcilier` as a work around

browser/src/Services/SyntaxHighlighting/Definitions.ts
```diff
@@ -1,11 +1,11 @@
...
 
 export interface IHighlight {
-    foregroundColor: string
-    backgroundColor: string
-
+    foreground: string
+    background: string
     bold: boolean
     italic: boolean
+    fallback?: string
```
bryphe: With the latest change, is this still needed?

browser/src/Services/Configuration/DefaultConfiguration.ts
```diff
@@ -115,40 +115,126 @@ const BaseConfiguration: IConfigurationValues = {
...
+                background: null,
+                bold: null,
+                italic: null,
+            },
         },
-        {
-            scope: "variable.parameter",
-            settings: "Identifier",
+        "meta.object": {
+            scope: ["meta.object.type"],
```
bryphe: Hmm, I don't understand how the `meta.object` relates to the scope here.  The previous scheme seemed simpler to me - and consistent with how the `tokenColors` in a theme is defined.

browser/src/Services/SyntaxHighlighting/Definitions.ts
Akin909: ah thats definitely not needed its was a point I missed, no good find and delete in project yet maybe another editor feature 😆

browser/src/Services/Configuration/DefaultConfiguration.ts
Akin909: haha as is always the way I skim read the docs but went through them a bit more closely now and not haphazardly applying styles to scopes against what the documentation recommends actually really improves highlighting 😆

bryphe:
FYI - I implemented 1 & 2 in #1484, and 3 in #1485 . I hope you don't mind me stepping in with those changes - I'm still working hard to stabilize the syntax highlighting to get it on by default and I'm concerned that if we don't address some of those issues up front it will be more challenging to untangle later. There's still a couple of to-dos on each but the 'plumbing' is there in terms of how I envisioned the system behaving. Let me know if you have feedback or if you'd rather integrate them into your changes here.

I really appreciate the work you've done here though - the hover looks _awesome_ with syntax highlighting and I'm stoked to get that in! 👍 Thanks for working with me on this!

Akin909:
@bryphe If at the end of the day `oni` ends up with syntax highlighting in the buffer and in the hover menu then thats well worth it 👍 I appreciate your helping re advice and the PRs. 

One thing that this brings to mind though is that I think some issues there are sometimes considerations that someone contributing might not be aware of, for example in this case, breaking the dependency from vim highlights as that led to *quite* a few tweaks which i'm happy to do to get some ** color in the hover box** 🤣 I think with issues that are mission critical or have specific implementation requirements there should be a label to indicate that the strategy re implementation might need to be discussed first? I tend to be a bit impetuous personally so a label against an issue that indicated it wasn't one to rush and implement willy nilly would be helpful. 

Granted this wasn't so much a standalone issue as new thing which I should probably have raised an issue for and discussed implementation first given that it wasn't a stretch to imagine there'd be many discussion points

Not sure what you think but I think it might save contributors from falling into traps they might not even know to look for or creating or worsening dependencies they didnt consider.

re ongoing work on this PR I will wait till #1485 and #1484 are completed merge those in or merge master if you merge them and see what needs to be changed in this PR to work with those and continue from there.

Akin909:
UPDATE: I've merge both of your PR's into mine, and will focus on remaining highlight quirks, regexp tweaks and will need to add more tokens to the default list as those generate all the tokens and there currently arent enough there to provide good levels of highlighting.

bryphe:
> One thing that this brings to mind though is that I think some issues there are sometimes considerations that someone contributing might not be aware of, for example in this case, breaking the dependency from vim highlights as that led to quite a few tweaks which i'm happy to do to get some color in the hover box

Definitely! I really appreciate that you're willing to work with me and take feedback 😄 

> I think with issues that are mission critical or have specific implementation requirements there should be a label to indicate that the strategy re implementation might need to be discussed first? I tend to be a bit impetuous personally so a label against an issue that indicated it wasn't one to rush and implement willy nilly would be helpful (exaggeration for effect I try never to be willy nilly.

This is challenging because it can be dependent on the approach and developer taking on the issue - for example, there might be cases where I have a solution in mind that doesn't need an in-depth architecture change, but it turns out that's not accurate... or what i thought was a big change ends up being simple, etc. In other words, two developers might look at the same issue but come out with very different ideas in terms of the implementation plan, so it's tough to know ahead of time, and not realistic for me to break down every issue to that level of detail anyway.

> Not sure what you think but I think it might save contributors from falling into traps they might not even know to look for or creating or worsening dependencies they didnt consider.

My preference is that if there are any unknowns, it's good to discuss them before embarking on the work. It can be hard to know ahead of time all the unknowns, and it's different for each developer too depending on what areas they are comfortable with - so there has to be some impetus on the developer to proactively voice these and get feedback on approaches. For example, I really like what @TalAmuyal did for #1415 - put together a sketch of the plan/interface, made sure we're on the same page, and then moved forward with it. Taking an incremental approach and breaking the PR into smaller chunks can be helpful here, too.

I tweaked our contributing guidelines in #1495 a bit and added a line here to that effect:
```
* If there is ambiguity in terms of design, architecture, or implementation, it's best to get feedback before implementing, to save both you and the maintainer time. If you're not sure, feel free to ask!
```
Let me know if you have feedback!

Another thing that I'd like to do is document our architecture more thoroughly, especially some of the design principles like _orthogonality_, _minimizing coupling_, _single responsibility principle_, etc. Some of these things are implicit or not obvious in the codebase so it'd be worth calling them out!

bryphe:
> UPDATE: I've merge both of your PR's into mine, and will focus on remaining highlight quirks, regexp tweaks and will need to add more tokens to the default list as those generate all the tokens and there currently arent enough there to provide good levels of highlighting.

Sweet, sounds good! Thanks @Akin909 - looking forward to having this 👍

Akin909:
Update: I've added some unit tests for the markdown functions (though tbh those aren't really linked to  marked I've just called it that) and will come back to it re. enzyme tests once I've pushed an enzyme PR I'm working on

Akin909:
Further update managed to fix the outstanding bug (that I'm aware of *tm*) which was that code blocks weren't highlighting which was actually not a commonly encountered issue as most of the markup is being rendered in `p` tags by `marked` but it was an issue with marked trimming whitespace from the text passed to the custom renderer which I've resolved. So believe this is good to go.

browser/test/MarkdownTests.ts
```diff
@@ -0,0 +1,91 @@
+import * as assert from "assert"
+import { StackElement } from "vscode-textmate"
+import * as markdown from "./../../browser/src/Editor/NeovimEditor/markdown"
+
+describe("Markdown Conversion Functions", () => {
+    it("Scopes to string fn should correctly convert scopes to classnames", () => {
+        const className = markdown.scopesToString(["token.scope.extension"])
```
bryphe: Thanks for the tests here! 👍

browser/src/UI/components/QuickInfoContainer.tsx
```diff
@@ -0,0 +1,54 @@
...
+        const hasDocs =
+            hasTitle &&
+            !!(
+                titleAndContents &&
+                titleAndContents.description &&
+                titleAndContents.description.__html
+            )
+
+        return (
+            <TokenThemeProvider
```
bryphe: Cool idea to separate the logic of getting the token colors from the logic of rendering

Issue #1407, opened 2018-01-31
Definition underline not correctly placed


Akin909:
Since very recently when the cursor rests on a `symbol` the definition underline is not located underneath the symbol sometimes it rests off to the side a little or by several words or is below by several lines or does not appear at all.

The errors I'm getting are mainly `react key warnings` and one regarding the `oni-layer-browser` plugin and `rxjs` (which I believe) is a separate issue  I was going to raise.

screenshot:

<img width="462" alt="screen shot 2018-01-31 at 11 44 38" src="https://user-images.githubusercontent.com/22454918/35621501-68a513a4-067c-11e8-9763-6705c80fc4ae.png">

Akin909:
To be closed once I confirm that this relates to changes to #1318 I'm running locally following removing using the document viewport.

Akin909:
Screen grab:

![underline](https://user-images.githubusercontent.com/22454918/35629232-fa49d380-0695-11e8-9d73-b872819d4ccb.gif)

Not sure if the definitions is one component that is constantly re-rendered in a different place or an array but if it is an array then this feels like the kind of bug would occur if keys are not set properly.

bryphe:
Ah nice catch, thanks for the issue @Akin909 !

bryphe:
Should be fixed now with #1454

Akin909:
indeed this does appear to be fixed, thanks, closing

PR #1454, opened 2018-02-07
Fix 1407: Definition underline doesn't always update correctly

bryphe:
__Issue:__ This appears to be a regression from #1314 - originally, the hover query and definition query were grouped together, but this bug started reproducing after that change.

__Defect:__ There is a race condition - in `LanguageEditorIntegration`, we assume that the definition will be reset and then a new one produced, but because of the store throttling, and how quick the definition query is in some cases, that isn't always true (and the position won't be updated).

__Fix:__ The ideal fix would be to address the race condition in `LanguageEditorIntegration`, but if we bring the delay back it should work in general. Without the delay, and with the fix, the go-to definition is actually a bit jarring because of the variable delay between it and hover, so restoring the previous functionality (using the same _delay_) seems like a reasonable solution for now.


bryphe:
Fixes #1407

PR #1456, opened 2018-02-07
Fix #1473: Run integration tests against local electron on developer machines

bryphe:
__Issue:__ One pain point with debugging the CiTests is that they run against the packaged build, and its a long iteration loop. It definitely makes sense for them to run against packaged builds on the Ci machines, but locally, it's much easier if they run against the already setup electron.

__Fix:__ Use environment variables to gate whether we should launch from the local install or packaged install. On CI machines, we should use the dist build, but on local machines, we use the local electron. This can be overridden locally by setting the `ONI_AUTOMATION_USE_DIST_BUILD` environment variable.

### DIFF
--- a/test/common/Oni.ts
+++ b/test/common/Oni.ts
@@ -7,7 +7,16 @@ const log = (msg: string) => {
     console.log(msg) // tslint:disable-line no-console
 }
 
-const getExecutablePath = () => {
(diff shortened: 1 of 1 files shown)


TalAmuyal:
Great job!

PR #1474, opened 2018-02-08
[ISSUE-1439]: Display a list of all failed tests at the bottom of the…

someguynamedmatt:
… CITest output

This is what I've come up with:

tests

Ideally, we would suppress the native output at the very bottom, but I wasn't sure how to go about doing that. I'm open to suggestions!

### DIFF
--- a/test/CiTests.ts
+++ b/test/CiTests.ts
@@ -5,7 +5,7 @@ import * as path from "path"
 
 import * as mkdirp from "mkdirp"
 
-import { Oni, runInProcTest } from "./common"
+import { IFailedTest, Oni, runInProcTest } from "./common"
 
 const LongTimeout = 5000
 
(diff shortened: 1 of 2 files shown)


bryphe:
Awesome, thanks @someguynamedmatt ! Looks like a big improvement over what we have today. Really appreciate your help!

> Ideally, we would suppress the native output at the very bottom, but I wasn't sure how to go about doing that. I'm open to suggestions!

We might able to do this with a custom [reporter](https://mochajs.org/#reporters). We actually have an example one here for code coverage: https://github.com/onivim/oni/blob/master/browser/testCoverageReporter.js

We could write a simple no-op one and just leverage the output we have in `after`, for the CI tests.

But that could be looked at separately - I'll bring this in now since it's a big improvement over our logging today! 👍 Thanks again for looking at this!

TalAmuyal:
Thanks @someguynamedmatt !!
I'm sure it will help me!

PR #1484, opened 2018-02-09
Syntax Highlight - Part 1 - Extract tokenColors into theme + merge with user config

bryphe:
This is splitting out a portion of #1443 - the part that sets the token colors. This adds the `tokenColors` to the current theme, and creates a `TokenColors` class whose responsibility is to be the 'source of truth' for the token colors, by coalescing/prioritizing token colors from the user's configuration and the current theme. I'd like to keep the `ThemeManager` focused on loading the theme in order to keep it simple. `TokenColors` has the responsibility of rationalizing the default token colors/theme token colors/user token colors, and notifying consumers of when the token color changes, and the current set of token colors.

This has a few benefits:
- Addresses the concern with `ThemeManager` being at different levels of abstraction for `colors` vs `tokenColors` 
- No chance of 'infinite loop' where a theme change -> token color change -> theme change -> token color change, etc, since the `TokenColor` is always downstream from `ThemeManager`
- Easier to potentially unit test the logic of `TokenColor` separately

This is a pre-req for the work to fix the syntax highlighting logic to not depend on highlights, which in turn can hopefully simplify some of the plumbing in terms of passing around the tokens in #1443 , as the syntax highlighting's dependency on vim highlight groups trickles down and causes issues further down the pipe.

To-dos:
- [x] Need to unwrap the colors from Vim
- [x] Add a simple test for TokenColors


browser/src/Services/TokenColors.ts
```diff
@@ -0,0 +1,90 @@
...
+        this._subscriptions.forEach(s => s.dispose())
+        this._subscriptions = []
+    }
+
+    private _updateTokenColors(): void {
+        const tokenColorsFromTheme = this._themeManager.activeTheme
+            ? this._themeManager.activeTheme.tokenColors
+            : []
+        const userColors = this._configuration.getValue("editor.tokenColors")
+        this._tokenColors = [
```
Akin909: @bryphe whilst I was looking at trying to do this I noted that generally spreading was leading to shallow merging and so for example is user had the following: 
```js
"tokenColors": [{
	"variable.other": {
		"bold": true,
}
```

and the default had specified:

```
"tokenColors": [{
	"variable.other": {
		"bold": false,
		"italic": true,
		"foreground": rebeccapurple,
}
```

I *think* at least in the case of objects you end up with one overwriting the other. In the case of spreading arrays
I *again think* that the result will be two `"variable.other"` tokens neither of which are quite right. I looked at using lodashes `merge` which works however is a bug magnet as it *`mutates`* the object or array so it can be very hard to debug since you have no reference to the original array or object.

A solution I planned on discussing was using `ramda's`, ramda is a functional utility lib so emphasis on non-mutation etc, and I believe with tree shaking its possible to import just the one file relating to the merge functionality you want without necessarily importing the whole lib, though depending on your feelings re functional programming I find it to be a safer tool than lodash since it never does things like mutating and has the same/ very similar functionality

bryphe: Oh ya, good point!

I was thinking that shallow merging may be acceptable for this, _if_ we go by order of precedence. 

For example, if the theme defines:
```
tokenColors: [{ scope: "variable.other", settings: { bold: true, foreground: "rebeccapurple" }]
```

And the user defines:
```
'editor.tokenColors': [{ scope: "variable.other", settings: { italic: true, foreground: "red" }]
```

We'd end up with:
```
 [{ scope: "variable.other", settings: { italic: true, foreground: "red" }, { scope: "variable.other", settings: { bold: true, foreground: "rebeccapurple" }]
```

And when we go to apply themes, we'd iterate through the array, find the first match, and apply those settings. So if we're applying highlight for `variable.other`, we'd find the first match, and use those settings (which would be the first index in the array). The downside with this approach is we wouldn't _inherit_ previous settings by doing the merge (ie, the `bold: true` wouldn't get set), but I think that's a reasonable constraint.

Let me know what you think though!

Akin909: I think having seen it in testing this I found the experience odd from user perspective aka, I wouldnt expect that half filling out a config setting or at least *I wouldn't realise it was half filled out* to suddenly apply my one change but to suddenly lose everything else aka the defaults or to have it not take my settings plus I think the overhead of someone using the tokens to have to wonder/work with duplicates is probably going to cause issues somewhere at some point, I think the expected result should be one object with my changes overriding the one field I filled out.

I defer to you here but this I think is an undesirable problem which can be fixed at worst by implementing a homebrewed deep merge aka when merging first check if theres a matching scope and update it otherwise just merge. I'm happy to add that if you agree to my PR?

bryphe: > , I wouldnt expect that half filling out a config setting or at least I wouldn't realise it was half filled out to suddenly apply my one change but to suddenly lose everything else aka the defaults or to have it not take my settings plus I think the overhead of someone using the tokens to have to wonder/work with duplicates is probably going to cause issues somewhere at some point

Makes sense! I think it's reasonable to add this functionality. It'd be a great candidate for a standalone PR scoped to implement that - ideally we could have an individual PR  that implemented  a unit test in `TokenColorsTest` documenting this behavior (specifying a token in both config and theme, and validating we get the correct result), and the implementation changes to make that work!

Also, as an aside, one potentially way to implement this would be to `reverse` the scopes and then `reduce`, merging in matching scope selectors settings, like:

```
const collectedSettings = tokenColorsInReverseOrder.reduce((currentSettings, newValue) => {
   if (!matchesScopeSelector(newValue)) { return currentSettings }
   return { ...currentSettings, ...newValue.settings }
}, DefaultTokenSettings)
```

But that's just an implementation detail - we could do a deep merge as you suggest too.

I'll keep it simple in this PR to unblock the remaining work, but I agree that it would be more convenient to have the override behavior.

PR #1485, opened 2018-02-09
Syntax Highlight - Part 2 - Use 'TokenColor' instead of highlight groups

bryphe:
This change completes the 'decoupling' of the highlight groups from syntax highlighting. This pushes the burden of converting between `TokenColor` <-> `HighlightGroup` to `NeovimInstance`, so that the remainder of the stack can be agnostic of highlight groups. In other words, in the Oni stack, all we care about is `TokenColor`, and its just the Neovim layer that cares about highlight groups.

This adds a class called `NeovimTokenColorSynchronizer` which is responsible for creating new highlight groups whenever there is a new token color, and maintaining a mapping between token color -> highlight group. We make sure this is up-to-date prior to setting any syntax highlights, but the cache is really only updated on first launch or when colorschemes change.

This is intended to be on top of  #1484 

Remaining work:
- [x] Set `real` colors in `_convertTokenStyleToHighlightInfo`
- [x] Fix `SyntaxHighlightingReconcilerTest` to update to new API
- [x] Add simple test case for `NeovimTokenColorSynchronizer` (verify that it caches + gets highlight group, and that same token color gives same highlight group)


bryphe:
> A comment re your comment about bold and italic yes nvim_hl command returns that if it is available otherwise it is undefined

Cool, thanks @Akin909 ! I'll update those to take into account.

> Also are the editor tokens going to be set to nothing in the default config not that its particularly relevant anyway and is the structure there going to be an array? (again this is just to confirm re how I might integrate this with my PR)

Right, the previous setup for `editor.tokenColors` was kind of a hack to get _something_ showing while we had a dependency on vim highlight groups. Now that we get a full set of token colors from `Neovim`, we don't need to set those - we can just let the user override them if they desire. In other words, that functionality is now superseded by the `VimHighlightToDefaultScope` map in #1484 . 

>  (again this is just to confirm re how I might integrate this with my PR)

For your PR, you'll probably want to get `tokenColors` from the `TokenColors` instance -that brings together tokens from configuration, the current theme, and any defaults (from vim), so it'd have the set you'd want as the base for hover (and then, apply any additional scopes/settings from the hover on top of that).

Akin909:
@bryphe did you find having more vimhighlights in the defaults problematic, I ask as I've gone in the opposite direction i.e i've added more groups as the few that were there match really infrequently with the token highlights so highlighting was *really* sparse without colors from vim which the hover has no access to if there are fewer defaults there'll be little to no highlighting, an alternative :S is generating the highlights based on the existing groups but there be quite noticeable gaps

Can I add a method to the token colors to request more highlights from vim for other ui so that it wont interfere with vim's highlighting but the hover can access the colors it'll need? don't think this will return us to the original problem as this wouldn't be a dependency for functionality just a richer experience whilst we still lack themes etc.

bryphe:
> did you find having more vimhighlights in the defaults problematic, I ask as I've gone in the opposite direction i.e 

Yep, when I used JavaScript files I'd see large blocks of highlighted text, so I'm going to stick with the existing set today. This makes the change more scoped anyway, as it's using the same set of highlighting we have today. We can look at expanding this set incrementally in additional PRs.

> i've added more groups as the few that were there match really infrequently with the token highlights so highlighting was really sparse without colors from vim which the hover has no access to if there are fewer defaults there'll be little to no highlighting, an alternative :S is generating the highlights based on the existing groups but there be quite noticeable gaps

Makes sense - in the meantime, we can use a specifically increased set of tokens for hover. It sounds like we were going to have a separate / larger set for hover anyway.

> Can I add a method to the token colors to request more highlights from vim for other ui so that it wont interfere with vim's highlighting but the hover can access the colors it'll need? don't think this will return us to the original problem as this wouldn't be a dependency for functionality just a richer experience whilst we still lack themes etc.

I think the best would be to add a method in hover to combine the `TokenColors` (which is from the theme + configuration) with an augmented set specifically for hover. In other words, keep `TokenColors` having the [single responsibility](https://en.wikipedia.org/wiki/Single_responsibility_principle) of coalescing the configuration + theme, and then have another object (or function) whose responsibility it is to reconcile that with specific hover styling. Maybe a `HoverTokenColors` object that has `TokenColors` as a dependency?

Akin909:
I've created a render prop that can be used in multiple locations `TokenThemeProvider` which augments the themes with token colors as classes which it passes to the render component so hopefully it can be used quite easily in a load of places without much work from another contributor as for the colors I had the exact same thing which led to pretty undesirable color mishmashes whilst I was trying to color the theme. I'll just create a function in the token theme render prop to just more tokens based on the existing ones, itll mean a much narrower set of color for now but itll still be much better than nothing 😆

bryphe:
@Akin909 - cool, that makes sense to me! Like, one place I could also see using the `TokenThemeProvider` is giving some nice syntax coloring in completion - and we'd want the augmentation there too. So that seems like a good place to augment the themes. 👍

Akin909:
@bryphe I've added a function to my PR to populate scopes which works well one thing which it's missing is a scope for strings. As these appear everywhere or often enough I'd like to add this *one* token to the defaults, it shouldn't add any weird colour issues (I've tried it previously) and the only token which use it are strings so as far vim highlighting is concerned nothing would change strings would still be coloured as strings but I'd have a hex value to use for the hover?

bryphe:
@Akin909 - sounds reasonable! I think adding tokens incrementally is good, that way if there is a regression or issue we'll notice it and be able to narrow it down.

bryphe:
Looks like this is green now, so I'll bring it in - hope it's not too bad to integrate with #1443 . Really excited for our new hover experience 👍

PR #1489, opened 2018-02-09
Feature/Add inner name component to truncate long names

Akin909:
This PR adds truncation to long file names rather than wrapping which leads to issues with padding, there seems to have been an attempt in the less to add text overflow etc, I've added a styled component to render the name as the explicit styles are needed on the component itself (which is why this wasnt already working) and also an explicit width is necessary for the text overflow to kick in but setting a static width forces all tabs to be at that width which is visually undesirable. This component renders a width *only* if a file name is longer than 50 chars and prevents wrapping of the text

output:

screen shot 2018-02-09 at 13 18 47

### DIFF
--- a/browser/src/UI/components/Tabs.less
+++ b/browser/src/UI/components/Tabs.less
@@ -77,7 +77,6 @@
     transition: opacity 0.25s;
 
     overflow: hidden;
-    text-overflow: ellipsis;
     user-select: none;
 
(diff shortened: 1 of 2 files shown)


bryphe:
Cool, thanks for fixing this @Akin909 !

Akin909:
💯 merging :)

PR #1497, opened 2018-02-10
Bugfix/add package json to prettierignore

Akin909:
Not sure what the best solution would be to this but since `npm/yarn` insist on formatting the packages .json this constantly leaves changed files post push or commit, which occasionally have to be stashed to change branches etc. or muddy the water in terms of knowing what i've actually meaningfully changed. A solution to get round this is to add new package.json files to a `prettierignore` that way there isn't a competition (ps not sure if some sort of glob would work, theres nothing in the docs about globbing)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+package.json
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,2 @@
 package.json
+vim/core/oni-plugin-typescript/package.json
--- a/vim/core/oni-plugin-typescript/package.json
+++ b/vim/core/oni-plugin-typescript/package.json
@@ -10,7 +10,10 @@
         "test": "npm install && tsc -p tsconfig.test.json && mocha --recursive ./lib_test/test"
     },
     "oni": {
-        "supportedFileTypes": ["javascript", "typescript"]
+        "supportedFileTypes": [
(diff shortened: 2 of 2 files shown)


TalAmuyal:
Great! was also looking for a solution.
Would've approved the PR if I had the courage to w/o @bryphe's permission :P

Akin909:
@TalAmuyal 😆 I know exactly what you mean still definitely feel a bit uneasy without @bryphe's input 😱, yeah has been bugging me for a while  hopefully this helps unfortunately without a glob itll be a manual process of specifying each new sub-package.json but at least a solution till the next package crops up

bryphe:
> @TalAmuyal 😆 I know exactly what you mean still definitely feel a bit uneasy without @bryphe's input 😱

Haha, I know the first few will be weird. But if it makes you feel better, I didn't even look at the files - if you guys are both good with the change then I'm good too 👍 I have noticed the issue when switching branches so it'd be awesome to have a fix! Will be fun to test out this new workflow!

> Would've approved the PR if I had the courage to w/o @bryphe's permission :P

Go for it! 😄

Akin909:
👍 Thanks @TalAmuyal

bryphe:
Woohoo! 🎉 Very nice 😄

PR #1498, opened 2018-02-10
Tabs show modified state.

### LINKED ISSUE #977: Tab dirty state doesn't show in tab

CrossR:
Some early work on getting tabs to show the modified state and fix #977.

Current issue is that I'm using the `buffer.modified` value to check if a buffer has been modified, which is what the `buffer` mode tabs use.

However, this seems to not be set when swapping between splits or between tabs. This also seems to be an issue for me if I use `"tabs.mode": "buffer"` as well though, so is this a regression or was it just not wired up like that in the first place?

To test this I'm doing as follows :
- Open Oni
- Open file
- Open new file with `:vsplit FILE2`
- Edit file 2
- Click on original file, and modified status will go away in the Oni visualised tabs. Neovim returns 1 as expected when using `:echo &modified`.

This happens in both modes, and on the latest release, so I don't think its me, but if someone could confirm its not any of my settings that would be useful.

Regardless, to do:

- [x] Check where I am calling `tabpagebuflist` makes sense.
- [x] Fix the modified status turning off when swapping. 
- [x] Add Unit tests for Tab modified status function
- [x] Add CI test for Tab modified (and Buffer if needed)


CrossR:
@bryphe, it looks like the current CI test way of pulling in the `config.js` settings didn't work for strings. I think I've fixed that here since I needed to use it, but am just pointing out the change since it isn't directly related to the title of the PR!

bryphe:
Ah thanks for the heads up @CrossR ! Shoot, I hit the same issue in #1354  -
https://github.com/onivim/oni/blob/6ac8763a8e127e78dce96642429444883361791a/test/common/runInProcTest.ts#L63

Sorry to duplicate that work, but your fix for it looks great. Thanks for checking, I agree it makes sense to bring it in with this change 👍  Really appreciate you writing the test for it too!

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -431,7 +431,17 @@ export class NeovimEditor extends Editor implements IEditor {
...
-        this._neovimInstance.on("tabline-update", (currentTabId: number, tabs: any[]) => {
+        this._neovimInstance.on("tabline-update", async (currentTabId: number, tabs: any[]) => {
+            const atomicCalls = tabs.map((tab: any) => {
+                return ["nvim_call_function", ["tabpagebuflist", [tab.id]]]
+            })
+
+            const response = await this._neovimInstance.request("nvim_call_atomic", [atomicCalls])
+
+            tabs.map((tab: any, index: number) => {
+                tab.buffersInTab = response[0][index]
```
CrossR: Need to check that `nvim_call_atomic` will always return in the order its called, which is what I'm expecting.

@bryphe, would you happen to know is this is the case?

CrossR:
This should be in a working state now, with passing Unit + CI tests.

I want to do a little more testing to make sure there is no odd edge cases I'm missing, but otherwise its done.

browser/test/Mocks/index.ts
```diff
@@ -170,9 +170,10 @@ export class MockEditor extends Editor {
 export class MockBuffer {
     private _mockHighlights = new MockBufferHighlightsUpdater()
     private _cursor = { line: 0, column: 0 }
+    private _modified = false
 
-    public get id(): string {
-        return "1"
+    public get id(): number {
```
CrossR: I swapped this to be a number, since the non-mocked buffer's ID is a number.

CrossR:
Done some more testing with this and didn't hit any hiccups, so I'll bring this in now!

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
bryphe: Sorry I missed this - that is my understanding though, that the order is preserved. We at least rely on that in a few other places too 😄

CrossR: I had a look over the documents + the PR that added `nvim_call_atomic` and I saw no mention it would return in any other order so I presume we are okay!

Issue #1488, opened 2018-02-09
Unhandled exceptions take down unexpected functionality


Akin909:
@bryphe  this is related somewhat to #1383 but I've noticed a pattern where I lose a certain bit of functionality sporadically whilst using oni i.e on startup or several hours later but in general a lot of these errors seem to propagate and take down some bit of functionality some where sometimes in a very unexpected way. I believe #1487 is one example, another is #1419.

Not sure what a good solution is but my brute force very inelegant knee jerk reaction is to wrap every function in the `NeovimEditor` for starters in try catches 😆

bryphe:
> @bryphe this is related somewhat to #1383 but I've noticed a pattern where I lose a certain bit of functionality sporadically whilst using oni i.e on startup or several hours later but in general a lot of these errors seem to propagate and take down some bit of functionality some where sometimes in a very unexpected way.

Yes, definitely frustrating! We still have a ways to go in terms of building up 'robustness'

> Not sure what a good solution is but my brute force very inelegant knee jerk reaction is to wrap every function in the NeovimEditor for starters in try catches 😆

Haha, unfortunately I don't think this is a tenable solution (unfortunately)

The main issue that happens is that, once an unhandled exception occurs, often the app can be in an undefined state. Imagine you get an exception right before setting some critical piece of state - catching it might keep the app from crashing at that point, but now the state is inconsistent, and you might get even _crazier_ issues due to that inconsistent state. The impact of that is you can spend your time debugging lots of different issues that have a common root cause!

When you compare to a runtime like C++ or C#, if there is an unhandled exception, the app just crashes / quits! In some cases, that might be more desirable.

Our best defense here is to put the errors _front and center_ - we should let the user know as soon as we hit these issues, because once there is an unhandled exception, all bets are off. We should give them the tools to quickly log (ideally, we'd have telemetry too so we could keep an eye on the most common unhandled exceptions, and prioritize them!)

In parallel, we should try and catch these errors as quickly as possible. It can be challenging to test all the cases due to the variety of configurations & vim plugins, but  as we find issues, it's helpful to build a `CiTest` or something to reproduce it (like the regression test added in PR #1309 for the window-opening-crash). These tests pay dividends because we know that error is covered and won't come back!

Today our automation was swallowing errors... and a lot of errors if they didn't have user impact were sort of being swept under the rug... but I think if we make them "front and center" and also monitor them in automation (#1491), we can make a push forward in terms of robustness.

Akin909:
[...]

PR #1499, opened 2018-02-10
Bugfix/on focus split error

Akin909:
@bryphe thanks to the new error ux I was coerced 😆 (actually think it is a great idea 🍾 though I wish it would only inform of an error once not every single time it occurs, maybe we could cache/storing error messages and only show the new modal if the error hasn't already been seen before, plus it specifically address an issue I raised 🚀 #1488 )
into looking into this bug which kept occurring whilst using splits.

Seems that setting menu items in the `sharedNeovimInterface` is async so I've change the calls to reflect that. Another issue which the bug raised which I havent found a solution for is that in the file explorer on opening a file `windowManager.focus` is called but with the `Oni.Editor` rather than a windowSplit id which is what it expects, except as far as I can see it the Oni.Editor has no knowledge of which split it belongs to which I imagine is a sensible thing for me to add I just have no idea where or how they interrelate since each split has a `private` inner editor and the editor has no idea about splits

EDIT: My temporary fix/hack was to pass in what I noted manually was the right split id which quietens down the errors. PS: if you agree re storing the messages i'm happy to look into that the red banners are a great call to action but also undismissable, 

TODO:

- [x] As well as the above this is a note to self that there is still a missing await as an error still occurs on initial vim enter 🤔

### DIFF
--- a/browser/src/Services/Explorer/ExplorerSplit.tsx
+++ b/browser/src/Services/Explorer/ExplorerSplit.tsx
@@ -108,7 +108,9 @@ export class ExplorerSplit {
         switch (selectedItem.type) {
             case "file":
                 this._editorManager.activeEditor.openFile(selectedItem.filePath)
-                windowManager.focusSplit(this._editorManager.activeEditor as any)
+                // FIXME: the editor manager is not a windowSplit aka this
+                // Should be being called with an ID not an active editor
+                windowManager.focusSplit("oni.window.0")
                 return
             case "folder":
                 const isDirectoryExpanded = ExplorerSelectors.isPathExpanded(
--- a/browser/src/Services/Sidebar/SidebarPane/SidebarPane.tsx
+++ b/browser/src/Services/Sidebar/SidebarPane/SidebarPane.tsx
@@ -35,13 +35,13 @@ export class SidebarPane {
         )
     }
 
-    public enter(): void {
+    public async enter(): Promise<void> {
         this._menuBinding = getInstance().bindToMenu()
 
         const widgets = this._store.getState().widgets
         const ids = flatMap(widgets, w => w.ids)
 
-        this._menuBinding.setItems(ids)
+        await this._menuBinding.setItems(ids)
 
         this._menuBinding.onCursorMoved.subscribe((id: string) => {
             this._store.dispatch({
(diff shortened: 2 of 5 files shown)


bryphe:
> @bryphe thanks to the new error ux I was coerced 😆 (actually think it is a great idea 🍾 though I wish it would only inform of an error once not every single time it occurs, maybe we could cache/storing error messages and only show the new modal if the error hasn't already been seen before, plus it specifically address an issue I raised 🚀 #1488 )

Awesome!! Thanks for fixing this 👍 I started to notice it too now once the notification was added, but didn't get a chance to investigate - really appreciate the fix. Hoping it'll prompt me to be better about catching/fixing issues up front too 😄 

The caching / storing error messages sounds like a great idea. It doesn't make sense to spam the user with the same message.

> Another issue which the bug raised which I havent found a solution for is that in the file explorer on opening a file windowManager.focus is called but with the Oni.Editor rather than a windowSplit id which is what it expects, except as far as I can see it the Oni.Editor has no knowledge of which split it belongs to which I imagine is a sensible thing for me to add I just have no idea where or how they interrelate since each split has a private inner editor and the editor has no idea about splits

This seems reasonable to me- there are still some changes happening in regards to the Window API (how we manage / control multiple windows from the API). Using the id directly should be fine, though (in the future we might have _more_ Editor windows but we don't need to tackle that pre-emptively).

Thanks again for fixing this! 💯

Akin909:
@bryphe I know this is all green now from a review perspective anyway but just wanted to point out that I made another change here lol, when the shared instance is initialised we save an `_initPromise` then await it shortly after not sure how or why this makes that much of a difference but rather than save the promise (note: we dont use the init promise for something) I just await the instances initialisation which fixes lingering issues where on startup the `MenuBinding.setItems` could not call `.eval` on the instance because it didnt exist yet.

bryphe:
>  I just await the instances initialisation which fixes lingering issues where on startup the MenuBinding.setItems could not call .eval on the instance because it didnt exist yet.

Cool, thanks for fixing that!

Akin909:
Haha seems I spoke to soon before now the bug is actually dead on startup and and subsequent split navigation although there was/is a more subtle issue at play here which I'm not sure I entirely understand when the `sharedNeovimInstance` is started the promise which should resolve once the neovim instance is initialised hangs interminably so that on navigating splits the first time after load the neovim instance is still `null` and the bug occurs just that once (now that it initialised) seems for some reason initialising a neovim instance doesn't resolve till sometime later its unclear when because the log show `starting...` till you move through windows and the bug occurs.

My solution just wraps that specific action in a check to see if the instance is initialised it doesn't solve the question of why the neovim instance start function doesnt resolve 🤔 till... not sure when or how but it does eventually

PR #1500, opened 2018-02-10
Stop waiting for Oni at end of install.

CrossR:
Tiny change for a tiny bug.

At the end of installing on Windows, the option to run Oni causes the installer to wait and not close until Oni is closed.
This allows the installer to close as expected.

Only other thing I noticed is there is currently a difference between Windows + other platforms, where Windows is a registered editor for less file types than we have in the `package.json`. Should probably update `BuildSetupTemplate.js` now so that it generates the filetype list on the fly from the `package.json` list, so we don't have multiple versions of the truth.

### DIFF
--- a/build/setup.template.iss
+++ b/build/setup.template.iss
@@ -33,7 +33,7 @@ Name: "addToRightClickMenu"; Description: "Add {{AppName}} to the right click me
 Name: "{group}\{{AppName}}"; Filename: "{app}\{{AppExecutableName}}"
 
 [Run]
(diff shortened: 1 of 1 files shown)


bryphe:
> At the end of installing on Windows, the option to run Oni causes the installer to wait and not close until Oni is closed.

Oh cool, now that you've mentioned this, I've noticed it before. Thanks for catching it and fixing!

> Should probably update BuildSetupTemplate.js now so that it generates the filetype list on the fly from the package.json list, so we don't have multiple versions of the truth.

That's a great idea 👍

PR #1505, opened 2018-02-10
Automation - Stability: Report running processes when tests start (and kill)

bryphe:
__Issue:__ Automation tests intermittently time out.

One contributing factor I've found in the logs is that, somehow, Oni/Electron is failing to close. The tests will fail if a second instance of Oni is open, so this change would let us detect any open process and potentially try to kill them before opening a new instance.


bryphe:
The `kill` process part might need to be part 2 - once this instrumentation is hooked up, we'll have a good sense if we have the right set of processes to terminate (or if this is even the issue)

Issue #246, opened 2017-02-22
No help on command line


xelra:
When typing `oni --help` or `oni -h` on the command line, no help appears that explains options. Instead it just starts ONI.

extr0py:
Good point, thanks for tracking this.

extr0py:
We can potentially do the following:
- add `minimist` library to `main.js`
- look for `-h, --help` options
- print out some canonical help text with our current supported options

keforbes:
Are there any flags/options Oni currently supports?  It seems to me the help would just be:
```
Oni [filename]
       launch Oni with or without a filename
```
I agree we should have some help output but what else would go in the help?

CrossR:
Looks like this was added now by @someguynamedmatt in #1509.
Cheers!

I'll close now then.

PR #1509, opened 2018-02-10
[ISSUE-246]: Add --help option to command-line oni

someguynamedmatt:
> based on #246 

- Added `--help` (also `-h`) to command line initialization
sample output:

args

### DIFF
--- a/cli/oni
+++ b/cli/oni
@@ -12,7 +12,17 @@ processOptions.env = Object.assign({}, process.env, { "LOCAL_ONI": 1 })
 
 var isWindows = os.platform() === "win32";
 
-var args = process.argv.slice(2);
+var argv = require("minimist")(process.argv.slice(2));
+var args = argv._;
+var version = require(path.join(__dirname, "..", "package.json")).version;
+
+if(argv.help || argv.h) {
+    process.stdout.write("ONI: Modern Modal Editing - powered by Neovim\n");
+    process.stdout.write(` version: ${version}\n`);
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks for taking this on @someguynamedmatt !

One thing I realized is that unfortunately this won't address #246 completely - the `cli/oni` script is _only_ used for running via `npm` (`npm link`) - I don't believe it's used when Oni comes from any of the packaged binaries unfortunately.

someguynamedmatt:
Ah, okay. I'll keep poking around then.

someguynamedmatt:
@bryphe Is there a trick to getting a packaged build locally? I have another solution, but I'd like to test it out before harassing you with PRs 😄

someguynamedmatt:
@bryphe Okay, this should be better. I've tested it on OSX by packaging the app, installing it locally, and handing it `--help`. Seems functional. I can't get the packaging to work on Linux, though I'd assume it would work as well. Let me know what you think. And, as always, suggestions welcome!

bryphe:
Looks good to me! I just tried it on Windows, and there is a bit of weirdness happening with `stdin`/`stdout` here, but I think this is a pre-existing issue:
image

(Somehow, the output gets printed _after_ the prompt - it's strange!)

bryphe:
But it makes sense to bring this in - we can tackle the windows issue separately. Thanks for the contribution, @someguynamedmatt ! 👍

PR #1473, opened 2018-02-08
Fix #1446: Explorer - add overflow handling to items

bryphe:
This prevents wrapping of items in the explorer menu, and truncates with ellipsis if the name goes outside the explorer UI.

We still need to get resize behavior (#1414), but this is similiar to how other editors handle it.

PR #1513, opened 2018-02-11
Bugfix/ Catch thrown error in tss parsing response

Akin909:
When parsing a response from the `tss` if the response fails, we currently reject with an error which isn't handled at all, this error seems to crop up if you rename a file or open a new file (ts) I think the `tss` is probably correctly not able to deal with request since it maybe doesn't know what happened not sure if theres a better solution here but seems we should at least catch the error, not sure whether more need to be done here but I just log that that occured 🤷‍♂️

### DIFF
--- a/vim/core/oni-plugin-typescript/src/TypeScriptServerHost.ts
+++ b/vim/core/oni-plugin-typescript/src/TypeScriptServerHost.ts
@@ -81,7 +81,12 @@ export class TypeScriptServerHost extends events.EventEmitter {
 
         this._rl.on("line", msg => {
             if (msg.indexOf("{") === 0) {
-                this._parseResponse(msg)
+                try {
+                    this._parseResponse(msg)
+                } catch (e) {
+                    // tslint:disable-next-line
+                    console.warn(`Error parsing TSS response: ${e}`)
+                }
             }
         })
     }


bryphe:
I'm on the fence about this one - I'm worried that if we sweep these exceptions "under the rug", we might miss valid bugs. I think that the exception being thrown in these cases is a real issue that we should try and understand. 

#1003 is an example of some similiar cases that were addressed when we upgraded to TypeScript 2.6. My preference would be to understand the issue a bit more before we suppress it - it'd be great if we could isolate it to a specific set of commands, or understand the circumstances. With #1003 , there were some changes in the protocol's expectations of when we call `"open"` vs `"change"`.

It'd be great to have test cases that exercise this directly. I think the ideal situation for this is to create test cases that can reproduce the error, and implement fixes to make it 'green'.

One possible way would be to create a `CiTest` that works against Oni.languages (`LanguageManager`) directly. We could construct a bunch of requests like:
```
await Oni.languages.sendRequest("textDocument/didOpen", { ... })
await Oni.languages.sendRequest("textDocument/didChange", { ... })
...
etc
```

And validate we get proper responses. This is something that would pay off greatly when we upgrade TypeScript (or if we decided to switch to another LSP), because we'd have a ton of confidence that things will work due to this coverage. We have only cursory, incidental coverage mainly in the auto complete test, which isn't ideal.

Akin909:
@bryphe definitely don't want to sweep any significant errors under the rug and I get what you mean re the actual root causes, I seem to be able to fairly consistently reproduce it by opening a new ts file moving from an open file in a project. Looking at the circumstance I'm not sure any of the actions above would cover this as it seems to be specifically about changing sessions

screen shot 2018-02-11 at 15 04 46

I think its more of a mismatch between how oni communicates changes like that with the `tss`, but just guessing. Although it is a bit "dodgy", I think it's inevitable that the tss or any other lsp will throw errors, I think even perfect dilligence cant prevent that given that things will likely change etc. my thinking was that where something bound to throw an error one day whilst we ought to be informed it should also not disrupt using oni, and I *think* that this is one of the errors that take down functionality as after it occurs I find I no longer get completions

Akin909:
I'll close this and if I come up with the actual solution I'll make a PR for that and add an associated test case. Although I'm having quite a bit of trouble running test locally (although I was gonna raise that separately so if I pick you brain and there might be solutions that could help anyone else its document somewhere)

bryphe:
Sounds good! 👍  I'd really prefer that we have an understanding of the issue + test coverage before we suppress this. Our TS test coverage is very lacking considering how much we depend on it, and how easy it is to break when changing `tsc` versions - making it fragile. Having test coverage and understanding the underlying issue can help us make an informed decision.

(FYI, I am planning on disabling the notifications (by default) for the _production_ build just for this one release on Monday - because there are still several bugs that have been existing already, but we haven't fixed yet. Just a one-time thing though for this release. Doesn't change the behavior of any of the issues though, just doesn't report it via notification. Wanted to give us one more time before it is always on - although we've made some awesome progress in fixing issues that had been ignored or missed previously!)

> Although I'm having quite a bit of trouble running test locally (although I was gonna raise that separately so if I pick you brain and there might be solutions that could help anyone else its document somewhere)

Hmm, let me know what issue you are facing! It should be a lot better now with #1473 because you don't need a packaged build - you can just run `npm run test:integration`. The one gotcha is all Oni instances need to be closed. Usually when I'm writing a test I also comment out all the tests in `CiTests.ts` so it just runs my one test.

PR #1514, opened 2018-02-11
Check there is files to open before attempting to open.

CrossR:
This was causing an error if the Open File window was open and closed without selecting anything.

At least the error messages being out front means stuff like this is easily caught now!


bryphe:
Nice, thanks for fixing this @CrossR !

> At least the error messages being out front means stuff like this is easily caught now!

Definitely! 👍

Issue #1524, opened 2018-02-11
Development/ Integration tests timeout on (this) MacOS


Akin909:
As per gitter integration test are hanging on *my* mac not sure if this is truly a platform issue possibly local environment. Here's a log of the output of the `yarn test:integration`  command
[out.log](https://github.com/onivim/oni/files/1714771/out.log)

bryphe:
Looks like the tests are getting stuck on this wait condition:
```
     Log.info("[AUTOMATION] Waiting for neovim to attach to editor...")
        await this.waitFor(
            () =>
                editorManager.activeEditor.neovim &&
                (editorManager.activeEditor as any).neovim.isInitialized,
            30000,
        )
        Log.info("[AUTOMATION] Neovim initialized!")
```

Could you please attach the output of `:CheckHealth` when you run Oni? It'd be really helpful if it is without a config. I'm wondering if there is any environment variable set or anything that could impact Neovim starting up.

Akin909:
@bryphe here's the output, is my init.vim loaded when the integration test are spun up? there's a lot of stuff in there so wouldn't suprise me if something in there broke something else.

```
health#nvim#check
========================================================================
## Configuration
  - OK: no issues found

## Performance
  - OK: Build type: Release

## Remote Plugins
  - OK: Up to date

## terminal
  - ERROR: command failed: infocmp -L
    infocmp: environment variable TERM not set
  - INFO: $XTERM_VERSION=''
  - INFO: $VTE_VERSION=''
  - INFO: $TERM_PROGRAM=''
  - INFO: $COLORTERM=''
  - INFO: $SSH_TTY=''

health#provider#check
========================================================================
## Clipboard (optional)
  - OK: Clipboard tool found: pbcopy

## Python 2 provider (optional)
  - WARNING: pyenv was found, but $PYENV_ROOT is not set.
    - ADVICE:
      - Did you follow the final install instructions?
      - If you use a shell "framework" like Prezto or Oh My Zsh, try without.
      - Try a different shell (bash).
  - WARNING: No Python interpreter was found with the neovim module.  Using the first available for diagnostics.
  - WARNING: provider/pythonx: Could not load Python 2:
    /Users/akinyulife/.pyenv/shims/python2 does not have the "neovim" module. :help |provider-python|
    /Users/akinyulife/.pyenv/shims/python2.7 does not have the "neovim" module. :help |provider-python|
    python2.6 not found in search path or not executable.
[...]

PR #1525, opened 2018-02-12
#1524 - Extra logging to help diagnose test timeout issue

bryphe:
This isn't a fix for #1524 , but hopefully can give us a bit more info or clues in terms of why the `SharedNeovimInstance` `nvim_ui_attach` call is failing!


bryphe:
@Akin909 , would you be up for running the `AutoClosingPairs` test against this branch, and sharing the logs? Would be interesting to see around this `nvim_ui_attach` call if there is any craziness..

bryphe:
I'll bring this in, since it could help with debugging issues in the future, too.

PR #1528, opened 2018-02-12
Code Coverage: Try alternate piping strategy

bryphe:
Try piping via cat to coveralls


bryphe:
Going to try switching from `coveralls` to `codecov`..

PR #1532, opened 2018-02-12
Add commandline and wildmenu mode on by default

Akin909:
@bryphe thanks for turning these on by default, as a side note as part of my alleged enzyme PR I'm looking to add some tests for these and will look at adding some integration ones as well once I have the integration tests working locally


bryphe:
Awesome, thanks @Akin909 ! I'll bring this in - and then start getting the 0.3.0 builds ready for release 👍  🎉

PR #1536, opened 2018-02-12
Feature: add file explorer item options

Akin909:
I went out on a limb and added a feature I *assumed* would be of value to other people which relates to the fact that clicking the root directory in the file explorer does nothing. My immediate thought was that if I didn't find what I was looking for then clicking that would move me one level up 🤷‍♂️ , I definitely appreciate that this is an assumption someone else might assume that clicking it would close everything ?? (my counter point there would be that that's the equivalent of closing the explorer 🌵 )I have no idea what the *correct* action should be but I though I'd put forward my proposed solution. Its a very small addition so not much lost If this doesn't seem like the way we'd like to go with this 🤞 

EDIT:  An additional command I can add to this PR is the reverse action aka navigate downwards and I an also create commands for both so they can be used by the navigator


CrossR:
One thing I will say is that given a UI of 
image

(ie the downward pointing arrow by the side) I would assume it would collapse the file tree.

Can totally see your point though that closing the file tree can be done by just hiding it (at least in its current form, if we allowed multiple workspaces that would be a different story), but if this is adopted I think the UI should change to reflect the action more easily.

Akin909:
I see your point @CrossR I think that's actually what other editors with multiple trees do, happy hear someone else's take as I definitely felt like this was something that might be unique to me (I've hit top folder like 5 times knowing full well that the last time it just showed "not implemented yet" ^^)

bryphe:
>  (I've hit the top folder like 5 times knowing full well that the last time it just showed "not implemented yet" ^^)

Ah ya I meant to fix that before the release... derp!

But ya, I think that if we change the functionality, we'd need to have a different icon there - otherwise it's confusing that the caret does one thing for the folders, but something different for the container.

The reason I used the `container` expand/collapse metaphor for that was I was thinking we might have additional items there (originally i had 'recent files', but maybe multiple workspaces? recent files? etc).

Having a gesture to navigate upward makes sense, though. It would be neat to have a way to expose other functionality though - like, when you hover over an item, have a set of buttons become available. (By hover, I mean both focus via the keyboard, and hover with the mouse). Potentially we could use `h` and `l` to navigate these additional options. And at the very least, we should expose that functionality as a `command` that can be `bound`. Maybe `backspace` would be a reasonable default binding when we are in the explorer window?

Akin909:
Begun refactoring to have a *small* menu above the icons, going to try and use the vim navigator to add `h` and `l` functionality. To select commands like `delete` and `rename`

PR #1538, opened 2018-02-12
Bugfix: Reduce log noise

TalAmuyal:
- Reduce the severity of certain unused Neovim events to `verbose` (originally was `warning`)
- Prettify `oni-plugin-markdown-preview`'s `package.json`

Benefit: reduce clutter for a user that wants to view logs and report the actual error.


bryphe:
Thanks for doing this, @TalAmuyal ! 👍 , makes sense to downgrade these - they tend to clutter up error reports (or act as red herrings for real issues).

PR #1550, opened 2018-02-13
Use new PluginManager API

TalAmuyal:


TalAmuyal:
@bryphe for some reason Travis's OS X machine really dislikes me.
The tests fail again for no good reason.
How can we improve it?

I really want to move on and work on more features, but it really feels that the CI process is an obstacle for me.

bryphe:
Ah ya, sorry about the issues @TalAmuyal . 

I believe the problems stem from the caching we have in TravisCI. The problem is, we need _some_ caching because downloading the oni-neovim-binaries and oni-ripgrep-binaries is throttled by github.

I'll try either:
- Seeing if there is a safe token we can use to authenticate with github (the token to download doesn't need much permissions)
- See if we can cache _just_ those packages that have binaries

bryphe:
Still investigating the cache failures.... but since it's not due to your change, I'll bring this in. I have another idea to try in terms of dealing with the binary downloads..

TalAmuyal:
Thank you very much, @bryphe !

bryphe:
Sure thing, @TalAmuyal ! I really appreciate your contributions.

I'm hoping #1606 and #1616 will address this. After that, there is still some test flakiness, but those should hopefully make the PR builds more reliable, especially in the cases where the dependencies change.

PR #1555, opened 2018-02-14
[issue-1548] Fixed issue where null Sidebar was being returned under certain confi…

someguynamedmatt:
…gurations which prevented quickOpen and commandMenu functionality

### DIFF
--- a/browser/src/Services/Sidebar/SidebarStore.ts
+++ b/browser/src/Services/Sidebar/SidebarStore.ts
@@ -57,11 +57,16 @@ export class SidebarManager {
         return this._store
     }
 
-    constructor(private _windowManager: WindowManager) {
+    constructor(private _windowManager: WindowManager = null) {
         this._store = createStore()
 
-        this._iconSplit = this._windowManager.createSplit("left", new SidebarSplit(this))
-        this._contentSplit = this._windowManager.createSplit("left", new SidebarContentSplit(this))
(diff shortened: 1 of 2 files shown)


someguynamedmatt:
@Akin909 Haha, yeah it’s hard to see when the commit messages get too long. @CrossR documented the bug yesterday. I don’t have merge permissions so anyone can merge this. Thanks!

bryphe:
Thanks for investigating this, @someguynamedmatt ! (And reviewing, @Akin909 👍 )

I think it makes sense to bring this in since it fixes the `Control+P` and `Control+Shift+P` functionality, which are pretty crucial.

One bug I did notice though that we can address separately is if I press `Control+Shift+F` (find in files), I get an error:
image

What we might want to eventually do is _always_ create the sidebar + sidebar content pane, but if `sidebar.enabled` is `false`, we just hide them (or they start hidden, and we only `show` them if `sidebar.enabled` is true`). Then, if a user does `Control+Shift+B` to toggle the sidebar, or `Control+Shift+F` to find in files, we can make it visible.

bryphe:
Logged #1562 to track this next issue, but I'll bring in this now and we can look at that later. Thanks again for your help here @someguynamedmatt and @Akin909 !

someguynamedmatt:
@bryphe I like that idea. Wouldn’t be too difficult to implement either.

Issue #1562, opened 2018-02-14
Sidebar: When 'sidebar.enabled' is false, get errors when trying to find in files


bryphe:
Following the discussion from #1555 :

One bug I did notice though, when `sidebar.enabled` is `false` (after #1555), and I try and 'find in files' with `Control+Shift+F`, I get an error:
![image](https://user-images.githubusercontent.com/13532591/36218374-f8ffe856-1168-11e8-93c7-d23761052d1d.png) 

What we might want to eventually do is _always_ create the sidebar + sidebar content pane, but if `sidebar.enabled` is `false`, we just hide them (or they start hidden, and we only `show` them if `sidebar.enabled` is true`). Then, if a user does `Control+Shift+B` to toggle the sidebar, or `Control+Shift+F` to find in files, we can make it visible.

someguynamedmatt:
Hmmm, @bryphe I can't recreate this on my Linux box. This is the config file I'm using (very lean for testing):

```
const activate = (Oni) => {
   console.log("config activated")
}

const deactivate = () => {
   console.log("config deactivated")
}

module.exports = {
   activate,
   deactivate,
   "editor.fontSize": "14px",
   "editor.fontFamily": "Droid Sans Mono for Powerline",
  //add custom config here, such as
  //"oni.useDefaultConfig": true,
  //"oni.bookmarks": ["~/Documents",]
  // "oni.loadInitVim": true,
  "sidebar.enabled": false,
  //"editor.fontSize": "14px",
  //"editor.fontFamily": "Monaco"
}
```

PR #1564, opened 2018-02-15
WIP: based on Issue-1562

someguynamedmatt:
### Allow for Oni to behave as expected when `sidebar.enabled` is set to `false`

> Based on [Issue #1562](https://github.com/onivim/oni/issues/1562)

When `sidebar.enabled` equals  `false` (i.e. the sidebar is never given a WindowManager) hitting `<s-c-f>` would throw an error

- Added a new key binding `<s-c-e>` 

- Historically, the sidebar was never initialized if the configuration file disabled it. Now, when setting `sidebar.enabled` to `false` the sidebar is created normally, but hidden from the user
    - This prevents the previously documented error and still allows for search/explorer use with the appropriate key commands: `<s-c-f>` (File search), `<s-c-e>` (Explorer), and `<s-c-b>` (open sidebar to last used option)

- Removed error previously documented by @bryphe 

@bryphe @Akin909 This is still a work-in-progress, the one thing I can't seem to work through is opening the `Explorer`, focusing on it (your cursor is in it), and then keeping if focused while closing it. 

When you do this you find that the main editor window has lost focus and Oni seems to be unresponsive. By manually clicking in the main window everything "goes back to normal". Focus should be regained automatically when the sidebar is closed. This issue exists in the current code as far as I can tell (only with the `Explorer` sidebar, `Search` behaves correctly). Maybe you guys have a better solution to giving the main window its focus back?

As always, open to suggestions. Thanks!

### DIFF
--- a/browser/src/Input/KeyBindings.ts
+++ b/browser/src/Input/KeyBindings.ts
@@ -52,6 +52,7 @@ export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configurati
         input.bind("<s-c-t>", "language.symbols.document")
         input.bind("<c-tab>", "buffer.toggle")
         input.bind("<s-c-f>", "search.searchAllFiles")
+        input.bind("<s-c-e>", "explorer.toggle")
 
         if (config.getValue("editor.clipboard.enabled")) {
             input.bind("<c-c>", "editor.clipboard.yank", isVisualMode)
--- a/browser/src/Services/Explorer/index.tsx
+++ b/browser/src/Services/Explorer/index.tsx
@@ -18,4 +18,16 @@ export const activate = (
     workspace: Workspace,
 ) => {
     sidebarManager.add("files-o", new ExplorerSplit(workspace, commandManager, editorManager))
+
+    const toggleExplorer = () => {
+        sidebarManager.toggleVisibilityById("oni.sidebar.explorer")
+    }
+
+    commandManager.registerCommand({
+        command: "explorer.toggle",
+        name: "Explorer: Toggle Visiblity",
+        detail: "Toggles the explorer in the sidebar",
+        execute: toggleExplorer,
+        enabled: () => !!workspace.activeWorkspace,
+    })
 }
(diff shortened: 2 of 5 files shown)


Akin909:
@someguynamedmatt this looks like a good fix out of curiosity what happens if I have the sidebar disabled then toggle the search quickly then finish the search, does this close the sidebar again, is this the issue you were describing re giving focus back to the editor window.

I think at the moment the issue with setting focus with the `windowManager.focusSplit(slit.id)` is that the Oni.editor does not currently have any knowledge of a `window.id` which is what the focusing function requires.

I discussed this very briefly with @bryphe, to me it seemed like each editor should have an associated window id which is what neovim do re buffers all having window numbers, especially for when the multiplexing PR comes in

someguynamedmatt:
Yeah that’s the issue exactly. When you close the sidebar right now you have to click the main buffer window to give it focus again. I’ve been looking for a solution and I think I’m getting a little closer but it’s definitely not easy.

I’m glad you like it so far.

bryphe:
Ah ya, right now it's hard to go from `Oni.editor.activeEditor` to the actual window id. 

Right now, the window manager creates and assigns an id in `WindowManager`'s `createSplit` method. It returns a `WindowSplitHandle` which would let us focus it, but that gets sort of tossed on the ground during initialization in the `startEditors` method.

It might be helpful in WindowManager to be able to get an `id` or the `WindowSplitHandle` object based on the split. Something like:
```
const windowSplit = windowManager.getSplit(Oni.editors.activeEditor)
windowSplit.focus()
console.log(windowSplit.id)
```

If we had that, it'd be pretty easy to bring the focus back when we're closing the sidebar.

In terms of implementation, the `WindowManager` maintains a dictionary of ids in `_idToSplit`, in order to implement that method, we could just look through the `IAugmentedSplitInfo`s, see which one matches the split passed in, and then return the id or a `WindowSplitHandle` object (which is really just a flyweight object that has the window id + talks back to WindowManager).

CrossR:
Thanks for the contribution! I updated the PR and added the focus bits over in #2343.

PR #1581, opened 2018-02-18
Feature/drag and drop files in explorer

### LINKED ISSUE #1406: Drag and Drop Files and Folders - Explorer
@bryphe I think it would be a very handy/ the norm in most editors to be able to resize a sidebar view by clicking and dragging, I've had a look at what is available and there are a few *React* components which provide this functionality.

An example is [`react-rnd`](https://github.com/bokuweb/react-rnd), [`re-resizable`](https://github.com/bokuweb/re-resizable) etc. which provide wrapper components with this functionality looking at `react-rnd` it doesn't look *too* difficult to implement this and potentially the values of the explorer can also be persisted when the editor is re-opened.

As a **Bonus Feature** there are also libraries for creating *draggeable* components, which potentially could be used to allow interactions like moving folders 😮 , not sure how much more complicated that functionality might be but can look into the resizing and then dragging for bonus points if reasonably straight forward

EDIT: PS I'm __V. likely__ underestimating the complexity of this task 🤣

Akin909:
This PR aims to fix #1406, using `react-dnd` it adds drag and drop functionality to the file explorer so files can be moved between folders, whilst a very gui feature this aims to simplify the process of moving things around for those off moments when there isn't a quick and easy vim command

### Todo

- [x] Make files in the root drop targets so files in subdirectories can be moved to the root

- [x] Bonus points: make folders drag and droppeable


Akin909:
@bryphe if you get an opportunity I'd like to get your feedback on this PR, at present I've got the drag and drop functionality working for files and folders 😄 but there are some ux things I thought might be nice or also might be overkill like a sound playing when a file is relocated, and also some sort of highlighting of the visual area for a drop.

The reason for this is that at the moment it doesn't feel very clear that a file has moved when you move it, it appears in the folder but and I preface this by saying its quite subjective, but it feels like its silent so it leaves a sort of uncertainty, till I check and verify it has moved.

I peeked at the vscode file explorer and noted that if you try to drop a file into a folder the folder area highlights, although that might be much harder in oni as the files and folders aren't in a common container I believe.

browser/src/Services/DragAndDrop.tsx
```diff
@@ -0,0 +1,117 @@
...
+const DropCollect = (connect: DND.DropTargetConnector, monitor: DND.DropTargetMonitor) => ({
+    connectDropTarget: connect.dropTarget(),
+    isOver: monitor.isOver(),
+    canDrop: monitor.canDrop(),
+})
+
+/**
+ * A component which can have items of a specific matching type dropped onto it
+ */
+@DND.DropTarget<IDroppeable>(({ accepts }) => accepts, DropTarget, DropCollect)
```
bryphe: Cool use of decorators! We havne't leveraged these much..

browser/src/Services/Explorer/ExplorerSplit.tsx
```diff
@@ -74,12 +78,60 @@ export class ExplorerSplit {
         this._store.dispatch({ type: "LEAVE" })
     }
 
+    public moveFileOrFolder = (source: Node, dest: Node): void => {
```
bryphe: I like that this is decoupled from the UI and primarily dependent on the state/selectors - means we could have some other ways to implement this behavior. It'd be cool to hook up a `yank` + `paste` workflow later on!

browser/src/Services/Explorer/ExplorerSplit.tsx
```diff
@@ -74,12 +78,60 @@ export class ExplorerSplit {
...
+
+    public moveFolder = (
+        source: ExplorerSelectors.IFolderNode,
+        destination: ExplorerSelectors.IFolderNode,
+    ) => {
+        if (source.folderPath === destination.folderPath) {
+            return
+        }
+        Log.info(`moving folders: ${source.folderPath} to ${destination.folderPath}`)
+        mv(source.folderPath, destination.folderPath)
```
bryphe: Is `mv` sync? Just wondering if there can be a race for the move call and the refresh action we dispatch.

browser/src/Services/DragAndDrop.tsx
Akin909: I kind of like them too but here it wasn't so much a choice as how all the documentation and examples are written out, I would have tried without but didn't want to complicate learning the new tool by doing everything differently from the docs 😄

browser/src/Services/Explorer/ExplorerView.tsx
```diff
@@ -49,36 +78,71 @@ export class NodeView extends React.PureComponent<INodeViewProps, {}> {
...
         switch (node.type) {
             case "file":
                 return (
-                    <SidebarItemView
-                        text={node.name}
-                        isFocused={this.props.isSelected}
-                        isContainer={false}
-                        indentationLevel={node.indentationLevel}
-                        icon={<FileIcon fileName={node.name} isLarge={true} />}
+                    <DragAndDrop
```
bryphe: This seems very intuitive! 👍

browser/src/Services/DragAndDrop.tsx
bryphe: 👍 Makes sense to me!

browser/src/Services/Explorer/ExplorerSplit.tsx
```diff
@@ -74,12 +78,60 @@ export class ExplorerSplit {
...
+        this._store.dispatch({ type: "REFRESH" })
+    }
+
+    public findParentDir = (fileId: string): string => {
```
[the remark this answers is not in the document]
Akin909: LOL 😆 nope nope it is not I was initially doing something more complex than necessary there and when I refactored I completely missed the obvious step there of just using `path.dirname` 👍 🤣

browser/src/Services/Explorer/ExplorerView.tsx
Akin909: 😄The library is really quite nice but I fumbled with it a bit initially spent a week making no progress because I didn't realise how the items for dragging and dropping *ought* to be labelled so decided to separate these out into render props that theoretically someone else could use that seemed a little bit more straightforward.

bryphe:
I just tried it out, the functionality is really nice!

> but there are some ux things I thought might be nice or also might be overkill like a sound playing when a file is relocated, and also some sort of highlighting of the visual area for a drop.

It's great that there is a border while we hover over the item - I like that. This is a good case where animations can be helpful - like having the source item fade out and the destination item fade in - it tells a story that the item has relocated (the destination item _fading_ or _popping_ in would be a good metaphor to help the user make sense of what just happened).  In addition, we might want to expand the destination folder - I noticed if I dragged a file into a folder, it was hard to track what happened.

A sound might be overkill (we don't really have any sounds today, we previously had a beep sound, but it's not there anymore AFAIK) - I'm not sure if any other tree / explorers do it?

A success notification might be helpful - like a big green checkmark + a message like "Successfully copied C:/oni/test.ts to C:/oni2/test.ts"

I like the changes you brought in with react-dnd, this will open the door for other cool stuff, like reordering tabs. Great work on the change! 👍

browser/src/Services/Explorer/ExplorerSplit.tsx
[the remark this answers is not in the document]
bryphe: Haha, i know the feel 😄

browser/src/Services/Explorer/ExplorerSplit.tsx
Akin909: From the docs all the `shelljs` commands I believe are sync they have a mega thread on their repo about adding async support but doesn't exist yet

Akin909:
@bryphe 👍 those would really improve the ux. I've added a success notification and expand the dir on drop, the outstanding thing now is to look into the animations

Akin909:
@bryphe had a go at using `react-transition` for the animations but unfortunately am at a loss for how to get **only** the file or folder that was just dropped or dragged animating. `DND` exposes a prop called `didDrop` which is fired just after a drop but by wrapping all drag and drop components in `<Transition >` they all animate on enter and leave 😖, not sure if you're aware of a away to animate just one component wrapped in a transition in and out based on a single prop without having the others animate.

bryphe:
> not sure if you're aware of a away to animate just one component wrapped in a transition in and out based on a single prop without having the others animate.

I think, if you're using the CSS transitions, we could add an additional className to the element - like `animation-enabled`, and then use a CSS selector inside the `styled-component` where the animation lives, to gate on having both the `enter` selector` and `animation-enabled`.

Ya, we definitely don't want all the items animating!

bryphe:
I'm okay splitting the animation out as a separate change, too - up to you @Akin909 !

Akin909:
@bryphe I'll have a go with what you suggest re the `css transition` and if Im successful I'll implement it here, if it seems to be dragging on or might be more involved than expected for some reason happy to split it out.

Akin909:
Just tried with the animations and no luck using the `classNames` probably just using the api wrong but think I'd rather give it another shot as part of another PR

bryphe:
Looks really nice, @Akin909 ! 👍 I like the notification a lot.

PR #1590, opened 2018-02-21
Fix tab creation issues some of the time.

CrossR:
Check that input parameters are arrays.

I'm not fully sure how this issue came about, calling `:tabnew` and `:tabclose` a few times works, and then at some point it just breaks....

This at the very least fixes the issue here. If we've got no array for the tabs, the only real thing we can do is return false since there is no buffers to have been edited yet.

Edit: As a primarily non-TS dev, is there a different way of writing my function parameters? I assumed that defining it with `number[]` either only an array would be passed in, or that it would be treated like c style langs and find would just run over the first element and quit if a single val was passed, but that evidently hasn't been the case.

### DIFF
--- a/browser/src/UI/components/Tabs.tsx
+++ b/browser/src/UI/components/Tabs.tsx
@@ -235,6 +235,10 @@ export const shouldShowFileIcon = (state: State.IState): boolean => {
 }
 
 export const checkTabBuffers = (buffersInTabs: number[], buffers: State.IBuffer[]): boolean => {
+    if (buffersInTabs instanceof Array === false || buffers instanceof Array === false) {
+        return false
+    }
+
     const tabBufs = buffers.filter(buf => buffersInTabs.find(tabBuf => tabBuf === buf.id))
 
     return tabBufs.some(buf => buf.modified)
--- a/browser/src/UI/components/Tabs.tsx
+++ b/browser/src/UI/components/Tabs.tsx
@@ -235,10 +235,6 @@ export const shouldShowFileIcon = (state: State.IState): boolean => {
 }
 
 export const checkTabBuffers = (buffersInTabs: number[], buffers: State.IBuffer[]): boolean => {
-    if (buffersInTabs instanceof Array === false || buffers instanceof Array === false) {
-        return false
-    }
-
     const tabBufs = buffers.filter(buf => buffersInTabs.find(tabBuf => tabBuf === buf.id))
 
     return tabBufs.some(buf => buf.modified)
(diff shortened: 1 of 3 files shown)


badosu:
@CrossR Is type checking on typescript not strict enough? As far as I can see there are type signatures that should have avoided this error.

CrossR:
I'm actually not fully sure, I would have assumed so too. The error ended up being that I hadn't considered the case of no buffers being passed over.

My assumption was in that case we'd get `[]` which when `find` was called on there would be nothing returned so we'd end up with a bunch of empty sets to finally return false. Instead we got passed a `0` which is how this happened. 

That said I'm not a JS/TS dev, so I can't tell what is the best way to protect against that. Its also possible the error is actually further up where we are using `any` and it just surfaced here.

CrossR:
Okay so because `buf` is defined as `any` it can't find the definition of `buffersInTabs` so that also becomes `any` meaning that it allows the `0` to be passed over.

Is there any advantage of using `any` here vs `State.ITabs`? Or just that it could raise issues elsewhere if it is also `any` there?

bryphe:
> Is there any advantage of using any here vs State.ITabs? Or just that it could raise issues elsewhere if it is also any there?

Would definitely prefer to use `State.ITabs`! In general, it's better to have a 'real' type as opposed to `any` - `any` is really a crutch. There's even a compiler option to disallow `any`, which would be cool to enable at some point. That'll help catch these issues upstream. 

My guess is that as you mentioned there is likely something happening upwards (perhaps when we call the `setTabs` action in the `tabline-update` handler in `NeovimEditor`) - we don't validate that the `response[0][index]` is a properly formed array, so really `tab.buffersInTab` could get set to anything here. The biggest challenge we have is for these items that are external to our type-system - like these RPC calls to Neovim. Ideally, we'd validate that they are correct at the boundary, use strong types, and then we can leverage TypeScript's type system to protect us the rest of the way.

browser/src/UI/components/Tabs.tsx
```diff
@@ -235,6 +235,10 @@ export const shouldShowFileIcon = (state: State.IState): boolean => {
 }
 
 export const checkTabBuffers = (buffersInTabs: number[], buffers: State.IBuffer[]): boolean => {
+    if (buffersInTabs instanceof Array === false || buffers instanceof Array === false) {
```
bryphe: Following from the discussion - I wonder if it makes sense to push this 'upward' to here:
https://github.com/onivim/oni/blob/44e162d359c6b2a669e5d6dbcef4589c4de90b56/browser/src/Editor/NeovimEditor/NeovimEditor.tsx#L456

Here, we can check if `response[0][index]` is a valid array, and if not, use an empty array. That way, everything 'downstream' using the `State.ITabState` typing can rely on the type system - that `tab.buffersInTab` is an actual array of numbers. Otherwise, we might end up having to duplicate this check in other places we use `tab.buffersInTab`.

CrossR:
Should I assume those Mac issues are to do with #1592 or similar? I'm assuming I didn't manage to break everything on just mac!

bryphe:
Thanks for the update, @CrossR ! Looks good to me.

> Should I assume those Mac issues are to do with #1592 or similar? I'm assuming I didn't manage to break everything on just mac!

Ah ya, still trying to figure out how to correctly cache those binaries. Sorry about that. I'll bring it in as-is while that is still being resolved.

bryphe:
Sorry about the OSX CI issues. I'll go ahead and bring this in now. Thanks for the fix, @CrossR !

PR #1613, opened 2018-02-23
Improve shell scripts

zapashcanon:
Hi,

I fixed various things spotted by ShellCheck using: `find -O3 . -type f -name '*.sh' -exec shellcheck -x {} \;`.

There are still these:

```sh
In ./oni.sh line 4:
    OS='Mac'
    ^-- SC2034: OS appears unused. Verify it or export it.


In ./oni.sh line 28:
ONI_CWD="$OPEN_DIRECTORY" open --new -a "$FULL_ONI_PATH" --args $*
                                                                ^-- SC2048: Use "$@" (with quotes) to prevent whitespace problems.
                                                                ^-- SC2086: Double quote to prevent globbing and word splitting.
```

For the first one, I think it'll be used in the future when Linux is supported.

For the second one, I'm not sure what's exactly expected here, replacing `$*` seems legit to me, but I'm not sure about adding quotes. Let me know what to do, I'll amend the commit if needed.

### DIFF
--- a/build/script/travis-build.sh
+++ b/build/script/travis-build.sh
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
-echo Travis build - detected OS is: $TRAVIS_OS_NAME
+echo Travis build - detected OS is: "$TRAVIS_OS_NAME"
 set -e
 
 node --version
 npm --version
 
 if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-  export DISPLAY=:99.0
+  DISPLAY=:99.0
+  export DISPLAY
   sh -e /etc/init.d/xvfb start
   sleep 3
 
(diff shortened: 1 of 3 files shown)


bryphe:
Hi @zapashcanon ,

Thank you for the PR! Appreciate you putting this together.

> I fixed various things spotted by ShellCheck

Do these have an end-user impact, or are they primarily hygiene fixes? If it's the latter, I'm inclined not to bring this in, because there is potential for a regression - we don't have good test coverage around some of these scripts (`oni.sh` is the one I am most concerned about).

TalAmuyal:
Nice, @zapashcanon!
I didn't know about `shellcheck`, definitely gonna use it on my scripts - thanks!

zapashcanon:
@bryphe, in `oni.sh`, one thing will have an end-user impact: if `self` contains white space, now it'll work. The other thing is explained [here](https://github.com/koalaman/shellcheck/wiki/SC2006). With the current change in `oni.sh`, I'm confident there won't be any regression.

bryphe:
Thanks @zapashcanon ! Appreciate the details and contribution. 👍 Bringing this in now!

PR #1618, opened 2018-02-23
Feature/improve error ux

Akin909:
This PR tones down the bright red of the notification box to rendering the title based on the level of the notification it also handles issues with sizing where a large error (several of which kept occuring 🙇 ) would cause the container to push the editor tab bar upwards causing jerkiness.

@bryphe I realise the intent was to have a very obvious error message so users log bugs, though I found that the large wall of red (which is what occurs if there is an error with a large message) very distracting (not sure if others feel the same). I think some red anywhere suddenly appearing will definitely catch the user's eye as would any kind of notification over the editor window so hopefully these changes don't run counter to your intent.

Also added a possibly vain attempt to only show new errors by checking if an error is in the saved/queued errors.

Screenshot to illustrate changes:

screen shot 2018-02-23 at 22 39 56

screen shot 2018-02-23 at 22 54 22

I've implemented it with less colour but also played with this idea (which I'm putting here just to show alternatives, *personally* still feel like I never really want a large coloured banner anywhere on my screen)

screen shot 2018-02-23 at 22 20 53


CrossR:
I was going to say perhaps we should use blue for Info messages as most apps I've used use Red/Yellow/Blue for Error/Warn/Info, and green for succeeding.

But I guess the issue with that or any colour is we are assuming the user knows the colours too. I know in other notification UI you usually get a little text box message too, that makes it explicit what the message is. Is something like that worth adding etc?

Either that, or possibly swapping the icons between the three, such that you have Cross/Exclamation Point/Info symbol?

Akin909:
@CrossR I think blue for info sounds like a great idea I couldn't quite figure out what it should be but now you mention that really clicks, as does a blue icon, maybe font awesome have an info icon (seems highly likely), for warn vs error though not sure what symbol would tell them apart

Re the small text box not sure I know what you mean, like a line maybe between the icons to say `warning` or `info` or `error`, happy to include something like that though I think red and yellow are kind of universal for something probably not good is happening

bryphe:
The change looks great, @Akin909 ! I agree, the big square red box is a bit much - especially since we still have a bunch of errors to fix! I like your first screenshot with the less color.

👍 to using blue for info. I think that would be great. The `info-circle` icon from font awesome would be good for this.

For the error, it might be worth having a different icon too - There is the `times-circle` icon which is a circle with an x in it that might be fitting for an error. I think if we have the colors + the icons it would be enough to differentiate.

From a styling perspective, the only item I'd like to add is a 4 pixel border on the left with the color (red/yellow/blue). 

Thanks for looking at this!

bryphe:
Ah and I should have fixed the Linux issue on the build - you might need to grab the latest `master`.

browser/src/Services/UnhandledErrorMonitor.ts
```diff
@@ -33,9 +33,12 @@ export class UnhandledErrorMonitor {
             this._onUnhandledRejectionEvent.dispatch(evt.reason)
         })
 
-        window.addEventListener("error", (evt: any) => {
+        window.addEventListener("error", (evt: ErrorEvent) => {
```
bryphe: Thanks for the fixes here 👍

Akin909:
Thanks for the feedback guys tweaked it and looks much nicer I think

screen shot 2018-02-24 at 00 27 18

bryphe:
Awesome, looks great @Akin909 !

Akin909:
woops just merged upstream to get latest test changes and seems I've lost approval 😆, also calling for tests although its primarily css (btw these kind of changes can be test with [`jest-styled-components`](https://github.com/styled-components/jest-styled-components) though not sure re the value but it is possible 🤷‍♂️. Though still blocked on #1552 with typing issues.

Akin909:
Also was thinking would a timeout to remove the notifications after some timelimit be desirable as I find that if an error occurs and in my case I've seen it before but I'm working on something else or dont know what to do with it they just persist indefinitely till I'm forced to remove it by 🤢 **clicking** which I personally find quite frustrating

TalAmuyal:
@Akin909, how about using sneak mode to press the X button?

Akin909:
@TalAmuyal haven't actually tried 😆 will give that a shot.

Although re the suggestion my thinking was along the lines of as a user who's been shown a notification (not as someone contributing to the project but just as a regular user) an error occured maybe i've filed it before maybe I'm at work and just need to get on with things, if the flow is modal pops up with error, I glance at it but wanna get back to work *subjective opinion* - I feel like I shouldn't have to do anything to make it go away.

I think about it a bit like when I get a notification on my phone the back light comes on I see the message, I'm not going to do anything about it right now and eventually it just goes away and my phone goes back to sleep.

Might be over thinking this been looking at ux and user flows a lot more recently at work so suddenly questioning every interaction I have.

bryphe:
> Also was thinking would a timeout to remove the notifications after some time limit be desirable as I find that if an error occurs and in my case I've seen it before but I'm working on something else or don't know what to do with it they just persist indefinitely till I'm forced to remove it by 🤢 clicking which I personally find quite frustrating

A timeout would be great for notifications in general - but I was thinking for errors it makes sense to keep them sticky.

> @Akin909, how about using sneak mode to press the X button?

Integrating with sneak mode would be awesome! 👍 This should be relatively straightforward - just need to wrap the `cancel` button in the `<Sneakable />` component

bryphe:
> woops just merged upstream to get latest test changes and seems I've lost approval 😆, also calling for tests although its primarily css which happens to be in js/ts (btw these kind of changes can be tested with jest-styled-components though not sure re the value but it is possible 

Ah ya, I just turned off the gate on `codecov/project`  - I wonder if there is a way for us to configure the threshold? I don't think it really makes sense to add tests for the css, since it's so easy to visually verify - I think that would actually just end up being a tax w/o providing benefit. Approving again 👍

bryphe:
Looks like the codecov checks are still blocking - is it cool if I bring this in, @Akin909 ? Or were there any other pending changes you were planning? Thanks for your work on this! 💯

PR #1619, opened 2018-02-24
Image Layers

bryphe:
This change leverages our 'Buffer Layers' API to create a layer for images that renders it above the binary text:

image

It mostly works, but there are a few remaining issues:

- [ ] Image layer is inadvertently being activated for _all_ buffers - needs to be fixed!
- [ ] Split out image layer to an extension
- [x] Split out setting of filetype + language contribution to separate change, and add unit test


TalAmuyal:
Looks nice and a really great feature 👍

PR #1630, opened 2018-02-25
Bugfix/Neovim instance can be undefined on startup

### LINKED ISSUE #1523: Bug/ Requests to NeovimInstance made before it is initialised.
This was discussed briefly but I believe in #1499  but I think that the way the `NeovimInstance` is initialised means that some functions which call methods of the instance can be run before initialisation is complete. There are several examples as I personally get these errors quite often, but they include `colorChanged` command sending a `nvim.request` call on oni's startup or sending input too early or #1499. They usually take the form `cannot complete request of null` or something like that.

Not sure how to go about tacking this but perhaps the root cause is that the NeovimInstance start promise at least in case of #1499 didn't seem to resolve properly (there seems to be a timing issue as the promise should be resolved before Oni starts or rather it should block till neovim starts) so methods expecting the neovim instance were being passed null.

I've never use `Promise.race` but I'm wondering if since it resolves as soon as the first promise does, then Oni is getting initialised before the neovim instance promise has resolved

Akin909:
@bryphe whilst looking into #1583 re `shellEnv` behaviour I think I've found the cause of the request of undefined error that can occur/ occurs on the editor startup, btw this is not a fix for #1583 as the cause of that though related maybe is still unclear.

Having the call to `shellEnv` occur asynchronously in order to cut down on startup time means that when the editor window comes into view the promise to startup a neovim instance is often not actually complete so if you click the editor or try to go into the sidebar this was what was leading to the error

screen shot 2018-02-25 at 11 10 55

By using the sync version of the shellEnv function the errors stop entirely as things block till the instance is actually initialised.

I know the concern was about performance, so I checked the profile logs

First is the async version with the associated error notifications and time stamps:
screen shot 2018-02-25 at 12 30 42

Sync version:
screen shot 2018-02-25 at 12 31 28

The difference in startup cost seems quite small (and tbh on multiple attempts varied a lot with the 300ms difference being the largest I observed) and I think especially at the cost of starting Oni up before the neovim instance is ready.

I think neovim is a crucial thing that needs to be ready before the editor appears otherwise thing's reliant on it begin to initiate functionality dependent on it before it exists (also I *think* its initial reliance on the import promise still makes that function async so while internally it blocks at the sync call it does all this in a promise.

I also looked at the `mergeSpawnOptions` function which uses `shellEnv` and is actually quite an expensive function.

screen shot 2018-02-25 at 12 22 00

It seems that almost any attempt by oni to create a process calls this function which

1.) Dynamically imports shell-env
2.) Calls the shell-env method async

resulting in the calls above, the first is for `nvim`, second the `lsp` and third the `sidebar instance`.

I moved the files functionality into a class and saved the dynamic import on the first call then also saved the value of the call to shell-env to use in subsequent calls to this function which results in

screen shot 2018-02-25 at 12 23 16

fixes #1523

### DIFF
--- a/browser/src/Plugins/Api/Oni.ts
+++ b/browser/src/Plugins/Api/Oni.ts
@@ -10,7 +10,7 @@ import { EventEmitter } from "events"
 
 import * as OniApi from "oni-api"
 
-import * as Process from "./Process"
+import Process from "./Process"
 import { Services } from "./Services"
 import { Ui } from "./Ui"
 
--- a/browser/src/Plugins/Api/Process.ts
+++ b/browser/src/Plugins/Api/Process.ts
@@ -1,5 +1,4 @@
 import * as ChildProcess from "child_process"
-import { memoize } from "lodash"
 
 import * as Oni from "oni-api"
 import * as Log from "./../../Log"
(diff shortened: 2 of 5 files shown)


Akin909:
Also just as an aside running this in a local build and its completely not discernible from a user perspective is anything is at all slower, tbh I *subjectively* think that things might seem slightly faster 😆

bryphe:
Thanks for all the details you went into measuring the performance, @Akin909 ! Appreciate it.

I was concerned about the switch to the `sync` call at first, but given that we are caching the environment variables so we wouldn't hit that later (like spinning up an LSP), I think that is pretty reasonable. Nice work!

bryphe:
(And thanks for all the help investigating #1583 💯 )

PR #1631, opened 2018-02-25
feature/ upgrade ocaml lsp with upstream fixes

### LINKED ISSUE #1571: Ocaml LSP no longer working ?with new v1.0.26
Since very recently the Ocaml language server no longer seems to work i.e. does not provide completions, hover, diagnostics etc.

The Oni console shows
screen shot 2018-02-17 at 23 14 20

Looking at the changes made in the most recent `lsp` upgrade its not clear why Oni's integration no longer works though It might be that oni needs to change its configuration for the lsp

Akin909:
Noted that the crashes in the `ocaml-language-server` have been fixed upstream in the most recent version. This PR fixes #1571 as the issues have been resolved in 0.27 (tried locally and functionality is restored).


bryphe:
Awesome, thanks @Akin909 for investigating + fixing this!

PR #1636, opened 2018-02-26
feature/ add sneakable functionality to notification buttons

Akin909:
This PR adds sneakability to the buttons in the notification ui so they can be responded to using the keyboard


Akin909:
@CrossR thanks for the review 👍, yeah the clicking was driving me up the wall 😆

PR #1637, opened 2018-02-26
Feature/add timeout to info notifications

Akin909:
Add a 20s expiry time to all notifications that **AREN'T** errors aka info notifications and warn notifications, discussed in #1618 

This PR will conflict with #1636 but will push a conflict fixed version depending on which goes first


CrossR:
This looks good too, but think if I accept it you pushing the merge conflict fix will cause the review to become undone, so may as well wait and accept it after that.

As an aside, is it worth me creating an issue about the notification size?
The boxes seem much bigger for me now: 
image

Looks like there is some wasted space for me, though I am also not sure if that is because of my monitor res or scaling etc (4K monitor with 150% scaling and Windows isn't so great at scaling so it wouldn't surprise me if it was due to that).

Akin909:
@CrossR the max width on the notifications is set to 30% of the available space which since I work primarily of a laptop seemed to always be squarish but I guess on a hi-res large monitor 30% could easily be enormous might be worth me using `30rem` instead which should roughly equate to 30characters wide based on the current font size, can add that to this once I push the merge fixes

bryphe:
> might be worth me using 30rem instead which should roughly equate to 30characters wide based on the current font size, can add that to this once I push the merge fixes

Yes, +1 to using a measurement like this - the 30% is pretty huge on my monitor too. 👍

Akin909:
Updated 👍, dont actually have a large screen to hand but 25rem should be a more constant value than the percentage

CrossR:
image

Looks a lot more reasonable to me!

browser/src/Services/Notifications/NotificationsView.tsx
```diff
@@ -181,42 +181,67 @@ const NotificationHeader = styled.header`
...
+    private _timer: any // NodeJS.Timer should work but there's a type conflict
+    private _lifetime = 20000
     private iconDictionary = {
         error: "times-circle",
         warn: "exclamation-triangle",
         info: "info-circle",
     }
 
+    public componentDidMount() {
+        if (this.props.level !== "error") {
```
bryphe: In terms of implementation - how about we add an `expirationTime` property to the `INotification`, and add an `Epic` to handle the timeout behavior? I'd like for a consumer of our `Notifications` API to be able to set an expiration time.

Ideally, the UI wouldn't need to know or care about this - it could be handled in our store layer.

bryphe: More specifically:

1) We could add a `setExpiration` method on the `Notification` API object here: https://github.com/onivim/oni/blob/9d685a98ccd9485678144be4eff2562bd3cf4352/browser/src/Services/Notifications/Notification.ts#L16
2) We could add an `expirationTime` property in the store here: 
https://github.com/onivim/oni/blob/9d685a98ccd9485678144be4eff2562bd3cf4352/browser/src/Services/Notifications/NotificationStore.ts#L24

And then we could add an `Epic` on the store that looks something like this:

```
const hideNotificationAfterExpirationEpic: Epic<NotificationAction, INotificationState> => (action$, store) =>
    action$.ofType("SHOW_NOTIFICATION")
    .filter((action) => action.expirationTime !== -1)
    .delay((action) => action.expirationTime)
    .map((action) => ({ type: "HIDE_NOTIFICATION", id: action.id }))
```

At least something like that. The cool thing is, the UI doesn't need to worry about adding timers / handling lifecycle events, and we get the benefit of being able to specify this in our API surface as opposed to having it baked into our UI code.

Akin909: @bryphe sounds cool when I was adding it it definitely seemed like something that should be configurable given that I arbitrarily chose the number.

Thanks for the snippets that's really helpful as I'm not as familiar with Epics as I'd like and I've seen them around I'll have a tinker locally and push that when I get it working

bryphe: > Thanks for the snippets that's really helpful as I'm not as familiar with Epics as I'd like and I've seen them around I'll have a tinker locally and push that when I get it working

Awesome, thanks @Akin909 ! Let me know if you need any help.

The `ExplorerStore` uses just a couple of epics, hopefully those are good examples:
https://github.com/onivim/oni/blob/30c186aec1496ac9a8f6aa2af93a2b9fc98e0646/browser/src/Services/Explorer/ExplorerStore.ts#L191

(There's also some initialization required to 'wire up' the epics - example here: https://github.com/onivim/oni/blob/30c186aec1496ac9a8f6aa2af93a2b9fc98e0646/browser/src/Services/Explorer/ExplorerStore.ts#L228)

Akin909: @bryphe what looked like a slam dunk given your detailed snippet, has become a bit of a head scratcher. This is what i've got
```ts
const hideNotificationAfterExpirationEpic: Epic<NotificationAction, INotificationsState> = (
    action$,
    store,
) => {
    let currentExpTime: number
    return action$
        .ofType("SHOW_NOTIFICATION")
        .filter((action: IShowNotification) => action.expirationTime === -1)
        .map((action: IShowNotification) => (currentExpTime = action.expirationTime))
        .delay(currentExpTime)
        .map((action: IShowNotification) => ({ type: "HIDE_NOTIFICATION", id: action.id }))
}
```
except the map function at then end there gets passed a number by the delay function, not the action and I've had to do this not ideal thing of saving the exp time to an external var because the delay function doesn't get passed the action passed into it.

Had a brief look at the docs and couldn't find the correct alternative methods easily will have another look in a bit but thought I'd pick your brain first

bryphe: Ah interesting, I thought `delay` might have an overload like:
```
 .delay((action) => action.expirationTime)
```

But I guess that's not the case? If not, I'm okay with how you have it set up right now. It is a side effect but at least it's contained

Akin909: I'm quite interested in getting it to work with the `Epic` would be very cool will definitely spend a bit more time on it but can merge as is and I refactor it to the Epic in another PR or do it here?

bryphe: I'd prefer the `Epic` implementation for this PR, but let me know if there are any blockers.

Akin909:
@bryphe managed to get it working with the observables 😄 🎇 and added the api method as well as removed the component side effect

browser/src/Services/Notifications/NotificationStore.ts
```diff
@@ -72,6 +79,20 @@ export const notificationsReducer: Reducer<IdToNotification> = (
...
 }
 
+const hideNotificationAfterExpirationEpic: Epic<NotificationAction, INotificationsState> = (
+    action$,
+    store,
+) => {
+    return action$
+        .ofType("SHOW_NOTIFICATION")
+        .filter((action: IShowNotification) => !!action.expirationTime)
+        .mergeMap(({ expirationTime, id }: IShowNotification) => {
```
bryphe: Nice! 💯

bryphe:
Sweeet, nice find @Akin909 ! Change looks great - approving now.

Akin909:
Thanks @bryphe was nice to get the opportunity to use `redux-observable`, seems I took down one of the unit tests but just fixed it

Issue #1575, opened 2018-02-18
How should workspaces be used?


hoschi:
Like, really, I don't get it :cry: 

Before this new feature the current directory of my shell was set as `cwd` for Oni, when I called `oni` from command line. Now it opens the last workspace it was in. This on its should be changed or controlled with a command line flag. Opening oni from a current directory, just to be in another directory and navigating again to it is pretty awkward.

Another problem is to switch workpaces at the moment. It seems the only way to do this is to issue the "Workspace: Open Folder" command or via the menu entry. This opens a unusable file dialog on Linux, don't know if it is better on other platforms. Unusable because it doesn't show me the last folders I opened with oni. It show instead recently used folders ... where I don't even know how they come from. I suspect they get gathered when I use the system file explorer (like `Explorer` on Windows or `Finder` on Mac). But since I only use the console this list doesn't match anything useful for me. Even it would work, I still find it way more natural to open the current working directory where I'm at the moment.

To still be able to use Oni, I did following workaround in my config:
```js
let argv = require('electron').remote.process.argv
let defaultWorkspace
let cdIndex = argv.indexOf("--cd")
if (cdIndex >= 0) {
	defaultWorkspace = argv[cdIndex+1]
}

module.exports = {
    // ... normal config
    "workspace.defaultWorkspace": defaultWorkspace,
}
```

and call oni with 'oni --cd `pwd`'. Just to realize that `defaultWorkspace=undefined` has the same effect as gathering the `cwd` via args :grinning: Is this a bug? :wink:

Akin909:
@hoschi I don't open Oni much from the cli so cant speak to the experience sorry its causing trouble though. 

I think a lot of the workspace functionality is still in progress and I'm currently reading up on how lsps use it but they seem to be a cool gateway into project wide lsp functionality, symbol search is just one example of a usecase.

 Not sure if this will be a fix for you but we recently added a `workspace.autoDetectProjectRoot` option which means that `oni` will change its active workspace to the *project root* directory of the buffer `oni` has open. It should provide similar functionality as vim rooter (which was the inspiration) so when you switch buffer Oni should check for a project marker like `.git` and switch to that folder as its root.

The options for this are `always`, `noworkspace` and `never`. No workspace means that will only happen if oni has no active workspace. I think the problem your having is related to oni persisting the last opened workspace which I found not to be useful unless it updated the workspace when I changed buffer so I've found this setting when used with `always` helpful.

[...]

Issue #1397, opened 2018-01-30
Quick Open uses relative path sometimes.


CrossR:
Occasionally, QuickOpen seems to use a relative path.

As far as I can tell, the repo is as follows:

* `Ctrl+Shift+P` and Close the current Workspace.
* Close Oni, then re-open.
* `Ctrl+Shift+P` and Open a new workspace.
* `Ctrl+P` and open a file.

When I do this, I'll have Oni trying to open `browser\src\Font.ts` instead of `F:\Git\oni\browser\src\Font.ts`.

Clicking in the Explorer works, so I assume the fix for always using the full path wasn't applied to the quick open window. Closing and reopening Oni fixes this for QuickOpen.

bryphe:
Thanks for testing this out, @CrossR , and finding a repro. Unfortunately, it doesn't seem to reproduce for me on my following those steps.

A couple of troubleshooting questions after the 4th step:
- What's the value of `process.cwd()` in the dev tools?
- What's the value of `Oni.workspace.activeWorkspace` in the dev tools?
- What's the value of `Oni.editors.activeEditor.neovim.request("nvim_command_output", [":pwd"]).then((out) => console.log(out))` run in the dev tools?

It seems like the path is getting out of sync somehow. When we set the workspace path in step 3, i'd expect it to be set for both Neovim and the workspace. That assumption might not always be true if users have different configurations for managing the active neovim directory - if that's the case, we might be safer figuring out the full path from the 'QuickOpen' experience and sending that, as opposed to assuming neovim is always in sync. But knowing the values of those bullets above can help us see where it is breaking down.

CrossR:
If I repeat the steps listed, the values are as follows:
* `F:\User Files\My Documents\Git\oni`
* `"F:\User Files\My Documents\Git\oni"`
* `C:\Program Files (x86)\Oni`

If I then close and re-open Oni, where the QuickOpen works again I get `"F:\User Files\My Documents\Git\oni"` for all 3.

I guess from that it looks like Oni is updating the location it uses, but Neovim isn't being updated since `:pwd` isn't correct.

CrossR:
This has changed for me, I now can't seem to open anything from QuickOpen as all paths are now relative, even after closing and reopening Oni.

Running the tests mentioned above gives the following:
-`F:\User Files\My Documents\Git\oni`
-`"F:\User Files\My Documents\Git\oni"`
-`C:\Program Files (x86)\Oni`

[...]

PR #1643, opened 2018-02-26
feature/ change vim directory if oni workspace changes

Akin909:
This PR relates to #1575 and #1397, at present if `Oni` updates its workspace directory nvim's directory remains unchanged meaning a user might attempt to execute an nvim command with the expectation that they are in the directory `oni` reports but actually they are in different nvim dir as Oni and nvim are out of sync.

The reverse currently does happen already aka. if vim switches its dir Oni's workspace switches its directory to match.


bryphe:
Looks good to me! Thanks @Akin909 👍

PR #1658, opened 2018-02-27
Dependency: Upgrade to oni-api 0.0.36

bryphe:
- Strongly type arguments to `openFile`


bryphe:
Needed one more API update

Issue #1626, opened 2018-02-24
How to run locally directly from the repo?


trusktr:
It's not very clear.

I tried initing sub modules, then `yarn install && npm run build && npm start` which opens the window, but the windows tells me to wait for Webpack and that I may need to restart it.

I keep trying to restart it but that message never goes away.

What's the official steps to run the editor from the git repo?

trusktr:
I'm on macOS by the way.

trusktr:
For now, I am able to use the `.app` for macOS, but still want to know how to start manually.

CrossR:
They would be the steps I'd use to get it running, just using `npm run start` instead which I think does the same thing.

It'll sit on the waiting for webpack for a second or so, but then moves on and refreshes.

I know some users had issues where we weren't refreshing after the compilation had finished, but I thought it had been sorted.

There is an issue here about it : #1345.
It might be worth trying the fix in there, and if that works, we can look at sorting that issue. (Though now that I'm reading that ticket again I realise @someguynamedmatt is having the issue still)

someguynamedmatt:
@CrossR yep, I’m still seeing that too. I tried to work with a fix which included `window.location.relod()` when the webpack bundle was loaded, by that didn’t totally work.

someguynamedmatt:
@trusktr Yeah, refresh the electron instance (like you would with”normal” Chrome) on Mac cmd-r

trusktr:
`npm start` by itself threw an error. Then I tried `npm run build && npm start` which got me further (to the state that tells me to refresh above, though refreshing leads me back to state that tells me to refresh, and refreshing never works).

trusktr:
On commit d7b1c4fd (which passes all CI checks), I tried `yarn install && yarn start` which results in the following output with two errors:

```
[1] Project is running at http://localhost:8191/
[1] webpack output is served from http://localhost:8191/
[0] 2018-02-26 20:07:23.555 Electron Helper[72054:2682633] Couldn't set selectedTextBackgroundColor from default ()
[1] ts-loader: Using typescript@2.7.1 and /Users/joe.pea/src/oni/browser/tsconfig.json
[2] 8:07:25 PM - Compilation complete. Watching for file changes.
[2]
[2]
[1] Hash: a9eb5d6420497f2c5421
[1] Version: webpack 3.5.3
[1] Time: 18103ms
[1]        Asset     Size                                                                          Chunks                    Chunk Names
[...]

PR #1663, opened 2018-02-28
Fix #1345 - Improve developer experience when starting from webpack

### LINKED ISSUE #1345: Development: `npm run start` requires an initial reload
To iterate quickly, we use `npm run start` which creates several 'watchers', spins up a second instance of Oni, and reloads that instance whenever source files change - live. This is great for seeing the result of changes quickly.

However, there is a bug on Linux/OSX - on the first load, the developer needs to _explicitly_ reload the page once compiliation finishes, ie `window.location.reload()` in the console.

It'd be better if the logic in our `index.html` is more robust - we have this code here to pick up the webpack-dev-server bundle:
```
    if(process.env.ONI_WEBPACK_LOAD) {
        path = "http://localhost:8191/bundle.js";
        const hotReloadElement = document.createElement("div");
        hotReloadElement.textContent = "Waiting for webpack compilation... You may need to reload the browser once webpack finishes compiling."
        document.body.appendChild(hotReloadElement)
    }
```

But it only tries once to load the bundle. It'd be better if we retried or listened for the load to be complete, and then reload. The result should be that we need no additional interaction past running `npm run start` to start seeing live changes.

bryphe:
A few fixes:
- Refactored some of the launch logic out to a bootstrap script, with a loading animation
- Ping the `webpack-dev-server` periodically, and once we have the script, ensure that the UI is starting up

In addition, this adds an alternative mechanism for debugging:
- `npm run build-debug`
- `npm run launch`

This doesn't give the hot-reload functionality, so you'd have to run `npm run build-debug` after each change, but it's nice to have an alternative in case there are webpack issues..


bryphe:
Should fix #1345 and #1626

PR #1665, opened 2018-02-28
Bugfix: Fix heisenbug where process.env changes during execution

bryphe:
__Issue:__ While debugging automation, I noticed cases where the `process.env` would be changing seemingly at random through the course of running Oni.

__Defect:__ The `mergeSpawnOptions` method has a side effect of setting `process.env`. We shouldn't need to update `process.env` during runtime, and doing so can have unpredictable results (aka, heisenbugs)

__Fix:__ Don't set `process.env` - instead, just pass it, along with other environment variables, to the child process.

### DIFF
--- a/browser/src/Plugins/Api/Process.ts
+++ b/browser/src/Plugins/Api/Process.ts
@@ -114,7 +114,7 @@ export class Process implements Oni.Process {
         const requiredOptions = {
             env: {
                 ...process.env,
-                this._env,
+                ...this._env,
                 ...originalSpawnOptions.env,
             },
         }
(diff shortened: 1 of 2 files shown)


bryphe:
@Akin909 - do you know if there was any code that depended on setting `process.env` based on shell arguments? My assumption was that we'd only need to pass it to other processes we spawn (like neovim, lsps, etc)

Akin909:
@bryphe I don't think anything depends on process.env being set, when I initially added that I assumed that function related exclusively to startup. I think as long as Oni still picks up the users env vars on startup thats all that matters.

bryphe:
Cool, thanks @Akin909 !

PR #1666, opened 2018-02-28
Learning Pane: Initial learning pane sidebar + experimental configuration setting

bryphe:
This just sets up some scaffolding / placeholder for the learning pane. Right now, this is gated by `experimental.learning.enabled`, and is off by default.


browser/src/Services/Configuration/DefaultConfiguration.ts
```diff
@@ -47,6 +47,7 @@ const BaseConfiguration: IConfigurationValues = {
     "wildmenu.mode": true,
     "commandline.mode": true,
     "commandline.icons": true,
+    "experimental.learning.enabled": true,
```
CrossR: The PR implies this should be off by default, so should this be true?

Issue #1627, opened 2018-02-24
Pleeease, show actual tabs, not buffer tabs.


trusktr:
The main thing I displike about Vim emulation in editors like Atom or Sublime is that they don't group window splits inside tabs.

Atom, for example, groups tabs inside of splits, which is completely a bad workflow that I like to avoid by using Vim. See this issue: https://github.com/atom/atom/issues/672

But, I noticed that in Oni vim, when I try to make a `:tabnew`, it doesn't show it in the tabs. 😢 

And, if I try splitting windows, then click on Oni's tabs, it will change the buffer of the active split. 😢 

This is moving away from Vim-style tabs and going towards Atom-style or Sublime-style tabs, which is 😢 .

It would be great for tabs to represent actual Vim/Neovim tabs.

Is there a setting for this?

 I really admire the project, but I can't live without the splits-inside-tabs workflow.

CrossR:
Hi, if you set `"tabs.mode": "tabs",` in your `config.js` you'll get back Vim style tabs.

If they aren't good enough, there is also    `"tabs.mode": "native",`, which is vim tabs as is, with no externalisation.

But as far as I know, setting `"tabs"` should be the same, just with Oni's prettier tab bar.

(ps to edit the `config.js` its in the `<Meta+Shift+P>` window under `User Config`)

badosu:
@trusktr Please see this page to make Oni behave more like Vim: https://github.com/onivim/oni/wiki/How-To:-Make-Oni-closer-to-bare-Vim-experience

trusktr:
> `"tabs.mode": "tabs"`

Cool, that's much better!

Seems like this should be the default, if Oni's goal is to make the transition from (neo)vim seamless. Would be great if `:tabnew` just worked as expected out of the box (and the docs explain the `tab` options well to make it flexible).

> there is also `"tabs.mode": "native",`

I tried this one too, but even if I `set showtabline=2`, the native tabline never shows up. Do I have to do something else?

CrossR:
>Seems like this should be the default, if Oni's goal is to make the transition from (neo)vim seamless. Would be great if :tabnew just worked as expected out of the box (and the docs explain the tab options well to make it flexible).

Thats where it gets a bit awkward, since Oni wants to get new people into Vim as well as giving a UI for those that want a vim experience without the limits of a terminal UI.

[...]

Issue #1563, opened 2018-02-14
When using `tabs.mode: "tabs"` the files from the explorer open in the same tab


magopian:
This is a follow-up of #1557, exploring ways to use tabs in oni, and close them.

The problem occurs when opening a file from the explorer: it opens it in the same tab as the one already active, instead of opening it in another tab.

bryphe:
Good point - I wonder if a better default, for both the file finder and explorer, is to open in a new tab?  (In the fuzzy-file finder, we have a `Control+T` shortcut for opening in a new tab)

magopian:
At least for my usage, it would definitely make sense. It seems I have no way to open a file in a new tab from the explorer. But then I also know I don't have a conventional use of tabs/buffers in the vim world, so please take that with a grain of salt.

bryphe:
This is fixed now with #1667 (at least, the default is now to open in a new tab)

Issue #1174, opened 2017-12-21
Default neovim tabs behaving weirdly


Bretley:
I'm not entirely sure what's wrong here but doing :tabnew then :tabprev hides the newly created tab... I'm going to try and hunt down some init.vim fixes

jordwalke:
I believe I’ve created an issue for this. Away from keyboard at the moment. Oni encourages use of buffers but I agree it would be nice if this tabs bug was fixed. I’m a heavy tabs user, so it’s currently blocking my use case. I’m not clear what the benefit of being buffer centric is.  The downside of tabs is that it makes it harder to implement docked panels - but Oni implements those itself anyways so I don’t know what the remaining trade offs are.

CrossR:
>I’m not clear what the benefit of being buffer centric is.

To me at least, I find being buffer centric more new user-friendly, in most cases.
There were times when I'd open `vim file1 file2 file3` and then be confused where the 3 files had gone, and more confused when I tried to exit and it complained that I had more files to edit.

Being buffer centric means that situation, and any plugins etc that mess and open up buffers, or new users that open up buffers accidentaly etc, can more easily swap back.

I might be biased since I took forever to get my head around the concepts of buffers.

That said... setting `tabs.mode` to `"tabs"`, should swap back to the Vim style tabs in Oni UI. (`"native"` gives the native Vim tabs and Vim tab UI if needed and there is an issue).

bryphe:
Thanks @Bretley for logging the issue!

> I'm not entirely sure what's wrong here but doing :tabnew then :tabprev hides the newly created tab... I'm going to try and hunt down some init.vim fixes

For this particular issue, as @CrossR mentioned, should work correctly with `'tabs.mode': 'tabs'`. At least it seemed to work OK for me:

![tabs](https://user-images.githubusercontent.com/13532591/34283701-2a7c81bc-e682-11e7-90f2-48e48dceeba6.gif)

Let me know if you're doing / seeing something different, @Bretley !

> I’m not clear what the benefit of being buffer centric is.

Yes, @CrossR described this well - buffers-per-tab is a similiar model to other editors like VSCode/Atom/Sublime. They don't really have a concept that maps to 'vim tabs'. IMO, 'tabs' in Vim are somewhat poorly named... they should be 'window layouts' or something (at least from my perspective)
 😄 

We had a few discussions in the past around this, some of these conversations may be relevant:
- #593 
- #586 
- #949 
- Also a bit in the PR here #922 

[...]

PR #1667, opened 2018-02-28
Switch to 'tabs.mode': 'tabs' by default

bryphe:
This changes the default value of `'tabs.mode'`: `'tabs'`. This is (hopefully) the culmination of several discussions around whether to default to `buffers` vs `tabs`:
- #1174
- #1563 
- #1627 

The main concern I had was in creating an experience that is familiar to users coming from Atom / VSCode - where there is a 1:1 tabs to buffers, so I defaulted that setting to buffers. What I've found though is that this leads to very confusing behavior for existing Vim users, and also that it may not be mutually exclusive. I believe that we can satisfy the 'users coming from Atom/VSCode/Sublime' case in 'tabs' mode, with minor tweaks, by just keeping 1 tab === 1 buffer by default from the QuickOpen/Explorer. 

The main changes are:
- Switch to tabs.mode setting of 'tabs' by default (and potentially deprecate the 'buffers' setting, if possible)
- Set up the QuickOpen selection behavior to open in a _new tab_ if the file is not already open, or an _existing tab_ if it is already open. Same for file explorer. This is different than the current default, which was using `:e` - opening the buffer in the existing tab. I'm interested if this will impact / break anyones workflow. `Control+T` in the QuickOpen selection behaves as it does today, always opening a new tab. 

In addition, @CrossR  has unblocked this by giving our 'tabs' mode parity in functionality with 'buffers' mode in #1498 - thanks @CrossR ! 💯

Given all the feedback around the 'buffers' default, I think pivoting and trying out the 'tabs' default makes sense for the next release - cc  @jordwalke  @badosu@hoschi, as I know you all were impacted by this. The goal here is to have a system that is intuitive for both vim users and new users, without sacrificing power or flexibility 👍


badosu:
I like this change, I think that if we have an editor capable of potentiating the VIM experience and deliver it with quality we'll be able to safely be able to experiment new ideas.

Akin909:
@bryphe when you say potentially deprecate the `buffers` setting does this mean by default all users will have to work with `tabs`, definitely not wanting to restart what have been many discussions, but it seems like anyone with a primarily buffer based workflow coming from vim would only be able to use tabs? 😞

Might just wait to see what the change ends up looking like since I interact very little/never with tabs 😟

bryphe:
> when you say potentially deprecate the buffers setting does this mean by default all users will have to work with tabs

Only if people aren't using it! I wish I had telemetry - it's hard to gauge how many users had explicitly set 'tabs' vs 'buffers'. But I won't deprecate it if its useful. 

> Might just wait to see what the change ends up looking like since I interact very little/never with tabs 😟

Ya, it'll be interesting to see if it works for you, or if you have to flip back to 'buffers' mode. I heard the confusion / issues around the 'buffers' default in those linked issues - I'm curious what the response will be when it is flipped to 'tabs'.

badosu:
I know that the buffer list navigation mode is popular among vim users as well, so I think it's good to not remove that.

bryphe:
Cool, I'll include this as a 'breaking change' in the release notes for next release - we'll give it a shot and see how it works 👍

Issue #1673, opened 2018-03-01
Image layers break split opening vim plugin functionality


Akin909:
I still rely heavily on vim plugins like `CtrlSF` and `undotree` both of these work by opening splits. Since the integration of the image layers these plugins all now open with a buffer layer on top of them attempting to open an image file, my thinking is that before opening atm, the image layer likely doesn't check that the file being opened in the buffer is an image file before rendering the buffer layer

Akin909:
I noted that quite nicely the buffer layer addition mechanism takes a filter fn which can be applied to the image buffer layer to only add a buffer layer if the buffer is an image filetype

Akin909:
fixed by #1676

PR #1676, opened 2018-03-01
bugfix/ add a filter for image layers

Akin909:
Currently image layers pop up whenever a split is made, or rather I observe this using any plugins that create splits. This PR aims to fix this which I raised as #1673, using the filter functionality of the add bufferLayer method, which is very cool re api design 👍.

I added a filter to only open the image layer for `jpg` and `png` formats (the only 2 that came to mind at the time)

### DIFF
--- a/browser/src/Editor/OniEditor/OniEditor.tsx
+++ b/browser/src/Editor/OniEditor/OniEditor.tsx
@@ -146,9 +146,10 @@ export class OniEditor implements IEditor {
             wrapReactComponentWithLayer("oni.layer.errors", <ErrorsContainer />),
         )
 
-        this._neovimEditor.bufferLayers.addBufferLayer(buf => {
-            return [".png", ".jpg"].includes(path.extname(buf.filePath))
-        }, buf => new ImageBufferLayer(buf))
+        this._neovimEditor.bufferLayers.addBufferLayer(
+            buf => [".gif", ".jpg", ".jpeg", ".bmp", ".png"].includes(path.extname(buf.filePath)),
+            buf => new ImageBufferLayer(buf),
+        )
     }
 
     public dispose(): void {
--- a/browser/src/Editor/OniEditor/OniEditor.tsx
+++ b/browser/src/Editor/OniEditor/OniEditor.tsx
@@ -6,6 +6,7 @@
  * Extends the capabilities of the NeovimEditor
  */
 
+import * as path from "path"
 import * as React from "react"
 
 import * as types from "vscode-languageserver-types"
@@ -145,7 +146,9 @@ export class OniEditor implements IEditor {
             wrapReactComponentWithLayer("oni.layer.errors", <ErrorsContainer />),
         )
 
-        this._neovimEditor.bufferLayers.addBufferLayer("image", buf => new ImageBufferLayer(buf))
+        this._neovimEditor.bufferLayers.addBufferLayer(buf => {
+            return [".png", ".jpg"].includes(path.extname(buf.filePath))
+        }, buf => new ImageBufferLayer(buf))
     }
 
     public dispose(): void {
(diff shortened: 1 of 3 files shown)


browser/src/Editor/OniEditor/OniEditor.tsx
```diff
@@ -145,7 +146,9 @@ export class OniEditor implements IEditor {
             wrapReactComponentWithLayer("oni.layer.errors", <ErrorsContainer />),
         )
 
-        this._neovimEditor.bufferLayers.addBufferLayer("image", buf => new ImageBufferLayer(buf))
+        this._neovimEditor.bufferLayers.addBufferLayer(buf => {
+            return [".png", ".jpg"].includes(path.extname(buf.filePath))
```
CrossR: If possible, it may be worth using the file types that are stored in the config : https://github.com/onivim/oni/blob/master/extensions/images/package.json

Then we'd only have 1 version of the truth, but I also don't know how easy it is to grab them values.

Akin909: @CrossR I'm not sure about pulling values out of one of the many packages.json in the oni editor but can copy the list across although one thing I notice is that it includes svgs which tbh I work with on occasion and I think unlike binary ones like png etc. svg is similar to xml so these should be editable in the editor not sure if the layer is toggleable but adding svgs there would make them as things currently are not possible to work with.

Akin909:
Actually thinking about it rather than having the hard coded list there it might be useful to some users to have the list of files the image layer opened for editable or (*not for this PR* a means of toggling it on or off) the reason I suggest this is that in one scenario I know of a user might want to edit the raw file, or process it in some way, this can definitely be the case with something like `svg` but may also be the case for some other file format for some other reason

bryphe:
> Actually thinking about it rather than having a list defined by oni it might be useful to some users to have the list of files the image layer opened for editable aka via their config

Yes, that would be cool!

> (not for this PR a means of toggling it on or off) the reason I suggest this is that in one scenario I know of a user might want to edit the raw file, or process it in some way, this can definitely be the case with something like svg but may also be the case for some other file format for some other reason

Definitely - this is actually where I want to go with the 'buffer layers' feature. It'd be cool if there was a UI  (+ settings) that would let you toggle layers on/off, disable layers, etc. The reason I included `friendlyName` as part of the `BufferLayer` interface was to support this scenario.

My hope is that we could build some really interesting layers on top of this - like a debug layer, code coverage layer, a VCS layer (showing changes), etc...

browser/src/Editor/OniEditor/OniEditor.tsx
```diff
@@ -145,7 +146,10 @@ export class OniEditor implements IEditor {
             wrapReactComponentWithLayer("oni.layer.errors", <ErrorsContainer />),
         )
 
-        this._neovimEditor.bufferLayers.addBufferLayer("image", buf => new ImageBufferLayer(buf))
+        this._neovimEditor.bufferLayers.addBufferLayer(
+            buf => [".gif", ".jpg", ".jpeg", ".bmp", ".png"].includes(path.extname(buf.filePath)),
```
bryphe: Could we make this a configuration setting? Maybe `editor.imageLayerExtensions`? That way, users can always override it or add new extensions.

browser/src/Editor/OniEditor/OniEditor.tsx
bryphe: I'd be OK ditching that extension for now and supplying this via a `config` setting, too. The only purpose that extension is set to the filetype, though, so if we go that route, we'd want to remove that extension also.

browser/src/Editor/OniEditor/OniEditor.tsx
Akin909: Just pushed the change 👍

Akin909:
@bryphe added in the config setting for this now 👍

bryphe:
Thanks @Akin909 !

PR #1677, opened 2018-03-01
feature/ add a check to open in a tab if in tabs mode otherwise open buffer

### LINKED ISSUE #1675: New file not opening using `tab drop` for buffers based workflow
@bryphe seems with the switch to tabs and using tab commands new files don't open correctly if working via a buffer based workflow aka not using tabs at all, hitting open command in the quick open does nothing.

~~> Trying to configure Vim to always have one file per tab will not be successful, and would remove much of the power of Vim. However, when you want to edit a file, it is easy to use :tabe instead of :e so that usually there is one file per tab. You can also launch files in new tabs under Windows and Unix.~~

[full link](http://vim.wikia.com/wiki/Using_tab_pages)

~~I haven't looked into all the discussions re using tabs but it seems the intent to switch to tabs is to use them as a one to one mapping with buffers.~~

~~I really don't want to bog us down with another long thread on tabs v. buffers but if I am not understanding things incorrectly we may be trying to use vim tabs in a way they aren't designed to be used~~

~~> The easiest way to think about tab pages in Vim is to consider them to be viewports, layouts, or workspaces.~~

~~The above explanation from vim wiki is what i've always understood about tabs aka trying to map a single buffer to a tab is contrary to their purpose. Anyway for fear of doing what I set out not to do aka cause a `tabs vs buffers` thread I'm not sure why tab drop doesn't work (dont know v. much about using tabs).~~

Would you be up for using a different set of quick open commands depending on mode rather than defaulting to opening things in tabs, via the quick open as even if I have things set to buffers mode re tabline if things are opening in tabs then regardless of that setting I'm actually primarily using tabs 😟, which is a workflow I'm largely unfamiliar with 😭

Akin909:
Relates and hopefully fixes #1675, I added a check to open with a regular edit command if a user is using buffers mode otherwise it uses the `tabedit` command which for all I know is probably not the desired effect, the `tab drop` command does nothing for me locally though I wasn't using tabs so not sure if it would have been working for tab users.

The main aim of this PR is to have the quick open open files depending on the users `tabs.mode` settings, happy to revert the tab drop change as it likely does exactly what was intended but seemed to do nothing locally so attempted a fix for that ^^

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -751,7 +751,7 @@ export class NeovimEditor extends Editor implements IEditor {
                 [Oni.FileOpenMode.NewTab]: "tabnew!",
                 [Oni.FileOpenMode.HorizontalSplit]: "sp!",
                 [Oni.FileOpenMode.VerticalSplit]: "vsp!",
-                [Oni.FileOpenMode.Edit]: tabsMode ? "tabe!" : "e!",
+                [Oni.FileOpenMode.Edit]: tabsMode ? "tab drop" : "e!",
                 [Oni.FileOpenMode.ExistingTab]: "e!",
             },
             {
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -745,12 +745,13 @@ export class NeovimEditor extends Editor implements IEditor {
         file: string,
         openOptions: Oni.FileOpenOptions = Oni.DefaultFileOpenOptions,
     ): Promise<Oni.Buffer> {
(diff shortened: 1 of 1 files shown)


CrossR:
From my brief understanding of `tab drop` it seems to be working, so may be worth leaving in.

Ie : 
```
:e LICENSE
:vsplit oni.sh
:split package.json
:tabe appveyor.yml
```

Calling `:tabe package.json` opens a new tab, calling `:tab drop package.json` swaps me back over to the "workspace" with my 3 splits in.

Akin909:
@CrossR thanks for the explanation had no idea what it did, happy to revert it since it works as it should

Akin909:
Reverted 👍 just checks tabs mode now

bryphe:
Cool, the fix looks good to me! Thanks @Akin909 and @CrossR  👍

PR #1682, opened 2018-03-01
Externalized UI: Implement Oni split mode

bryphe:
This implements basic split rendering for Oni (this isn't on by default, is only available with `oni.splits.mode`: `'oni'`, and there isn't an entry point to hit this other than calling directly in the console)

But it lets you do crazy stuff like this:
image

I don't even know what is going on there 😄


bryphe:
- Finish wiring up commands for splits from window motions
- Fix bug with split direction (44c1e4e8)
- Fix bottom margin

bryphe:
I'll bring this in for now, but keeping the configuration setting off (`editor.split.mode` needs to be set to `'oni'` to enable this functionality). There are still lots of bugs / issues to be fixed before this becomes useful (like you can't close a split right now without #1703), and we don't properly sync the `shada` state between `nvim` instances, which means stuff like marks and yanks don't get persisted between splits, which would be very confusing.

PR #1687, opened 2018-03-02
Fix #1681 - Paths with spaces

### LINKED ISSUE #1681: Unable to open files from paths with spaces on Windows
Sometime between 5add3ff2d4a8e6379755bb98ff0d276040fce795 and 6bd62431333f8830ea073d879fb1afa7589ea827, the ability to open a file from a path that contains spaces on Windows stopped working. I now get 3 different tabs, with 2 out of the three being picked up as images too.

image

for the file path `"F:\User Files\My Documents\Git\oni\README.md"`

I get the following in the console on selection: `[NeovimInstance] Executing command: :tab drop F:\User Files\My Documents\Git\oni\README.md`.

This behaviour is the same for QuickOpen and the explorer, but File > Open still works.

After/part of a fix we should probably add a CI test for this.

bryphe:
Looks like the move to change the way we open files caused a regression in paths with spaces - thanks @CrossR for catching it!

This fixes the issue by escaping spaces when we send the vim command for opening files, and adds a ci test to verify that we don't break this functionality down the road.

In addition, there was another problem - we weren't getting information on the buffer loaded via `tab drop` (which made it hard to verify in the test!), so this required a change to `init.vim` to make sure we're listening to the proper event (`BufWinEnter`).

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -403,6 +403,9 @@ export class NeovimEditor extends Editor implements IEditor {
         this._neovimInstance.autoCommands.onBufEnter.subscribe((evt: BufferEventContext) =>
             this._onBufEnter(evt),
         )
+        this._neovimInstance.autoCommands.onBufWinEnter.subscribe((evt: BufferEventContext) =>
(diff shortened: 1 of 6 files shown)


bryphe:
Fixes #1681

PR #1688, opened 2018-03-02
Fix case where image layer was added to files w/o extension

bryphe:
__Issue:__ Our image layers were being added to files without an extension

__Defect:__ Our 'language contribution' was overly aggressive, and would assign any file without an extension that first 'contributed' language it could find.

__Fix:__ Only apply these contributions to files with extensions


CrossR:
I've noticed the Welcome screen flashes up the image layer when opening Oni, I assume this will fix that?

bryphe:
@CrossR - yes, should fix that too 👍

PR #1697, opened 2018-03-02
bugfix/ check if linecontent before executing get token at

Akin909:
The `getTokenAt` function in the `bufferManager` fails at the substring portion of the function, despite a check to see if the value is falsy. This fails I think because the value is truthy but not a string somehow 😕, I can't find where the value being passed in is incorrect as I cant reliably reproduce this error though its usually on startup, to guard against this I added a check to see if the `substring` exists on the `lineContents` var passed aka if its a string, if not the fn returns null.

Note the function's parameters are all typed an there are no type warnings so it is seemingly always being passed a string but somehow the value it is passed can fail to have a `.substring` method 😕


bryphe:
> This fails I think because the value is truthy but not a string somehow 😕

Hmm, can we track this is down? I'm worried this could cause issues elsewhere. Do you happen to have a callstack or a repro?

Akin909:
@bryphe been looking at this bug for a couple of days since it usually happens on startup and kills the hover and definition but not consistently and its hard to trace the origin from the callstack as the code is minified so hard to tell where its coming only realised it was getTokenAt because the error occurs often enough and I recognised the line despite its being garbled

bryphe:
If it happens frequently on startup, it might be worth building with `npm run build-debug` and seeing if you can repro it - that builds an un-minified script so you can get better callstacks 👍

Akin909:
@bryphe it's definitely the desperate rather than the *proper* fix I've had it on and off for a while now, have to restart oni to get lsp features back once it occurs, so since it was such a pain I gave up and added that haven't been able to find the root cause, think motivation wise its got the best of me since I cant seem to track it down, on the other hand this is definitely not an ideal fix, maybe we can leave this here for a little bit and I can take a look again in a couple of days

Akin909:
@bryphe seemed to have "lucked out" and run into a few of these in a row, seems that the `this.getLines()` function calls the `nvim.getLines` which apparently can return `string[]` (correct) or an array that looks like this `[1, "invalid buffer id"]`, so the value being passed into the linecontents functions is both truthy and also not a string

Akin909:
The reason the `getLines()` function is failing is that the bufferId its passing to neovim is invalid i.e. in this case neovim is passed a `bufferId` of 1 but the buffer checking `blist` or `Oni.editors.activeEditor.getBuffers()` returns the correct number which is 4(not relevant), not sure how or why oni has a reference to the wrong id

Akin909:
A tweak I can make to this PR is check that the result of `getLines` doesn't include a number if it does can throw and error or less dramatically return null from that function instead?

bryphe:
Thanks for drilling into the issue, @Akin909 ! Ya, seems like a couple of issues

> A tweak I can make to this PR is check that the result of getLines doesn't include a number if it does can throw and error or less dramatically return null from that function instead?

Throwing seems reasonable in this case. It's definitely misleading that we're returning that array!

> The reason the getLines() function is failing is that the bufferId its passing to neovim is invalid i.e. in this case neovim is passed a bufferId of 1 but the buffer checking blist or Oni.editors.activeEditor.getBuffers() returns the correct number which is 4(not relevant), not sure how or why oni has a reference to the wrong id

Hmm ya I wonder where it is getting the `1` from?

Akin909:
@bryphe the one is coming from the `startify` buffer that my plugin opens when `oni` loads basically the `bufferToId` object seems to hang on to a never ending list of buffers (not currently relevant but I think potentially a future issue is that, the `bufferList` oni uses doesn't refresh when buffers change aka if the set of open buffers change oni doesn't remove them from the list, I haven't noted this to be an issue but it does mean in terms of the object oni checks for buffers it has a lot stale info. it isnt an issue atm as buffer ids are unique thankfully.)

None of this unfortunately clarifies why the bufferId being passed in is the startify buffer id. Flow to reproduce (for me) open oni, startify window pops up (keep in mind that blist doesn't actually register this buffer), enter a session then error can occur my thinking is that somehow the current buffer hasn't switched before the call to getTokenAt and getLines is made like a race condition and since neovim itself seems to have no record of the startify buffer *somehow* it throws an error as invalid buffer

Akin909:
From the interactions I've had with the neovim api it seems that it follows the pattern of returning `1` as an exit code to indicate failure generally, if this is generally the pattern, I'm not certain that it is, I'm wondering given its a point of uncertainty whether it might be a good idea to wrap at least certain commands in the neovim instance, so that rather than call them directly you call a method which call the `request` checks if the value has a "failure code" and returns `null` sort of formatting the correct response so there isnt loads of adhoc logic to guard against the api's return values.

A potential pretty big issue with this would be if some valid response happened to begin with 1, in which case this wrapper method would think an error had occured :S, although from what light to moderate reading of the api docs I think it theres a pattern to all responses 😟

Akin909:
Looking into `type guards` in typescript I've added types to the request method and a check function using that which I hope gives better type safety, seemed whilst adding it that it could potentially be generalised to other api methods based on what their fail responses return, although this would probably be a lot of work but could be gradual having neovim api types which we pull from oni's api and update if methods change.

bryphe:
> If this is generally the pattern, I'm not certain that it is, I'm wondering given its a point of uncertainty whether it might be a good idea to wrap at least certain commands in the neovim instance, so that rather than call them directly you call a method which calls the request checks if the value has a "failure code" and returns null sort of formatting the correct response so there isn't loads of adhoc logic to guard against the api's return values.

Yep, it'd be nice to push this to a more common layer. It'd be great if we could push this into the `Session` layer where we directly handle the msgpack interaction with Neovim:
https://github.com/onivim/oni/blob/0307da1691a1c1cede5219c63bf0f10a93231783/browser/src/neovim/Session.ts#L87

browser/src/Editor/BufferManager.ts
```diff
@@ -148,13 +151,24 @@ export class Buffer implements IBuffer {
...
             start,
             end,
             false,
         ])
-        return lines
+
+        // Neovim does not error if it is unable to get lines instead it returns an array
+        // of type [1, "an error message"], we only check the first on the assumption that
+        // that is where the number is placed by neovim
+        const isStringArray = (array: MaybeStringArray): array is string[] =>
```
bryphe: I wonder if, as you suggested, it makes sense to move this to a shared layer?

We could move this logic to `Session.ts` where we handle getting the `response` from Neovim:
https://github.com/onivim/oni/blob/0307da1691a1c1cede5219c63bf0f10a93231783/browser/src/neovim/Session.ts#L56

And check if we have an array that is:
- 2 items
- First item is a number
- Second item is a string
And in that case, we dispatch an error.

Otherwise, we'll have all these handlers having to implement their own logic for handling errors

Akin909:
@bryphe went for a more conservative approach rather than assuming all errors return the same result I'm about 80% certain but cant find written documentation to back me up so I went for a method which I hope we can gradually expand upon i.e the expected type is passed in at the call site goes to `Session.ts` where I check if the array is one of a particular type all the while passing the specified type argument down.

Atm I just check the `nvim_get_buf_lines` function returns what is expected otherwise I reject the promise. Lemme know what you think.

Akin909:
Just read your comment a bit more closely 😆 did you mean to but some sort of check in the switch statement itself i.e. put a check in there and *? return null instead?* of the result if the length is two and the first letter is a number, happy to do this if you'd like I got a bit skittish and went for a more conservative solution since I worry about what might break and how if there are some unknown exceptions to this rule

browser/src/neovim/Session.ts
```diff
@@ -80,11 +92,26 @@ export class Session extends EventEmitter {
         })
     }
 
+    // Neovim does not error if it is unable to get lines instead it returns an array
+    // of type [1, "an error message"] **on Some occasions**, we only check the first on the assumption that
+    // that is where the number is placed by neovim
+    public isNeovimError = <T>(val: T, methodName: string) => {
+        switch (methodName) {
+            case "nvim_buf_get_lines":
```
bryphe: It doesn't really make sense to need to check this here. The `Session` layer should be agnostic of the specific methods available on nvim (you could think of it is a _general_ msgpack Session, not just for Neovim).

I was hoping that, at either this layer or `NeovimInstance`, we could find a _general_ way to detect errors. If that's not the case, it'd be better to have the error handling local in the `getLines` method as we did before.

Akin909:
@bryphe I've reverted changes here back to just the `getLines` function. I logged a few different responses back in `Session.ts` it seems that although there does seem to be a pattern of returning `[1, "Error Message"]` it doesn't seem to be universal I think there must be some way of checking generally for errors but I think putting any one size fits all solution in Session.ts is gonna cause issues, tbh the return values are soooo varied.

*suggestion re future issue not for this pr*
Not sure what you think about it but if you'd understandeably rather keep `Session.ts` agnostic then I wonder if having a class/service  which had specific knowledge of the correct format of neovim responses based on the request which could be imported into session.ts and could filter responses e.g. `NeovimErrorDetector/Handler.checkResponse(res)` aka move specific method checking into a unified location and have it check all responses and error if the format is incorrect.

bryphe:
> it seems that although there does seem to be a pattern of returning [1, "Error Message"] it doesn't seem to be universal I think there must be some way of checking generally for errors but I think putting any one size fits all solution in Session.ts is gonna cause issues, tbh the return values are soooo varied.

Sure, let's special case it for now then - hopefully we can find a way to generalize it later..

> suggestion re future issue not for this pr
Not sure what you think about it but if you'd understandeably rather keep Session.ts agnostic then I wonder if having a class/service which had specific knowledge of the correct format of neovim responses based on the request which could be imported into session.ts and could filter responses e.g. NeovimErrorDetector/Handler.checkResponse(res) aka move specific method checking into a unified location and have it check all responses and error if the format is incorrect.****

Hmm, yes, perhaps in `NeovimInstance`? Which has an instance of `Session` but it's definitely coupled to the neovim request / response format - so that might be a good place to generalize error handling.

PR #1701, opened 2018-03-02
Bugfix/ incorrect use of dom purify which was breaking <Components />

Akin909:
Whilst adding syntax highlighting I added in `DOMPurify` for another measure of safety on top of `marked's` inbuilt sanitization. The implementation however causes safe `html/xml/jsx` to be omitted from documention blocks resulting in docs with empty codeblocks and completely remove all titles from `vue` files hover info.

I've tweaked the way `DOMPurify` is called to preserve non-offensive html-esque syntax. 

Note: for future reference I've discovered an issue on *one* occasion with what the `lsp` sends, whilst the `lsp` dictates that documentation should be sent as markup/markdown in the case of `Observable` documentation html is sent, marked does not convert this to markdown so the whilst it sanitizes it and then in our cases it also passes through dom purify it is left in the hover docs, a potential solution would be to run a check for html after purification and run an html to markdown parser but I think this would only be worth while if it occurs often as I believe the documentation writers for `Observable` are ?in error?

EDIT: checked the same documentation in `vscode` and it also renders the raw html, which makes me think that the `Observable` docs arent written according to spec

### DIFF
--- a/browser/test/MarkdownTests.ts
+++ b/browser/test/MarkdownTests.ts
@@ -1,4 +1,5 @@
 import * as assert from "assert"
+import { unescape } from "lodash"
 import { StackElement } from "vscode-textmate"
 import * as markdown from "./../../browser/src/Editor/NeovimEditor/markdown"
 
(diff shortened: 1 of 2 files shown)


bryphe:
Thanks for catching this, @Akin909 ! 

I wonder if we could add a minimal unit test for this case in `convertMarkdown` - is there a small markdown string we could pass into exercise this case, and verify it is getting rendering correctly?

> Note: for future reference I've discovered an issue on one occasion with what the lsp sends, whilst the lsp dictates that documentation should be sent as markup/markdown in the case of Observable documentation html is sent

One tricky thing about the LSP implementations is they don't always adhere to the spec, so we have to be a bit flexible. I do believe though that some markdown specs allow HTML - https://github.github.com/gfm/ allows for some HTML.

browser/test/MarkdownTests.ts
```diff
@@ -88,4 +89,18 @@ describe("Markdown Conversion Functions", () => {
 
         assert.ok(tag.replace("\n", "").trim() === expected)
     })
+
+    it("should render safe jsx untouched ", () => {
```
bryphe: Sweet! Thanks for adding these 👍

Akin909:
Hmm not sure the test failure here is related to what I've changed 😿, just pushed again to rerun the test but they seem to be failing for different reasons or rather this time the CI test all timed out

PR #1705, opened 2018-03-03
Feature/change token format

Akin909:
Following discussions in #1443 to match oni's `editor.tokenColors` to match with `vscode`'s syntax to allow for cross compatibility with `vscode` themes this PR changes the theme format to match as well as *aims* to merge the user, default and theme tokens with priority and as a test adds the editor tokens from `vscode/onedark.theme.json`

Made a lot of progress here and this PR now uses a `TokenScorer` to determine which of the available scopes in a list should render for a symbol in a line, this change means that syntax highlighting nearly matches vscode/atom (not accounting for treesitter)
<div>
<h3>Oni</h3>
screen shot 2018-09-04 at 23 15 46
<h3>VSCode</h3>
screen shot 2018-09-05 at 08 17 27
</div>

Todo:
- [x] Ensure token merging works as expected.
- [x] Update theme token provider to fill in any necessary gaps as the imported theme does not cover many token scopes which the hover depends on


Akin909:
Granted tests pass, I believe this is done (for now... from reading around there are *loads* of tweaks that could be made to improve how we highlight in the long run) 


-> side note for my future self or anyone who wants to look at it in `markdown.ts`, where `symbols` are defined we dont apply any ranking logic but instead pass the *all* matching tokens to each symbols so *all* symbols have several applicable class name -> starting point for optimising the hover box at least

CrossR:
I've given this a quick look over, but is there anything specific you want a second look at?

Akin909:
@CrossR I think my main goal was to avoid breaking any of the syntax highlighting whilst allowing for the use of editor tokens from vscode themes, so that we can copy their editor tokens into our oni themes and they should match the highlighting (almost*) means with this PR anyone making a port of a theme can get really good highlighting by borrowing the tokens.

* there are quite a few quirks with how the highlighting is applied which I find hard to quantify prior to and despite this PR (I didnt alter the logic) it still occasionally disappears, or is partially applied, on sourcing vimrc as one example, one change I did make was converting the `hi oni_highlight_x guifg={ourColor}` to have a `hi!` which I believe forces the highlight

Akin909:
Will post a little message on discord anyone who uses `onedark` should see the new tokens and maybe might notice if anything is off although I think the reason for the bugs, although not at all sure, is that the highlighting is sort of cached and or rather we save the lines we have highlighted but I'm not sure if that's enough for vim as there are a lot of things that could alter or clear highlighting in vim and if we believe the line is highlighted then it won't reapply it (or at least I think this is a possible explanation for why sourcing your vimrc will clear all textmate highlighting)

Issue #1593, opened 2018-02-21
TypeScript: Hook up auto-import functionality


bryphe:
Splitting out from #1326 - integrate 'auto-import' functionality from TypeScript. This should generalize to other completion providers, too.

- Implement `additionalTextEdits` and `textEdit` for `CompletionItems` as defined by the completion request in the language server protocol: https://microsoft.github.io/language-server-protocol/specification
- Update our `TypeScriptServerHost` to send `additionalTextEdits` for auto-imports. This will involve settting the `includeExternalModuleExports` for the completions request, hooking up entry details resolution to get information about the additional code actions as part of the details request, and resolving those code actions to text edits. Looks like we'll need to plumb the 'source' through somehow, as this is needed when calling to get details.

Akin909:
@bryphe re. the `additionalTextEdits` can I clarify, from the docs you linked it seems as though there ought to be text edits returned as part of the completion items as the interface includes it as an optional key, as far as you are are should these be returned sometimes without having to change the args sent to the lsp (I've briefly logged the response and never seem to get any `textEdit` fields back in the completion responses).

bryphe:
> can I clarify, from the docs you linked it seems as though there ought to be text edits returned as part of the completion items as the interface includes it as an optional key, as far as you are are should these be returned sometimes without having to change the args sent to the lsp

Sure thing - I don't believe there are any additional capabilities we need to specify (looking at the [language server specification](https://microsoft.github.io/language-server-protocol/specification), specifically the `completion` capabilities in the `TextDopcumentClietnCapabilities`). 

> (I've briefly logged the response and never seem to get any textEdit fields back in the completion responses).

Hmm, which language server were you testing it out with? For typescript / javascript, it's expected, since we haven't wired up the ability to pass up the `additionalTextEdits` yet. So we'd have to check with the language server you're using, if it supports it.

Akin909:
@bryphe I hadn't quite got the flow correct in my head I found that where I was looking was downstream/ or upstream 🤔 in the LS but in the actual `typescript` client I found that the result is being trimmed so going to have a closer look at that, I'm just investigating a little bit for now as there are a lot of unknowns here for me as I've not seen much of this part of the code base.

bryphe:
Cool, would be great to have help here! 👍  Let me know if you have any questions.

The `typescript` standalone server actually implements its own protocol here: https://github.com/Microsoft/TypeScript/wiki/Standalone-Server-%28tsserver%29

[...]

PR #1717, opened 2018-03-06
feature/switch typescript language server to theia-ide

Akin909:
Following discussion with @bryphe in #1593 this PR switches the default `lsp` for typescript to [`theia-ide/typescript-language-server`](https://github.com/theia-ide/typescript-language-server), more details re. reasoning behind the switch can be found in the linked issue.

@bryphe I've essentially replicated the steps I took to setup the LSP and haven't touched Oni's ts lsp.

**Note**:
* Currently there are issues with how oni handles `rename` functionality which are exacerbated with integrating this LSP as I believe Oni might need to make some changes to align with the LSP implementation

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -187,6 +187,8 @@ const BaseConfiguration: IConfigurationValues = {
     "language.ocaml.languageServer.arguments": ["--stdio"],
     "language.ocaml.languageServer.configuration": ocamlAndReasonConfiguration,
 
+    "language.typescript.languageServer.command": "typescript-language-server",
+    "language.typescript.languageServer.arguments": ["--stdio"],
     "language.typescript.completionTriggerCharacters": [".", "/", "\\"],
     "language.typescript.textMateGrammar": {
         ".ts": path.join(
--- a/package.json
+++ b/package.json
@@ -217,6 +217,7 @@
         "shelljs": "0.7.7",
         "styled-components": "^2.3.0",
         "typescript": "2.7.1",
+        "typescript-language-server": "^0.1.9",
         "vscode-css-languageserver-bin": "^1.2.1",
         "vscode-html-languageserver-bin": "^1.1.0",
         "vscode-jsonrpc": "3.5.0",
@@ -225,10 +226,6 @@
         "vscode-textmate": "3.2.0"
     },
     "devDependencies": {
-        "@types/marked": "^0.3.0",
-        "@types/react-dnd": "^2.0.34",
-        "@types/react-dnd-html5-backend": "^2.1.8",
-        "@types/redux-batched-subscribe": "^0.1.2",
         "@types/classnames": "0.0.32",
         "@types/color": "2.0.0",
         "@types/detect-indent": "^5.0.0",
@@ -245,6 +242,8 @@
         "@types/mocha": "2.2.33",
         "@types/msgpack-lite": "0.1.4",
         "@types/node": "8.0.53",
+        "@types/react-dnd": "^2.0.34",
+        "@types/react-dnd-html5-backend": "^2.1.8",
         "@types/react-dom": "16.0.3",
         "@types/react-motion": "0.0.23",
         "@types/react-redux": "5.0.12",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1469,6 +1469,10 @@ combined-stream@^1.0.5, combined-stream@~1.0.5:
   dependencies:
     delayed-stream "~1.0.0"
 
+command-exists@^1.2.2:
+  version "1.2.2"
+  resolved "https://registry.yarnpkg.com/command-exists/-/command-exists-1.2.2.tgz#12819c64faf95446ec0ae07fe6cafb6eb3708b22"
+
 commander@2.11.0:
   version "2.11.0"
   resolved "https://registry.yarnpkg.com/commander/-/commander-2.11.0.tgz#157152fd1e7a6c8d98a5b715cf376df928004563"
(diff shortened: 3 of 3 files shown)


bryphe:
Awesome, thanks @Akin909 for making the branch available! I wanted to try it out, will save me some time 💯

bryphe:
Ya, looks like it works pretty well! I noticed diagnostics are missing too unfortunately - I wonder if they'd be open for a PR to implement that? The rename and go-to definition issues might be on our end - need to check it out.

Also, FYI, if you set this:
```
    "language.typescript.rootFiles": ["tsconfig.json", "package.json"],
```

It shouldn't restart the process as frequently (so should perform better when switching files in the same workspace)

bryphe:
I think the best bet for us would be to see if we could help them implement the diagnostics, which is the main gap - and then integrate this in lieu of our in-proc strategy. From there, we could help implement the 'auto-import' functionality - looks like they have an issue tracking it already:
https://github.com/theia-ide/typescript-language-server/issues/18

Akin909:
@bryphe I'm getting error markers and `tss` error and warning messages, or do you mean something different re diagnostics?
Thanks for the heads up re the root files setting will give that a go 👍

Akin909:
Not sure the root files option actually makes things faster 😞

bryphe:
> Not sure the root files option actually makes things faster 😞 

Hmm, are you still seeing the server restart when switching files after setting that? Perhaps it's a different issue.

> also re diagnostics not sure all the issues with errors have gone away although the behaviour seems more realtime unless switching file when the lsp is restarting so old errors wait till its booted back up to clear

Oh, cool! I'm glad it actually has diagnostics. I wasn't seeing them, but perhaps there is some pathing issue or something on Windows.. need to check further. I checked again in their code and it seems they are publishing diagnostics, so thats promising

If we can figure out that switching issue, seems like it's definitely worth bringing in! 👍 Then we can see if we can work with them to add the auto-import functionality

Akin909:
@bryphe re. the restarting of the `lsp` I initially looked at our client to see if it was triggering a restart however the flag `shouldServerRestart` we use doesn't seem to be malfunctioning but the lsp *is* restarting though I'm not clear as to why.

I've noted the following series of events
screen shot 2018-03-12 at 15 04 47

A `didClose` event is sent and the LSP responds by shutting down the lsp.

bryphe:
Thanks for the investigation, @Akin909 ! I'm stoked to integrate this and get out of the business of maintaining our own TypeScript layer 👍 

> A didClose event is sent and the LSP responds by shutting down the lsp.

The trace helps a lot. I think that the fault may be ours - we may be sending `didClose` too aggressively. According to the spec here: https://microsoft.github.io/language-server-protocol/overview we should be sending it when it's closed (when the editor is no longer tracking changes). However, we send it whenever we switch documents - so it might be that we need to change that behavior.

As a _temporary troubleshooting measure_, we could try commenting out this line:
https://github.com/onivim/oni/blob/7f15b1cb5a61423d148e719cd448442341cf92b4/browser/src/Services/Language/LanguageManager.ts#L65

Just to see how the language server behaves. If we find it's better after removing that, we might need to change the event (like, hook to an `onBufferClose` event versus the `onBufferLeave` event we listen to currently).

Akin909:
@bryphe thanks for the heads up was having difficulty figuring out what the significance of that sequence of events was, will give commenting that section out a shot

Akin909:
Another issue which I think my be contributing to lagging or the server functionality cutting out is an error which seems to occur on trying to getDefinition

screen shot 2018-03-15 at 12 22 49

Seems to be referenced at Microsoft/TypeScript#21818 but cant see any information that might help mitigate this 🤕

Akin909:
@bryphe started working on this again and I've managed to fix the rename bug, as well as the restarting bug - this required changing the logic we use to find a root dir as it was unfortunately not returning the correct directory causing the `shouldServerRestart` check to fail.

The one outstanding issue I believe with this PR is `lineCount` error above with the `textDocument/Definition` it only occurs after a while of use and actually had a look in the `tss.js` which is the whole server (its 5mb) and I think it must be that we are sending some incorrect data at some point but I'm still not familiar enough to know how this works aka the definitions appear normally but at some point something problematic get sent to the `tss` would you have any ideas or suggestion re where to look?

Akin909:
Update on this I found: https://github.com/Microsoft/TypeScript/issues/16456 which seems to be a very similar (not the same issue) and the cause of it seems to be the language server slowly getting into a state of confusion because of problematic input/ casing issues. I'm thinking that somewhere along the way where post document details like line and column numbers perhaps some of this info is off (I know neovim sends line info as 0 based and perhaps it might relate to that?)

bryphe:
> @bryphe started working on this again and I've managed to fix the rename bug, as well as the restarting bug - this required changing the logic we use to find a root dir as it was unfortunately not returning the correct directory causing the shouldServerRestart check to fail, as well as adding a onBufferDelete event to the editor to trigger the document close event.

Sounds like good progress! Thanks @Akin909 .

>  I'm thinking that somewhere along the way we post to the language server, details like line and column numbers perhaps some of this info is off (I know neovim sends line info as 0 based and perhaps it might relate to that?)

Hmm, the `sendRequest` and `sendNotification` are the central places we send stuff here:
https://github.com/onivim/oni/blob/f507d7235a9a489817891b96922df2b4cb8ad97f/browser/src/Services/Language/LanguageClient.ts#L91
https://github.com/onivim/oni/blob/f507d7235a9a489817891b96922df2b4cb8ad97f/browser/src/Services/Language/LanguageClient.ts#L111

We log the request names, but you might want to log the `args` properties here.

What I'd really like is a way for us to test the `LanguageClient` _independently of Oni_ there's really no dependency on the UI here. We could create a `LanguageClient`, and open some example TypeScript project / files programmatically (via appropriate `sendRequest` calls), and then validate that we get proper completions / definitions / etc.

Akin909:
Thanks for the heads up with those I will do some logging and try and figure out whats going on.

Re. testing the lsp independently that definitely seems v. useful/necessary, perhaps we can have another issue to outline how that might work? e.g. from what i've seen most lsps can be run from the command line so perhaps a  process could be spawned and we use that to send various request and check the responses to ensure that things come and go as expected as you suggest. Although sounds like it might require a service as opposed to just spawning a process as a mock `textDocument` would be needed as well as content to send etc, or perhaps the service would require actually files to be created etc.

I think with this PR i'd like to focus on just getting `theia-ide` to work so that I can get back to the original goal which was to actually look at helping theia set up auto-import functionality once we know its compatible with oni and is what we are going to be/ can use.

Akin909:
Raised the issues regarding testing the lsp as #2246

wesselvdv:
Any idea when this is going to land?

Akin909:
@wesselvdv there've been quite a few errors crop up in trying to integrate this lsp, so has taken *way* longer than I initially hoped. There've have been quite a lot of very nice improvememts to that client since which would be awesome to integrate with oni. Tbh can't give an eta ill update the client and have another go with it to see if im still encountering the error which kills the lsp client (not sure what exactly is going on there i.e. is it an oni error or the client is erroring)

PR #1718, opened 2018-03-06
Fix #1715 - Snippet crashes occasionally on insert

### LINKED ISSUE #1715: Moving through snippet placeholders crashes editor
Whilst tabbing through snippet placeholders (very cool functionality btw) I occasionally (well almost always eventually get an editor crashing error). Apologies for not enough detail will look into it more when I get a moment.

screen shot 2018-03-04 at 22 15 49

bryphe:
__Issue:__ Frequently when inserting snippets, the whole screen goes red with an error message... this is a horrible experience because it means that some edits may be lost 😦 

__Defect:__ There were a couple of issues contributing. First, the `BufferLayer` infrastructure isn't robust when we try and ask about _pixel positions_ for cells that are off the screen - that's where we get the `screenX` error. In the snippet case, we were measuring a position at row `1` to use as the height value, but if we were scrolled past that... we'd hit this error.

__Fix:__ Make the measurement routines more robust - if we don't have the position, at least pass `NaN` so that the computations later don't crash. In addition, plumb an alternative way to get the height of a cell - this is something that would be useful in general for buffer layers, so we should extend our API to support this.

__Future Work__: 
- Extend the `BufferLayerRenderContext` to include `fontPixelWidth` and `fontPixelHeight` properties

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimBufferLayersView.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimBufferLayersView.tsx
@@ -19,6 +19,8 @@ export interface NeovimBufferLayersViewProps {
     activeWindowId: number
     windows: State.IWindow[]
     layers: State.Layers
+    fontPixelWidth: number
+    fontPixelHeight: number
 }
 
 const InnerLayerStyle: React.CSSProperties = {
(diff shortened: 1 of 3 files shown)


bryphe:
Fixes #1715

PR #1727, opened 2018-03-07
[WIP] Add QuickOpen  Refine option.

CrossR:
This adds a basic refinement option to the QuickOpen by allowing multiple search terms, where each sequential filters the options further.

So far, its only added to the `regex` strategy, since the `fuse` one looked a little more confusing, but if we think its suitable I can add that too.

I also don't see any tests for the QuickOpen except for the `open` and `close` ones, so I should probably look into adding the regex and then this filter logic to it. Though thinking about it, is that because the filter sorting isn't deterministic? Either way, I can at least make a Unit test for this bit of code.


bryphe:
This is awesome, @CrossR ! Thanks a lot for implementing this. Will be really useful for my workflow... I can't count the times I start typing `index` and realize there's a ton of matches..

> So far, its only added to the regex strategy, since the fuse one looked a little more confusing, but if we think its suitable I can add that too.

My hope is that we can remove the `fuse` strategy and just use the `regex`. It's essentially 'deprecated' right now. I just wish I had telemetry to see if anyone is using it (they'd have to explicitly set it in their config)

> I also don't see any tests for the QuickOpen except for the open and close ones, so I should probably look into adding the regex and then this filter logic to it. Though thinking about it, is that because the filter sorting isn't deterministic? Either way, I can at least make a Unit test for this bit of code.

Would be great to have a unit test here for this new logic! The filtering logic in this new method would be deterministic and a good fit for a unit test. 

And speaking of test, there's some new functionality here that I haven't really shared out - there's a helper to create a test file (or switch to a test, if it's already there):
image Just an FYI 👍

Akin909:
@bryphe not quite the telemetry you're probably looking for but I still use the fuse strategy personally haven't found either `fuse` or `regex` to be close enough to the experience of searching with `fzf` (probs using it wrong) but will definitely be looking again with the refine option though.

CrossR:
Any ideas on some other tests that this would need? I've added the more obvious ones.

Is it worth me adding this for the fuse method? Its currently still the default as far as I can tell, so this feature wouldn't be surfaced by default. Or maybe its a second issue + PR to check the `regex` strategy and then make it the default. (Though I've been using it as long as its been out and have had no issues).

bryphe:
> Any ideas on some other tests that this would need? I've added the more obvious ones.

This is a great set of cases, @CrossR ! 100% code coverage on the diff - can't ask for better than that! 👍 They do a nice job of documenting the expected behavior, too.

> Is it worth me adding this for the fuse method?

I don't think it's necessary, at least as part of this change. I actually meant to flip the default to `regex` and see if that impacted anyone's workflow - my impression is that the `regex` behavior is more in-line with user expectations, especially once we add this refinement in! Perhaps we can do that for the next release (0.3.2) - makes sense to have this refinement behavior the default.

I'd actually like to look into deprecating the `'fuse'` strategy, if it turns out that it isn't used once it's not the default - would be a good thing to track via #1730 

Nice work @CrossR !

PR #1733, opened 2018-03-07
Fix #685 - Enable autocompletion and quickinfo for config.js

### LINKED ISSUE #685: Enable autocompletion and quickinfo for config.js
Editing the configuration is non-trivial because it involves some coding, as well as domain knowledge on the Oni API and what it providers.

We should make it easier to explore this interface via quick info and completion. At the minimum, we should hook up `Oni.d.ts` so that it gets picked up and resolves. We may have to special case this and bake-in some optimizations, but it is already asking a lot for users to configure their editor through code - this process should be streamlined as much as possible.

bryphe:
This is the remaining work to enable autocompletion / quick info in the config.js (which is now a `config.tsx`).

This moves the _default_ configuration strategy to use TypeScript, which allows us to bring the full capabilities of the language service + the oni api typings with configuration, resulting in a much better experience:
image

image

This fixes a bug in the distribution build where we don't supply typings properly for `oni-api`, and adds a CiTest to validate it build-over-build

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,9 +34,14 @@
             "!examples",
             "!lib_test",
             "!test",
+            "!node_modules/oni-api",
             "!node_modules/typescript"
         ],
         "extraResources": [
+            {
+                "from": "node_modules/oni-api",
+                "to": "app/node_modules/oni-api"
+            },
             {
                 "from": "node_modules/typescript",
                 "to": "app/node_modules/typescript"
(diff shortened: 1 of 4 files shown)


bryphe:
Created this document on migration, too:
https://github.com/onivim/oni/wiki/Migrating-Configuration-to-TypeScript

PR #1747, opened 2018-03-08
Cross Oni Movement

### LINKED ISSUE #67: Multi-monitor support

CrossR:
Potentially fixes #67.

I've not actually got 2 monitors, so I've written this such that it should work broadly moving across any Oni instances. Plus I don't think the screen number actually matters since I think if you have 2 1080p monitors, the top left of your primary is 0,0 and then the other ones are either `-1920,0` or `1920,0`, so this approach (ie using the current location) should work.

That said! I don't want to put a bunch of effort in based on vague assumptions I've had from reading StackExchange and GitHub issues, so if someone with 2 monitors could test it, that would be great. Currently, I've not bothered hooking up the movement keys, I'm just using the `Ctrl+Shift+P` menu to add Up, Down, Left, Right movements.

Assuming that works on multiple monitors, since it works for me on one at least:
- [x] Check issue with `windows` being correctly emptied, since currently if you close a Window, this stops working due to an object being destroyed.
- [x] Actually link this up up the to unhandled movement option.
- [x] Move the code out of main and into a separate file (+ tidy it)
- [x] Unit test it, and spend a little time seeing if CI tests are feasible.
- [x] Deal with minimised windows? Ie currently if you have 3 windows of oni, with the middle minimised, the middle will open up again.
- [x] Check the selection logic, and make sure it prioritises the smallest distance of the current movement.
- [x] Move logic out of OniEditor.

If not, well at least I only spent an hour or so on this to avoid other work.


bryphe:
This is awesome, @CrossR ! Once we have this I might never need to leave Oni... I'll put browser and terminal in one window and code in the other haha

> so if someone with 2 monitors could test it, that would be great. 

Tested it out, worked perfectly moving right/left between my two monitors 👍 

>  Unit test it, and spend a little time seeing if CI tests are feasible.

Thanks for thinking about the tests!

The CiTests may be tricky - we don't have any tests that launch multiple Oni instances today, and in automation, we don't force a single instance of the `main` process (so this actually wouldn't work, unless we made further changes to add a setting to allow that). I think a unit test for the window-switching logic would be sufficient, and we could document a manual test case here:  https://github.com/onivim/oni/blob/master/test/Manual.md#manual-test-cases

It would be nice to have a full CiTest eventually, but I'm not sure it's the right tradeoff of value vs time at the moment (up to you though). 

Other bullet points look great - excited to have this in!

CrossR:
It looks like the error I get isn't due to the `windows` list not being correctly updated on close, but rather when the initial Oni window (that is, the first Oni window to be opened) is closed, the `windows` list gets deleted entirely.

Need to look into why that is happening, and how to fix it...

CrossR:
It seems that for some reason, when the main Window is closed, the first item in the `windows` array becomes unreferenced...for some reason. Which then causes issues since I need to check over all the values in that array. I'll poke around it for a bit tomorrow, since I've not had that much time today to check.

main/src/main.ts
```diff
@@ -84,7 +84,9 @@ ipcMain.on("move-to-next-oni-instance", (event, direction: string) => {
 // Keep a global reference of the window object, if you don't, the window will
 // be closed automatically when the JavaScript object is garbage collected.
 let windows: BrowserWindow[] = []
-let mainWindow: BrowserWindow = null
+const activeWindow = () => {
```
bryphe: Solid! Thanks for fixing this 👍

CrossR:
I've added tests for all the easily testable parts. The remaining function that hasn't been tested doesn't return anything, so we'd need a CI test that can check the focus of a window, which sounds pretty awkward.

Only thing left here I think is I want to check is that where I've added the subscription makes sense in `OniEditor`?

Oh and the original ticket mentions adding commands for Expand Up/Down etc, but I think that is distinct enough to be done in a separate PR. Could probably also do with making the movement into an action that users can bind, such that a user can use it without having to move to the edge of the window with their own keybinds.

main/src/WindowManager.ts
```diff
@@ -0,0 +1,155 @@
...
+            return true
+        }
+    }
+
+    return false
+}
+
+// Given a window, check if it is the best window seen so far.
+// This is determined by the difference in X and Y relative to the current window.
+export function checkWindowToFindBest(
```
bryphe: Very cool! Great set of unit tests covering this functionality.

package.json
```diff
@@ -175,8 +177,9 @@
         "pack:win": "node build/BuildSetupTemplate.js && innosetup-compiler dist/setup.iss --verbose --O=dist",
         "test": "npm run test:unit && npm run test:integration",
         "test:integration": "npm run build:test && mocha -t 120000 lib_test/test/CiTests.js",
-        "test:unit": "npm run test:unit:browser",
-        "test:unit:browser": "npm run build:test:unit && cd browser && electron-mocha --renderer --require testHelpers.js --recursive ../lib_test/browser/test",
+        "test:unit": "npm run test:unit:browser && npm run test:unit:main",
+        "test:unit:browser": "npm run build:test:unit:browser && cd browser && electron-mocha --renderer --require testHelpers.js --recursive ../lib_test/browser/test",
+        "test:unit:main": "npm run build:test:unit:main && cd main && electron-mocha --renderer --recursive ../lib_test/main/test",
```
bryphe: Thanks for setting this up! 💯

browser/src/Editor/OniEditor/OniEditor.tsx
```diff
@@ -170,6 +171,10 @@ export class OniEditor implements IEditor {
             buf => extensions.includes(path.extname(buf.filePath)),
             buf => new ImageBufferLayer(buf),
         )
+
+        windowManager.onUnhandledMove.subscribe((direction: string) => {
```
bryphe: Just saw your comment - there is one problem with the location here. In the 'multiplexing' scenario (when `editor.split.mode` is `'oni'`), we will potentially have _multiple_ `OniEditor` instances - so that means this would get fired multiple times.

One option would be to plop this code in `browser/src/index.tsx`, after we've created the `WindowManager`. Or add an `activate` method in `MultiProcess`, like we have for some of the other services, pass in the `windowManager` as a dependency, and wire up the event there. Let me know if that isn't clear

bryphe:
> I've added tests for all the easily testable parts. 

Awesome! Thanks for wiring up unit tests for our `main` process - big step forward to have some of that code covered under test. This is a great example of a change that improves the quality of the code while implementing new functionality.

> The remaining function that hasn't been tested doesn't return anything, so we'd need a CI test that can check the focus of a window, which sounds pretty awkward.

Agree 100% - I don't think a CiTest is necessary for this. 

> Only thing left here I think is I want to check is that where I've added the subscription makes sense in OniEditor?

Ah ya, good catch - there might be an issue in the 'multiplexing' case - I left a more detailed comment on the file. But we probably will want to move it to a place that only gets initialized once.

> Oh and the original ticket mentions adding commands for Expand Up/Down etc, but I think that is distinct enough to be done in a separate PR. Could probably also do with making the movement into an action that users can bind, such that a user can use it without having to move to the edge of the window with their own keybinds.

Yep, makes sense to have that be a separate PR 👍 

Nice work! 💯

CrossR:
Assuming no failures on the CI side of things, this should be done now.

bryphe:
Awesome! Code changes look great 👍 Approving now

PR #1749, opened 2018-03-08
Fix command line icons config option.

CrossR:
I've still got an old config on my dev VM, and realised I was getting externalised icons there, and not on my actual PC.

@Akin909, I'm assuming that we want this enabled by default? The default config option was updated at least, so I am assuming it should be!


Akin909:
@CrossR thanks for catching that yeah that was the plan 👍

PR #1761, opened 2018-03-09
Fix tabs so that they use the correct id when closing/focusing

### LINKED ISSUE #1760: Tabs behave weirdly (closing & moving between tabs)
Please watch this screen recording first: http://recordit.co/hYs2Yfbfxx

**What happens**:
- Clicking on a tab sometimes select the tab next to it
- Clicking the close button on a tab sometimes closes the tab next to it
- Sometimes clicking on either the tab or the close button does nothing

**Environment**:
Oni: Built on 1743b9a3 (shows as 0.3.1)
OS: Mac OS High Sierra 10.13.3 (17D47)

rgehan:
This PR fixes #1760

### DIFF
--- a/browser/src/UI/components/Tabs.tsx
+++ b/browser/src/UI/components/Tabs.tsx
@@ -303,7 +303,7 @@ const getTabsFromVimTabs = createSelector(
         allBuffers: State.IBuffer[],
     ) => {
         return tabState.tabs.map((t: State.ITab, idx: number) => ({
-            id: t.id,
+            id: idx + 1,
             name: getIdPrefix((idx + 1).toString(), shouldShowId) + getTabName(t.name),
             highlightColor: t.id === tabState.selectedTabId ? color : "transparent",
             iconFileName: showFileIcon ? getTabName(t.name) : "",
--- a/test/CiTests.ts
+++ b/test/CiTests.ts
@@ -26,6 +26,7 @@ const CiTests = [
     "Editor.BufferModifiedState",
     "Editor.OpenFile.PathWithSpacesTest",
     "Editor.TabModifiedState",
+    "Editor.CloseTabWithTabModesTabsTest",
     "MarkdownPreviewTest",
     "PaintPerformanceTest",
     "QuickOpenTest",
--- a/test/ci/Common.ts
+++ b/test/ci/Common.ts
@@ -22,6 +22,11 @@ export const getElementByClassName = (className: string) => {
     }
 }
 
+export const getElementsBySelector = (selector: string) => {
+    const elements = document.body.querySelectorAll(selector)
+    return elements || []
+}
+
 export const createNewFile = async (
     fileExtension: string,
     oni: Oni.Plugin.Api,
--- a/test/ci/Editor.CloseTabWithTabModesTabsTest.ts
+++ b/test/ci/Editor.CloseTabWithTabModesTabsTest.ts
@@ -1,0 +1,43 @@
+import * as assert from "assert"
+import * as Oni from "oni-api"
+
+import { getElementsBySelector } from "./Common"
+
+export const test = async (oni: Oni.Plugin.Api) => {
+    const getTabsCount = () => getElementsBySelector(".tabs .tab").length
+    const waitForTabCount = count => oni.automation.waitFor(() => getTabsCount() === count)
+
+    const openTab = () => {
+        const initialTabCount = getTabsCount()
+        oni.automation.sendKeys(":tabnew")
+        oni.automation.sendKeys("<cr>")
+        return waitForTabCount(initialTabCount + 1)
+    }
+
+    const closeLastTabWithMouse = () => {
+        const tabs = getElementsBySelector(".tabs .tab")
+        const closeButton = tabs[tabs.length - 1].querySelector(".corner.enable-hover")
+        closeButton.click()
+        return waitForTabCount(tabs.length - 1)
+    }
+
+    await oni.automation.waitForEditors()
+
+    // 1. Create two tabs, then close the last on
+    await openTab()
+    await openTab()
+    await closeLastTabWithMouse()
+
+    // 3. Open another tab
+    await openTab()
+
+    // 4.1 Attempt to close it
+    await closeLastTabWithMouse()
+}
+
+// Bring in custom config.
+export const settings = {
+    config: {
+        "tabs.mode": "tabs",
+    },
+}


browser/src/UI/components/Tabs.tsx
```diff
@@ -83,13 +83,13 @@ export class Tabs extends React.PureComponent<ITabsProps, {}> {
...
         }
 
-        const tabs = this.props.tabs.map(t => {
+        const tabs = this.props.tabs.map((t, i) => {
             return (
                 <Tab
                     key={t.id}
                     {...t}
-                    onClickName={() => this._onSelect(t.id)}
-                    onClickClose={() => this._onClickClose(t.id)}
```
rgehan: `i+1` as Neovim tabs' indices seem to start at 1 (shameful)

browser/src/UI/components/Tabs.tsx
```diff
@@ -83,13 +83,13 @@ export class Tabs extends React.PureComponent<ITabsProps, {}> {
...
+                    onClickName={() => this._onSelect(i + 1)}
+                    onClickClose={() => this._onClickClose(i + 1)}
```
rgehan: `i + 1` as Neovim tabs' indices seem to start at 1 (shameful)

rgehan:
It seems like Codecov requires a test for this change.
I don't see any Component test in the app, where should I add one ? Maybe under `test/ci` ?

bryphe:
Thanks for fixing this, @rgehan ! 👍 I'm really impressed with how quickly you picked up the code base!

> It seems like Codecov requires a test for this change. I don't see any Component test in the app, where should I add one ? Maybe under test/ci ?

It would be great to have an an integration test for this scenario (ideally, one for both 'buffers' mode and 'tabs' mode), since it is such a critical scenario to get right. A good example test that is similar is one @CrossR wrote to exercise the 'dirty indicator': https://github.com/onivim/oni/blob/757f9b12ade524b9117f2e93bcdf627e44120682/test/ci/Editor.TabModifiedState.ts#L16

Here's some notes on writing / running integration tests: https://github.com/onivim/oni/wiki/Development#integration-tests

One question I had - does this change work correctly with both `'tabs.mode'` set to `'tabs'` and `'buffers'`? 

We will be moving to `'tabs.mode'` set to `'tabs'` by default in 0.3.1, so I want to make sure that will work OK after the change (as this UI handles both settings - it is agnostic to whether the mode is buffers/tabs). 

If we do need to differentiate behavior, we handle these events differently depending on 'buffer-mode' vs 'tab-mode' here: https://github.com/onivim/oni/blob/757f9b12ade524b9117f2e93bcdf627e44120682/browser/src/Editor/NeovimEditor/NeovimEditor.tsx#L854

rgehan:
> It would be great to have an an integration test for this scenario (ideally, one for both 'buffers' mode and 'tabs' mode), since it is such a critical scenario to get right. A good example test that is similar is one @CrossR wrote to exercise the 'dirty indicator':

I'll try to add one for this two cases then

> One other question I had - does this change work correctly with both 'tabs.mode' set to 'tabs' and 'buffers'?

It definitely works in `tabs` mode, but I'm haven't tested it extensively in `buffers` mode. I'll check this.

rgehan:
I've been working on adding tests to that fix, and I'm a bit surprised by something.

I was running my integration test again and again, and it looked like my fix wasn't taken into account and that the bug was still present.

I started suspecting that the CI tests weren't recompiling the whole app, and were instead using a previous build.

After inspecting the `build:test*` scripts, I think I'm correct. I then tried deleting my `lib/` folder, and now the tests won't even start, comforting me in this idea.

@bryphe Is this a normal behavior? Shouldn't the tests recompile the app? I guess having to recompile for the tests would slow down Travis' build, but we could leverage the `CI` environment variable to only re-build on dev envs

bryphe:
> I was running my integration test again and again, and it looked like my fix wasn't taken into account and that the bug was still present.

Hmm, how were you building / testing? Were you using `npm run start` (webpack dev server) or `npm run build`/`npm run build-debug`?

> I started suspecting that the CI tests weren't recompiling the whole app, and were instead using a previous build.

This is correct. `npm run test:integration` uses the last built binaries, builds the test integration tests, and runs - but doesn't build the app bits.

>@bryphe Is this a normal behavior? Shouldn't the tests recompile the app? I guess having to recompile for the tests would slow down Travis' build, but we could leverage the CI environment variable to only re-build on dev envs

I think normal behavior is relative 😄 But it's pretty common for projects to split up these steps:
- _product build_
- _test build_
- _lint_
- _running tests_

(although i see various pivots / variations - like sometimes build steps are combined, or unit tests are run along with build, etc), in the interest of keeping each of those _iteration loops_ small. 

Just to give you an idea of a 'worst case' - the integration tests used to only run against our _distribution_ bits - meaning you'd need to build, pack, and the pack step takes a while - it could take around ~10 minutes to test the changes! We changed that in #1456 so that the integration tests will run against a local electron build (not requiring the pack step) on local developer machines - but still require a local build via `npm run build` prior to running the tests.

But apologies that this wasn't intuitive and cost you extra time 😦 At the minimum, I updated the wiki here with a note to run `npm run build` prior to get the changes picked up: https://github.com/onivim/oni/wiki/Development#integration-tests

Let me know if you have other ideas and how we can make it easier! I am open to tweaking the scripts too, if that makes it more intuitive. The end goal is to make it as easy & pain-free to contribute as possible 👍

rgehan:
> But apologies that this wasn't intuitive and cost you extra time 😦 At the minimum, I updated the wiki here with a note to run npm run build prior to get the changes picked up: onivim/oni/wiki/Development#integration-tests

Don't worry about that, I didn't lose that much time 😄 If it can help future-me or future-anyone not to fall into the same trap it's fine!

rgehan:
It is still a work in progress, I need to fix/test this when `tabs.mode` is set to `buffers`

bryphe:
👍 The test for tabs looks great!

rgehan:
Thanks :smile:

What is the proper behavior for the `tabs.mode: buffers` setting? Because I've encountered everything and I'm not sure what is the correct one now.

I think that clicking on a file in the Explorer opened it in the only tab currently opened, but I'm not sure about that.

By looking at [NeovimEditor.tsx@754](https://github.com/onivim/oni/blob/master/browser/src/Editor/NeovimEditor/NeovimEditor.tsx#L754) I think it would use `:e!` which would be coherent with what I said before.

rgehan:
@bryphe I checked if it was working in `buffers` mode, and it doesn't. It seems like it is even broken on master.

I investigated a bit, and it looks like the `getActiveBufferId()` selector doesn't return the correct value every time.

If you want to try this by yourself, open 5-6 files and select/delete some of them. After that, clicking on a tab won't focus it, yet the content of the file will be displayed in the editor.

Do you have any idea about what could be causing that?

browser/src/UI/components/Tabs.tsx
```diff
@@ -46,9 +46,10 @@ export interface ITabsProps {
     visible: boolean
     tabs: ITabProps[]
 
+    shouldUseVimTabs: boolean
```
bryphe: Ideally, the `Tab` component shouldn't need to worry about whether we are using `tabs` or `buffers` - it just gets a bunch of `Tab` objects with `id`, `name`, etc, and hands that back up later.

It looks like the latest fix _only_ impacts `tabs` - what if we move the fix here?
https://github.com/onivim/oni/blob/766fc30133aea41551295c74d458cdec29da6790/browser/src/UI/components/Tabs.tsx#L306

We create the data model for `<Tab />` with either `getTabsFromVimTabs` or `getTabsFromBufferTabs`, depending on the `tabs.mode`.

A minimal fix would be to change `getTabsFromVimTabs` so that it uses `id: t.id + 1` - that way, we wouldn't need a new prop or to change any logic in `<Tab />` (and the buffer strategy stays the same as it is today)

bryphe:
> @bryphe I checked if it was working in buffers mode, and it doesn't. It seems like it is even broken on master.

Thanks for the investigation here! Yes, definitely possible there are existing bugs here... If this doesn't make `master` worse or change behavior in the `tabs.mode: 'buffers'` case, it's fine - I'm happy to keep the fix scoped for tabs 👍 (Especially since it is the default, now!)

I did leave one comment - I think we could streamline the fix just a bit. Let me know what you think.

Thanks for all your work on this, @rgehan ! Much appreciated

browser/src/UI/components/Tabs.tsx
rgehan: That does makes sense. I wasn't too fond of changing the logic all around the file. I'll try and make the changes you suggested.

rgehan:
I think it would be better to keep the scope of the PR restricted to `tabs` mode only. I'll create an issue for the `buffers` mode.

Isn't codecov a bit too restrictive? Because right now, I think all my changes are pretty much covered by tests, but codecov is even less happy than before 😛

bryphe:
> I think it would be better to keep the scope of the PR restricted to tabs mode only. I'll create an issue for the buffers mode.

Perfect! Thanks @rgehan .

> Isn't codecov a bit too restrictive? Because right now, I think all my changes are pretty much covered by tests, but codecov is even less happy than before 😛

Yep, unfortunately it doesn't pick up code coverage in our `CiTests` currently (just the unit tests). So we don't gate check-ins with it at the moment - just use it as a guideline. It would be nice if we included the CiTests in our coverage collector... the % would be much higher 👍 

The latest change looks great! Will bring it once it goes through the Ci build.

bryphe:
Cheers @rgehan - thanks for making the change and logging the issue! 💯 

I'm really happy we have this test to ensure the scenario doesn't regress moving forward. Great work! I'm bringing it in now.

Issue #1502, opened 2018-02-10
Update Windows File Associations


CrossR:
Currently, the File Associations for Oni live in two places : `package.json` and `build\BuildSetupTemplate.js`.
This causes them to become out of line, so we should update `BuildSetupTemplate.js` to instead pull in the extensions from our `package.json`.

This should mean updating the existing `fileExtensions` array in `BuildSetupTemplate.js` to be built dynamically, such that we have a single version of the truth.

CrossR:
Just brought this is now.

Cheers @mgbennett!

PR #1769, opened 2018-03-10
Dynamic file associations

mgbennett:
This commit removes the hardcoded mappings of fileExtensions from BuildSetupTemplate.js and reuses the mappings in package.json as per #1502.

There's a meaningless empty commit as I didn't realize at first that the change the valuesToReplace to not use strings as keys was done by a precommit hook and thought it was leftover from some earlier fiddling.


CrossR:
Thanks for the first contribution @mgbennett !

Nice to have it all in one place now, as well as having those extra bindings being brought in.

I'll bring this in now.

PR #1775, opened 2018-03-12
feature/add truncation to working dir path

Akin909:
The statusbar item which shows the working directory is not currently limited and can have any length depending on the working dir path. This means it dominates the left hand side of the status bar. This PR adds a limit and a text overflow to truncate it if it is too long.


bryphe:
Thanks for adding this, @Akin909 ! Looks good to me 👍

PR #1791, opened 2018-03-13
[WIP] Fix issue where token color is not being overridden

bryphe:


bryphe:
Was trying out the `editor.tokenColors` option, and saw that it was busted. Fixing it so it has parity with the documentation - this is important for the usability of our textmate highlighting feature.

Issue #1478, opened 2018-02-08
[RFC] Use `.config/oni` for configuration instead of `.oni`


badosu:
Currently oni uses the `~/.oni` folder to store local configuration.

While this is a common model to use, these days many providers use the `~/.config/oni` model instead.

This avoids cluttering of the home folder while also organizing all local configuration files, as you can see electron even creates one by default:

```
~ ❯❯❯ tree .config/oni
.config/oni
├── Cache
│   ├── data_0
│   ├── data_1
│   ├── data_2
│   ├── data_3
│   └── index
├── GPUCache
│   ├── data_0
│   ├── data_1
│   ├── data_2
│   ├── data_3
│   └── index
├── Local Storage
│   ├── chrome-devtools_devtools_0.localstorage
│   └── chrome-devtools_devtools_0.localstorage-journal
├── Preferences
└── Settings

3 directories, 14 files
```

This would also unify both configuration folders.

CrossR:
As far as I know, this is also how Neovim works, so makes even more sense to me to do it this way.

bryphe:
👍 , makes sense to me.

TalAmuyal:
That is not exactly correct (though I do like the idea).
You describe the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html) which requires things like cache to be in a seperate dir.
It seems to be implemented by [this package](https://www.npmjs.com/package/xdg) so it might be easy to implement compliantly (I think it supports Windows as well).

badosu:
Thanks @TalAmuyal, I'll take a stab at this!

PR #1792, opened 2018-03-13
Use xdg compliant path for config.js/config.tsx files

rgehan:
This PR attempts to move the config folder to an XDG-compliant path: `~/.config/oni` as discussed in https://github.com/onivim/oni/issues/1478

:information_source: This change should probably be mentioned in a changelog/documentation so that users are not confused as to where their config went

### DIFF
--- a/browser/src/Services/Configuration/UserConfiguration.ts
+++ b/browser/src/Services/Configuration/UserConfiguration.ts
@@ -36,5 +36,5 @@ export const getUserConfigFolderPath = (): string => {
 
     return Platform.isWindows()
         ? path.join(Platform.getUserHome(), "oni")
-        : path.join(Platform.getUserHome(), ".oni")
+        : path.join(Platform.getUserHome(), ".config/oni") // XDG-compliant
 }


CrossR:
Given this a test on Windows, and it defaults to `~\.config\oni\config.tsx`, which isn't ideal.

On Windows (and currently) we have it stored in `%APPDATA%\oni\config.tsx`, which is the default for Windows config files.

I'd be tempted to leave Windows how it is currently, since there is no issue, and only apply this fix to Linux/Mac.

My knee-jerk reaction too is that we might not even need a package for it, since we could just change the string from `.oni` to `.config/oni`. That said, if we imagine we will need to use the cache or data folders etc in the future, its probably worth including it for when we need it in the future. If we do use a package, perhaps we can find/write one that correctly deals with Windows, though I expect that is because its left out of the spec.

All that said its depends on what @bryphe thinks?

rgehan:
I've looked into the source of the package, and for what we want to do, it now seems a bit overkill.

All it does is concatenate `os.homedir()` with `.config/oni`.

I'm now curious as to why `os.homedir()` isn't correct on Windows though

CrossR:
>I'm now curious as to why os.homedir() isn't correct on Windows though

Well, it is returning the correct User home, its just that Config files aren't stored in the User home on Windows. Ie it is correctly returning `C:\Users\Ryan\` which is the home folder of my user, but config files are stored elsewhere in `%APPDATA%`.

rgehan:
Then using that package is probably a bad idea. Its advantage would have been working cross-platform, but if that's not the case, we have no need for it.

bryphe:
Thanks @rgehan and @CrossR for looking at this! 

Agreed - if there isn't any special logic we need or special cases to handle, the package seems overkill - it seems like we could use the XDG-compliant path in this method here for the non-windows case:
https://github.com/onivim/oni/blob/c20fe4090d4bb8979bfe164385f52243a50d7c63/browser/src/Services/Configuration/UserConfiguration.ts#L37

(Switching the `.oni` to `.config/oni`).

rgehan:
@bryphe I updated the commit already, and did what you're just suggesting :)
Thanks @CrossR

bryphe:
Looks good to me, @rgehan - thanks for making the change! I marked it was `breaking change` so that I remember to document it in the 0.3.2 release notes. Bringing it in now.

badosu:
Good one @rgehan, thanks!

PR #1795, opened 2018-03-13
Edit user config in a VSCode-like split view

rgehan:
This PR attempts to modify the way the "user config" is currently edited so that:
- It is shown side-by-side with a reference file showing all the possible options
- The reference file is annotated/commented and not just a raw JSON file
- The reference file is readonly

### What is actually implemented
- The split view is displayed with the default configuration values

### What needs to be done
- The reference file needs to be annotated/commented. This might have to wait for the new *Config API* to be implemented
- The reference file needs to be readonly

### DIFF
--- a/browser/src/Services/Configuration/ConfigurationEditor.ts
+++ b/browser/src/Services/Configuration/ConfigurationEditor.ts
@@ -14,6 +14,7 @@ import * as Log from "./../../Log"
 import { EditorManager } from "./../EditorManager"
 
 import { Configuration } from "./Configuration"
+import { DefaultConfiguration } from "./DefaultConfiguration"
 
 // For configuring Oni, JavaScript is the de-facto language, and the configuration
 // today will _always_ happen through `config.js`
@@ -106,9 +107,24 @@ export class ConfigurationEditManager {
             }
         }
 
-        this._editorManager.activeEditor.openFile(normalizedEditFile, {
+        // Create the buffer with the list of all the available options
+        await this._createReadonlyReferenceBuffer()
+
+        // Open the actual configuration file
+        await this._editorManager.activeEditor.openFile(normalizedEditFile, {
+            openMode: Oni.FileOpenMode.VerticalSplit,
+        })
+    }
+
+    private async _createReadonlyReferenceBuffer() {
+        const referenceBuffer = await this._editorManager.activeEditor.openFile("", {
             openMode: Oni.FileOpenMode.NewTab,
         })
+
+        // Format the default configuration values as a pretty JSON object, then
+        // set it as the reference buffer content
+        const referenceContent = JSON.stringify(DefaultConfiguration, null, "  ")
+        await referenceBuffer.setLines(0, 1, referenceContent.split("\n"))
     }
 
     private async _transpileConfiguration(


bryphe:
Awesome, thanks for thinking about this @rgehan ! Now that we have completion in the programmatic portions, this is really the last hurdle IMO to having a complete, robust configuration story.

> How would you go about opening a file in readonly mode? I currently use :view (like so: :tabnew! | view [file] for example)

That seems reasonable. We might want to add an option to the `openFile` API on the editor, like:
```
editorManager.activeEditor.openFile("some config description", { openMode: Oni.FileOpenMode.VerticalSplit, readonly: true })
```
or alternatively:
```
const buffer = await editorManager.activeEditor.openFile("some config description", { openMode: Oni.FileOpenMode.VerticalSplit" })
await buffer.setReadOnly(true)
```
(In either case, under the `openFile`/`setReadOnly` API, we'd use something like `:view` / set the readonly flag on the buffer via the `msgpack` API / send some VimL, like `:set nomodifiable`)

> Should we enforce readonly-ness at the editor-level? Because as far as my knowledge of Vim goes, it is always possible to save/edit a readonly file by using a ! symbol

We could look into this later - at least initially, we could rely on Vim's read-only mode to start, and use that to build this configuration flow. If we find that it isn't enough, we can always add additional readonly-ness down the road. P

> Should we even open a file? Isn't there a way to pass Neovim a "string" and not bother with a file? We could use a temporary file as a last resort but I don't like it.

We actually don't need to open a file! We could create a new split window, and drop contents in the buffer. We do this, for example, if the user is customizing snippets - if it's a new file, we'll drop in some defaults:
https://github.com/onivim/oni/blob/c20fe4090d4bb8979bfe164385f52243a50d7c63/browser/src/Services/Snippets/UserSnippetProvider.ts#L96

So a potential option would be to generate this 'on-the-fly' by serializing the current set of configuration values (which we can get off the configuration option).

Vim has a way to create a 'scratch buffer', which has no file associated with it, with these commands:
```
:setlocal buftype=nofile
:setlocal bufhidden=hide
:setlocal noswapfile
```
We might be able to leverage that for this scenario (along with nomodifiable).

> What to display as a reference? I was thinking about showing a formatted version of:

I think there are two pieces of information that is useful here:
- The _default value_ (which we get from `DefaultConfiguration.ts`)
- The _documentation comments_ (which are in `IConfiguration.ts`)

But perhaps a short-term fix would be to port the comments, so that both pieces are available in `DefaultConfiguration.ts`.

Longer-term, I'm not sure the way we are defining configuration options scales out to plugins - since plugins aren't going to be able to add to `IConfiguration.ts` or `DefaultConfiguration.ts`

 I was thinking about a new API, like:

```
Oni.configuration.defineConfigurationValue("myPlugin.myConfigurationValue", /* default */ 10, /* Description*/ "This is a configuration value that sets myConfigurationValue for my plugin. Must be a number between 9 and 11")
```

Then it would be easy to programmatically get the information for all configuration values - even ones provided by plugins. (I'm also looking at it to smooth out the configuration-reload story - it'd be nice to know which configuration values support refreshing, and which ones require a reload of the editor, to let the user know via a notification that they need to restart the editor)

That's a bit of a tangent though. Hope that helps give you some ideas. Let me know if you have any other questions.

So to summarize - seems like there are 3 challenges:
- Opening a file in a vertical split - the UserSnippetProvider has a similiar example
- Making the buffer read-only - I think I like having an option to `setReadOnly` on the buffer - could live somewhere here: https://github.com/onivim/oni/blob/c20fe4090d4bb8979bfe164385f52243a50d7c63/browser/src/Editor/BufferManager.ts#L302
- Figuring out _what_ goes in that vertical split - this is the hardest one I think 😄 Using `DefaultConfiguration.ts` for now seems like the best option, if we include the comments there too. Long-term, maybe we'll have a better way to generate this.

Sorry for the wall of text! Thanks again for exploring this, @rgehan 💯

rgehan:
@bryphe I've fiddled a bit with your suggestions and have encountered a few little issues:
- Splitting the window after having run `:set nomodifiable` makes the newly opened buffer non-modifiable too
- `:set nomodifiable` affects the active buffer, meaning that for `setReadOnly` to work, I have to call it just after I created the buffer; which is not that much of an issue now, but could be later on.

The only solution I can see, is to pass a `readOnly` flag to `openFile()`, and conditionnaly call `setlocal nomodifiable` in the method directly. It is less clean than a `setReadOnly` but I guess it would work right away.

Am I mistaken there?

bryphe:
> I’ve looked a bit into Nvim API, and it seems the simplest way to achieve that would be to use nvim_buf_set_option

Yeah! Seems like we could implement `setReadOnly` by calling `nvim_buf_set_option` for `modifiable`.

rgehan:
Except for the readonly (see #1806), it works pretty well.

It looks like that right now:
capture d ecran 2018-03-15 a 19 22 20

You've suggested I move/clone the comments from `IConfigurationValues.ts` to `DefaultConfiguration.ts`, but it would only be useful if we were to show the raw file as a reference.

The problem is that in `DefaultConfiguration.ts` there are a few values that are computed at runtime, such as:
```
{
  "debug.showNotificationOnError": process.env.NODE_ENV !== "production"
  "language.html.languageServer.command": htmlLanguageServerPath
  "terminal.shellCommand": os.platform() === "win32" ? "cmd" : "bash"
  // etc
}
```

Showing the file, unformatted and raw, would mean that all these values aren't that useful (but we'd have the comments at least).

Right now, I'm just showing a formatted version of what is exported by `DefaultConfiguration`. It is not decorated by comments and all, but it represents the base configuration. IMO, it is a bit better than showing "code".

bryphe:
Looks great, @rgehan ! Having the available values there is a huge step forward.

> Right now, I'm just showing a formatted version of what is exported by DefaultConfiguration. It is not decorated by comments and all, but it represents the base configuration which is IMO, a bit better than showing raw "code".

This makes sense. Showing the values is preferred to the code, as those are what are important to the user when configuring.

> @bryphe What do you think about this?

I'm a big fan of incrementally adding functionality - given that, it makes sense to me to bring in as-is now, and we can continue to improve it. The longer-term goal is to have that new "config API", which, once it's there, we could add those comments into this buffer. Don't think it's worth waiting though 👍 

Even without comments, this is much better than our current scenario - it at least gives the user things to search for, and the ability to see what options are available.

One other idea I had (that can come later) is that we could create a custom _buffer layer_ - a _buffer layer_ lets us render all kinds of custom UI on top of buffer text (it's basically a react component we render, but it's aware of buffer positions). We could add some conveniences like a way to copy configuration values over (sort of like VSCode). Something we could think about and incrementally add later.

bryphe:
The code change looks good to me! We can implement the `setReadOnly` functionality in a separate change. 

Nice work @rgehan , really appreciate the contribution! 💯

rgehan:
@bryphe In this case, I think the PR's good for now.
I'll come back to it in the next few days to add the readonly; once I've figured out why I cannot use a yarn linked `oni-api`

bryphe:
> @bryphe In this case, I think the PR's good for now.

Perfect! Bringing it in now. Thank you for the contribution, @rgehan !

> I'll come back to it in the next few days to add the readonly; once I've figured out why I cannot use a yarn linked oni-api (#1806)

Sounds good. I saw #1806 - unfortunately I haven't been able to get the linking working either. I wasn't sure if its a `yarn` issue or what... I would manually copy the files over from my local `oni-api` repo, but its not an ideal workflow. Would be great to know what's happening.

PR #1811, opened 2018-03-15
Tidy up and remove options from default vim config.

CrossR:
I've not moved any of the other values to the core conf just yet, so as to minimise the changes that the default and core config go through.

Eventually, is the plan to remove all options from the default config? Seems like we could achieve everything there in the Oni layer, which makes it more front and central from a customisation point of view.

### DIFF
--- a/vim/default/bundle/oni-vim-defaults/plugin/init.vim
+++ b/vim/default/bundle/oni-vim-defaults/plugin/init.vim
@@ -5,12 +5,6 @@ set smartcase
 set splitright
 set splitbelow
 
-" Default tab settings
-set tabstop=4
-set shiftwidth=4
-set softtabstop=4
-set expandtab
-
 " Turn off statusbar, because it is externalized
 set noshowmode
 set noruler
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for cleaning up these defaults, @CrossR ! Looks good to me 👍

bryphe:
> Eventually, is the plan to remove all options from the default config? Seems like we could achieve everything there in the Oni layer, which makes it more front and central from a customisation point of view.

Yep! Ideally we wouldn't need to 'eject' to a VimL config (at least for the majority of settings), and be able to handle all the configuration in the `config.tsx`. I think we're getting pretty close... and the less defaults we overwrite in `VimL`, the easier it will be for users to bring over their vim config w/o conflicts.

CrossR:
I've tagged this as a breaking change, since despite it bringing the values back to vim defaults, its possible someone could be used to that functionality.

badosu:
> Yep! Ideally we wouldn't need to 'eject' to a VimL config (at least for the majority of settings), and be able to handle all the configuration in the config.tsx

I think it would be great if we could leverage this to remove the tab configuration as well, so that unless it's specified in `config.ts` it should not be set. To keep the 4 spaces default we could deploy this configuration in `config.ts` by default so it's explicit to the user.

If the user removes the configuration, the `init.vim` one is used.

bryphe:
> I think it would be great if we could leverage this to remove the tab configuration as well, so that unless it's specified in config.ts it should not be set. To keep the 4 spaces default we could deploy this configuration in config.ts by default so it's explicit to the user.

Yes - I think we're close to having that, with #1288 👍

Issue #1767, opened 2018-03-10
Fuzzy finder highlighting would look better with bold instead of underlining


rgehan:
I find the underlining of separate letters a bit hard on the eye. It obfuscates the words a bit.

Let's compare how VSCode looks vs. how Oni looks:

![capture d ecran 2018-03-10 a 18 00 43](https://user-images.githubusercontent.com/793402/37244693-291b61e6-248d-11e8-9282-da8a03929cff.png)

![capture d ecran 2018-03-10 a 18 00 51](https://user-images.githubusercontent.com/793402/37244694-2ad957c2-248d-11e8-9f58-820ffbe9874f.png)

I think bolding the matches would look a bit more polished

bryphe:
👍 , I'm open to changing this

If someone is up for taking a look at this - the component that handles the highlighting is around here:
https://github.com/onivim/oni/blob/3c88c847e3c55ac0f23bb945c9709deeb9021d02/browser/src/UI/components/HighlightText.tsx#L42

chenglou:
Sublime Text bolds + uses another color: 
<img width="592" alt="screenshot 2018-03-13 22 37 36" src="https://user-images.githubusercontent.com/1909539/37385003-259926be-270f-11e8-94ab-e0c4d4ee2abd.png">

badosu:
Seems like this is fixed in the latest release.

PR #1814, opened 2018-03-15
Remove annoying underline & color menu highlightings

rgehan:
Quick fix for #1767.

It would now look like this:
capture d ecran 2018-03-16 a 14 23 35

It uses the `highlight.mode.normal.background` theme color

### DIFF
--- a/browser/src/Services/Menu/Menu.less
+++ b/browser/src/Services/Menu/Menu.less
@@ -61,22 +61,12 @@
             .label {
                 margin: 4px;
                 padding-right: 8px;
-
-                .highlight {
-                    font-weight: bold;
-                    text-decoration: underline;
-                }
             }
 
             .detail {
                 font-size: @font-size-small;
                 color: @text-color-detail;
                 flex: 1 1 auto;
-
-                .highlight {
-                    font-weight: bold;
-                    text-decoration: underline;
-                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


browser/src/Services/Menu/Menu.less
```diff
@@ -64,7 +64,7 @@
 
                 .highlight {
                     font-weight: bold;
-                    text-decoration: underline;
+                    color: #608be1;
```
bryphe: It would be cool to have this pulled from the current theme - could we move the `.highlight` style to a `styled-component` and use a theme color?

We could potentially add a `& .highlight { ... }` style here: https://github.com/onivim/oni/blob/80e16ee3f09d5173da7f3d181ca8f92389e98102/browser/src/Services/Menu/MenuComponent.tsx#L205

And we could grab some theme color - either a new one, or reuse one, like:
```
    & .highlight {
        font-weight: bold;
        color: ${p => p.theme["highlight.mode.normal.background"]};
    }
```

bryphe:
Screenshot looks great! We should pull the color from a theme color to ensure that it will generalize across all the colorschemes. We could either create a new one, or reuse one of the `highlight.*` colors. I think `highlight.mode.normal.background` is the closest to whats in the screenshot.

browser/src/Services/Menu/Menu.less
rgehan: I really like that way, it's cleaner thanks!

rgehan:
@bryphe I've taken your remarks in consideration and made the necessary changes. It's ok for me!

bryphe:
Bringing it in now-

PR #1819, opened 2018-03-16
Re-added autoread work around

CrossR:
This is just to bring back in the work-around for autoread that was seemingly removed before.

If this was intentional, then feel free to close this PR!

There is a linked Github issue to why this is used, since it does look a bit odd. Ideally it would be updated  in core, but until then this seems to be the best/common way of doing it.


bryphe:
Thanks for catching this and bringing it back, @CrossR ! 💯

Issue #1737, opened 2018-03-07
VimNavigator: Allow editing the folder structure in explorer


CrossR:
Having the ability to edit files and directories in the explorer using the vim commands would be very useful.

I'm not fully sure what makes the most sense but I see a few different things we'd need:
* New/Add File
* New/Add Folder
* Rename item
* Move Item
* Delete Item

Delete being `d` or `x` would make sense, same for using the normal `y`+`p` for moving/copying etc.
`a` or `i` possibly for adding a new file? I'm less sure on that, and how you can distinguish between a file and folder. Rename probably makes sense to be `r`.

Some parts of this have already been added by @Akin909, so we could potentially just call the code he has already written, at least for moving files etc.

That is all of the top of my head, if anyone has used the in Vim options more than I have, I'd be interested to know what you think makes more sense. We could also nick ideas from something like Ranger, which is a text based file explorer that has vim-style keybinds.

Akin909:
@CrossR this seems like a really cool idea and much more *vim* than dragging and dropping 😆, yes the `moveFileOrFolder` function could definitely be re-used here all in all depending on how exactly the `vimNavigator` works this could potentially be easier to implement dragging and dropping since there are no moving parts. Was thinking of something similar/but not quite the same in #1536, but this I think would remove the need for a menu or icons just some visual indication when yanking or pasting.

badosu:
I think this is a great idea, but I suggest adding a prefix character so that's not too easy to make mistakes. e.g. `<leader>d` to delete a file.

This prefix character can be used on all file changing commands.

Akin909:
I think atm `Oni` can't take chorded input like `<leader>d` although the more oni commands we rack up the more of a need I think we have for it but think there's already an issue tracking it #1055. A simple alternative would be binding `<ctrl-d>` etc only whilst in the explorer.

Akin909:
Regarding this functionality one thing I'm currently not clear on re possible implementation @bryphe is how or if the `vimNavigator` deals with `y`, `p` etc. commands?

bryphe:
This would be awesome! If we can address these bullets in our explorer, we'll be in a great place to having full parity with NERDTree/netrw.

> Regarding this functionality one thing I'm currently not clear on re possible implementation @bryphe is how or if the vimNavigator deals with y, p etc. commands?

[...]

Issue #1652, opened 2018-02-27
buffer-like features for GUI File Explorer


trusktr:
I really like that in NERDTree it's just a buffer, so I can hit `/` and search for text in it to move the cursor there quicker. I miss this in File Explorer, where I feel more limited in that sense.

Akin909:
@badosu this hasn't actually implemented this yet, although don't know if some version of this was possible prior to the explorer changes but its one of the remaining things I was planning to look at as well as ironing out some kinks

PR #1839, opened 2018-03-17
Feature/add file/folder manipulation commands to explorer

Akin909:
Following discussion in #1737 this PR adds the ability to yank and paste files and folders using `y` and `p` as defaults (a user can remap this) these are also gated by checking that the explorer split is active.

yank

### Todo
- [x] Add epic to clear yank register after timeout if user never pastes
- [x] *Stretch goal* try and get animation working for moving nodes in explorer
- [x] *Very nice to have...* Multi-select aka make yank register an array and move multiple items.
- [x] Add `h/l` movements to expand and collapse a dir
- [x] Localise bindings to explorer pane
- [x] Add undo functionality
~- [ ] Add Sidebar to `Oni.Plugin.Api` to correctly type `isExplorerActive`~

EDIT: Added multi-select 😆 

yank-two

EDIT: Animation finally ^^
explorer-animation


Akin909:
Added functionality to toggle the currently yanked item if `y` is pressed on it again so a user can deselect if they selected the item by accident.

Akin909:
Interesting follow up here would be adding a an `undo register` with a `u` command so that if the user executes a move/or future rename etc. we store these in redux e.g the type of action aka move and the subjects and `undo` the action i.e. carry it out in reverse

bryphe:
Woah, this is so cool @Akin909 ! 💯 Can't wait to try it out!

Akin909:
Added undo functionality currently only for the implemented `paste` actions bound to the key `u` so a user can yank some files move them to a directory and then, change their mind and hit `u` to undo the last move and all files will be returned to their original directories. 😄 

Lack of transitions is an ongoing pain as though initially seeming to work if the refreshDirectory action is triggered a whole new set of react nodes are created which means these are created at the same time as the transition component so no transitioning happens 😞

Akin909:
Blocked atm by #1857, as since a recent commit now unable to actually move through the explorer menu at all

Akin909:
Could use some input here @bryphe @CrossR.

We currently dont have an easy api for taking input from a user at will other than the `notifications` I'm wondering if for the rename functionality we would want to reuse the notifications but add in an input, or maybe use the `rename symbol` ui.

Although I think I either way my preference would be to leave those changes out of this PR.

Also I was going to add a notification to the delete functionality here, so if a user hits `explorer.delete` a notification pops up to confirm that that's what the user intended.

Lastly I was toying with what I thought would be a cool idea which is to hang on to the file details and contents post deletion and allow a user to use the `undo` functionality I currently have for yank and paste and use that to undo the deletion i.e. we create a file with the same name and populate it with the file contents we held onto.

Akin909:
Thought it through and decided to leave the rename out, to make this PR simpler, what I have added though is undo functionality for the delete command.

@bryphe, @CrossR the way that this works is that I move the deleted folder to the `tempdir` and if they undo I move it back, and if not accessed for the duration of the `tempdir` pre clean things then actually get deleted.

CrossR:
>Thought it through and decided to leave the rename out, to make this PR simpler, what I have added though is undo functionality for the delete command.

Makes sense, rename is a lot more awkward on the UI side!

>@bryphe, @CrossR the way that this works is that I move the deleted folder to the tempdir and if they undo I move it back, and if not accessed for the duration of the tempdir pre clean things then actually get deleted.

Sounds good, and I've seen there is a config option for it too which is one thing I did wonder about.
How does the undo-ing actually work? Does it follow on undo tree?

I.e. I delete 1 then 2, then 3. So undo brings back 3, then 2 then 1?

Only thing I am wondering is perhaps we could have a filesize limit/force delete version too? I often delete big CSVs so moving them around probably wouldn't be the easiest. Probably something for a further refinement of the feature though and not here!

browser/src/Services/Configuration/IConfigurationValues.ts
```diff
@@ -189,6 +189,11 @@ export interface IConfigurationValues {
     // 'zero-latency' mode typing, and increases responsiveness.
     "editor.typingPrediction": boolean
 
+    // Files deleted in the explorer can be persisted for the duration
+    // of the session meaning that deletion can be undone is this is set
+    // to true
+    "explorer.persistDeletedFiles": true
```
CrossR: Should be `boolean` here I think.

Akin909: woops sloppy copy paste there 😆 thanks 👍

Akin909:
Re. rename I'm a bit unclear on whether to reuse the notifications or if that would be pushing them beyond their primary function and if not the rename ui as it is I think could use some polish 🤷‍♂️ 

> Sounds good, and I've seen there is a config option for it too which is one thing I did wonder about.
How does the undo-ing actually work? Does it follow on undo tree?

> I.e.

> I delete 1 then 2, then 3. So undo brings back 3, then 2 then 1?

Re. the undoing yep just like that essentially a stack/list:
```js
[{ pastedThings, type: "PASTE" }, { deletedThing, type: "DELETE"}]
```

So its entirely linear if I paste then undo the paste is undone and the paste action is removed from the list the next thing that would then get undone is the delete.
if I do something undoable inbetween that will become the next thing to undo so its last in first out.

Good idea re file size tbh I was thinking there are bound to be cases where we should handle things differently but couldn't come up with any examples based on how I work, ill add in a clause to say if a file is over..(not sure how big would be problematic here?) then we just delete it, and maybe an extra command thats just an un-undoable delete.

bryphe:
Really cool stuff @Akin909 ! I like all the thinking around undo for the delete scenario - really helps to build confidence in working with the explorer. I'm excited for all these changes - the current explorer is very limited, but this steps it up greatly in terms of usefulness 💯 

> Re. rename I'm a bit unclear on whether to reuse the notifications or if that would be pushing them beyond their primary function and if not the rename ui as it is I think could use some polish 🤷‍♂️

IMO, I don't think Notifications are the right 'UX' here - because it kind of forces you to switch context. If you start an 'add file' action, and it takes you over to the other side of the screen, that's a bit jarring. 

Also, there's a lot of value in having the add be _inline_ - because you have some visual assurance of where your addition is going in the tree (like, is it going in the right directory?)

What I'd recommend using for this is our `TextInputView`: https://github.com/onivim/oni/blob/bd8188f6750f79f028b55291525c42280dd730c5/browser/src/UI/components/LightweightText.tsx#L20

This is a common textbox that powers the rename experience, the menu, and the browser address bar. It also automatically grabs focus when it is visible, and retracts it when it's unmounted. So we could use this for both the 'rename' and 'add' experience - for rename, if a node is in 'rename' mode, we could render this `<TextInputView />` instead of the div/span. Likewise, for the add experience, we could create a wrapper component that hosts the `<TextInputView />`

The other good thing about this vs the rename is that the `<TextInputView />` is agnostic about styling, so we can configure whatever style we want that fits in well by tweaking the container.

Hope that helps with some ideas! It seems like this is pretty close to being ready, so it might be worth implementing the add / rename functionality in a separate PR, just so this one isn't blocked 😄 Excited to start using it!

> Good idea re file size tbh I was thinking there are bound to be cases where we should handle things differently but couldn't come up with any examples based on how I work, ill add in a clause to say if a file is over..(not sure how big would be problematic here?)

Yes, great point about the file size @CrossR ! Might be worth making this configurable? Like `explorer.maxUndoFileSizeInBytes` or something.

Akin909:
@bryphe apologies for the sudden rename which has increased the surface area of this PR, I renamed the `ui-tests` to `jest-tests` as part of a migration strategy I thought that at least for **unit tests** contributors could try adding new tests as `jest` tests, which down the line will reduce the amount of tests that will need to be converted.

Akin909:
@bryphe @CrossR think I'm done with this PR (if you want to try it out locally), if I work on it any longer im sure I'll keep finding more and more things to tweak/add although it will probably conflict with master once #2077 is merged in.

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -744,7 +744,7 @@ export class NeovimEditor extends Editor implements IEditor {
 
         document.body.ondrop = ev => {
             ev.preventDefault()
-            ev.stopPropagation()
+            // ev.stopPropagation()
```
bryphe: Will this break drag-and-drop on the editor? It might be worth separating this out into a full fix for #2070 - I'm assuming it was probably commented out for testing 😄

Akin909: @bryphe yes apologies I mean to remove it actually it fixes the *explorer* drag and drop issues but not sure of its impact on dragging files onto oni was just trying to see if this was the cause of the regression

browser/src/Services/Explorer/ExplorerSplit.tsx
```diff
@@ -212,26 +261,85 @@ export class ExplorerSplit {
...
+        if (!selectedItem) {
+            return
+        }
+        this._store.dispatch({ type: "DELETE", target: selectedItem, persist: false })
+        this._sendDeletionNotification(selectedItem.name)
+    }
+
+    private _sendDeletionNotification(name: string): void {
+        this.sendExplorerNotification({
+            title: `${name} deleted`,
```
bryphe: This notification looked a bit awkward to me - with the tile saying like "File1.txt deleted" and "File1.txt removed" in the body. The other notifications have a more general title - like "File moved" and then the body would have specifics, like: "File test.txt was moved to E:/oni/test_directory successfully".

Perhaps we could change this to:
```
title: "File deleted",
details: `${name} was deleted successfully`
```

browser/src/Services/Explorer/ExplorerStore.ts
```diff
@@ -109,6 +179,105 @@ export const rootFolderReducer: Reducer<IFolderState> = (
...
+        case "YANK":
+            return {
+                ...state,
+                yank: [...state.yank, action.target],
+            }
+        case "PASTE":
+            return {
+                ...state,
+                paste: action.target,
+                undo: [...state.undo, action],
```
bryphe: Cool! This reminds me of the undo example in the redux docs 😄  Tried it out and it works well, I like it!

browser/test/Services/Explorer/ExplorerStoreTests.ts
```diff
@@ -49,4 +63,38 @@ describe("ExplorerStore", () => {
...
+            await TestHelpers.waitForAllAsyncOperations()
+            // three because an init action is sent first
+            await assert.ok(actions.length === 3)
+            const clearedRegister = !!actions.find(action => action.type === "CLEAR_REGISTER")
+            assert.ok(clearedRegister)
+        })
+    })
+
+    describe("Store utility helper tests", () => {
+        it("Should not add an action if it isn't set to persist", () => {
```
bryphe: Thanks for adding tests here!

browser/src/Services/Explorer/ExplorerView.tsx
```diff
@@ -63,6 +66,31 @@ interface IMoveNode {
     }
 }
 
+const NodeTransitionWrapper = styled.div`
```
bryphe: Nice! Glad you were able to get this working 👍 Good example for how we can incorporate animations across our UI.

package.json
```diff
@@ -870,7 +870,7 @@
         "shell-env": "^0.3.0",
         "shelljs": "0.7.7",
         "styled-components": "^2.3.0",
-        "typescript": "2.7.1",
+        "typescript": "^2.8.1",
```
bryphe: Just curious - was there a part of this change that required moving to TypeScript 2.8?

jest.config.js
```diff
@@ -1,14 +1,16 @@
 module.exports = {
+    bail: true,
     verbose: true,
     collectCoverage: true,
-    collectCoverageFrom: ["**/*.{tsx}", "!**/node_modules/**", "!**/dist/**"],
-    setupFiles: ["<rootDir>/ui-tests/jestsetup.ts"],
+    coverageDirectory: "<rootDir>/coverage/",
+    collectCoverageFrom: ["**/jest-tests/*.{tsx}", "!**/node_modules/**", "!**/dist/**"],
```
bryphe: Does this enable code coverage for our `tsx` files now?

package.json
Akin909: That was the result of a very long story involving trying to neaten the typing of the epics since there are a lot of quirks that result in type casting, I left it in as I thought it would spare someone the trouble of having to update ts later on (has proven stable with no issues)

bryphe:
Just tried it out, I like it a lot @Akin909 ! Great to see the animations working 👍 

> @bryphe apologies for the sudden rename which has increased the surface area of this PR, I renamed the ui-tests to jest-tests as part of a migration strategy I thought that at least for unit tests contributors could try adding new tests as jest tests, which down the line will reduce the amount of tests that will need to be converted.

Sounds reasonable to add new tests in the jest strategy. Would prefer a separate PR though for it just to keep the size down - this PR is already big! 😄 

Overall things look good to me though. Appreciate you adding test coverage for some of the new scenarios. A few things I wanted to check on:
- Was commenting out the `evt.stopPropagation` intentional? It's not clear to me if that would be a complete fix for #2070, and if not, it'd be better to move it to a separate PR
- The text for the delete notification looked a bit out-of-place to me compared to the other notifications we had - I left a suggestion in the PR comments.
- A few other random questions too (around typescript/jest).

Really great to see the progress here! Big step into making our explorer a usable replacement for NERDTree. Thanks for all your work here @Akin909 ! 💯 

Now we just need search (#1652) and the ability to add files/folders to the explorer, and it'll have everything I need 😎

jest.config.js
Akin909: On going experimentation I though on reading the docs that just adding that might be enough basically `jest` works with istanbul built in and can output coverage reports even in ts straight forwardly apparently but not sure if they are being picked up or im missing a step somewhere.

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
bryphe: Cool, that's what I figured! Ya, in that case, I think it makes sense to remove it for now, just so we can evaluate the impact on dragging files into the editor as part of a complete fix for #2070

browser/src/Utility.ts
```diff
@@ -246,9 +246,13 @@ export function ignoreWhilePendingPromise<T, U>(
...
 }
 
-export function checkIfFileExistsSync(filename: string): boolean | Error {
+export function checkIfPathExists(
+    fileOrFolder: string,
+    type: "file" | "folder" = "file",
+): boolean | Error {
     try {
-        return fs.statSync(filename).isFile()
+        const stats = fs.statSync(fileOrFolder)
```
DeltaEvo: Awww sync code :cry:

browser/src/Services/Explorer/ExplorerFileSystem.ts
```diff
@@ -24,12 +32,21 @@ export class FileSystem implements IFileSystem {
...
         exists(path: string): Promise<boolean>
     }
+    private _backupDirectory = path.join(tempdir(), "oni_backup")
+
+    public get backupDir(): string {
+        return this._backupDirectory
+    }
 
-    constructor(nfs: typeof fs) {
+    constructor(nfs: typeof fs, _promisify: typeof promisify) {
```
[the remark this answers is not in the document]
Akin909: @DeltaEvo `jest` isn't resolving `promisify` since its used inside of the class, so I'm passing it in explicitly so I can use it in the test.

My most recent changes btw are a work in progress as I need to resolve changes to the filesystem.

browser/src/Utility.ts
Akin909: @DeltaEvo my changes here are a work in progress, I'm switching things over to use the async methods since your PR just working on resolving the issues with tests, once I push local changes that call will be async

browser/src/Services/Explorer/ExplorerFileSystem.ts
```diff
@@ -58,4 +75,77 @@ export class FileSystem implements IFileSystem {
...
+     * Delete a file or Folder
+     *
+     * @name deleteNode
+     * @function
+     * @param {ExplorerNode} node The file or folder node
+     */
+    public deleteNode = (node: ExplorerNode): void => {
+        switch (node.type) {
+            case "folder":
+                rm("-rf", node.folderPath)
```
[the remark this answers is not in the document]
Akin909: I can have a look at how those commands work as an alternative

Yes I noted unfortunately that all calls were sync and theres a thread on their repo to add async version but they haven't implemented these

browser/src/Services/Explorer/ExplorerFileSystem.ts
```diff
@@ -58,4 +75,77 @@ export class FileSystem implements IFileSystem {
...
     }
+
+    /**
+     * Delete a file or Folder
+     *
+     * @name deleteNode
+     * @function
+     * @param {ExplorerNode} node The file or folder node
+     */
+    public deleteNode = (node: ExplorerNode): void => {
```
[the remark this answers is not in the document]
Akin909: @DeltaEvo do you think it will make a difference to the functionality here? they point to the right `this` in this scenario

browser/src/Services/Explorer/ExplorerStore.ts
```diff
@@ -221,11 +482,18 @@ const expandDirectoryEpic = (fileSystem: IFileSystem): Epic<ExplorerAction, IExp
     })
 
 export const createStore = (fileSystem?: IFileSystem): Store<IExplorerState> => {
-    fileSystem = fileSystem || new FileSystem(fs)
+    fileSystem = fileSystem || OniFileSystem
```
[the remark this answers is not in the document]
Akin909: good idea

browser/src/Services/Explorer/ExplorerFileSystem.ts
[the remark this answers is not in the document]
DeltaEvo: Yeah i know, the behavior is not different, the only thing different is that you can't set another this for this function
I was asking just because we are in a class and using class functions seems less "hacky"

browser/src/Utility.ts
DeltaEvo: :heart:

browser/src/Services/Explorer/ExplorerFileSystem.ts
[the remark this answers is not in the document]
Akin909: Tbh using arrow functions in classes is something I do primarily because of all the hiccups with losing the correct this bindings in react class component methods.

Personally when I use an arrow function in a class which is almost paradigmatic in react (to avoid multiple lines of [this binding in constructor functions](https://codeburst.io/good-bad-and-ugly-parts-es6-class-arrow-functions-and-react-1ead4fa2c763)), it's primarily to signal that I intended for a function to be used with the this of where it was defined.

I like to avoid situations where `this` is changed as that for me has lead to more bugs than not.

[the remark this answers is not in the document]
parkerault: I will vouch that using arrow functions to avoid manual binding of class methods is idiomatic react, FWIW. 😄

parkerault:
Sorry to butt in, but if you are adding yank and paste, maybe `dd` to delete a file would also be appropriate? This looks awesome BTW.

browser/src/Services/Explorer/ExplorerFileSystem.ts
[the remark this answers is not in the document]
DeltaEvo: Uw okay thanks for the explanation, so in this case can you change the three methods that are in this class to use arrow functions too ? :heart:

Akin909:
@parkerault this pr includes delete functionality sorry the name doesn't indicate that its currently bound to `d` but can be rebound however users would like, there been some discussion in the original issue re the bindings as oni currently doesnt take chorded input like `dd` which would have been my first choice.

browser/src/Services/Explorer/ExplorerFileSystem.ts
```diff
@@ -58,4 +80,89 @@ export class FileSystem implements IFileSystem {
...
+     * Delete a file or Folder
+     *
+     * @name deleteNode
+     * @function
+     * @param {ExplorerNode} node The file or folder node
+     */
+    public deleteNode = async (node: ExplorerNode): Promise<void> => {
+        switch (node.type) {
+            case "folder":
+                await remove(node.folderPath)
```
DeltaEvo: fs-extra dosen't let us to set the inner fs module used for it (Seems to be a pr pending https://github.com/jprichardson/node-fs-extra/pull/558), so we are always removing file on disk ignoring the fs arg, it was why I was sugesting rimraf because it let us specify the fs instance to use

For mv however i don't find a good module for it, but it could be implemented pretty easily if we upgrade electron (Give us fs.copyFile) https://github.com/sindresorhus/move-file/blob/master/index.js#L9-L34

Akin909: Thanks for the feedback, I actually had a look at alternatives like `rimraf` and co initially, but *decided* to go with `fs-extra` because not only is it a well-typed, maintained package, it gives us all of `fs` functionality already promisified, adds on approx 12 new v. useful `async` methods not just for this PR but for upcoming changes like `rename`, `add` and the filesystem watcher related functionality.

I think a green PR on the v. active repo is a good sign that they will implement this functionality so we aren't likely to be missing this for long, vs. the trade off using several different packages to patch together the same functionality that this *one* provides.

Also a big goal of this PR is to get people to use the `persistable delete` I implemented which means actually moving files to the tmp dir (possibly configurable later) as opposed to deleting them from the drive hopefully the actually delete would be used only when users really want a file gone in which case I think it is valid to have it actually removed from the system.

Tbh switching packages from `shelljs` was actually not trivial and led to an almost complete rewrite of this PR because of redux-observable and how it handles promises and while any further changes might not be that big i'm looking to close this out once I've added more tests so if it feels necessary later to change things then that can come when/if a change seems warranted.

Akin909:
@bryphe I've made the requested changes, and did a bit of a refactor after changes to the explorer file system caused some unexpected observable issues.

* I've moved the fs methods into the `ExplorerFileSystem` and out of the `ExplorerSplit` that component now just dispatches actions which are co-ordinated by epics (added epics for most of the register (yank, paste etc) functionality).

I've left the jest changes in as I think I ought to have included them in the original PR and while I ideally like to keep all unrelated changes out if I can, I've found that I often intend to make a small pr for something like that for example but inevitably forget (hence my sneaking it in here)

Similar as above with the TS version, but also I added it to use the new `conditional types` as I wanted to make the use of multiple action types in the `ExplorerStore` simpler but I dont think I understand them well enough to apply it yet. On the other hand a small but (*for me*) very welcome change it brings is the addition of `numeric separators` aka `60_000` vs `60000` I've found them soooo beneficial to readablility and used them in a few places here.

Akin909:
Update need to change some error handling in the epics due to a quirk with the way the are currently handled (almost there but not quite)

CrossR:
Started to use this branch as my daily version to give it a look over and its worked well so far!

bryphe:
Excellent! Changes look great to me 👍 

This is awesome stuff... feels like our explorer is really growing up thanks to all your work, @Akin909 ! 💯

PR #1852, opened 2018-03-19
Highlight text crash fix

### LINKED ISSUE #1842: Can't use numbers in Oni menus
If I use numbers in an Oni menu, there is an error, which causes all of Oni to crash.
Its due to the highlighting expecting a string, and instead receiving a number and calling split on that.

I've got a PR (and attempted some React tests @Akin909!) ready to fix this first bit, but does it make more sense to just set the non-string object to `""` or attempt a `.toString()`? Feels like we should assume they give a string, and fail otherwise, since attempting to stringify could leave us in a bad state if we get an object.

Second issue here I guess is potentially why a `.split("")` manages to take out the entire UI.

Example:

```
var makeTestMenu = function () {
        var testMenu = oni.menu.create();
        var menuItems = [
            {
                icon: "window-restore",
                detail: 2,
                label: 1
            },
            {
                icon: "window-restore",
                detail: 2,
                label: 2
            }
        ];
        testMenu.show();
        testMenu.setItems(menuItems);
        testMenu.onItemSelected.subscribe(function (menuItem) {
            if (menuItem) {
                console.log(menuItem.detail);
            }
        });
    };
    oni.input.bind("<s-c-x>", makeTestMenu);
```

CrossR:
Partially fixes #1842.

If a string isn't passed in, we set it `""`.
Seems this is the best option since otherwise we would need a messy check I think to check the input, especially when its called in the users' config so we have no control over it.

I've attempted to add some React tests....though I'll be the first to admit I've never used them, so hopefully they look okay!

### DIFF
--- a/ui-tests/HighlightText.test.tsx
+++ b/ui-tests/HighlightText.test.tsx
@@ -21,21 +21,46 @@ const initialState = {
 }
 
 describe("<HighlightTextByIndex />", () => {
-    const HighlightTextIndexComponent = <HighlightTextByIndex {...initialState} />
-
     it("renders a shallow instance of the component", () => {
-        const wrapper = shallow(<HighlightTextByIndex {...initialState} />)
-        expect(wrapper.length).toEqual(1)
+        const component = shallow(<HighlightTextByIndex {...initialState} />)
+
+        expect(component.length).toEqual(1)
     })
 
-    it("renders the correct text", () => {
-        const wrapper = mount(<HighlightTextByIndex {...initialState} />)
-        expect(wrapper.text()).toContain("highlight text")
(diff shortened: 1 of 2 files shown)


ui-tests/HighlightText.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+        const component = mount(<HighlightTextByIndex {...testState} />)
+
+        // Check the correct text is there
+        expect(component.text()).toContain("no highlight text")
+
+        // Check the structure is correct
+        expect(component.text()).toHaveLength(17)
+
+        // Check only 4 chars were highlighed
```
CrossR: Copy-paste error!

PR #1858, opened 2018-03-20
Enhancement(small)/ update title bar background color for one-dark

Akin909:
This is I hope a boon to all `onedark` users I've noted that at least on macos where the title bar is set to a color the onedark theme has muddy brown title bar which I think *personally* isn't really in keeping with the theme, or how the title bar is rendered in other editors where I've seen the theme used

#### Oni default:
screen shot 2018-03-20 at 10 13 08

#### Proposed change: 
screen shot 2018-03-20 at 10 12 15

#### Vscode: (as an example, in the proposal I opted for a darker title which is what atom does)
screen shot 2018-03-20 at 10 15 55

It's a little hard to see the difference due to the quality of the images but I think if you switch to onedark in one and compare it to atom or vscode the difference is obvious. This is a minor color tweak ive been using locally for months to achieve the same effect.

### DIFF
--- a/extensions/theme-onedark/colors/onedark.json
+++ b/extensions/theme-onedark/colors/onedark.json
@@ -5,7 +5,7 @@
         "background": "#1E2127",
         "foreground": "#ABB2BF",
 
-        "title.background": "#1F1F1F",
+        "title.background": "#1E2127",
         "title.foreground": "#ABB2BF",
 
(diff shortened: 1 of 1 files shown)


bryphe:
Looks great! Thanks for making the change, @Akin909 👍

Issue #1860, opened 2018-03-20
Explorer filepath not set when called from command line


importjake:
Calling `oni .` for example from within a project does not set the explorer to be active on the supplied directory. Example (I was calling `oni .` in a project directory of mine named `cryptrack-client`) but when oni opened the explorer was active on my previous open project (a typescript workshop).

![screen shot 2018-03-20 at 8 24 31 am](https://user-images.githubusercontent.com/3849309/37664572-ce7948dc-2c18-11e8-93be-092b9c6c25c1.png)

CrossR:
Hi,

The command line calling of Oni isn't quite where it should be yet...
There is a comprehensive list of proposed changes over in #1707, which I think addresses the issue you have here.

Setting `  "workspace.autoDetectWorkspace": "always",` will get you part of the way there - It will auto change the workspace to the current files project, but I think currently it doesn't support opening folders. Ie `oni .` won't change the explorer workspace until you load a file. `oni LICENSE` would auto change it.

We could use this ticket to track the small change of having that option work with folders, and use the other for the bigger changes that are needed to the commandline.

badosu:
Closing this in favour of https://github.com/onivim/oni/issues/2163 and #1707 to track cli usage of Oni.

Issue #1707, opened 2018-03-04
RFC: Command line usage


13k:
Disclaimer: beware, wall of text ahead.

## Abstract

Command line usage is pretty much ingrained in my workflows, as I suspect is for most vim users. Currently, the `oni` command lacks much of what would make it good for such usage.

I'm new to both neovim and oni, but I've used vim for more than 12 years and I want to switch to a modern version of vim. I tried both atom and vscode with vim emulation extensions, but it doesn't live up to expectations. 

I have installed the rpm release package in Fedora, and everything I say here is in regards to using `oni` from the command line when installing it from release packages (not the npm package, neither from a dev checkout).

## Motivation

I always start editors from the command line, mostly because:

a) I use the terminal when working on a project. I have tmux setup with sessions spawning panes and windows accordingly for each project. There's also a myriad of other command line tools that are used in development workflows.
b) To use the same environment in the editor as the environment set up in `a)`. For example, variables set up according to several environment managers (pyenv, rbenv, nvm, gimme, rustup, etc). This enables development tools (debuggers, linters, language servers, etc) to be correctly detected, on a per-project basis.

## Concerns

The release package installs a symlink to the electron binary in the system `$PATH`, bypassing `cli/oni` and spawning the electron app directly.

So my main issues are:

1) Regardless of `cli/oni` [being changed to spawn the electron process in the background](https://github.com/onivim/oni/pull/150), it's never used and the editor is always launched in the foreground;
2) Command line options seem to be ignored or don't exist at all (correct if I'm wrong);
3) Workspace detection seems non-intuitive

## Proposal

Starting the editor from command line should be treated as a separate concern from the regular app launch. When starting from the command line, it should:

* Dettach to background by default, since it's not a terminal application;
* Provide command line options with help;
* Provide a command line option to keep in the foreground (named `foreground` for reference), like `gvim -f`. This makes it possible for `oni` to be used in command line workflows that requires spawning an editor and waiting for termination (like `git mergetool` or setting `VISUAL=oni -f`);
* Provide command line option to open a buffer at a specific location (named `goto` for reference), similar to `vim +line <file>`;
* Apply specific workspace detection (see below);

### Workspace detection

When detecting workspaces, `$PWD` should be taken into consideration. Therefore, detection logic could be:

* if no arguments are given: open an empty buffer, detect workspace using `$PWD`. If it fails, don't use any workspace
[...]

PR #1866, opened 2018-03-20
Change navigateToProjectRoot to accept folders.

CrossR:
Proposed partial fix for #1860.

This would allow `oni .` to load the current folder, whereas currently it needs a file instead.
This means with `"workspace.autoDetectWorkspace": "always",` opening with `oni .` would load the current folder.

Behaviour when `"workspace.autoDetectWorkspace"` is not `"always"` should not be affected.
I could stick some tests around this, but I think due to its interfacing with the file system, they would need to be CI tests.

This would go along side any bigger changes that are needed in #1707.

### DIFF
--- a/browser/src/Services/Workspace/Workspace.ts
+++ b/browser/src/Services/Workspace/Workspace.ts
@@ -130,16 +130,11 @@ export class Workspace implements IWorkspace {
 
     public navigateToProjectRoot = async (bufferPath: string) => {
         const projectMarkers = this._configuration.getValue("workspace.autoDetectRootFiles")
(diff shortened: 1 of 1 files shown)


CrossR:
Yep, works fine for me on Windows, I'll stick a copy of it on my Arch or Ubuntu VMs and test there, but works fine on Windows at least.

`oni .` sets the workspace to that folder (assuming there is a project root file in there).

CrossR:
I've given this a test on my Arch box now and it works fine, so I'll bring this in now.

PR #1879, opened 2018-03-21
Crop command details in command menu when too long

### LINKED ISSUE #1560: Shortcut hints on the command pallet got cropped when the window size is small
(Just saw Oni's "What's new in 0.3.0" video and realized that the command pallet had been long available. 😉 Really helpful video for the slow-witted me.)

When I started to explore Oni's shortcuts with command pallet, I noticed that some shortcut hints are not fully displayed:

When I entered full-screen, the issue is solved, so there're probably some problems with the overflow strategy.

Cryza:
This fixes #1560 by truncating the details and inserting an ellipsis ("...") if the complete information does not fit into the command line width.

I decided against migrating these files from `.less` to `styled-components` as the change was so simple and I wanted to get it out of the way first. I will migrate the `.less` files within another PR. I hope that is okay :)

### DIFF
--- a/browser/src/Services/Menu/Menu.less
+++ b/browser/src/Services/Menu/Menu.less
@@ -67,6 +67,9 @@
                 font-size: @font-size-small;
                 color: @text-color-detail;
                 flex: 1 1 auto;
+                overflow: hidden;
+                text-overflow: ellipsis;
+                padding-right: 8px;
             }
         }
     }


bryphe:
@Cryza - looks great! 👍 

> I decided against migrating these files from .less to styled-components as the change was so simple and I wanted to get it out of the way first. I will migrate the .less files within another PR. I hope that is okay :)

For sure, this is perfect. Makes sense to keep it scoped 😄 

Bringing it in now - thanks for the contribution!

PR #1882, opened 2018-03-21
Fix drag and drop.

### LINKED ISSUE #471: Feature request: Open a file by dragging the file to the panel

CrossR:
Fix #471.

I've also updated how files are opened such that it is more inline with other editors - ie open the file in a new tab, rather than in the existing one, except if the current tab is empty.

Is this even possible to automatically test?
There is a `TODO:` above this code for it to be refactored out into a react component, at which point I assume we could use the  react testing stuff to test it, but for now I've no idea how this could be tested.

### DIFF
--- a/main/src/menu.ts
+++ b/main/src/menu.ts
@@ -38,12 +38,8 @@ export const buildMenu = (mainWindow, loadInit) => {
         })
     }
 
-    const executeVimCommandForMultipleFiles = (
-        browserWindow: BrowserWindow,
-        command: string,
-        files: string[],
-    ) => {
-        executeMenuAction(browserWindow, { evt: "open-files", cmd: [command, files] })
+    const openMultipleFiles = (browserWindow: BrowserWindow, files: string[]) => {
+        executeMenuAction(browserWindow, { evt: "open-files", cmd: [files] })
     }
 
     const executeOniCommand = (browserWindow: BrowserWindow, command: string) => {
@@ -165,8 +161,7 @@ export const buildMenu = (mainWindow, loadInit) => {
                     dialog.showOpenDialog(
                         focusedWindow,
                         { properties: ["openFile", "multiSelections"] },
-                        files =>
-                            executeVimCommandForMultipleFiles(focusedWindow, ":tabnew ", files),
+                        files => openMultipleFiles(focusedWindow, files),
                     )
                 },
             },
--- a/vim/core/oni-core-interop/plugin/init.vim
+++ b/vim/core/oni-core-interop/plugin/init.vim
@@ -110,16 +110,6 @@ function OniCommand(oniCommand)
     call OniNotify(["oni_command", a:oniCommand])
 endfunction
 
-function OniOpenFile(strategy, file)
-     if bufname('%') != ''
-         exec a:strategy . a:file
-     elseif &modified
-         exec a:strategy . a:file
-     else
-         exec ":e " . a:file
-     endif
- endfunction
-
 augroup OniClipboard
     autocmd!
     autocmd! TextYankPost * :call OniNotifyYank(v:event)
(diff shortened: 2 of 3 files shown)


Akin909:
@CrossR as a buffers user having this always open in a tab would *personally* steer me away from using this feature 😭, when it was working I think it was opening a new buffer could it take into account the users' settings? 🙇

CrossR:
Sure!

I did start refactoring it to use the `openFile` method that uses the `Oni.FileOpenOptions` options, but decided against it for a small change, but hadn't realised the issue for buffers users.

CrossR:
Refactored it to use the function that takes user config into consideration now!

Since I was there, I removed the old function that didn't and hooked the initial startup + Menu > Open file to use that as well, so they all take the users config into account as well. That also let me delete the VimL function it called.

main/src/menu.ts
```diff
@@ -38,12 +38,8 @@ export const buildMenu = (mainWindow, loadInit) => {
         })
     }
 
-    const executeVimCommandForMultipleFiles = (
-        browserWindow: BrowserWindow,
-        command: string,
-        files: string[],
-    ) => {
-        executeMenuAction(browserWindow, { evt: "open-files", cmd: [command, files] })
+    const openMultipleFiles = (browserWindow: BrowserWindow, files: string[]) => {
```
CrossR: Command wasn't needed here since the new function takes the users config into account.
Also renamed it since `executeVimCommandForMultipleFiles` that only called `"open-files"` felt misleading.

CrossR:
Only issue I've realised is that since we use `:e!` for the buffers mode, if a user started to edit the initial `[No Name]` buffer and then opened a file with the explorer or dragged a file, it would hide that buffer. Don't seem to have that issue with `tabs` mode.

That said, it seems like a separate issue since the `[No Name]`  buffer is still there, its in `:buffers` and  swappable to, so it just looks like we aren't surfacing it properly. I've noticed a second issue with them than the isDirty label is flaky on them too, so its probably something similar. I assume that is also what is causing the `buffers` mode to get into weird states occasionally, where no tab is highlighted, or the wrong tab is.

I can raise a second issue for that.

Akin909:
@CrossR do you mean that with the `:e!` command the newly opened file would be focused? is so that seemed like desirable functionality? an alternative is the `:badd` command which adds a buffer to the bufferlist without opening it if that isn't the desired outcome

Re behaviour of the `[No Name]` buffer I never use or edit it so not familiar with their behaviour, as to the buffers mode being in weird states I've not noticed that except once in a while an error is thrown relating to `activeTabPage.activeWindow.id` where it sometimes cant find it its really uncommon and I added a check for it a while ago but it still resurfaces very weirdly.

Think all in all there still a few issues which buffers, buffer events etc.

CrossR:
I've written it currently to work as follows:
* If there is a file open, then open  first file in a new buffer/tab.
  * This could be easily given a config option if necessary, to disable it.
* If there isn't a file open, then open first file in the current buffer/tab.
* Any other files open in a new buffer/tab.

Which works fine, except in the case where no file is open, but an edit has been made to the `[No Name]` buffer. In that case, it will hide the buffer causing it to look gone, when really its just not being shown.

Ie :

* Open Oni
* Edit the `[No Name]` buffer whilst in `tabs.mode` as `buffers`.
* Drag a file on.
* The file you drag on will take focus, which is fine. But the `[No Name]` buffers is "gone".

That said, that is currently the case if you make edits in Oni and then manually enter `:e! FILE`, so I think it can be picked up in a separate issue.

vim/core/oni-core-interop/plugin/init.vim
```diff
@@ -110,16 +110,6 @@ function OniCommand(oniCommand)
     call OniNotify(["oni_command", a:oniCommand])
 endfunction
 
-function OniOpenFile(strategy, file)
```
bryphe: Sweet! That's great we can handle this in the JS layer now 👍

bryphe:
Awesome, thanks for the fix @CrossR ! 👍 I like seeing stuff in our core `init.vim` go away 😄

Issue #1414, opened 2018-02-01
Sidebar: Allow resize


badosu:
Allow resizing of the sidebar grapping the separator.

bryphe:
Makes sense, thanks for the suggestion @badosu . Once the API work in #1412 is in, this shouldn't be too difficult to integrate.

badosu:
An update: this is how the Achievements pane is looking for me on the Sidebar. I am unable to resize it:

![screenshot from 2018-04-04 09-06-31](https://user-images.githubusercontent.com/347552/38306624-89d112e8-37e7-11e8-9747-71ffe20933ba.png)

badosu:
Since https://github.com/onivim/oni/issues/2205 was merged, let's use this issue to track resizing the sidebar via grabbing the separator.

PR #1893, opened 2018-03-22
Bugfix: Input being swallowed when switching between multiplexed editors

bryphe:
__Repro:__
- Set `editor.split.mode` to `'oni'`.
- Restart Oni
- Edit user config
- Try and make changes to the configuration file - specifically, type in characters that would engage autoclosing pairs functionality (`(`, `<enter>`, etc)

__Expected:__ Should function as normally without being in oni split mode
__Actual:__ Some of the input gets sent to the leftmost editor, making it impossible to do anything in the right editor.

This bug only occurs when `editor.split.mode` is set to `'oni'` (so off by default), and is a class of issues - things that are using `activeEditor` will only handle the _original_ editor. 

This is fine and the code is working as designed, in the non-multiplexed world, since we only ever had one editor. The bug is that, in the 'multiplexed' world, we can now have multiple editors, and the feature hasn't been updated to handle it. 

In this case, the `anyEditor` is a convenient proxy that always points to the _currently_ focused editor - so the fix here is to make the `AutoClosingPairs` functionality multiplexed-aware by switching to `anyEditor` vs `activeEditor` in places that are deferring functionality. For example, for input bindings that might get executed later, we should use `anyEditor`, since the actual editor might be different depending on which editor has focus (and the bindings apply across all editors).

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -757,6 +757,10 @@ export class NeovimEditor extends Editor implements IEditor {
 
         this._neovimInstance.autoCommands.executeAutoCommand("FocusGained")
         this.checkAutoRead()
+
+        if (this.activeBuffer) {
+            this.notifyBufferEnter(this.activeBuffer)
+        }
     }
 
     public checkAutoRead(): void {
--- a/browser/src/Editor/OniEditor/OniEditor.tsx
+++ b/browser/src/Editor/OniEditor/OniEditor.tsx
@@ -188,9 +188,9 @@ export class OniEditor extends Utility.Disposable implements IEditor {
 
     public enter(): void {
         Log.info("[OniEditor::enter]")
-        this._neovimEditor.enter()
-
         editorManager.setActiveEditor(this)
+
+        this._neovimEditor.enter()
 
         commandManager.registerCommand({
             command: "editor.split.horizontal",
--- a/browser/src/Services/AutoClosingPairs.ts
+++ b/browser/src/Services/AutoClosingPairs.ts
@@ -162,7 +162,7 @@ export const activate = (
                 subscriptions.push(
                     inputManager.bind(
                         pair.open,
-                        handleOpenCharacter(pair, editorManager.anyEditor, true),
+                        handleOpenCharacter(pair, editorManager.activeEditor, true),
                         insertModeFilter,
                     ),
                 )
@@ -171,14 +171,14 @@ export const activate = (
             subscriptions.push(
                 inputManager.bind(
                     pair.open,
-                    handleOpenCharacter(pair, editorManager.anyEditor, false),
+                    handleOpenCharacter(pair, editorManager.activeEditor, false),
                     insertModeFilter,
                 ),
             )
             subscriptions.push(
                 inputManager.bind(
                     pair.close,
-                    handleCloseCharacter(pair, editorManager.anyEditor),
+                    handleCloseCharacter(pair, editorManager.activeEditor),
                     insertModeFilter,
                 ),
             )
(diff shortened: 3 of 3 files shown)


bryphe:
Updated the fix - the `onBufferEnter` method is actually correct, because it is called on buffer enter (so the activeEditor is always going to be the one currently focused). However, listening to the `onBufferEnter` event has to happen for the `anyEditor`, so that we pick up the buffer enter event for _all_ editors. That event wasn't getting fired correctly (we need to fire it on focus too, between the multiplexeed editors, so that it behaves the same as non-oni splits today).

parkerault:
@bryphe I was testing this out this afternoon and aside from the issue with input going to the wrong editors, the resize commands don't work (`C-w C-<` and `:vertical resize n`), and when you have multiple oni splits open, `C-w C-q` will close all the splits and the oni window, which is not what I would expect.

bryphe:
@parkerault - sweet, thanks a lot for testing it out!

Ah yes, there are still some resize issues. We need to make our 'Oni-controlled' windows resizable. There's #1412 and #1414 that are tracking similiar things.

One unknown in my head is how we'll handle these sorts of cases:
oni-screenshot-1521769633255

Nothing to stop the user from having some crazy combination of 'oni splits' and 'vim splits'! 😄 Will be an interesting problem to rationalize how that sizing logic should work in that world..

The `C-w C-q` case is a good catch - I don't think we have an issue tracking the `C-w C-q` case - would be great to have an issue. I started this project here: https://github.com/onivim/oni/projects/13 to track the remaining work / blockers to enable this setting by default.

parkerault:
Since I haven't looked into how Oni integrates with nvim I'm not sure how much control you have, but is it possible to hijack all splits and make them oni splits when they are created by vim? Otherwise I guess you are always going to be limited to user-created splits. Is the use case for oni tabs primarily to avoid confusing vim newbies?

parkerault:
@bryphe I created an issue for the split closing bug here: https://github.com/onivim/oni/issues/1899

bryphe:
Sweet, thanks @parkerault ! I added it to the externalized UI project

> Since I haven't looked into how Oni integrates with nvim I'm not sure how much control you have, but is it possible to hijack all splits and make them oni splits when they are created by vim?

I think this is tricky business - if we hijack splits, vim plugins that try to open splits won't be aware of the new split we create (since it's in a new neovim process).

>  Is the use case for oni tabs primarily to avoid confusing vim newbies?

There's a few reasons - one is it is more like the way atom / vscode / etc behave, so it is easier on newbies. But also, one thing you'll notice, is that things like our scrollbar, error adorners, etc are only visible on the _active window_ - this is because certain metrics we need to position our UI are only available for the active window. So by 'multiplexing' we can actually provide a better experience - if our windows are 1-to-1 with the vim windows, we can have a more consistent experience across the splits.

We usually split our user personas into two cases - the 'new user' who maybe doesn't care as much about the esoteric vim functionality, and the 'pro vim user' - who knows vim in and out and doesn't want to change their workflow. For the latter case, we're okay with potentially having them turn off settings - in fact @badosu put together this great [bare vim config](https://github.com/onivim/oni/wiki/How-To:-Make-Oni-closer-to-bare-Vim-experience) for us.

So we might be able to strike a compromise, perhaps change `C-w C-<` to by default resize the _oni split_ if `editor.split.mode` is `'oni'` (and have the native vim behavior otherwise). We probably wouldn't be able to hook the `:vertical resize n`, but that's sort of the 'pro vim' use case, so it may be ok to just leave that as-is. 

It'd also be nice to have draggable handles for the window splits, too, for the 'new user' case 😄 Users coming from VSCode/Atom would definitely expect it.

But nothing set in stone... still need to figure out how we can best balance that!

bryphe:
I'll bring this in now since it unblocks the input issue - still more work to do in #1412 and #1899 to get this stabilized, though! Thanks again for trying it out, @parkerault !

PR #1896, opened 2018-03-22
Configuration: Add configuration setting / metadata API

bryphe:
This change adds a `registerSetting` API to register a configuration setting, along with metadata. The metadata will be used to help populate the configuration helper page we have. It will also help in terms of deciding when we need to reload oni vs can handle the configuration update in the current session, with the `requireReload` property.

I'm picturing additional pieces of metadata being useful in the future, too. One piece in particular (which is stubbed out here, but not implemented) is a _merge strategy_ - when multiple configuration providers specify a value (for example, user config vs workspace config), the current strategy is by default to overwrite. However, depending on the configuration setting, there might be different ways to resolve this - this allows specifying a custom lambda to reconcile the settings.

### DIFF
--- a/browser/src/Services/Configuration/Configuration.ts
+++ b/browser/src/Services/Configuration/Configuration.ts
@@ -111,7 +111,7 @@ export class Configuration implements Oni.Configuration {
 
     private _settingMetadata: { [settingName: string]: IConfigurationSettingMetadata<any> } = {}
     private _subscriptions: {
-        [settingName: string]: Event<IConfigurationSettingValueChangedEvent<any>>[]
+        [settingName: string]: Array<Event<IConfigurationSettingValueChangedEvent<any>>>
     } = {}
 
     public get editor(): IConfigurationEditor {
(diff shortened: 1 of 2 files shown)


CrossR:
Is this intended for plugins, such that they can register a setting?
Or is it intended that all the settings move to this so we have descriptions for everything?

bryphe:
Great question - I should've called this out in the description.

I'd like to incrementally move our settings to use this. The `IConfigurationValues`/`DefaultConfigurationValues` doesn't scale to plugin settings is one reason - but also, it'd be useful to know whether a reload is enabled for configuration settings so that we can help prompt the user and make it clear when they reload the config or not. 

The settings I'd like to move over to start are ones where we can set `reloadRequired` to `false` (like `tabs.mode`) - this way, we can provide a better experience to users when they update their config, and let them know via a notification or something if they need to reload oni to pick up the change.

Plus, as we move settings w/ descriptions, we can get documentation across the board to populate in our new config experience that @rgehan implemented.

PR #1905, opened 2018-03-23
[WIP] Update to support 32 bit builds.

CrossR:
Update to support 32bit binaries.

Depends on both https://github.com/onivim/oni-release-downloader/pull/1 and https://github.com/onivim/oni-neovim-binaries/pull/2. I'll need to update the dependency and then do some testing with the binaries after that.

Tested making both on my machine, but will still need to check the compiled binaries after.
Also still need to update the build scripts, but that won't matter until the above PRs have been merged.


bryphe:
Looks great, @CrossR ! Thank you very much for taking this on. I'll go check out those other PRs now.

bryphe:
I just updated packages:
- `oni-release-downloader@0.0.9`
- `oni-neovim-binaries@0.1.1`
So we'll need to update the `package.json` for it to work.

I wonder if we'll have to update `oni-ripgrep` too?
https://github.com/onivim/oni-ripgrep/blob/da2e385aaaded84d6872c450d268555d411e0f1a/lib/postinstall.js#L10

I believe the binary is `x86`, but we might need to explicitly add a `win64` section to the `postinstall.js` there.

bryphe:
Looks like we might have to update the `appveyor.yml` to pick up the new commands - right now, it runs `npm run dist:win`. We might need to switch it to a script that's aware of the environment variable - the `upload:dist` command has an example of using a nodejs script.

CrossR:
Updated this now, along with a few other scripts that used the old values.

Once (if) this builds correctly, can download them and test them on my PC and VM, as well as other things like the QuickOpen/Search etc.

build/script/appveyor-test.ps1
```diff
@@ -0,0 +1,28 @@
+# Output useful info for debugging.
```
bryphe: Nice! Thanks for factoring this out to a script. 👍

CrossR:
The build failed due to a few typos, but still passed on AppVeyor.

I'm guessing powershell is like bash in that it continues on errors by default, so I need to check the error code after each command, and exit with a non-zero error code if that is the case.

CrossR:
Yeah, I wasn't  able to use Ripgrep in the 32bit binary...
I'll take a look in a bit about sorting that. Will git etc be fine?

bryphe:
Sorry I missed this, @CrossR ! I just brought in your PR for `oni-ripgrep` and published `0.0.4` to `npm` - that should be set.

Seems like we're in good shape after that! 👍

CrossR:
Brought that change in now, so hopefully its sorted!

Last outstanding issue to be sorted in this PR or one soon after is updating the Powershell script to exit and throw an error if the test fails, so this doesn't happen : https://ci.appveyor.com/project/bryphe/oni/build/1.0.2877/job/j5sbnv6827yemott

bryphe:
> Brought that change in now, so hopefully its sorted!

Woohoo! Awesome, thanks @CrossR 👍 

> Last outstanding issue to be sorted in this PR or one soon after is updating the Powershell script to exit and throw an error if the test fails, so this doesn't happen : https://ci.appveyor.com/project/bryphe/oni/build/1.0.2877/job/j5sbnv6827yemott

Ah ya, good catch! We had that issue with the OSX/Linux builds for a while too - there'd be an error but it'd still happily report a pass.

CrossR:
Updated the script now to exit with an error on failure. It correctly failed when I broke the CI tests, so looks good.

Not sure about the issue on Linux, since it passed every test and only failed on the Binary download from the cache as far as I can tell?

Setting up a 32 bit VM at the mo to do some testing as well.

bryphe:
> Updated the script now to exit with an error on failure. It correctly failed when I broke the CI tests, so looks good.

Excellent! This is awesome - thank you for all your work on this, @CrossR ! Big step forward to have real `x86` and `x64` builds instead of our crazy hybrid build we have today 😄 

> EDIT: Tested and it works! I'm at home for a bit during my month off, so testing it on a laptop with an i3/4GB of RAM and inside a VM didn't give me the best look into the performance etc, but the QuickOpen + Search works at least.

Woohoo! I anticipate that using the correct `x64` build of electron on an `x64` machine (vs the `x86` we had) may actually be slightly faster, due to the optimized instruction set... not sure if it will be a tangible difference though. At least in some benchmarks: https://www.digitalcitizen.life/google-chrome-64-32-bit-performance-comparison the x64 build was a bit faster - not sure how it relates to the electron builds though. Looking forward to testing & trying it out!

Thanks again for all your help, @CrossR ! 💯

PR #1906, opened 2018-03-23
Achievements: Show 'notification bubble' to guide users to the learning pane

bryphe:
We've got an achievement that shows up on first launch, which is pretty cool... But it might not be clear to users how it's useful or what they can do with it.

I've added a notification bubble that pops up on the sidebar learning pane's icon - so that hopefully it will help bring users over there (and showcase the learning / achievements).

sidebar-notification

It's pretty subtle but I hope it will help new  users get on-boarded. Will be more useful when we have more content in that pane 😄


bryphe:
The API is also generalized, such that anything could add a notification to the sidebar - `Oni.sidebar.setNotification("pane-id")`

PR #1918, opened 2018-03-24
Add default configuration for Haskell LSP

Tehnix:
This adds a default configuration for the Haskell LSP, HIE.

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -193,6 +193,11 @@ const BaseConfiguration: IConfigurationValues = {
     "language.ocaml.languageServer.arguments": ["--stdio"],
     "language.ocaml.languageServer.configuration": ocamlAndReasonConfiguration,
 
+    "language.haskell.languageServer.command": "stack",
+    "language.haskell.languageServer.arguments": ["exec", "--", "hie", "--lsp"],
+    "language.haskell.languageServer.rootFiles": [".git"],
+    "language.haskell.languageServer.configuration": {},
+
     "language.typescript.completionTriggerCharacters": [".", "/", "\\"],
     "language.typescript.textMateGrammar": {
         ".ts": path.join(


Akin909:
@Tehnix, thanks for the PR 👍, Although i'm wondering whether this would go better in the documentation wiki re LSPs we have tended to put information re how to integrate new LSPs [`here`](https://github.com/onivim/oni/wiki/Language-Support), the reasoning behind this as @bryphe has pointed out is that the more lsps Oni ships with the bulkier the application, I think having it in the wiki like for `golang`, `c++` and `c-sharp` would be really great though with some installation instructions?

Tehnix:
That would also make sense, yeah :) I'll add it to the wiki page instead, and close this PR then.

Tehnix:
Perhaps a bit too quick on that one. I saw that `pyls` the python LSP is configured in the `DefaultConfiguration`, but still requires one to install it yourself.

Perhaps a mix of adding the configuration here, and then install instructions in the wiki entry? 

Let me know what you think.

CrossR:
We've got `pyls` and `clangd` defined in the code already, so I don't see any issues with adding another?

If we were bundling the executables that would be a bit different, but I think just including the config pre-entered seems fine to me.

Akin909:
tbh atm its a few lines and having it means a haskell user who has the lsp downloaded will have everything work automagically which is great.

Not sure what the long term plan is though, aka going forward will we have defaults for most user languages in which case if it isn't clear that were setting specific defaults then it could raise issues if our defaults clash with what a user ends up setting e.g. a user sets 2/4 of the lsp config opts since they want to use a (different) lsp for a language than our default then don't realise we've set another two defaults to something that's incompatible with their choice 🤷‍♂️.

I'm happy with this merged into the default config or as documentation will defer to whatever the prevailing preference is.

bryphe:
💯 , thanks for putting this together @Tehnix , and thanks for merging it in @Akin909 !

I didn't realize there is a haskell language server... I just read a book on haskell, I'll have to try this out 👍 Very cool!

> Not sure what the long term plan is though, aka going forward will we have defaults for most user languages in which case if it isn't clear that were setting specific defaults then it could raise issues if our defaults clash with what a user ends up setting

Long-term, once our plugin story is more mature, it'd be cool if we had language-specific plugins. It'd be great if they could bundle the language servers as well (like VSCode's plugins). It's a nice experience if you just add one plugin with minimal configuration and you're all set 👍 

Shorter-term, I think this approach is great - especially for cases like this where there is one primary language server. One thing I'd like to add to make this a bit nicer is to have some sort of message / URL that we can point a user to if the language server fails to start (for example, if the user doesn't have it installed). It'd be cool if we could put up a notification like "I see you're opening a Haskell file - Oni works great with the haskell language server. Click here to learn more.". Potentially we could add configuration settings like `language.<language-identifier>.languageServer.helpText` and  `language.<language-identifier>.languageServer.helpUrl`. Then, if a language server fails to launch, we look at the config and see if those are available, and if so, pop up a notification.

Preconf:
@Tehnix, @Akin909  Hello and thank you very much for adding support for Haskell Language Server in Oni. Unfortunately I am not able to get it to work. 

Oni version: 0.3.4-1
OS: Arch Linux
stack and HIE have been installed from the distro's repositories

First of all upon loading/creating a Haskell file I am getting errors:

1. Received notification for - Unhandled Rejection: Unable to find root csproj file
2. Uncaught (in promise) Unable to find root csproj file
3. I can also see a warning: No grammar found for language: haskell
4. Another warning: Error from tss: /bin/sh npm command not found

Not sure why is looking for a c# project file....

My directory structure is really simple for testing purposes:

Projects
  - .oni/config.js - custom config
  - hello.hs --> single haskell file

The contents of the configuration file are:

```
const activate = (oni) => {
   // access the Oni plugin API here
   
   // for example, unbind the default `<c-p>` action:
   oni.input.unbind("<c-p>")

   // or bind a new action:
   oni.input.bind("<c-enter>", () => alert("Pressed control enter"));
};

module.exports = {
    activate,
   "ui.colorscheme": "hybrid",
   "ui.animations.enabled": true,
   "ui.fontSmoothing": "auto",
  "language.haskell.languageServer.command": "hie",
  "language.haskell.languageServer.arguments": ["--lsp"],
  "language.haskell.languageServer.rootFiles": [".git"],
  "language.haskell.languageServer.configuration": {},
}
```

Any pointers to get this running properly would be greatly appreciated. I can provide you with more info if required so please let me know.

Tehnix:
I'm fairly certain HIE won't work without a project file, i.e. a .cabal file (and possibly a stack.yaml). Are you able to get HIE working with other editors (e.g. VSCode) in the same project?

Akin909:
@Tehnix thanks for fielding that I dont know HIE at all, @Preconf what I can say real quick re the `csproj` error you're seeing is its a hang over from the initial implementation and the error message is actually from the function which tries to identify the project root I have a PR which should fix how that function works so itll be less confusing/will work better

Tehnix:
@Akin909 No worries, it's a pretty HIE specific question :)

@Preconf Heh, I feel a bit bad that we don't have any "setup" or "getting started" in the HIE repository. For the best experience, I recommend the following:

- Use stack in general for managing projects
- Build haskell-ide-engine with `make build-copy-compiler-tool`
- HIE should only work in projects, you can maybe do `stack new HelloWorld simple`
- Set your Oni command to `"language.haskell.languageServer.command": "stack exec -- hie"`
- Set your oni root files to `"language.haskell.languageServer.rootFiles": [".git", "stack.yaml", "*.cabal"]`

Should get you going in the right direction :) The reason for me recommending `make build-copy-compiler-tool`, is that the GHC version you built HIE with needs to match the GHC version in your project, else you'll get some strange errors (if it even works). You can check with `hie --compiler`.

Preconf:
@Tehnix Awesome, thank you very much for the detailed instructions! Hopefully I can get it working :+1:

PR #1922, opened 2018-03-25
Feature/highlight config reference

Akin909:
Following on from #1795 I've added a method to set the `reference` split to a scratch buffer and set syntax highlighting to `json` (this is mainly because the lsp throws an error if set to `js` or `ts` for example as the contents of the split aren't valid js or ts).

I've set the buffer to open as `nomodifiable` as well.

The outstanding thing is to add the method `setScratchBuffer` to the `oni-api` also I added a `scrollbind` to the window as I thought it'd be nice to have the reference scroll with the user though I can easily see ppl objecting to this, open to removing it or making it configurable.

Also tweaked the set lang method as after a lot of debugging seems that the command `nvim_buf` was not actually working.

EDIT: opened onivim/oni-api#37 to add new buffer methods

screen shot 2018-03-26 at 00 03 51

### DIFF
--- a/browser/src/Editor/BufferManager.ts
+++ b/browser/src/Editor/BufferManager.ts
@@ -42,6 +42,7 @@ import { TokenColor } from "./../Services/TokenColors"
 import { IBufferLayer } from "./NeovimEditor/BufferLayerManager"
 
 export interface IBuffer extends Oni.Buffer {
+    setLanguage(lang: string): Promise<void>
     getCursorPosition(): Promise<types.Position>
     handleInput(key: string): boolean
     detectIndentation(): Promise<BufferIndentationInfo>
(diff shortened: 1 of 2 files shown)


bryphe:
Very cool! 💯 Having the highlight makes a huge difference here 👍

browser/src/Editor/BufferManager.ts
```diff
@@ -175,11 +177,30 @@ export class Buffer implements IBuffer {
...
-            language,
-        ])
+        await this._neovimInstance.command(`setl ft=${language}`)
+    }
+
+    public async setScratchBuffer(): Promise<void> {
+        // set the open buffer to be a readonly throw away buffer, also add scrollbind
+        // may need a config option
+        const calls = [
+            ["nvim_command", ["setlocal buftype=nofile"]],
```
bryphe: Cool, thanks for implementing this!

bryphe: And nice use of `nvim_call_atomic`  👍

bryphe:
> The outstanding thing is to add the method setScratchBuffer to the oni-api also I added a scrollbind to the window as I thought it'd be nice to have the reference scroll with the user though I can easily see ppl objecting to this, open to removing it or making it configurable.

As a short-term, feel free to add that method to `IBuffer` in `BufferManager`. These are like 'api candidates' - I'm planning on moving items there to `oni-api` anyway. At least, it means you don't have to wait on `oni-api` to get this in.

> Also tweaked the set lang method as after a lot of debugging seems that the command nvim_buf was not actually working.

Nice find! That probably caused some issues..

PR #1933, opened 2018-03-26
Add an example of binding to an internal command

Himura2la:
I use Ctrl+Shift+P for screenshoting and have no any ability to pass it to Oni.

The shortcuts configuration for internal features is not detailed at all, I think it would be a great addition.

1. Adding the example and keywords to search in the code
2. Reproducing the VS Code behavior

### DIFF
--- a/configuration/config.default.js
+++ b/configuration/config.default.js
@@ -9,6 +9,7 @@ const activate = oni => {
     //
     // Add input bindings here:
     //
+    oni.input.bind("<f1>", "commands.show")
     oni.input.bind("<c-enter>", () => console.log("Control+Enter was pressed"))
 
     //


bryphe:
Hi @Himura2la - thank you for the PR!

> The shortcuts configuration for internal features is not detailed at all, I think it would be a great addition.

Agree, we definitely need a lot of help here!

I like the idea of having an example of using a command, for sure. It'd also be nice if we could show the list of available commands (similiar to how @rgehan added our list of configuration settings in #1795)

I wonder if the `F1` binding will be confusing, though, just because prior to the editing the configuration, the behavior would be different - `F1` brings up the neovim help by default.

Maybe another example could be opening the devtools - `Oni.input.bind("<f11>", "oni.debug.openDevTools")`? I think that one wouldn't conflict, and still demonstrate an example.

Let me know what you think - thanks again for the PR and for supporting the project! 💯

Himura2la:
No problem, I'll change this and try to add the list of commands. 
Just thought that F1 is not commonly used for help now, because it's a very convenient button to press. Those who need help usually search for the exact topics with ":help", but I absolutely understand your wish not to break the existing behavior, that should be a much more weighted decision.

Himura2la:
I added a link to source code in wiki: https://github.com/onivim/oni/wiki/API-Examples#list-of-commands-and-defaults 
This info looks non-static and I did not introduce another place to maintain, which nobody knows about by this.

bryphe:
> I added a link to source code in wiki: https://github.com/onivim/oni/wiki/API-Examples#list-of-commands-and-defaults

Excellent, thanks for adding that!

> This info looks non-static and I did not introduce another place to maintain, which nobody knows about by this.

Ya, makes sense 👍 

And the PR looks good - I'll bring it in now. Thank you for the contribution @Himura2la & for improving our documentation!

Issue #1932, opened 2018-03-26
Register Oni as an editor for all file types, so it shows up in “Open with…” on macOS


mikl:
I know very little of macOS app development, so this is a suggestion for now…

A nice “Quality of Life” improvement for users on macOS would be for Oni to register itself as an editor for everything, so it always shows up in “Open with…” like I have a bunch of other apps doing in this screenshot:

![screen shot 2018-03-26 at 17 13 02](https://user-images.githubusercontent.com/10263/37915175-31685ece-3119-11e8-9af1-aa30fc8823ef.png)

I don’t know the mechanism for this, but it appears to be something any app can do (even if it makes very little sense, as is the case with “Day One” or “Notes.app” in this example), and it would be a huge help for Oni users on the Mac, since otherwise you have to select “Other” and find the app manually for every new file type.

CrossR:
Seems reasonable! I think we have it setup for certain app types (as listed in the `package.json`) but not generally to appear in that "Open With" context menu.

I know the same functionality was added to the Windows binary so I expect we'd want feature parity when possible. (Only reason I didn't add it to Mac was since I'm unable to test it there and on Windows its done with registy keys which isn't applicable outside Windows).

Akin909:
@mikl thanks for the opening this so we can track this 👍. We actually use `electron-builder` 's config settings to register `oni` as a default filetype but we have to manually specify the filetypes and I think at present that list is just 

```json
"fileAssociations": [
            {
                "ext": "js",
                "name": "javascript",
                "role": "Editor",
                "isPackage": true
            },
            {
                "ext": "json",
                "name": "JSON",
                "role": "Editor",
                "isPackage": true
            },
            {
                "ext": "jsx",
                "name": "JSX",
                "role": "Editor",
                "isPackage": true
            },
            {
                "ext": "py",
                "name": "Python",
                "role": "Editor",
                "isPackage": true
            },
            {
                "ext": "go",
                "name": "Go Lang",
                "role": "Editor",
                "isPackage": true
            },
            {
                "ext": "re",
                "name": "Reason",
                "role": "Editor",
                "isPackage": true
            },
            {
                "ext": "vim",
                "name": "VimL",
                "role": "Editor",
                "isPackage": true
            },
            {
                "ext": "ts",
                "name": "Typescript",
                "role": "Editor",
                "isPackage": true
            },
            {
                "ext": "tsx",
                "name": "Typescript",
                "role": "Editor",
                "isPackage": true
            },
            {
[...]

PR #1936, opened 2018-03-26
Support more file types

mikl:
First attempt at addressing issue #1932, explicitly accepting more file extensions.

There’s a few more [tricks from Atom’s plist we might use](https://github.com/atom/atom/blob/master/resources/mac/atom-Info.plist#L1537) but these require patching the plist after generation which is not quite as trivial as this change, so I’ll leave that for another day.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,76 +77,704 @@
         },
         "fileAssociations": [
             {
-                "ext": "js",
-                "name": "javascript",
+                "name": "ADA source",
                 "role": "Editor",
-                "isPackage": true
+                "ext": [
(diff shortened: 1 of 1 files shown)


Akin909:
@mikl that looks great 👍, didn't realise we could make use of an `ext` property, that makes things much more convenient.

bryphe:
Very cool! Thanks a lot, @mikl - looks great 👍 

Planning on getting a new release next week (4/2) - so this'll be in for it. Really appreciate the contribution!

Bringing it in now-

Issue #1125, opened 2017-12-14
Check for init.vim at startup?


keforbes:
I just had a thought for a feature request, not sure if it'd be helpful or annoying.  At startup, Oni could check to see if `init.vim` exists in the default location.  If it does, we could prompt "init.vim found! Would you like to enable it?"

I was just thinking there are some Vim veterans who try out Oni but then get frustrated when their settings aren't respected (because `oni.loadInitVim` is `false` by default).  I think the default opinionated config makes perfect sense for newcomers but if `init.vim` already exists we could probably assume the user is Vim-savvy and might not want our defaults.  This should mitigate issues which led to the [FAQ](https://github.com/onivim/oni/wiki/FAQ) being created.

Of course, if we add this prompt we'd probably also want to write a config `promptToEnableInit: false` or something so we don't prompt on subsequent Oni launches.

bryphe:
Ah ya, I like it - thanks for the suggestion @keforbes !

I think it would be a good use case for #98 - having a notification pop up. And as you mentioned it would likely save a lot of trips to the FAQ 👍

bryphe:
Would be great to have a few different 'health checks' here - would be nice to guard against plugin incompatibilities like #1163

CrossR:
This feels like its becoming more and more important now.

As I see it it would work something like : 

* Pop up with a button if its not been set it the user config.
* On accept: Set to `true`.
* On close without accepting: Set to `false`.

Which looks like we'd need to add buttons/interaction of some form in the Notifications, setting config (I feel we have that but not 100%) and checking if a value exists in the users config vs the default.

Akin909:
That api is really familiar to me as react native does a very similar thing with their `Alert` api, plus its a really flexible way add a notification and to also render it 👍

PR #1937, opened 2018-03-26
Feature/notification buttons

Akin909:
Following the excellent suggestion @bryphe made in #1125 I've added the button api to the notifications so that buttons can be rendered with text and an associated callback.

Screenshot:
screen shot 2018-03-26 at 21 24 33

EDIT: should hopefully be another small-ish PR once this is integrated to check for a user's `init.vim` and prompt them if it isn't loaded.


bryphe:
Wow, thanks for the quick PR, @Akin909 ! Looks great!

Akin909:
@bryphe 👍 won't be able to merge this one myself though due to the timeout issue with the `CItests`

PR #1942, opened 2018-03-27
Feature: add prompt to use init vim if one is found

### LINKED ISSUE #1125: Check for init.vim at startup?

Akin909:
This PR aims to fix #1125, by showing a notification to prompt the user to user their `init.vim` then reload if `oni` for changes to take effect. If they say no the value of `loadInitVim` is set to false.

screen shot 2018-03-27 at 11 48 34

### Todo
- [x] Confirm that if user has no option set the persisted option is used when oni reloads - this does not seem to be the case in the dev build


browser/src/Utility.ts
```diff
@@ -245,3 +245,15 @@ export function ignoreWhilePendingPromise<T, U>(
 
     return ret
 }
+
+export function checkIfFileExistsSync(filename: string): boolean | Error {
```
bryphe: Would `fs.existsSync` work here instead?

browser/src/neovim/NeovimInstance.ts
```diff
@@ -543,6 +545,21 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
         return this.command(`e! ${fileName}`)
     }
 
+    /**
+     * getInitVimPath
+     */
+    public getInitVimPath() {
+        const rootFolder = Platform.isWindows()
+            ? path.join(process.env["LOCALAPPDATA"], "nvim") // tslint:disable-line no-string-literal
```
bryphe: It looks like this might be duplicated from the logic in `openInitVim`? Perhaps we could refactor this such that the existence check is in a separate method - like `doesInitVimExist`? Then, we could re-use the `getInitVimPath` in `openInitVim`, so we don't have to duplicate this logic.

browser/src/Utility.ts
Akin909: @bryphe I was initially going for something async and noted that `fs.exists` was being deprecated so assumed that was the same for `fs.existsSync`, but just checked and thats not the case so can tweak that, another perhaps over optimisation was that I wanted to ensure that the path not only existed but was  specifically a file on the *very* unlikely chance the function was used on a path that did exist but wasn't a file

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -254,6 +254,35 @@ export class NeovimEditor extends Editor implements IEditor {
             notification.show()
         })
 
+        const initVimPath = this._neovimInstance.getInitVimPath()
+        const exists = Boolean(initVimPath)
+        const initVimInUse = this._configuration.getValue("oni.loadInitVim")
+        if (exists && typeof initVimInUse === "undefined") {
```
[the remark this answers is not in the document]
bryphe: Ah, I think I see what's happening - we're using `undefined` to track the state where we haven't done this check yet, and `false` to explicitly say that we've opted out. I guess that's why the config change below is necessary. That seems a bit confusing to me though - using `undefined` as a sentinel is a bit scary.

Could we perhaps instead have an explicit value, tracking if we've done this check? Like, in the case where the user presses `yes` or `no`, we  set an explicit value, like `this._configuration.setValue("_internal.hasCheckedInitVim", true, true /*persist*/)`. Then, we use that to decide whether or not we should do this check - instead of the implicit `undefined`/`false`/ differentation.

bryphe:
Thanks for putting this together, @Akin909 ! 👍 The notification is looking good, this will be really nice to help users migrate their config over.

Had a couple of items I was checking on - let me know if you have any questions.

It would be nice to have a `CiTest` too to cover this

browser/src/neovim/NeovimInstance.ts
Akin909: I noted that and was getting itchy to refactor as you suggested but I noted theres an interim step in the `openInitVim` where it actually uses the root folder to `mkdirp(rootfolder)` so it seemed to need this interim piece of data which I guess a common method could still provide but tweaking for that seemed like increasingly diminishing returns can still move in that direction if you'd prefer

browser/src/Services/Configuration/IConfigurationValues.ts
```diff
@@ -86,7 +86,7 @@ export interface IConfigurationValues {
     // By default, user's init.vim is not loaded, to avoid conflicts.
     // Set this to `true` to enable loading of init.vim.
     // Set this to a string to override the init.vim path.
-    "oni.loadInitVim": string | boolean
+    // "oni.loadInitVim": string | boolean
```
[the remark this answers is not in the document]
Akin909: Was gonna leave a note since its a big change essentially with a default config setting as false its currently not possible to determine if the user's config has set the value to `false` or if `oni's` default did. The only way I found to actually tell the difference was to have either an explicit user setting or no setting. I guess if each option had info re where it was set that could help.

As it stands if the default remains then basically if the user has their config set to false then there's no way to check so the notification always appears if the setting is false since theres a false value but no distinction

Akin909:
@bryphe thanks for taking the time to review 👍, was gonna hold off on PR's since I seem to be the one mac user who cant get any integration tests to run (had a go at the potential solution re. the proxy setting you recommended in #1540 but no joy), figured this would need one as well so should have held off but was tempted to implement it 😆.

Will hold of here till I can figure out what to do about `chromedriver` and get tests working, so I can add a `ciTest`.

browser/src/Services/Configuration/IConfigurationValues.ts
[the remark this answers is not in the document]
bryphe: Ah I see, I'd prefer the check to be explicit - leaving the functionality of `oni.loadInitVim` as-is. IMO, it's not easy to explain `undefined` states of config values to users - so I'd prefer to be explicit, and use a separate explicit setting to check if we've shown the init.vim - prompt - like `prompt.hasShownInitVimMessage`.

browser/src/Utility.ts
bryphe: I see - if it has extra functionality, then I'm okay with keeping it in. Thanks for the explanation! That is confusing about  `fs.exists` being deprecated vs `fs.existsSync`.

browser/src/neovim/NeovimInstance.ts
bryphe: I'd recommend refactoring this a bit - to me, `getInitVimPath` means it will return the path, and _not_ necessarily do a check to see if it exists or not. Plus, if we refactor, we can re-use the logic - and that logic for finding the `init.vim` had a few different bugs over time, so I think we should try and centralize it:
```
    /**
     * getInitVimPath
     */
    public getInitVimPath() {
         // Use path from: https://github.com/neovim/neovim/wiki/FAQ
        const rootFolder = Platform.isWindows()
            ? path.join(process.env["LOCALAPPDATA"], "nvim") // tslint:disable-line no-string-literal
            : path.join(Platform.getUserHome(), ".config", "nvim")
        const initVimPath = path.join(rootFolder, "init.vim")
        return initVimPath
    }
```

Then, we could use this in `openInitVim`:
```
    public openInitVim(): Promise<void> {
        const loadInitVim = this._configuration.getValue("oni.loadInitVim")

        if (typeof loadInitVim === "string") {
            return this.open(loadInitVim)
        } else {
            const initVimPath = this.getInitVimPath()
            const rootFolder = path.dirname(initVimPath)
            mkdirp.sync(rootFolder)

            return this.open(initVimPath)
        }
    }
```

And have a third method that checks if it exists, so we can decide if we should show the prompt:
```
public doesInitVimExist(): boolean {
    const initVimPath = this.getInitVimPath()
    try {
            return checkIfFileExistsSync(initVimPath) ? initVimPath : null
      } catch (e) {
            return null
      }
}
```

Akin909:
@bryphe I've made the changes you suggested above although an issue I'm having in the ciTest I'm writing is that notifications do not show up at all. I've tried just rendering a test one on load in the neovim editor (which is what I've done previously to see the notifications) but they seem not to appear in ci version.

I set the config to `notifications.enabled: true` but this doesn't seem to help.

test/ci/initVimPromptNotificationTest.ts
```diff
@@ -0,0 +1,38 @@
...
+        assert.strictEqual(title, "init.vim found")
+    }
+
+    // Validate that the notification element appears
+    assert.ok(!!notification)
+}
+
+export const settings = {
+    config: {
+        "oni.loadInitVim": false,
```
bryphe: We might need to add `"_internal.hasCheckedInitVim": false` here, otherwise it might pick up the 'persisted' settings. Perhaps that's why its not showing up?

bryphe:
Change looks good to me @Akin909 - I'm wondering if the notification not showing up is due to `_internal.hasCheckedInitVim` flag being set to `true` already? Since it is persisted...

I'm not 100% sure if the persistent settings for automation are isolated from normal runs (they _should_ be, but that might not be the case).

badosu:
@Akin909 Does this prompt appears if `oni.loadInitVim` is true?

Akin909:
@bryphe turns out I think I was being an idiot forgot to build before running the ci test, but added the flag for good measure.

@badosu if the initVim is set to `true` in the config then a user will never see this. It shows only if the user has never seen it before and they have the `loadInitVim: false` setting. Though theoretically if a long time use temporarily turns off the init.vim then the **first** time they reload oni with that setting they will be asked if they'd like to turn it on

Akin909:
@bryphe a bit stuck here as locally the `ciTest` for this functionality works but the test seems to fail in CI not sure what's different about the environment, seems the node isn't appearing but i'm not sure why is there anything specifically different about how the CI tests run?

CrossR:
@Akin909, I may be missing something but do you make a `init.vim` on the CI server/pass an override etc?

Since you check if one exists, it'll work instantly for you since you have an `init.vim` but the CI servers don't, and I don't see any code that would make one in the CI test at least.

I assume that `        if (initVimPath && !initVimInUse && !hasCheckedInitVim) {` is failing due to that.

bryphe:
Good catch @CrossR , that does sound like the issue!

We have one other test that needs an `init.vim`: https://github.com/onivim/oni/blob/a99544bdffbceb07c86a1f8c0da614ef7c2e4f52/test/ci/Neovim.InvalidInitVimHandlingTest.ts#L59

And in that case, we write it specifically to a temp path.

For this test, we could do something similiar - ideally we could write the `init.vim` to a temporary location, and then pass Oni the `$MYVIMRC` environment variable with the path (this would minimize conflicts in other tests).

Something like in the test:
```
export const settings = {
    env: {
        "MYVIMRC": createInitVimAndReturnPath(),
    }   
}
```

We only pass the config file to `env` today, though - we'd need to plumb this value through to where it gets used: https://github.com/onivim/oni/blob/a99544bdffbceb07c86a1f8c0da614ef7c2e4f52/test/common/Oni.ts#L101

Akin909:
@bryphe @CrossR 👍 💯 thanks that's really helpful I completely forgot that we'd need an `init.vim` for the CI for this test 😆 I'll look at adding that.

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -254,6 +254,43 @@ export class NeovimEditor extends Editor implements IEditor {
...
+        const initVimInUse = this._configuration.getValue("oni.loadInitVim")
+        const hasCheckedInitVim = this._configuration.getValue("_internal.hasCheckedInitVim")
+
+        if (initVimPath && !initVimInUse && !hasCheckedInitVim) {
+            const initVimNotification = notificationManager.createItem()
+            initVimNotification.setLevel("info")
+            initVimNotification.setContents(
+                "init.vim found",
+                `We found an init.vim file would you like Oni to use it?
+                this will result in Oni being reloaded`,
```
CrossR: "this" should be capitalised here.

CrossR:
Glad this is working now! I expect this will really cut down on the confusion and number of issues we get.

bryphe:
Latest set of changes look good to me! 👍 Thanks for updating the `CiTest` infrastructure to let us plumb environment variables through. Appreciate all your work on this, @Akin909 !

PR #1946, opened 2018-03-27
Fix #1924 - Part 1 - Make tutorials clickable and sneakable in the sidebar

### LINKED ISSUE #1924: Tutorial improvements
Just a few small notes on the tutorials, so they are written down somwhere:

* Currently, the tutorials themselves are not clickable or sneakable, which makes discovery of using them hard.
* The end screen of the tutorial doesn't respond to different screensizes, meaning it can overflow on certain machines.
* Currently, the `init.vim` seems to be loaded, which possibly isn't wanted. (Similarly, some are editable which may not be wanted).
* Swapping between tutorials doesn't work - Ie open tutorial 2 and select tutorial 1, it is re-focuses tutorial 2.
* The first tutorial has 10 lines, and can only fit 9 on screen, meaning its possible to scroll off and end up with the goal looking reached, when in reality its on line 1 and the cursor is on line 2. (See below).
image

Sorry for dumping a bunch of them! I'm actually really excited about this feature, and all the things above are small changes.

bryphe:
This fixes the first issue in #1924 - making the tutorials clickable and sneakable.

We were using a common `SidebarItemView` control, that is used in all sorts of sidebar experiences - from the explorer, plugins, etc. However, that control didn't handle the sneak/click action in a common way.

This change refactors the `SidebarItemView` to make that common functionality, and plugs in sneak/click to open the appropriate tutorial.

### DIFF
--- a/browser/src/Plugins/PluginSidebarPane.tsx
+++ b/browser/src/Plugins/PluginSidebarPane.tsx
@@ -16,6 +16,8 @@ import { VimNavigator } from "./../UI/components/VimNavigator"
 
 import { PluginManager } from "./../Plugins/PluginManager"
 
+import { noop } from "./../Utility"
+
 export class PluginsSidebarPane implements SidebarPane {
     private _onEnter = new Event<void>()
     private _onLeave = new Event<void>()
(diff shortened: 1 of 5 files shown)


bryphe:
Also, it turns out there are a bunch of places where clicking was implicitly a no-op (because it wasn't handled before). For those, I just made it explicitly a no-op - but we'll want to add better functionality there, eventually.

PR #1947, opened 2018-03-27
Browser: Enable by default

bryphe:
The browser feature is still pretty ripe - but I think it's in a good spot to get feedback. It's really an opt-in feature, as you only get to it via the command palette today. Having it in the wild will help track down bugs and figure out the priorities for the next set of work.

This removes the `experimental.browser.enabled` flag and replaces it with `browser.enabled`.

### DIFF
--- a/browser/src/Services/Browser/index.tsx
+++ b/browser/src/Services/Browser/index.tsx
@@ -66,8 +66,22 @@ export const activate = (
 
     const activeLayers: { [bufferId: string]: BrowserLayer } = {}
 
+    const browserEnabled = configuration.registerSetting("browser.enabled", {
+        requiresReload: false,
+        description:
+            "`browser.enabled` controls whether the embedded browser functionality is enabled",
+        defaultValue: true,
+    })
+
+    configuration.registerSetting("browser.zoomFactor", {
+        description:
+            "This sets the `zoomFactor` for nested browser windows. A value of `1` means `100%` zoom, a value of 0.5 means `50%` zoom, and a value of `2` means `200%` zoom.",
+        requiresReload: false,
+        defaultValue: 1,
+    })
+
     const openUrl = async (url: string, openMode: Oni.FileOpenMode = Oni.FileOpenMode.NewTab) => {
-        if (configuration.getValue("experimental.browser.enabled")) {
+        if (browserEnabled.getValue()) {
             url = url || configuration.getValue("browser.defaultUrl")
 
             count++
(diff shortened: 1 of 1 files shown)


justinmk:
How would a vimscript plugin (e.g. the `gx` mapping) open a URL in the browser?

bryphe:
@justinmk - I was thinking that it should be possible for us to know when there is a `URL` passed to a buffer. For example, it'd be great if this could work with `:vsp https://neovim.io`. It seems `netrw` already knows about http/https and attempts to run `wget` or `fetch`.

Potentially, the way Oni could set this up is by adding the browser 'buffer layer' whenever the buffer starts with http/https. I believe this would work with the `gx` mapping, too.

The downside is that `netrw`'s default behavior on Windows for http/https isn't great - like if I run `:vsp https://neovim.io/`:
browser-buffer-layer

I get the error message at the bottom. So even if we show the browser layer, we still get that unfortunate error message.

But I think this would be a natural entry point to have the browser experience shown, and would be pretty natural coming from other features like `term://`. I was looking into if it was possible to disable netrw's default behavior, just for the `http` and `https` protocols, but I didn't have any luck.

I'll put together a PR of hooking up the 'buffer layer' based on the filter, too, and see what you think.

justinmk:
> it'd be great if this could work with :vsp https://neovim.io

I use `:edit {url}` (or `:vsp {url}`, etc.) to view the raw HTML, I don't want to render it in a browser. I use this for viewing logs, etc.  `:edit` (and its family of related commands) should edit.

If a user wants to render a webpage, why not support this from the "omni box" (what do you call the ctrl-shift-p dialog?). Just like a web browser supports searches (and commands) from its address box, Oni could support URLs in its command box.

> The downside is that netrw's default behavior on Windows for http/https isn't great

Are you using the Nvim windows build from https://github.com/neovim/neovim/releases ? We ship with `curl.exe` on Windows, and `:r !curl -L https://...` should work.  If `gx` doesn't look for curl (on Windows) then it's just a netrw bug (we can fix this).

> I was looking into if it was possible to disable netrw's default behavior, just for the http and https protocols, but I didn't have any luck.

It's provided by the `au Network BufReadCmd` autocmd group:

```
    ...
    http://*  exe "sil doau BufReadPre ".fnameescape(expand("<amatch>"))|call netrw#Nread(2,expand("<amatch>"))|exe "sil doau BufReadPost ".fnameescape(expa
    file://*  exe "sil doau BufReadPre ".fnameescape(expand("<amatch>"))|call netrw#Nread(2,expand("<amatch>"))|exe "sil doau BufReadPost ".fnameescape(expa
    https://* exe "sil doau BufReadPre ".fnameescape(expand("<amatch>"))|call netrw#Nread(2,expand("<amatch>"))|exe "sil doau BufReadPost ".fnameescape(expa
    ...
```

But I would not recommend bothering to change the behavior of `:edit`, except to improve its success rate at actually pulling the source of a URL :)

bryphe:
> I use :edit {url} (or :vsp {url}, etc.) to view the raw HTML, I don't want to render it in a browser. I use this for viewing logs, etc. :edit (and its family of related commands) should edit.

That's fair! The only UI entry point today as a result of this PR is in our command palette (the `Control+Shift+P` experience):

image

So I think that's a good stepping stone. The feature is still alpha quality anyway, so it's okay for it to be a bit tucked away.

> If a user wants to render a webpage, why not support this from the "omni box" (what do you call the ctrl-shift-p dialog?). Just like a web browser supports searches (and commands) from its address box, Oni could support URLs in its command box. 

Yes, I like this suggestion. Right now, it's a two-step process - you have to open the browser window, and then type the address. It'd be cool if we could just handle that inline directly in the command palette, like `brow duckduckgo.com`. #1960 tracks this enhancement.

> Are you using the Nvim windows build from https://github.com/neovim/neovim/releases ? We ship with curl.exe on Windows, and :r !curl -L https://... should work. If gx doesn't look for curl (on Windows) then it's just a netrw bug (we can fix this).

It's strange, I have like a 50% success rate. Sometimes it works, and sometimes it fails with an error message 😄 I'll take a look later when I have a bit more time and get more details.

Then to answer your earlier question:
> How would a vimscript plugin (e.g. the gx mapping) open a URL in the browser?

This is poorly documented but we actually expose oni commands on the VimL side via `:call OniCommand`. As an example, the browser registers a bunch of commands here:
https://github.com/onivim/oni/blob/8186d0ba5754282b81464748355ad702d5a4f8e5/browser/src/Services/Browser/index.tsx#L89

So for example, I could call `:call OniCommand("browser.openUrl.verticalSplit")`

The one thing we are _missing_ today is the ability to send arguments via these commands. For the `gx` case, we'd probably want something like:
`:call OniCommand("browser.openUrl", <url-to-pass>)` Tracked this in #1961

bryphe:
Bringing this in now - but still tracking some of the other pieces in #1961 & #1960

PR #1949, opened 2018-03-27
Prototype: Add browser layer when the buffer is http or https

bryphe:
In line with the way we'd open a `term://` buffer, I think it would make sense to have the 'browser' available when we open an `http://` or `https://` buffer. Like this:

browser-buffer-layer2

@justinmk - what do you think about this?

The experience is a bit weird and rocky with the way `netrw` jumps in - it might be nice for Oni to disable `netrw`'s handling of `http` and `https` by default, or perhaps when the browser is enabled.

### DIFF
--- a/browser/src/Editor/OniEditor/OniEditor.tsx
+++ b/browser/src/Editor/OniEditor/OniEditor.tsx
@@ -52,6 +52,7 @@ import { NeovimEditor } from "./../NeovimEditor"
 import { SplitDirection, windowManager } from "./../../Services/WindowManager"
 
 import { ImageBufferLayer } from "./ImageBufferLayer"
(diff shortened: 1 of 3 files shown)


justinmk:
Why not use the command (ctrl-shift-p) dialog as the entry point for this (instead of `:edit`, `:vsp`, etc.)? 

Changing the existing functionality of `:edit {url}` (view the raw source of a URL) means that functionality has to be found somewhere else.

bryphe:
Thanks for the feedback, @justinmk ! 

> Changing the existing functionality of :edit {url} (view the raw source of a URL) means that functionality has to be found somewhere else.

Seems reasonable. I'll close this out.

PR #1968, opened 2018-03-29
Achievements Layer: Clean up background / foreground colors

bryphe:


browser/src/Services/Learning/Achievements/AchievementsBufferLayer.tsx
```diff
@@ -27,8 +27,8 @@ export interface ITrophyCaseViewState {
 }
 
 export const TrophyCaseViewWrapper = withProps<{}>(styled.div)`
-    background-color: ${p => p.theme.background};
-    color: ${p => p.theme.foreground};
+    background-color: ${p => p.theme["editor.background"]};
+    color: ${p => p.theme["editor.foreground"]};
```
Cryza: With `p` being a rather undescriptive parameter name and since you are obviously looking for brevity, what do you think about using destructuring here?
`color: ${{ theme } => theme["editor.foreground"]};`

bryphe: Ah ya, I could see this being unclear - thanks for looking this over, @Cryza ! And appreciate the suggestion.

Less about brevity on my end and more about just slamming out these changes 😄 

All things being equal, I don't have a strong preference either way - but it probably makes sense to use the more verbose `${ props => ... }` syntax, since that's the predominant style in the code today (IMO)

cc @Akin909 in case you have feedback too

Akin909: @bryphe I've tended towards `props` so far since I though for anyone less familiar with `css-in-js` it'd be confusing to destructure or use `p` but I much prefer *personally* the destructuring syntax

bryphe: Tried it out, but it needed to be:

```
color: ${({theme}) => theme["editor.foreground"]};`
```

The nested `()` in there in addition to the destructuring seemed intense to me, so I went the `props` route. Thank you both for the feedback!

PR #1980, opened 2018-03-29
Refactor FileIcon to styled-components

Cryza:
This is one of several smaller upcoming PRs with the goal to replace all remaining `.less` files with `styled-components`.

- Move the animation code from `Tabs.less` into `FileIcon.tsx`
- Replace `additionalClassNames` prop with the (also more restrictive, but better encapsulated) `playAppearAnimation`


bryphe:
👍  Thanks for splitting this out, @Cryza !

Issue #1917, opened 2018-03-24
Populate quickinfo from Neovim side (via OniCommand('...'))


Tehnix:
Is there a way to populate the quickinfo tooltip/popup from the Neovim side? I know we can send some Oni commands via `call OniCommand('...')`, but it would be cool if this could e.g. be extended to take in arguments, and then pass that to the command it activated. 

That way we could make existing Neovim plugins play nicely with Oni. For example, I have set up [a haskell layer](https://github.com/Tehnix/spaceneovim-layers/blob/master/layers/%2Blang/haskell/config.vim#L186) which shows the type information on `CursorHold`, which I currently display in the command output line, because of lack of options.

CrossR:
Now that `OniCommand` has been updated to allow the passing of arguments, we should be able to support this, at least once the Oni API has been updated to expose the popup settings.

We can use this ticket to track adding it.

PR #1984, opened 2018-03-30
Extend OniCommand

### LINKED ISSUE #1961: VimL: Update our `OniCommand` function to accept an argument
Pulling this out from #1947 - this is poorly documented, but we expose a function `:call OniCommand` that allows for running commands registered in Oni.

For example, from my `init.vim`, I could bind a key to: `:call OniCommand("browser.openUrl.verticalSplit")`.

However, this is rather limited today, because it doesn't take any arguments. For the browser scenario, it'd be great if we could pass a string argument that the Oni layer can then process - it would be far more useful. (Especially in the context of the case given in #1947 - mapping `gx`).

CrossR:
Fix #1961.

Update `OniCommand` to take any number of arguments.
Currently, restricted to strings... not sure if that is an issue.

Added a quick CI test too.

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -406,9 +406,9 @@ export class NeovimEditor extends Editor implements IEditor {
         )
 
         this.trackDisposable(
-            this._neovimInstance.onOniCommand.subscribe(command => {
-                const commandToExecute = command[0]
-                const commandArgs = command.length > 0 ? command.slice(1) : []
+            this._neovimInstance.onOniCommand.subscribe(context => {
+                const commandToExecute = context.command
+                const commandArgs = context.args
 
                 commandManager.executeCommand(commandToExecute, commandArgs)
             }),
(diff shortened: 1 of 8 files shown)


CrossR:
@bryphe , do you know if/how to add a function for `gx` to call? I'd like to test that with this, but I'm not sure how to remap it.

bryphe:
Awesome work on this, @CrossR ! 

Nice test case, great way to exercise the end-to-end functionality - gives me confidence that it won't be regressed 👍 

> @bryphe , do you know if/how to add a function for gx to call? I'd like to test that with this, but I'm not sure how to remap it.

I'm not sure if there is an option to configure it - it might be we just need to override it. `:nnoremap gx :echo 'hello'<CR>` seems to take over the default behavior. I looked at `:help netrw-gx` and it seems that you can specify a _program_ to handle the URL, but not a Vim command.

We could try something like:
```
:nnoremap gx :call OniCommand("browser.openUrl", expand("<cWORD>"))<CR>
```
@justinmk might have ideas too 😄 

Thanks for putting this together, @CrossR  - I think this will open up for some interesting ways plugins could start to leverage. For example, this helps to enable #1917 - having a command to show the quickinfo tooltip from VimL.

PR #1991, opened 2018-03-30
Refactor Error.tsx to styled-components

Cryza:
Another one of my `.less`=>`styled-components` PRs

- Move error styling (squiggles and error markers on the right side) to `styled-components`
- Simplify styling of error markers (outcome should still be the same, I made sure not to change the appearance)

Some styles within `Error.less` were already unused, most notably everything inside `.error`, so that could safely be removed as well.


Cryza:
I made the following assumptions while replacing everything from the width: 200px to the nested icon styling:

- the 200px width, flex and the complex positioning and styling for the icon seems to be a leftover from when there seemed to have been text here. Since that code is gone, this should be gone, too
- the active state seems to have been removed in an earlier change, the class is never set

Please tell me if any of this was wrong ✌️

Cryza:
I made the following assumptions while replacing everything from the width: 200px to the nested icon styling:

- the 200px width, flex and the complex positioning and styling for the icon seems to be a leftover from when there seemed to have been text here. Since that code is gone, this should be gone, too
- the active state seems to have been removed in an earlier change, the class is never set

Please tell me if any of that was wrong ✌️

Cryza:
I am pretty sure now that all the styling I removed while refactoring was leftovers from when there was a way to activate the error markers, which apparently also displayed the error text. If this was removed intentionally in favor of e.g. Showing the error message in the quick info, I am pretty sure we can safely merge this PR.

@bryphe I guess you have the best overview here, can you please confirm or correct?

bryphe:
> I am pretty sure now that all the styling I removed while refactoring was leftovers from when there was a way to activate the error markers, which apparently also displayed the error text. If this was removed intentionally in favor of e.g. Showing the error message in the quick info, I am pretty sure we can safely merge this PR. 

Sorry for the slow reply! Yep, that was all vestigial stuff from when we had actual text displayed in my component. Thanks for removing it and investigating 👍  

Looks good to me, @Cryza ! Bringing it in now - thank you very much for doing this! 💯

Issue #1972, opened 2018-03-29
Oni can only be closed by right clicking the dock Icon to fully close the app


Akin909:
Was not entirely sure whether to raise this issue as since changes to the `browserWindow` management in #1747 I've found `oni` windows behave more predictably and reliably, thanks @CrossR. 

Wasn't sure I wanted to instigate any tweak given the new found freedom from weird bugs, but `electron` leave a comment in the main.js relating to macOS and closing windows as macOS has some peculiar behaviour regarding closing apps but in general most mac apps fully close if `cmd-q` is pressed but with oni atm as I believe a reference to the/a window ?remains somewhere the app doesn't actually quit only the window closes

Akin909:
Actually reading about this online it seems to be a bit of a debate but I *think*/most of my mac apps fully close when I hit `cmd-q`, save the snippet's code seems to contradict that but the comment seems to support it 😕 ❓

jwmann:
I also have this issue.

Most MacOS apps will fully "Quit" when you press `Cmd+q` and it will be fully terminated.
However, the difference between MacOS and say Windows is that in most apps, if you click the "X" on the window in MacOS, the app doesn't "Quit" it will still be open, just with no window.
In Windows, this will terminate the app.

ackey:
I believe I fixed this issue in #2404

brunobely:
Just downloaded Oni for the first time (v0.3.6) and `cmd+q` does not quit.  Sometimes (I cannot really tell what triggers it) it closes the window I have open, but the app does not fully quit until I right-click Oni on the dock and hit Quit

PR #2001, opened 2018-03-31
Add accelerator to Exit menu entry

tom-james-watson:
Currently you can quit Oni with Command+Q, but only if a window is
currently open. If you close all windows, e.g. by `:q`ing all tabs, the
app is left running but Command+Q will no longer quit the app, and you
have to press the Oni->Exit menu item.

Adding this accelerator here allows the app to be closed via the the
standard shortcut regardless of whether there are currently any windows
open.

### DIFF
--- a/main/src/menu.ts
+++ b/main/src/menu.ts
@@ -250,6 +250,7 @@ export const buildMenu = (mainWindow, loadInit) => {
             },
             {
                 label: "Exit",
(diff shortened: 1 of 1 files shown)


Akin909:
@tom-james-watson I just raised #1972 to track this, didn't realise we were missing an accelerator for that 👍, thanks that's a major bug bear of mine resolved

Issue #1998, opened 2018-03-31
Add 'surround.vim' as a default plugin


bryphe:
From discussion on discord, I believe it makes sense to include `surround.vim` in our set of default plugins.

A few things needed here:
- Add `vim-surround` as a submodule in `vim\default` : https://github.com/tpope/vim-surround
- Add an attribution and link in the `Bundled Plugins` section in our `README.md`

badosu:
Do you think there's a way we can expose an option to not load the bundled plugins? It would be good if instead of it being on or off for all plugins we had a hook on `activate` or somewhere else that specifies the bundled plugins with the default ones explicit on the default config.js.

E.G.

``` js
const activate = (oni) => {
  oni.plugins.load("wellle/targets.vim");
  oni.plugins.load("tpope/vim-commentary");
  oni.plugins.load("tpope/vim-unimpaired");
  // ...
};

```

bryphe:
@badosu - this is great feedback! I agree, I think this would be nicer than the `oni.useDefaultPlugins` option we have, as it is much more explicit. IMO It's not really clear how the `oni.useDefaultPlugins` behaves, what the ramifications of changing the value are, etc, and it changes if we decide to bundle a new plugin...

Actually, this is quite a bit in-line with what I'm starting to do with the `plugins` configuration option here in #1903 - the idea is you'd have a setting like:

```
"plugins": [
  "welle/targets.vim",
  "tpope/vim-commentary",
  "tpope/vim-unimpaired",
  "oni-plugin-live-preview",
  "oni-plugin-power-mode",
   etc...
]

The setting is actually there in master, but it is not fully implemented yet, so I don't recommend using it (yet). But I think semantically it is close, and with some relatively minor tweaks I believe it could fill this use case.

> Another idea is to not expose on config.js but if the hook is defined to override the default one, so this way we achieve 2 good things: users can have an very straightforward package bundler for when you don't want to inherit your init.vim or use NeoBundle, Plug, etc..; and we allow for minimalists to have a way to not load too much cruft.

Agree 100%. What do you think of using the `plugins` configuration setting for this? By default, we could populate it with our initial set of plugins, but the user could always scrap some.

The other consideration I had was deciding when we should check for updates for plugins. I was thinking about having an API call, like: `Oni.plugins.checkForUpdates()` that could be run in the `activate` method. Do you have any ideas there?

[...]

PR #2003, opened 2018-03-31
added vim-surround as a default plugin

muream:
First pull request ever here, I'm not too comfortable with git either so I just learned about submodules, which are pretty cool!
If i'm doing anything wrong, please tell me!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,9 @@
 [submodule "vim/default/bundle/vim-unimpaired"]
 	path = vim/default/bundle/vim-unimpaired
 	url = https://github.com/tpope/vim-unimpaired
+[submodule "vim/default/bundle/vim-surround"]
+	path = vim/default/bundle/vim-surround
+	url = https://github.com/tpope/vim-surround.git
 [submodule "vim/core/typescript-vim"]
 	path = vim/core/typescript-vim
 	url = https://github.com/leafgarland/typescript-vim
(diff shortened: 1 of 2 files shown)


badosu:
To be frank, I think it would be better to have _less_ bundled plugins instead of more, but I understand why one would want to have that plugin.

muream:
I did it since I saw [this issue](https://github.com/onivim/oni/issues/1998) I should probably have stated this in the PR.
I do agree that to many plugins is a bad idea but one of Oni's goals is to lower the bar for new users. Anyone coming from VSCode, Atom, etc. will expect a way to surround text with quotes, parenthesis, etc.

bryphe:
Looks great, @muream ! Thanks for taking this on!

@badosu - I like the feedback you left on #1998 - being more explicit about the set of plugins we bring in as part of the default configuration (as opposed to the opaque `oni.useDefaultPlugins` setting we have today...). My plan is to start looking at this as part of a more complete plugin management story for the next release.

So I think it's reasonable to bring this in now, to be bundled with our existing set of 'default' plugins - and then orthogonally we'll work on a better story for being explicit and managing that default set as we delve into plugin management.

bryphe:
Will bring this in now. Thanks again @muream ! 👍 Congrats on the first PR! This might finally give me a reason to learn vim-surround 😄 

In parallel, we'll continue investigating ways to improve our default-plugin story as part of a holistic plugin management scenario.

muream:
Thanks @bryphe ! Hopefully you can expect more and more PRs from me!

Issue #1266, opened 2018-01-10
QuickOpen does not prioritize matches accordingly


badosu:
In this example I have:

- `src/store/index.js`
- `src/router/index.js`

If I type `storeindex` this is what I get:

![screenshot from 2018-01-10 10-46-14](https://user-images.githubusercontent.com/347552/34773619-7f8b3452-f5f3-11e7-8698-d68800fbac43.png)

I would expect index on store to be more important than a file that does not have store in the path.

CrossR:
Have you tried with `"editor.quickOpen.filterStrategy"` set to `"regex"`? It would fix that specific issue, since `storeindex` would only match on the second one.

Though it may be too strong for what you want generally, such that we might just want an optional sort applying to the standard `"fuse"` as well if possible.

badosu:
@CrossR Does the `regex` strategy work as a reliable fuzzy-finder strategy?

CrossR:
I've found it quite nice, I'm fairly used to just typing a bit of the name of the folder I want, and then the file I want. If I'm less sure and just type the name of the file, it seems to work well still.

badosu:
Seems like it's working out, I wonder why this is not the default strategy.

bryphe:
Awesome! I was hoping the `regex` strategy would be more intuitive / address some of the issues you had, @badosu 

> Seems like it's working out, I wonder why this is not the default strategy.

I was actually considering switching this to be the default (i'm using it as well). If both you and @CrossR are using it and it's working out, might be worth swapping the default for next release (and seeing if anyone misses the old behavior 😄 )

hoschi:
Hm with `regex` strategy the dir delimiter `/` doesn't work. Probably not escaped?

E.g. when opening general files which exist in more than one dir like `src/features/tasks/index.stories.js` I can't type `tasks/in` anymore. 

Also with regex some matches are very weird. Why does it highlight `xxxRCxxxEAT...` when my search starts with `act`?
![2018-01-11-152559_416x149_scrot](https://user-images.githubusercontent.com/163128/34830348-e27f8cd4-f6e3-11e7-87ef-c4c551edd986.png)

bryphe:
> Hm with regex strategy the dir delimiter / doesn't work. Probably not escaped?

Good catch, ya, I think it isn't escaped. We should fix this.

> Also with regex some matches are very weird. Why does it highlight xxxRCxxxEAT... when my search starts with act?

[...]

PR #2009, opened 2018-04-01
[WIP] VSCode QuickOpen Scoring/Sorting

### LINKED ISSUE #1829: Improve scoring logic of QuickOpen
The results returned back from the QuickOpen now are much better with the `regex` filter, but they aren't sorted the best.

This can lead to confusing results where you've typed the exact name of a file and its still not the first result.
image

One idea was to bring over the VSCode QuickOpen scoring logic which can be found here : https://github.com/Microsoft/vscode/blob/master/src/vs/base/parts/quickopen/common/quickOpenScorer.ts

This would give us a scoring system that we could then sort the results based on, and it could live in its own repo, like the [Snippets provider code](https://github.com/onivim/vscode-snippet-parser).
Apart from that, we could attempt to write our own, or check an equivalent bit of logic from Atom.

I'm open to attempting to move the VSCode stuff over into a module, but would appreciate some heads up on how I should do the move. The most awkward bit looks to be the `comparers`, which links into a bunch more stuff in that file and so on. Should I be grabbing everything and just copying it over? The snippets code is only two files, so wasn't too sure.

CrossR:
Fix #1829.

Very WIP bit to import the VSCode scorer, so it can be used with the QuickOpen menu, to hopefully make the sorting of the QuickOpen a bit smarter.

This is pretty much a POC at this point, and if its seems like its going well, I'll spin it out into its own Module, so its a bit more seperate. Probably also worth at that point tidying up the `Utility` file to somewhat mirror the original file structure so its easier to apply updates in the future.

I've got it combined now since its easier to work with for now.

If it looks good, should hopefully hook up the highlights as well, and also sort the regex multiple terms filter to work with it a bit nicer.


CrossR:
Currently I'm using `score` from VSCode, though it looks like I might get better results by moving to `scoreItem` which in turn calls score as part of the process.

CrossR:
Swapped to scoreItem with a few wrappers to make it easier for Oni to call out to.

Few changes I've made so far:
* I've removed the cache for now, but we could add it back. That would mean bringing the code into the QuickOpen class though, and passing over the cache to the regexfilter func.
* I've dropped the fallback comparer for now, but it looks easy enough to add back so I might try that just for some extra scoring.
* The filtering I added breaks, which isn't great. I've got a few ideas for fixes though.

Initial results do look good though, or at least typing in a files name gives that file which is a good start.

CrossR:
I've stuck a PR up on the API project since this will need score to be added to the API.
Also, how would I go about refactoring this into a seperate module?

The folder is self-contained, and only functions from the `OniQuickOpenScorer` file are used in the Oni side.

CrossR:
I noticed I was still getting bad/sub-optimal sorting occasionally, so I'm testing out removing the regex filter too.... We could still use it and offer up two version though.

I realised well what is the point of filtering then scoring since if the score is 0...then it doesn't match and it can be filtered that way. So I've tested adding that and it looks good so far. Any opinions @bryphe? It would be simple enough to undo the stuff I've done and offer both a regex and a vscode option.

Below, I've added a few test searches that map over to the tests that were done over in #1266.
They are now much closer to FZF / CtrlP and VSCode. There is a slight difference though which I'm looking into.

image1
image2
image3

browser/src/Services/QuickOpen/RegExFilter.ts
```diff
@@ -39,73 +40,44 @@ export const regexFilter = (
...
-
-        return {
-            ...fo,
-            detailHighlights,
-            labelHighlights,
+    // Since the VSCode scorer doesn't deal so well with the spaces,
+    // instead rebuild the term in reverse order.
+    // ie `index browser editor` becomes `browsereditorindex`
+    // This allows the scoring and highlighting to work better.
+    const vsCodeSearchString =
```
bryphe: Thanks for thinking about this! Awesome that the 'refine' functionality will still work 👍

bryphe:
Awesome progress, @CrossR ! I'm glad that the 'refine' functionality still works, I love that capability 👍 

> I realised well what is the point of filtering then scoring since if the score is 0...then it doesn't match and it can be filtered that way. So I've tested adding that and it looks good so far. Any opinions @bryphe?

This seems reasonable to me - I think it simplifies things if we can just rely on the score, vs filtering _then_ scoring. I guess the only consideration would be performance (if doing a pre-filter saves the scoring some time), but that seems like a premature optimization right now IMO

It'll be really great to have a first-class scoring algorithm here - will make this much more generally useful. I like that you compared with the results in the thread - seems like we're close to parity with FZF/CtrlP - hopefully one less plugin Oni users need 😉 

> I've stuck a PR up on the API project since this will need score to be added to the API.
Also, how would I go about refactoring this into a seperate module?

> The folder is self-contained, and only functions from the OniQuickOpenScorer file are used in the Oni side.

We have a couple options here. For now, we could just bring this in as-is, which seems reasonable. The other is we could factor it out to a separate module, like `oni-quickopen-scorer`, publish that package to `npm`, and bring that package in. It'd be extra work that way to set up the module, so I'll leave it up to you - I'm cool either way (we can also refactor it out to a module later, incrementally).

And I just saw your update to the `oni-api` - I'll bring that in now.

Thanks for all your work on this! 💯

bryphe:
FYI - just published `oni-api@0.0.42` with your changes in!

CrossR:
Swapped to a seperate filter now, called "vscode". Need to update the default config with it still.
Reverted the RegEx file back and also copied the RegEx filter tests over to the VSCode filter + extended them for the new highlights + scoring logic.

I think it makes sense to have this bit in its own module, plus its a bit ugly with having the lint ignores at the top of the file, which could be removed if it was in its own module... Only worry is that its probably a bit of a pain for me to publish since it would then require me transferring a git repo over to Oni and a npm module, so perhaps it makes more sense for that to be done later/by @bryphe ?

CrossR:
Fixed an issue where identical scores weren't being sorted by length now, and also removed some old regex cleaning code that was causing searches with file extensions to fail or act funny, and added a few unit tests to check both of those cases.

Swapped the default over now, so I think this is done.

bryphe:
Wow, just tried this out with some problematic cases I've had in the past and it feels _great_! Thank you @CrossR for all your work on this, it's a huge improvement. 👍 

I really like that the refine is still around and path separator characters work, too. Awesome work 💯

Issue #1952, opened 2018-03-27
enhancement: bind h,j,k,l to scroll the browser split


Akin909:
Because of the amazing sneakability already available one of my first thoughts was to try and figure out how to scroll the browser without having to move my mouse.

I wonder if it'd be possible to bind the vim navigation keys to scrolling left right up and down, not sure if the `webview` api gives access to that sort of functionality

bryphe:
@Akin909 - yep! I was thinking about this too in the context of `g` and `GG` (to go to the top and bottom).

Lots of possibilities looking at the webview documentation: 
https://electronjs.org/docs/api/webview-tag

You could use the `sendInputEvent` to simulate a scroll event (like a mouseWheel event)

Lots of other possibilities on the API too. For example, you can start a search session (maybe binding `/` would be a good starting point for that?), take a screenshot, etc.

What we'd probably want to do to start is create _commands_ for each of the gestures we want to support, like:
 - `browser.scrollDownOneLine`
- `browser.scrollUpOneLine`
- `browser.scrollToTop`
- `browser.scrollToBottom`
- `browser.startSearch`
- `browser.cancelSearch`

A good place to look is here:
https://github.com/onivim/oni/blob/5678f045df06acdb80a3ff6e26c1f8a9c082c648/browser/src/Services/Browser/index.tsx#L131

Then we could add default bindings for them. These commands already have an `enabled` filter, so we could bind keys to them without conflicting with other bindings.

PR #2012, opened 2018-04-01
Feature/add keyboard binding to browser

Akin909:
Add `h,j,k,l` binding to the browser layer to scroll left, down, up, right. Relates #1952 

TODO:

- [x] Find a better gate to check if browser is active - any ideas @bryphe  as this needs to be in the `Keybindings.ts` but you have a variable in `Browser/index.tsx` which is a record of active layers by id.


CrossR:
>Find a better gate to check if browser is active - any ideas @bryphe as this needs to be in the Keybindings.ts but you have a variable in Browser/index.tsx which is a record of active layers by id.

Not any help, but just agreeing it would be nice. 
If the UI elements surfaced a "isSelected" bool via the API that would be great. I've found myself wanting that for both the browser and the Explorer/Sidebar, either for Tests or keybinds.

Akin909:
@CrossR that would be nice to have a general mechanism to query to know what the selected part of the editor was at any given time 👍, there is a mechanism for the sidebar though or rather in #1839, I found the pieces to put one together though a more general API method would be nice.

@bryphe just read the comments you left in #1952 more thoroughly and it answers the above question 😆

Akin909:
@bryphe regarding the search functionality I plan to leave that out for the time being since I think that will actually probably be a bit more involved than the commands I've added so far so I plan on making a separate pr for that

bryphe:
This is super cool @Akin909 ! Just tried it out, I really like it... can't wait to use it in my workflow 👍 

> @bryphe regarding the search functionality I plan to leave that out for the time being since I think that will actually probably be a bit more involved than the commands I've added so far so I plan on making a separate pr for that

Sounds good to me!

> Not any help, but just agreeing it would be nice.
If the UI elements surfaced a "isSelected" bool via the API that would be great. I've found myself wanting that for both the browser and the Explorer/Sidebar, either for Tests or keybinds.

Ya, it would be nice to have this - I guess we just need to figure out what form the API should take?

PR #2013, opened 2018-04-01
Build: Test removing appveyor cache

bryphe:
Remove the appveyor cache folder - check to see if we got the proper nvim binaries for the `x64`/`x86` case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -948,7 +948,7 @@
         "mkdirp": "0.5.1",
         "mocha": "3.1.2",
         "nyc": "^11.4.1",
-        "oni-release-downloader": "0.0.8",
+        "oni-release-downloader": "^0.0.10",
         "opencollective": "1.0.3",
         "optimize-js-plugin": "0.0.4",
         "prettier": "^1.10.2",
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,7 +24,7 @@ install:
   - node --version
   - npm --version
   # install modules
-  - yarn install
+  - yarn install --verbose
 
 artifacts:
     - path: dist/*.exe
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,9 +9,6 @@ branches:
     - master
     - /^release.*/
 
-cache:
-    - "%LOCALAPPDATA%\\Yarn"
-
 platform:
     - x86
     - x64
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -948,7 +948,7 @@
         "mkdirp": "0.5.1",
         "mocha": "3.1.2",
         "nyc": "^11.4.1",
-        "oni-release-downloader": "0.0.8",
```
CrossR: Looks like this is the issue!
`0.0.8` doesn't have any code to deal with x64 so all windows builds are `win32`.
Could that cause the issues we've seen?

I was checked the individual packages for this, but not here apparently.

bryphe: Ah excellent! 👍  Makes sense now - the `peerDependencies` in the other packages just specify their 'requirement' for a package, and this is where we actually acquire the dependency to use. It's confusing, there's probably a better way to handle this..  Glad we got it going, thanks @CrossR ! 🎉

CrossR:
```
[00:02:21] verbose 111.357 appveyor env var is currently set to: true.
[00:02:21] platform env var is currently set to: x64.
[00:02:21] running in appveyor, updating platform as needed
[00:02:21] downloading binary for win64 platform.
```

Well at least it makes sense now!
Do you want me to submit a PR to the release downloader to remove the additional logging now?

And this PR can be used to bring in the updated package?
Wish there was some level of logging for `yarn` between `--verbose` and nothing... We needed the output but not every moved file.

bryphe:
> Well at least it makes sense now!

Phew, ya 😄 

> Do you want me to submit a PR to the release downloader to remove the additional logging now?

That's okay, I don't think that's necessary - the logging will be useful to have long-term, just in case we run into issues later.

> Wish there was some level of logging for yarn between --verbose and nothing... We needed the output but not every moved file.

Exactly... It would be nice to have an option to show the output of the spawned tasks, but not the internal work yarn is doing shuffling files. I think I'll leave the `verbose` in, because I'd like to have the logging available, but if there is a better option for making that output available too, I'm open to it.

So fingers crossed, but I think with this PR in we should be all set 🤞 Thanks for all your help @CrossR , I'm excited to finally have a _proper_ set of Windows builds for this release 😄

CrossR:
Seems to all be sorted now! Logs and the nvim version are 64bit on my machine, and still say 32bit on the other build.

My `bash` issue has also gone now too.

PR #2016, opened 2018-04-02
Feature/auto-prefix browser url if needed

### LINKED ISSUE #1951: enhancement: browser url needs to be prefixed with "http/https"
Wanted to try out the `browser` 🎉 @bryphe it is amazing,
in the interest of providing possibly useful feed back one of the first things I noted was that I tried to type `www.google.com` in the address bar but this failed with a cannot get resource. 

I tried again with `https://google.com` which works fine. I think it could just be me but my instinctive response was to use the address bar in the same sloppy fashion in which I use google chrome. I'm wondering based on the implementation if its possible to check for a starting `https` or `http` and if not present then to prefix the address which is searched?

Akin909:
Add auto-prefixing of browser layer url input so if user passes in `apple.com` which would normally fail. This is now converted to `https://apple.com`.

Fixes #1951

### DIFF
--- a/browser/src/Services/Browser/BrowserView.tsx
+++ b/browser/src/Services/Browser/BrowserView.tsx
@@ -5,7 +5,6 @@
  */
 
 import * as path from "path"
-import * as Url from "url"
 
 import * as React from "react"
 import styled from "styled-components"
@@ -183,22 +182,25 @@ export class BrowserView extends React.PureComponent<IBrowserViewProps, IBrowser
         )
     }
 
+    public prefixUrl = (url: string) => {
+        // Regex Explainer - match at the beginning of the string ^
+        // brackets to match the selection not partial match like ://
+        // match http or https, then match ://
+        if (url && !/^(https?:)\/\//i.test(url)) {
+            return `https://${url}`
+        }
+        return url
+    }
+
     private _navigate(url: string): void {
         if (this._webviewElement) {
-            const prefixedUrl = this._prefixUrl(url)
+            const prefixedUrl = this.prefixUrl(url)
             this._webviewElement.src = prefixedUrl
 
             this.setState({
                 url,
             })
         }
-    }
-
-    private _prefixUrl = (url: string) => {
-        const urlObj = new Url.URL(url)
-        const { protocol } = urlObj
-        console.log("protocol: ", protocol)
-        return url
     }
 
     private _goBack(): void {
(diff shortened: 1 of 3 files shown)


ui-tests/BrowserView.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    IBrowserViewState,
+} from "../browser/src/Services/Browser/BrowserView"
+import { Configuration } from "../browser/src/Services/Configuration"
+
+const mockEvent = new Event<void>()
+
+// Using the disable life cycle methods here as the CDM calls sneak which is
+// an external dependency I'm not trying to test here
+
+describe("<BrowserView /> Tests", () => {
```
bryphe: Nice, thanks for adding a test to cover this component, @Akin909 !

browser/src/Services/Browser/BrowserView.tsx
```diff
@@ -182,9 +182,20 @@ export class BrowserView extends React.PureComponent<IBrowserViewProps, IBrowser
...
+        if (url && !hasValidProtocol.test(url)) {
+            return `https://${url}`
+        }
+        return url
+    }
+
     private _navigate(url: string): void {
         if (this._webviewElement) {
-            this._webviewElement.src = url
+            this._webviewElement.src = this.prefixUrl(url)
```
bryphe: Awesome! I noticed though that if I navigate multiple times to different websites, like:
`apple.com` <-- works
`onivim.io` <-- doesn't work

I believe that we may need this logic where we set `webviewElement.src` at `_initializeElement` below, as well. Perhaps we could just refactor that to call `this._navigate(this.props.initialUrl)`, so that it can pick up this logic too?

Akin909: @bryphe seemed to be related to using `https` rather than `http` as a prefix when attempting to access `onivim.io` going to and from `apple.com`, `google.com` etc. all seemed to work fine.

I've tweaked it to use http by default in case there are more cases like that, and call this._navigate in initElement

Also changed the way initElement is called as were getting a warning re. impurity in the render and antipatterns I thought that could be contributing but were still getting it (it doesn't really change functionality  and with async rendering is soon to be no longer true)

ui-tests/BrowserView.test.tsx
Akin909: 👍 think theres still quite a bit more testing that could be added there I can look at in other PRs

bryphe:
Looks great, @Akin909 ! Hope you don't mind, I'll bring this in so I can get it in for the `0.3.2` build 👍 Thanks for taking this on!

PR #2034, opened 2018-04-04
Feature/ add menu command to open new window

Akin909:
### Problem
To open a new window (on mac at least) a user must right click the dock Icon and select open a new window.

### Solution
Added a command to the command palette to open a new window by adding a method to the `multiprocess` which sends an event via the renderer to the main process asking for a new window.


bryphe:
Sounds good! Thanks for adding this, @Akin909 👍 Tried it out on Windows and it works great there too.

PR #2036, opened 2018-04-04
Bugfix/syntax highlighting failure post setLines

### LINKED ISSUE #1838: Syntax highlighting fails when the buffer text is replaced using set lines
Whilst working on #1540 I noted that when calling the oni API method `setLines` the lines put into oni were not being syntax highlighted, the effect or lack of effect is not constant and seems to apply only when the layout of the lines is changed.

@bryphe noted in #1540 that this might relate to buffer updates not being trigger on setting Lines

Akin909:
Fixes #1838, by checking the new buffer update against the version of the old state if the version has changed then the highlighting is re-applied

### DIFF
--- a/browser/src/Services/SyntaxHighlighting/SyntaxHighlightingReducer.ts
+++ b/browser/src/Services/SyntaxHighlighting/SyntaxHighlightingReducer.ts
@@ -129,14 +129,15 @@ export const linesReducer: Reducer<SyntaxHighlightLines> = (
             const updatedBufferState: SyntaxHighlightLines = {
                 ...state,
             }
+            // console.log("state: ", state)
 
             for (let i = 0; i < action.lines.length; i++) {
                 const oldLine = updatedBufferState[i]
                 const newLine = action.lines[i]
 
-                if (oldLine && oldLine.line === newLine) {
-                    continue
-                }
+                // if (oldLine && oldLine.line === newLine) {
+                //     continue
+                // }
 
                 updatedBufferState[i] = {
                     tokens: [],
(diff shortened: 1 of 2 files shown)


browser/src/Services/SyntaxHighlighting/SyntaxHighlightingReducer.ts
```diff
@@ -134,7 +134,10 @@ export const linesReducer: Reducer<SyntaxHighlightLines> = (
                 const oldLine = updatedBufferState[i]
                 const newLine = action.lines[i]
 
-                if (oldLine && oldLine.line === newLine) {
+                // check if the buffer version has changed and if so
+                // update the line - rather than check if specific line
+                // is changed
+                if (oldLine && oldLine.version === action.version) {
```
bryphe: Does making this condition:

```
if (oldLine && oldLine.version >= action.version) {
```

not work? I'm thinking it may be a bit safer, as we likely don't need to re-apply updates for an old version.

bryphe:
Thanks for making the change, @Akin909 ! 👍 

Looking at the build logs, it looks like some unit tests may need to be updated with this change, as well-

Akin909:
@bryphe 👍 the comparison was an oversight re `>=` vs `===` I actually put in `===` found that it worked with the prettier plugin but the comparison in your snippet was what I intended.

Yeah didn't realise that i'd broken the unit tests till this went through `ci` I'll have a look at tweaking them to match the new comparison

bryphe:
Looks great, @Akin909 ! Thank you! 💯

Issue #1568, opened 2018-02-16
Very sensitive scrolling on macbook pro


powermosfet:
When scrolling using the trackpad on my 2017 Macbook pro, the sensitivity is so high that the only thing i can reliably do is going to the start or end of the file.

Going slowly seems to make it even worse, i can scroll through a 500-line file with about 2cm of movement.

Other text editors on my system do not have this problem.
Other electron apps on my system do not have this problem.

josemarluedke:
I have the same issue here.

badosu:
Probably related to https://github.com/onivim/oni/issues/405, as the scrolling motion can only happen in full line height steps at the moment.

This could probably be mitigated by configuring the threshold amount of scroll needed to advance one row, but the ideal solution would be smooth scrolling.

chenglou:
If there's smooth scrolling with bounce back then it'd be fantastic!

eemeli:
This is not a dupe of #405, but a different issue. I'm experiencing the same; on my MacBook it's practically speaking impossible to scroll accurately with the touchpad. At the level where the scrolling action is detected, the view already jumps about 20 lines. An amount of scrolling that in a browser window moves the view about one line moves the view 50-100 lines in Oni.

aquova:
I too am having the same issue. The screen runs so rapidly that it makes scrolling with the trackpad somewhat unfeasible.

bryphe:
This seems like a really bad issue 😦 I wish I had a Macbook Pro to reproduce it on! Smooth scrolling is a ways out, but for the time being, it seems like we should be able to throttle the scroll motion. 

We're likely not picking up the scroll delta correctly here, or perhaps we need to account for some other values. 

If someone has a Macbook and can reproduce this, this code would be the place to look:
https://github.com/onivim/oni/blob/c601fad812ba14b515ab3d8604358e3eee531b87/browser/src/Input/Mouse.ts#L36

Specifically, it'd be helpful to log out the `evt.deltaY` here - what are the values we're getting?
```
            if (evt.deltaY) {
                if (evt.deltaY < 0) {
                    scrollcmdY += `ScrollWheelUp>`
                } else {
                    scrollcmdY += `ScrollWheelDown>`
                }
                this.emit("mouse", scrollcmdY + `<${line},${column}>`)
            }
```

bryphe:
[...]

PR #2042, opened 2018-04-05
Implement scroll delta threshold to improve macOS scrolling experience.

samdoiron:
Fixes https://github.com/onivim/oni/issues/1568

Scroll events are emitted very frequently on macOS. Previously,
we were moving a single line on every scroll event, which made scrolling
on macOS way too fast.

The value `100` is taken from Windows measurements given here:
https://gist.github.com/bryphe/43caa81bbec16d90d7b47f084014774b

### DIFF
--- a/browser/src/Input/Mouse.ts
+++ b/browser/src/Input/Mouse.ts
@@ -6,6 +6,7 @@ import { IScreen } from "./../neovim"
 // Handle modifier keys
 export class Mouse extends EventEmitter {
     private _isDragging = false
+    private _scrollDelta = 0
 
     constructor(private _editorElement: HTMLDivElement, private _screen: IScreen) {
         super()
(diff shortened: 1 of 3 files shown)


bryphe:
Awesome, thanks for taking this on @samdoiron ! Really appreciate it 👍 

The fix looks great to me, however, I realized there is one case I missed - I forgot to link @CrossR 's gist in the issue (I edited it so the link is correct now, but it wasn't there before): https://gist.github.com/CrossR/a9b8b973a53f66ba78f53cd8c47fc547

It turns out that in some cases, if the `evt.deltaMode` is `1`, it gives us the `deltaY` in _lines_ instead of _pixels_. Some more info here: https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode
(And it can even be set to  _pages_ if it is `2`).

So we'll need to take that case into account, otherwise @CrossR won't be able to scroll anymore 😄 

We could add a helper function like:
```
private _normalizeScrollDeltaToPixels(delta: number, deltaMode: number): number {
    switch(deltaMode) {
           case 0:
              return delta
          case 1:
              return delta * this._screen.fontHeightInPixels
          case 2:
              return delta * this._screen.fontHeightInPixels * this._screen.height
          default:
             return delta
    }
```

And then use that in the code above, like:
```
            const normalizedDelta = this._normalizeScrollDataToPixels(evt.deltaY, evt.deltaMode)
            if (normalizedDelta) {
                 this._scrollDelta += normalizedDelta
                 ...
```

Just a thought - sorry I missed including that data point in the original issue. Let me know if you have any questions. Thanks again for working on this!

samdoiron:
Not a problem! I'm a fan of the effort.

Re: the CI failure. Not sure what is causing that, and I suspect it isn't this change. Error logs reference a missing nvim.exe.

bryphe:
Looks great, @samdoiron ! Thanks for accounting for `deltaMode` and appreciate you looking at the CI failures - I agree, it doesn't look like it's due to this change. It seems that the `nvim` binary is failing to download (likely be a GitHub throttling issue). 

I'll take a look at that separately, and bring this in the meantime, since OSX/Linux passed 👍

Awesome contribution, this will be a major usability improvement for MacBook users. Thanks @samdoiron! 💯

Issue #2035, opened 2018-04-04
Tutorial Proposal: Copy and Pasting - `y`, `p`


keforbes:
Building on the `d` delete tutorial and `w` motion tutorial explaining "nouns" and "verbs", we should have a tutorial for copy and pasting.  Possible commands should include:
`yw` - yank a word
`yy` - yank a line
`p` - paste after cursor
`P` - paste before cursor
`dw` -> `p` - show deleted characters go into paste buffer

Note that `p` and `P` are most easily understood when pasting lines.

bryphe:
Great suggestion @keforbes ! Might be a good place to introduce `xp` (transpose a character)  and `ddp` (transpose a line), too.

keforbes:
Good ideas, we should definitely include those features as well.  I'm hoping I can help implement some of these tutorials myself; the API you made for them seems pretty straightforward.  After Copy and Pasting I'd like to add a Visual Mode tutorial but I don't know if it's worth creating a separate issue for that; I'm hoping I can implement both.

You guys did a phenomenal job on this feature by the way; it's truly amazing.  I went a couple months without paying attention to Oni then I come back to find this incredible tutorial page with great visual feedback.  Well done!

bryphe:
> You guys did a phenomenal job on this feature by the way; it's truly amazing. I went a couple months without paying attention to Oni then I come back to find this incredible tutorial page with great visual feedback. Well done!

Thanks @keforbes ! 👍 Good to see you again and glad you like the feature! I think there is a lot of potential for us to "lower the bar" to modal editing with a bit more content... and maybe some fun vimgolf stuff down the road.

> Visual Mode tutorial but I don't know if it's worth creating a separate issue for that;

A visual mode tutorial would be excellent. It would showcase / explain how "selection" (which is really just visual mode) in Oni is different from other editors. Plus showing off some of the powerful stuff like blockwise visual mode would be neat.

Might be worth having a separate issue for it, just because people might comment with some ideas.

> I'm hoping I can implement both.

That would be awesome! 😃 Hit me up if you have any questions.

keforbes:
> Plus showing off some of the powerful stuff like blockwise visual mode would be neat.

Personally, I think blockwise visual mode should come in a later tutorial.  I'm thinking advanced tutorials would include things like: blockwise visual mode, named registers, marks, and macros.  I want to do basic visual mode next because I think it builds on the Copy & Paste tutorial since you can delete, copy and paste visually-selected text.

[...]

PR #2045, opened 2018-04-05
Add Tutorial: Copy & Paste

keforbes:
This creates a Copy & Paste tutorial, dependent on a new `WaitForRegisterState.tsx` class I created which validates the correct contents are found in the unnamed register.  I've overloaded the constructor so named registers can use this class in later, more advanced, tutorials.

As part of this Pull Request, I also reordered some of the other tutorials.  I moved the SearchInBufferTutorial to be the last "motion" tutorial which is then followed by the "delete" tutorials.  I think this flows better since it means the "delete" tutorials flow into the "copy & paste" tutorial which utilizes delete (and the concept of "nouns" and "verbs").

Also, one minor issue.  On my system, when copying, if I hit `y` and then wait for the editor to show `OPERATOR` mode, and then hit `w`, the `WaitForRegisterState` check works perfectly.  But if I hit `yw` quickly, it just sits there and waits for me to hit any other key before it recognizes the register was updated.  I did this development within a VM so maybe I'm introducing unnecessary lag/delay but I wanted to mention it in case others can reproduce it.  Although if you *can* reproduce it, I don't think the fix would be in my code.


keforbes:
Fixes #2035.

bryphe:
> This creates a Copy & Paste tutorial, 

Just tested it out - very nice 😄 

> dependent on a new WaitForRegisterState.tsx class I created which validates the correct contents are found in the unnamed register. I've overloaded the constructor so named registers can use this class in later, more advanced, tutorials.

Awesome, that's great! Perfect if we add a tutorial for using the registers (or even perhaps macros, or the visual mode registers...). Thanks for making it reusable.

> As part of this Pull Request, I also reordered some of the other tutorials. I moved the SearchInBufferTutorial to be the last "motion" tutorial which is then followed by the "delete" tutorials. I think this flows better since it means the "delete" tutorials flow into the "copy & paste" tutorial which utilizes delete (and the concept of "nouns" and "verbs").

Makes sense! I imagine organizing them will be sort of incremental as we add more and identify natural groupings.

> Also, one minor issue. On my system, when copying, if I hit y and then wait for the editor to show OPERATOR mode, and then hit w, the WaitForRegisterState check works perfectly. But if I hit yw quickly, it just sits there and waits for me to hit any other key before it recognizes the register was updated. 

I actually noticed this too - for me, if I yank, it doesn't register as a success until I move the cursor.

I believe the issue is with the code here: 
https://github.com/onivim/oni/blob/d51ca84b9916120679a45ff8808070a0d0ddc563/browser/src/Services/Learning/Tutorial/TutorialGameplayManager.ts#L67

Right now, we only `tick` on a very limited set of events - mode changed, cursor moved, and buffer changed. In the case of yanking, it doesn't fall into that narrow set of categories, and therefore we never tick - so the `WaitForRegisterState` never gets a chance to say its complete. If I move the cursor, it forces a tick, and then we hit the completion criteria.

I think that as the tutorials get more involved, we'll need a 'real' game loop, like:
```
window.setInterval(() => {
   this._tick()
}, 250)
```

(The alternative would be that we could keep hooking events... but that might get tricky!)

The only challenge there is we want to make sure to clear that interval when the tutorial is over or the buffer is closed. We could track this in a separate issue, as you mentioned, it's not your tutorial code at fault.

bryphe:
BTW, we tweaked our 'contributing' guidelines a bit - as a collaborator, feel free to merge in your own PRs.

They just require approval from one other collaborator - and you also have the ability to approve PRs. I approved this one so you're good to go. 👍 

Let me know if you have any questions!

keforbes:
> BTW, we tweaked our 'contributing' guidelines a bit - as a collaborator, feel free to merge in your own PRs.

I still prefer to keep you in the loop so I'm not sneaking anything in without your knowledge. :smile:

keforbes:
I'll pull this in since the fix for #2050 will resolve any lingering issues here.  As long as #2050 is fixed before the next release, no users will notice this oddity of the yank tutorial.

bryphe:
> I still prefer to keep you in the loop so I'm not sneaking anything in without your knowledge. 😄

Haha, you actually don't have to worry about that - github will actually enforce that one other collabarator  'approves' it - and if both a submitter and collaborator are cool with the change, I'm cool with it too 👍 But of course I'm happy to be looped in if there are any questions or concerns.

#2049 is a good example of a PR going in without waiting on me. It's awesome to have the help reviewing PRs - means I'm not a bottleneck 😄

PR #2049, opened 2018-04-06
feature: add config option to have sidebar open by default or not

Akin909:
My most common action on starting `oni` is immediately closing the file explorer as although I use it, I find having this open from the beginning quite incovenient *personally*. This change gives a user the ability to set the explorer to be closed when oni launches until a user choses to have it open


badosu:
Thank you very much @Akin909!!

PR #2061, opened 2018-04-07
feature/ add textmate grammar for java

Akin909:
Finally figured out how to add new text mate grammars so this is likely the first of several PR's to add grammars so Oni can syntax highlight more file types

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -213,6 +213,10 @@ const BaseConfiguration: IConfigurationValues = {
             "TypeScriptReact.tmLanguage.json",
         ),
     },
+    "language.java.textMateGrammar": {
+        ".java": path.join(__dirname, "extensions", "java", "syntaxes", "Java.tmLanguage.json"),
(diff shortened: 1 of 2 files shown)


bryphe:
Awesome! Thanks @Akin909 👍

Akin909:
@bryphe I was planning on adding a few of these but given that as you pointed out in #2038, we can just use the bundles straight from textmate can I just use the unconverted (tmBundle) files rather than going `bundle` to `cson` to `json`?. (PS not planning on doing a crazy amount as I believe you'd rather keep oni Lean and were looking to setup plugins but seemed there are a few languages that I think ought to work out of the box?)

bryphe:
> @bryphe I was planning on adding a few of these but given that as you pointed out in #2038, we can just use the bundles straight from textmate can I just use the unconverted (tmBundle) files rather than going bundle to cson to json?.

Awesome! I've been porting over the ones from VSCode (which are already converted 👍 ): https://github.com/Microsoft/vscode/tree/master/extensions Thank you for doing this!

> (PS not planning on doing a crazy amount as I believe you'd rather keep oni Lean and were looking to setup plugins but seemed there are a few languages that I think ought to work out of the box?)

I think at least these are relatively small, and provide a much better experience out-of-box, so I'm cool with adding them for popular languages 👍 They're less intense than NPM packages which often have a long dependency chain!

PR #2074, opened 2018-04-10
Bugfix/use max width for tab names

Akin909:
Realised that the strategy I added to conditionally truncate file length was missing cases as it would conditionally apply a width if the text was over 50chars.

This PR simplifies things and just uses a css `max-width` to add text overflow

BEFORE:
screen shot 2018-04-09 at 08 26 55

(under 50chars but way too long to fit into the tab, icon is overflowing)

AFTER:
screen shot 2018-04-10 at 07 37 35

Also added some initial test for the tabs components

### DIFF
--- a/browser/src/UI/components/Tabs.tsx
+++ b/browser/src/UI/components/Tabs.tsx
@@ -58,7 +58,7 @@ export interface ITabsProps {
 }
 
 const InnerName = withProps<{ isLong?: boolean }>(styled.span)`
-    ${p => p.isLong && `width: 250px;`};
+    max-width: 18em;
     white-space: nowrap;
     text-overflow: ellipsis;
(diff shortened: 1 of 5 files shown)


browser/src/UI/components/Tabs.tsx
```diff
@@ -134,13 +134,16 @@ const TabWrapper = styled.div`
     animation: ${TabEntranceKeyFrames} 0.1s ease-in forwards;
 `
 
+interface IChromeDivElement extends HTMLDivElement {
```
bryphe: Cool, thanks for defining this!

bryphe:
Looks great to me! Thanks for catching this and fixing it, @Akin909 👍

PR #2077, opened 2018-04-10
Remove sync calls in ExplorerFileSystem

DeltaEvo:


bryphe:
Thanks for making this PR, @DeltaEvo 💯 The code change looks great! Will definitely improve the perf of expanding large folders, since the UI won't be blocked anymore.

I checked out the build logs and it looks like there is a unit test failure:
```
  1) ExplorerStore
       SET_ROOT_DIRECTORY
         expands directory automatically:
     AssertionError [ERR_ASSERTION]: Validate expanded folders is set
      at Object.<anonymous> (/home/travis/build/onivim/oni/lib_test/browser/test/Services/Explorer/ExplorerStoreTests.js:41:20)
      at fulfilled (/home/travis/build/onivim/oni/lib_test/browser/test/Services/Explorer/ExplorerStoreTests.js:7:58)
      at <anonymous>
```

You can run the unit tests via `npm run test:unit:browser` or `npm run debug:test:unit:browser` (to open in an electron window for debugging). I also like to use the grep option, like: `npm run debug:test:unit:browser -- --grep Explorer`

I think the unit test worked fine in the sync case, but not in this async case now - it seems like the test isn't waiting for all the pending promises to finish.

Since the `ExplorerStore` is focused on testing the functionality of the store/epic, and not on the specifics of the filesystem implementation, perhaps it is reasonable to just add a mock for `IFileSystem` instead:
```
        const MockFileSystem = {
            readdir: async (fullPath: string) => { return [{type: "file", fullPath: filePath} as ExplorerState.FolderOrFile]},
            exists: async (fullPath: string) => { return true },
        }

        store = ExplorerState.createStore(MockFileSystem)
```

Otherwise, if we use the existing strategy (which is an ExplorerFileSystem + 'memory-fs'), we'll need to hook up a way for the test to know when the `readdir` call has completed, so it that it can `await` it prior to checking for the store state. That might be more complicated, but let me know what you think.

Thanks again for the change and your work on this, @DeltaEvo  - really appreciate it! Sorry it ended up with these extra investigations 😳

DeltaEvo:
@bryphe finally i kept the memory file system, because it can be useful later :smile:

DeltaEvo:
Uw we seems to have a strange build failure, I will try to retrigger one by closing and reopening the PR

browser/test/Services/Explorer/ExplorerStoreTests.ts
```diff
@@ -38,8 +61,9 @@ describe("ExplorerStore", () => {
             })
 
             await TestHelpers.waitForAllAsyncOperations()
+            TestHelpers.tick(0) // execute setImmediate of memory-fs callbacks
+            await Promise.all(explorerFileSystem.promises)
```
bryphe: Awesome, thanks for adding an explicit way to wait for those operations to complete! 👍

bryphe:
Just tried out the latest changes, look great @DeltaEvo ! Thanks for troubleshooting the test and figuring out a way to keep the MemoryFileSystem around 😄 

> Uw we seems to have a strange build failure, I will try to retrigger one by closing and reopening the PR

Ah ya, looks like it was an intermittent issue on OSX - re-ran the build and everything looks good. Awesome work - bringing it in now!

PR #2078, opened 2018-04-10
Stop propagation of event in File Explorer to avoid multiple click events

DeltaEvo:
#2077 is needed to see this bug

### DIFF
--- a/browser/src/Services/Explorer/ExplorerView.tsx
+++ b/browser/src/Services/Explorer/ExplorerView.tsx
@@ -42,6 +42,12 @@ const scrollIntoViewIfNeeded = (elem: HTMLElement) => {
     // tslint:disable-next-line
     elem && elem["scrollIntoViewIfNeeded"] && elem["scrollIntoViewIfNeeded"]()
 }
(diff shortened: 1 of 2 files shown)


browser/src/Services/Explorer/ExplorerView.tsx
```diff
@@ -42,6 +42,12 @@ const scrollIntoViewIfNeeded = (elem: HTMLElement) => {
     // tslint:disable-next-line
     elem && elem["scrollIntoViewIfNeeded"] && elem["scrollIntoViewIfNeeded"]()
 }
+const stopPropagation = (fn: () => void) => {
```
bryphe: Nice use of a higher-order function to share the code between the handlers! 👍

bryphe:
@DeltaEvo , thanks for fixing this too as part of #2077 ! Change looks excellent. Will bring this in as soon as the window build finishes 👍

bryphe:
Bringing this in now! Thanks for the contribution, @DeltaEvo ! 💯

PR #2080, opened 2018-04-10
Add config option for default action of QuickOpen.

CrossR:
Added a config option for the default  action of the QuickOpen being to open in the current buffer.

Does it make sense to also have a `openInTabByDefault` or something similar? Since currently `Enter` and `Ctrl-Enter` does the same in `"buffers"` mode. (Though that was the case before this PR). May make sense to leave that until anyone actually asks for that.


Akin909:
Looks good 👍 not sure what you mean re an openInTabDefault, is that for `buffers` mode? if so then I think that'd be a great addition I occasionally use tabs on a big monitor so having the option to open in a tab as opposed to a second redundant option would be good

CrossR:
> not sure what you mean re an openInTabDefault, is that for buffers mode? if so then I think that'd be a great addition I occasionally use tabs on a big monitor so having the option to open in a tab as opposed to a second redundant option would be good

Yep!
Currently we do have all the options a user would want, (`Enter` + `t/s/v/ctrl` for New tab, Horizontal, Vertical, current buffer), but `Enter` always opens in a new tab or in the current buffer (for `tabs` vs `buffers`). Its just that for `buffers` `enter` and `ctrl+enter` do the exact same thing - `:e!`, which seems a bit redundant.

Perhaps what would be more useful would be exposing the `Oni.FileOpenMode` option to the config, such that we don't need a few different config options, and a user can instead just pass over a `FileOpenMode` and have that used? ie `"editor.quickOpen.defaultOpenMode": Oni.FileOpenMode.Edit,` would than achieve what this PR was for.

Then a user could just bind `Enter` to whatever and `Ctrl-Enter` to something else. That would keep `Enter` as the default, and `Ctrl-Enter` as the "alternative", whilst meaning we don't need a few config options for things that are basically the same.

bryphe:
Thanks for taking this on @CrossR ! 💯 

> Perhaps what would be more useful would be exposing the Oni.FileOpenMode option to the config, such that we don't need a few different config options, and a user can instead just pass over a FileOpenMode and have that used? ie "editor.quickOpen.defaultOpenMode": Oni.FileOpenMode.Edit, would than achieve what this PR was for.

I like this idea a lot! I think it would be more clear to have that explicit control and simplify things a bit. It helps because then we don't have to get in the business of trying to handle all the use cases between buffers/tabs - we can pick a reasonable default but the user can override with whatever they'd like.

> Then a user could just bind Enter to whatever and Ctrl-Enter to something else. That would keep Enter as the default, and Ctrl-Enter as the "alternative", whilst meaning we don't need a few config options for things that are basically the same.

Would that mean we'd have two configuration settings, like:
```
"editor.quickOpen.defaultOpenMode": Oni.FileOpenMode.Edit,
"editor.quickOpen.alternateOpenMode": Oni.FileOpenMode.VerticalSplit,
```
?

Or just that we'd expect the user to rebind `<C-enter>` if they wanted to override? I'm thinking perhaps if we switch to the `FileOpenMode` strategy, it might make sense to have them both be explicit settings.

Thanks again for the work on this, @CrossR ! 👍

badosu:
@bryphe @CrossR Just chiming in with the way I've used similar functionality earlier (with CtrlP):

- `<C-v>`: opens a vertical split
- `<C-t>`: opens a new tab
- `<C-h>`: opens an horizontal split
- `<Enter>`: opens in current buffer, and if for some reason can't (due to file modified and not written) opens a vertical split.

CrossR:
Okay, work calmed down a tiny bit so I've done the few changes I wanted to make here.
Works great for the `config.tsx` users since they get auto complete (if they know to use `Oni.FileMode` but hopefully we can address that for all config options elsewhere), though I expect its a bit more awkward for people using the `config.js` still.

Though, I'll update the docs at least, to make it a bit easier.

CrossR:
Looks like I can't merge this as the Mac test has failed due to a timeout in the Macro tests....
I'd assume that my changes in the Oni side of things can't have broken that though.

PR #2083, opened 2018-04-12
Specify optional 'baseVimBackground' for themes

bryphe:
Our theming system today doesn't handle differentiating between dark/light versions of themes very well. The current implementation has some issues - notably, when a colorscheme is set, we check that the colorscheme's name matches the name of our active themes `vimBaseTheme` - if they don't match, we assume the user picked a random colorscheme, and we'll use our inferred colors. 

With our implementation today, we don't reliably detect this, and we don't do anything to set the `background` to `light` / `dark`, which many Vim themes require.

Also, without being able to set the `background` from the metadata, it's awkward to convert existing vim themes - we end up having to wrap them in separate `init.vim` files - it'd be great to not need to do that!

This adds a `baseVimBackground` property for themes, so that themes can explicitly set `light` or `dark`, without needing to create an additional VimL wrapper. It also shows how this works for the solarized theme - we can delete the light/dark versions of the VimL wrappers, and just specify this property in the theme. The downside is this exposes a bug with the `solarized8_light.json` colors - it looks like most of the colors are dark (it was actually using the _inferred_ theme, because the `solarized8` colorscheme reported didn't match what was specified in theme).

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -44,6 +44,7 @@ import { Configuration, IConfigurationValues } from "./../../Services/Configurat
 import { IDiagnosticsDataSource } from "./../../Services/Diagnostics"
 import { Overlay, OverlayManager } from "./../../Services/Overlay"
 import { SnippetManager } from "./../../Services/Snippets"
+import { IThemeMetadata } from "./../../Services/Themes"
 import { TokenColors } from "./../../Services/TokenColors"
 
 import * as Shell from "./../../UI/Shell"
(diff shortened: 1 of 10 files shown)


parkerault:
Wow, that was fast! 😄

bryphe:
@parkerault , haha, thanks! 😄 Just wanted to get the idea out and see if it'll work to unblock the scenario. 

If this works out, I'll add a test case here to build out our 'safety net' and validate switching between light/dark theme and make sure it picks up the right token colors 👍

parkerault:
I noticed that a lot of the icons are designed for a dark background; I think at some point someone is going to have to create another set of icons for light backgrounds or this will continue to be an issue: 

image

parkerault:
Also, the status bar language server and prettier buttons:

image

bryphe:
> I noticed that a lot of the icons are designed for a dark background; I think at some point someone is going to have to create another set of icons for light backgrounds or this will continue to be an issue

Ah ya, actually, the icon theme specifies light/dark versions I believe: https://github.com/onivim/oni/blob/master/extensions/theme-icons-seti/icons/seti-icon-theme.json

(There are alternate colors specified for light, ie):
```
        "_json_light": {
            "fontCharacter": "\\E046",
            "fontColor": "#b7b73b"
        },
        "_json": {
            "fontCharacter": "\\E046",
            "fontColor": "#cbcb41"
        },
````

We just don't use them at the moment. But since we have the light/dark information now - we potentially could make a better decision here 👍

PR #2098, opened 2018-04-16
Use Vim shell option as default value for terminal.shellCommand

DeltaEvo:
Currently the Oni Terminal is different from neovim Terminal (Invoked with `:terminal`), the shell is not the same on both options, oni use a configuration setting `terminal.shellCommand` and Neovim use the `shell` variable

I changed only the default value of `terminal.shellCommand` but should we be aligned with the Neovim behavior and use `shell` variable ?
And if yes do we remove the `terminal.shellCommand` ?

### DIFF
--- a/browser/src/Services/Terminal.ts
+++ b/browser/src/Services/Terminal.ts
@@ -15,13 +15,12 @@ export const activate = (
     configuration: Configuration,
     editorManager: EditorManager,
 ) => {
-    const getTerminalCommand = () => {
-        const terminalCommand = configuration.getValue("terminal.shellCommand", "")
-        return `term://${terminalCommand}`
-    }
+    const openTerminal = async (openMode: Oni.FileOpenMode) => {
+        const terminalCommand =
+            configuration.getValue("terminal.shellCommand") ||
+            (await editorManager.activeEditor.neovim.callFunction("nvim_get_option", ["shell"]))
 
-    const openTerminal = (openMode: Oni.FileOpenMode) => {
-        editorManager.activeEditor.openFile(getTerminalCommand(), { openMode })
+        editorManager.activeEditor.openFile(`term://${terminalCommand}`, { openMode })
     }
 
     commandManager.registerCommand({
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -410,7 +410,7 @@ const BaseConfiguration: IConfigurationValues = {
     "tabs.showIndex": false,
     "tabs.wrap": false,
 
-    "terminal.shellCommand": os.platform() === "win32" ? "cmd" : "bash",
+    "terminal.shellCommand": null,
 
     "ui.animations.enabled": true,
     "ui.colorscheme": "nord",


CrossR:
I think the `terminal.shellCommand` should be left in, since I can see a number of situations that people would prefer to overload it potentially. Ie, I'm mainly a Windows user and to my knowledge, `:terminal` defaults to `cmd` for me, but I almost always use WSL `bash` instead.

Similarly, in the future I could see it being useful to be able to pass any shell over and have a the ability to pick a certain shell. I've added something similar to my Oni config for now, but I could see other people finding it useful as well.

Changing the default though makes sense to me, such that users of shells other than `bash` get that loaded and so on.

badosu:
I think we should adhere to the principle of least surprise, which in this case would be:

- If `terminal.shellCommand` not specified by the user use vim's `shell` variable
- Use `terminal.shellCommand` if specified

The advantage of using the `shell` variable fallback is that vim already has logic to search for the user's shell and we inherit the user's config without additional configuration required for Oni.

DeltaEvo:
Personaly I think that `terminal.shellCommand` should have the same value of the `shell` variable in vim  (So if `terminal.shellCommand` set `shell` variable to it's value) :thinking:

Idk what is the best solution here :joy:

badosu:
@DeltaEvo 

> Personaly I think that terminal.shellCommand should have the same value of the shell variable in vim (So if terminal.shellCommand set shell variable to it's value) thinking

Yes, this implementation would have the behaviour I mentioned.

CrossR:
>If terminal.shellCommand not specified, use vim's shell variable
>Use terminal.shellCommand if specified

Sorry if it wasn't clear, thats what I mean too!

DeltaEvo:
Ow xD, so everyone is on the same line, I will implement this then, thanks :heart:

badosu:
Great! This also is expected on some vim plugins, fzf for instance: https://github.com/junegunn/fzf.vim/issues/229

badosu:
Btw, great catch! I just tested `:terminal` in my installation and it open bash, which is unusable for me since I depend on many setting from my `.zshrc`

DeltaEvo:
I changed the implementation thanks @badosu & @CrossR :heart:

browser/src/Services/Terminal.ts
```diff
@@ -15,26 +13,21 @@ export const activate = (
     configuration: Configuration,
     editorManager: EditorManager,
 ) => {
-    const getTerminalCommand = () => {
```
badosu: Is this function only used in this scope?

DeltaEvo: Yes, it's not exported or anything

browser/src/neovim/NeovimInstance.ts
```diff
@@ -471,6 +471,11 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                     // set title after attaching listeners so we can get the initial title
                     await this.command("set title")
 
+                    const shell = this._configuration.getValue("terminal.shellCommand")
```
badosu: So this is overriding the `shell` variable which currently would mislead the user thinking it's just for the terminal functionality. We have two options, to keep this behaviour but change the namespace to not include terminal or to only use the shell for the terminal functionality.

If using the second approach we would keep compatibility with the previous way, it would be a single line change to evaluate `set shell` in the `DefaultConfiguration.ts` file.

DeltaEvo: Aw, it what this beavior I was thinking with
```
So if terminal.shellCommand is set, set shell variable to it's value
```
I wanted to align the behavior of oni's command to open terminal and :terminal command on neovim, I was surprised by the fact that the two was not equivalent

So it's like your comment, you want only to use shell as fallback :thinking: 

Having two value for this `terminal.shellCommand` for oni and `shell` for other neovim plugins seems more misleading for the user, because the shell used in oni could be different if he use a Neovim plugin that spawn a shell, it was why i wanted to aligned the two behaviors generally an user only have one shell, so why 2 different configurations ?

DeltaEvo:
O_o wow such strange build failure https://travis-ci.org/onivim/oni/jobs/367200492, @bryphe seems to be an edge case in `oni-release-downloader`

browser/src/neovim/NeovimInstance.ts
badosu: Yes, sorry I meant the shell on the terminal invocation.

I agree that having multiple shell behaviours is confusing, but that's what's implied in the `terminal.shellCommand` from what I understand and that's the current behaviour.

As I mentioned, we can keep current behaviour, the only change required for that would be to fetch the `shell` variable on the default configuration. If we change it then it would be a global setting and the namespace would have to be renamed accordingly.

I think @bryphe may have an insight on this issue.

bryphe: I would actually prefer to keep the `shell` setting and `terminal.shellCommand` separate. On Windows, `set shell=powershell` doesn't work well, but I want to use `powershell` as my  terminal shell in Oni (since it has git enhancements and stuff!). So I sometimes use the `!` commands like `:!rimraf /some/dir`, and I'd like them to use `cmd.exe`, but if I open a terminal - I want powershell.

(Note that you can actually set the `shell` variable via `vim.setting.shell` today, too).

I look at `terminal.shellCommand` as just a short-hand for running `term://{some terminal command}`, and I believe there are use cases to have a separate terminal shell vs ex mode shell.

So my preference would be:
- Set `terminal.shellCommand` to null by default
- If `terminal.shellCommand` is null / undefined, and we're opening from the command palette, we just open using `term` (which will use the default shell)
- If `terminal.shellCommand` is defined, and we're opening from the command palette, we'd open with `term://${terminal.shellCommand}`

That solves my use case where I can specify 'powershell' on my windows machine for opening a terminal, but still use the default shell `cmd.exe` for `!` commands.

bryphe: (And by default, since `terminal.shellCommand` isn't defined, you get the default Vim behavior unless you explicitly override it).

browser/src/Services/Terminal.ts
```diff
@@ -15,26 +13,21 @@ export const activate = (
...
-        editorManager.activeEditor.openFile(getTerminalCommand(), { openMode })
+    const openTerminal = (vertical: boolean) => {
+        editorManager.activeEditor.neovim.command(`:${vertical ? "v" : ""}sp! | terminal`)
     }
 
     commandManager.registerCommand({
         command: "terminal.openInVerticalSplit",
         name: "Terminal: Open Vertical",
         detail: "Open a terminal emulator in a vertical split",
-        execute: () => openTerminal(Oni.FileOpenMode.VerticalSplit),
```
bryphe: Is there a reason we need to change this? (Or a case it's not working in today?)

The reason I ask is I'm working on a 'multiplexed' mode - you can enable it today with `"editor.split.mode": "oni"` (but it's still experimental). This lets you open editor windows in new tabs:
multiplexed-windows




So using the `openFile` API with that open mode will do `vsp`/`sp` if we aren't in that multiplexed mode... but if we are in that multiplexed mode, it'll use this new type of split.

(I actually showcase this in our intro video on the website, too! https://www.youtube.com/watch?v=uLpbxfq0IXM&feature=youtu.be&t=40s)

badosu: The current implementation replaces the whole `shell` variable on vim side. We can just open `term://` (which is going to happen if `terminal.shellCommand` is null or eval `set shell` on the `terminal.shellCommand` configuration setting. This should be just a single line change for the whole PR.

DeltaEvo: It's because I don't open a file here, i'm using a command, and oni's api currently doesn't have a wait to spawn a command with a FileOpenMode

browser/src/neovim/NeovimInstance.ts
DeltaEvo: Okay, so I will implement this later (it's 2:53 for me, so it's bed time :joy: )

bryphe: Haha sounds good! That's super late, get some sleep! 😄

bryphe:
Latest changes look great @DeltaEvo - thanks for accounting for all the feedback! ❤️ 

Bringing it in now! 👍

PR #2099, opened 2018-04-16
Regression from #2078: Event can be null when Sneak navigation is used

DeltaEvo:
I broke navigation with sneak mode with #2078, because the onClick handled is called without arguments


bryphe:
Ah, thanks for catching this @DeltaEvo ! 👍 Fix looks great!

PR #2019, opened 2018-04-02
[WIP] Update intro video with 0.3.2 features

bryphe:
- [ ] Show achievements
- [ ] Show creating browser window
- [ ] Show creating terminal
- [ ] Start 'my-react-app'
- [ ] Make text change
- [ ] Add setting logo
(show usual js stuff - language service + snippet)
- [ ] Close browser
- [ ] Close terminal
- [ ] Show interactive tutorial

Issue #2094, opened 2018-04-14
Expand config metadata settings with default values.


CrossR:
We've spoken about this in a bunch of places I feel, but I couldn't find an exact issue to do with it....
Now we metadata associated with the config options, I think the last big piece of that puzzle is the addition of some way of specifying the valid values for that option.

Currently, we just silently fail in the background with default values, so if you make the very easy mistake of settings your tab mode to `tab` rather than `tabs` you default to buffers! Which for when there is only a few options, seems a pretty bad user experience. If instead an error was raised of "`tab` isn't valid" that is a lot better, and if there is autocomplete, even better again.

VSCode does something similar (or I assume they do), which allows the config experience to be a lot nicer and include auto complete and equivalent.

I see a few big blockers for this though:

- [ ] Not much of the config is currently using this new metadata API. We'd need to refactor most of the config options to do so. I'm not against doing some of it, I find it fairly relaxing sometimes to just write bits up and what not!
- [ ] Surface the appropriate metadata to the user, like the description and stuff.
- [ ] Solving the UI problem for when an input has no set list of values. Ie for `oni.loadInitVim` we can surface `true`/`false`, but we also need to surface some way of saying "Hey you can also pass a path here".

bryphe:
Thanks for logging the issue, @CrossR ! Agree, this is the next step we need.

> Not much of the config is currently using this new metadata API. We'd need to refactor most of the config options to do so. I'm not against doing some of it, I find it fairly relaxing sometimes to just write bits up and what not!

That would be awesome! Maybe we can do this incrementally - there are a _lot_ of places we'd need to add. But the good thing is the new API also lets us figure out if we can 'hot load' a config setting or not, so we can have a better experience around letting users knowt o reload, too.

> Surface the appropriate metadata to the user, like the description and stuff.

I actually have a prototype for this, for the very few config settings we have at the moment.

PR #2104, opened 2018-04-17
[Prototype] #2019 - Part 2 - Output some documentation for configuration settings

bryphe:
We have a new API for registering configuration variables w/ metadata (like documentation, whether the setting is hot-reloadable, etc), but we aren't really using it yet.

This change checks to see if there is metadata defined as part of the configuration registration, and outputs in the reference buffer:
image

A couple things remaining:
- [ ] Filter out the 'activate' and 'deactivate' (undefined) settings. These were automatically filtered out in the previous `JSON.stringify` strategy.
- [ ] Add a Ci Test - this can add a configuration value w/ metadata, and validate it shows up in the reference buffer.

A next step later is that we don't have a setting for the allowed set of values for the configuration setting - we should potentially add that as part of the metadata.

### DIFF
--- a/browser/src/Services/Configuration/ConfigurationEditor.ts
+++ b/browser/src/Services/Configuration/ConfigurationEditor.ts
@@ -15,6 +15,7 @@ import { EditorManager } from "./../EditorManager"
 
 import { Configuration } from "./Configuration"
 import { DefaultConfiguration } from "./DefaultConfiguration"
+import { flatMap } from "./../../Utility"
 
 // For configuring Oni, JavaScript is the de-facto language, and the configuration
 // today will _always_ happen through `config.js`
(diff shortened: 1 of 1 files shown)


CrossR:
Thinking about this, could a different version of this output markdown?

That is, this (alongside #2094) could be used to auto generate our config docs potentially.
If we have the comments, potential values and config name being output to a file, we have everything we need to make the doc page automatically.

All we'd need is a slightly different function that instead of outputting the JSON file, outputs markdown in some fashion.

On a different  note, if some examples of how to comment the docs are given, I could get started on commenting the current commands soon, to help this be worth while.

bryphe:
> Thinking about this, could a different version of this output markdown?

Definitely! We could even have a CiTest that generates the markdown:
```
const values = oni.configuration.getValues()
Object.keys(values).forEach((configurationName) => {
       const defaultValue = values[configurationName]
       const metadata = oni.configuration.getMetadata(configurationName)
     
       appendConfigurationDescriptionToMarkdown(configurationName, defaultValue, metadata.description)
})
```

We just need the `appendConfigurationDescriptionToMarkdown` function, and render it in place of the JSON, or output to a file, etc.

I think we have everything we need to make that happen!

> On a different note, if some examples of how to comment the docs are given, I could get started on commenting the current commands soon, to help this be worth while.

Definitely. I think the only examples we have so far are in the browser feature:
https://github.com/onivim/oni/blob/3bb8412b0d0839e45a51ba579109a094721d310b/browser/src/Services/Browser/index.tsx#L104

We'd basically need to start moving configuration values out of `IConfigurationValues.ts`/`DefaultConfiguration.ts` into these `configuration.registerSetting` calls.

I think a good starting point would be the `autoClosingPairs.enabled` and `autoClosingPairs.default` settings. We could move these from `DefaultConfiguration.ts` and `IConfigurationValues.ts` into the `activate` method of the `AutoClosingPairs` service, and use the `configuration.registerSetting` call to initialize and add some documentation to them.

Would be awesome to have help, the more configuration values we have migrated to this new API, the more benefit we'll get from auto-generating the docs 👍 Keep me posted if you have any questions!

PR #2106, opened 2018-04-17
Fix relative module loading for configuration

DeltaEvo:
Currently if we want to load a module used by oni in our configuration (React for example), we need to use an absolute path because node module resolution algorithm only traverse directories up to find a node_module folder. This make it more difficult to share our onivim configuration between different computers

To fix this with this commit, we read the content of the configuration and wrap it in a "commonjs executing function" to use the same path as current oni's files and so the same algorithm


bryphe:
I'm working on checking the caching behavior in #2111 , but I don't believe the build failures are due to this change 👍 

Code change looks good to me - cool idea @DeltaEvo !

bryphe:
Build passed after re-running it - bringing it now! Thanks @DeltaEvo ! 👍

PR #2109, opened 2018-04-17
Update to webpack@4

DeltaEvo:
## Before

- `npm run build`: 64239ms
- `npm start`: 38832ms
- `du -h lib/browser`: 4.9M

## After

- `npm run build`: 27127ms
- `npm start`: 35269ms
- `du -h lib/browser`: 4.9M

Note: babili is not needed anymore since webpack4 has a minifier by default (uglify-es) and optimize-js optimization is old, since browsers have fixed this bug (and the plugin optimize-js-plugin has not been update to webpack 4)


DeltaEvo:
Uw @Akin909 this dosen't seems to have fixed it, do you have an idea to why this build is failing ?

Akin909:
@DeltaEvo Hmm seems to be looking for the `sanitize` method on `marked` which we use to render markdown but that is in no way related to the prettier plugin, we dont use or require marked anywhere near the prettier one, but the prettier plugin is built right after the `oni-markdown-plugin` so not sure if the error is coming from elsewhere but appearing in that test.

If you run the plugin locally does it work correctly or does it throw the same/ a similar error?

DeltaEvo:
@Akin909 locally the plugin seems to work, i will try to look into `oni-markdown-plugin` (I was thinking that marked was used by prettier :joy: )

DeltaEvo:
Well, tests saved us from a very strange bug, thanks @bryphe and @Akin909 (and all other) for these tests :heart:

browser/src/Editor/NeovimEditor/markdown.ts
```diff
@@ -4,8 +4,7 @@ import * as marked from "marked"
 import * as Log from "./../../Log"
 import { IGrammarPerLine, IGrammarToken } from "./../../Services/SyntaxHighlighting/TokenGenerator"
 
-// tslint:disable-next-line
-const { default: DOMPurify } = require("dompurify")
+import * as DOMPurify from "dompurify"
```
DeltaEvo: I defeated you little bug:stuck_out_tongue:

bryphe:
Nice catch, @DeltaEvo ! Glad you were able to figure out the issue 👍 And ya, definitely glad we had test coverage around that - nice work on the CiTest @Akin909 !

bryphe:
Just looked at the snippets test issue - looks like there is a crash here:
```
        const detectIndentPromise = import("detect-indent")

        const [bufferLines, detectIndent] = await Promise.all([
            bufferLinesPromise,
            detectIndentPromise,
        ])

        const ret = detectIndent(bufferLines.join("\n"))
```

The `detectIndent` method can't be found. I can repro this just by adding a snippet (don't need to run the test, just try expanding the `import snippet` in a typescript/javascript file)

bryphe:
Looks like the new `import` behavior is working a bit differently:
image

Before, `detectIndent` was a function, but now it has a `default` member - I wonder if there is a difference in how it resolves/splits async imports.

A quick fix that resolved this locally was to just do:
```
        const ret = (detectIndent as any).default(bufferLines.join("\n"))
```
Unfortunately, it requires a cast to `any` - the typings for `detect-indent` assume the function is exported, which is out of sync with what the `import` is doing. Not sure if this is a webpack bug, or something we are doing wrong. I'm okay with the `any` fix though, just might need a comment to explain why its needed 😄

DeltaEvo:
Thanks for investigating at this @bryphe this new behavior is described here https://medium.com/webpack/webpack-4-import-and-commonjs-d619d626b655
To avoid using any and so loose typing, I think i will remove the dynamic import on `detect-indent`, it's a fairly small module so using a dynamic import on it seems to add more overhead that something else :)

Akin909:
@DeltaEvo I agree re. `detect-indent` 👍, this branch looks good now and all tests are passing so will bring this in if you're done, and quick check with @bryphe if you have any objections?

browser/src/Editor/NeovimEditor/markdown.ts
Akin909: Yeah I had some weird behaviour with that module (hence the bizarre import strategy) which the new changes to webpack seem to have normalised

PR #2115, opened 2018-04-18
Feature/add rename file functionality

Akin909:
Added rename (undoeable) functionality depends on #1839,

Outstanding
===

- [x] Add new epic tests
- [x] Add notification for rename

rename


browser/src/Input/KeyBindings.ts
```diff
@@ -119,6 +133,17 @@ export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configurati
 
     input.bind("<s-c-b>", "sidebar.toggle", isNormalMode)
 
+    // Explorer
```
badosu: Great keybindings! :clap:

browser/src/Input/KeyBindings.ts
```diff
@@ -119,6 +133,17 @@ export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configurati
...
+    input.bind("d", "explorer.delete.persist", isExplorerActive)
```
badosu: What is the difference between `delete` and `delete.persist`?

CrossR: I think `delete.persist` keeps the deleted file around for some amount of time in a temp folder, such that you can undo it (as long as its under the file size limit).

`delete` deletes it straight away.

Basically...an Oni recycle bin for those who want it. For those who don't they can rebind `d` to `delete` and instantly delete. May be worth adding that to your minimal config? Though I'm not sure what NerdTree etc do tbh, its been a while.

badosu: This seems fine to me, thanks @CrossR. Usually alternative actions I prefer the upcased char instead of Control, but that's fine.

bryphe:
Looks amazing, @Akin909 ! Great work! 👏

bryphe:
Bummer, looks like this test is failing on Windows:
```
[00:06:39] FAIL ui-tests\NodeView.test.tsx
[00:06:39]   ● Test suite failed to run
[00:06:39] 
[00:06:39]     The specified module could not be found.
[00:06:39]     \\?\C:\projects\oni\node_modules\keyboard-layout\build\Release\keyboard-layout-manager.node
[00:06:39]       Error: The specified module could not be found.
[00:06:39]       \\?\C:\projects\oni\node_modules\keyboard-layout\build\Release\keyboard-layout-manager.node
[00:06:39]       at Runtime.requireModule (node_modules/jest-runtime/build/index.js:328:31)
[00:06:39]       at Object.<anonymous> (node_modules/keyboard-layout/lib/keyboard-layout.js:4:31)
```

It's not clear to me why this module is being required - I would guess it's being brought in as a dependency for `TextInputView`, but `TextInputView` doesn't actually depend on it..

Akin909:
I managed to fix this locally by running npm rebuild but I think I need to have jest ignore that module

Akin909:
Seems jest is resolving imports in the explorer view file which is leading to that module which is used in the Keyboard layout by way of the vim navigator :O, I'm going to attempt a last ditch attempt at mocking instead otherwise I've upped the coverage of the actually missed bit in the diff and might hold of on an enzyme test here till I can figure out whats going on

bryphe:
Cool, thanks for investigating and fixing the `jest` issue @Akin909 ! Change looks excellent - I played around with renaming files and folders via the explorer, worked great 👍

Govinda-Fichtner:
@bryphe I was wondering where I can find the documentation for the keyboard shortcuts for this new functionality? I could not find it here: https://onivim.github.io/oni-docs/#/./oni.wiki/Shortcuts 

In https://github.com/onivim/oni/pull/2130 some shortcuts are documented directly at the beginning of the issue and they are working: <c-e> and <c-f>. Is it possible to remap them? 

It would really be nice if this kind of information would be available as part of the release information as well. Maybe I just did not look at the right places for this information.

I would also offer working on the documentation for Oni, but at least I need to know where to look up that information.

Akin909:
@Govinda-Fichtner apologies I haven't had time yet to write up documentation for the new explorer functionality hopefully if I get a moment later on today I'm going to write an explainer for the new features, yes they can all be remapped based on the command names.

If you check back in later today I'll have written out the commands.

Govinda-Fichtner:
@Akin909 This is awesome! I was really looking forward to new the file explorer stuff. Having more functionality like Nerdtree is an important step for me.

Govinda-Fichtner:
@Akin909 Where did you add the documentation for the file explorer shortcuts?

CrossR:
Hey @Govinda-Fichtner, in another issue @Akin909 had mentioned his mac broke so I wouldn't be surprised if he's catching up on stuff.

For now (and I realise its not the best) you can check the code directly and hopefully the commands are fairly self explanatory.

https://github.com/onivim/oni/blob/75e2b74770bdbcbb884af348b05de19e2f6d45f1/browser/src/Input/KeyBindings.ts#L134-L147

The difference between `delete` and `delete.persist` is that `delete` will place the file in a temp location as long as it isn't too big to be undone with `u`, whereas `delete.persist` will permanently delete the file. 

As for remapping, you'd do something like this in your Oni config

```
oni.input.bind("<c-r>", "explorer.rename")
```

That said, I think we need to improve the filtering for user binds. As you can see in the default binds, the binds only work when the explorer is active, so we should probably offer an easy way to replicate that functionality in the user config, since currently I think its a bit of a pain.

Akin909:
@CrossR 👍 still working on a loaner mac so definitely behind on stuff, @Govinda-Fichtner I added the key bindings that are available to the shortcuts [here](https://github.com/onivim/oni/wiki/Shortcuts) but haven't written up more of an explainer of the functionality yet will do in a little bit once I get my machine back

bryphe:
> That said, I think we need to improve the filtering for user binds. As you can see in the default binds, the binds only work when the explorer is active, so we should probably offer an easy way to replicate that functionality in the user config, since currently I think its a bit of a pain.

Good point, @CrossR !

The `commands` have the option of specifying a filter, too. When we evaluate an input character, we first  look at the filter provided by the user `bind`, and then we look at the `enabled` function on the `command` (if its a string command that's passed). That's really the one advantage of having a string `command` registered vs just having a function - when registering this string `command`, an `enabled` function can be optionally provided to specify when the command when should be active. This is a perfect use case of leveraging that `enabled` filter so that the user doesn't need tow orry about it.

So to make this more ergonomic, so that the user doesn't need to know to add `isExplorerActive` to all the explorer bindings, we should push that `isExplorerActive` filter over to where we register the commands in `ExplorerSplit.tsx`: 
https://github.com/onivim/oni/blob/75e2b74770bdbcbb884af348b05de19e2f6d45f1/browser/src/Services/Explorer/ExplorerSplit.tsx#L124

If we add that `isExplorerActive` predicate to the filter for the commands, then the user will just be able to do:
```
input.bind("r", "explorer.rename")
```

If the `enabled` filter returns `false` for `explorer.rename`, then that will be skipped / not evaluated.

bryphe:
FYI, I ported this out to an issue in #2194 👍

Issue #1506, opened 2018-02-10
Enhancement/handle directory changes outside oni


Akin909:
If the directory is changed outside oni, whilst outside oni and theres an attempt to navigate within it in the file explorer an error is thrown that the directory does not exist. Not sure what the correct response should be but I think a potentially valid one would be to move upwards till the first valid directory is found. 

Say I create a project then change my mind or rename it Oni should hook into the focus gained command to check whether the active dir is still present in the file system if not it should *maybe* just navigate upwards the reason I suggest this is because its hopefully less back tracking than say just navigating to the home dir which would be easier but I think a bit annoying as you now have to go searching all the way down for where you were I'd think as a user you probably just want it to back out till the next reasonable location.

Akin909:
@bryphe regarding this issue the simplest possible solution I can see would be checking if the file exists on `focusGained` in oni but this makes the assumption that this change isn't happening inside oni via a commandline command to move a file or dir.

I was thinking and wanted to run past you the idea of oni having a `fileSystem` watcher, potentially as a service to stick with the principle of one thing with a single purpose, which could then provide this functionality to areas of the app that would want to know about the status of the file system for example to file explorer the workspace etc. any thoughts?

Re. the implementation I've read around using `fs.watch` and its associates but have found that it comes with a lot of issues generally a lot of which are macOS specific alot of others more general plus its intensive on the user's CPU, does not recursively watch files and was apparently responsible for `Sublime` having issue with not watching files properly. Two alternates the first is `vscodes` solution and seems to be a well established typed solution is [chokidar](https://github.com/paulmillr/chokidar) the other is [`@atom/watcher`](https://github.com/atom/watcher) which is atom's home grown apparently highly performant bleeding edge alternative (side note, tried using this earlier and lead to a node_modules implosion...), I'm happy to proceed in whatever direction would serve oni best.

bryphe:
Thanks for thinking about this and putting together this proposal, @Akin909 .

> I was thinking and wanted to run past you the idea of oni having a fileSystem watcher, potentially as a service to stick with the principle of one thing with a single purpose, which could then provide this functionality to areas of the app that would want to know about the status of the file system for example to file explorer the workspace etc. any thoughts?

I think this sounds reasonable. I know I've hit this where I have a folder expanded in the explorer, I save a new file, and I don't see it in the explorer until I collapse/expand the folder. Would be great to have a story for that!

[...]

PR #2116, opened 2018-04-18
Feature/filesystem watcher

Akin909:
@bryphe due to webpack changes in #2109 🎉 issues I was having getting `chokidar` the FileSystem watching package we discussed in #1506 finally works as expected 💯.

This PR is a an initial implementation of the `FileSystemWatcher`, Its basically a thin layer around chokidar event listeners which I in turn use to create event listeners we subscribe to in the explorer

Result is that creating new files or deleting new files causes the explorer to update correctly

watched


bryphe:
Very cool @Akin909 ! 👍

browser/src/Services/FileSystemWatcher/index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    get allWatched(): chokidar.WatchedPaths {
+        return this._watcher.getWatched()
+    }
+
+    get onChange(): IEvent<IFileChangeEvent> {
+        return this._onChange
+    }
+
+    get onMove(): IEvent<IFileChangeEvent> {
```
bryphe: I like the abstraction a lot - and it seems like it'd be relatively straightforward for us to have a 'mock' implementation for tests, if we need it 👍

browser/src/Services/FileSystemWatcher/index.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    private _onAdd = new Event<IFileChangeEvent>()
+    private _onAddDir = new Event<IStatsChangeEvent>()
+    private _onMove = new Event<IFileChangeEvent>()
+    private _onChange = new Event<IFileChangeEvent>()
+    private _defaultOptions = { ignored: "**/node_modules" }
+
+    constructor(watch: IFSOptions = {}) {
+        this._workspace = Workspace.getInstance()
+        this._activeWorkspace = this._workspace.activeWorkspace
```
bryphe: I'm wondering - would it make sense to hook the `onDirectoryChanged` event on `Workspace`? In that case, we'd probably want to remove the existing listeners from the old directory, and start listening to the 'new' directory.

Issue #1252, opened 2018-01-08
[UI] Bold fonts slightly cut off


rococode:
![](https://i.imgur.com/hIKpdGG.png)

It seems that bold characters get slightly cut off: in the image above you can see the `m` at the end of `from` looks like it's missing a lil bit at the end.

![](https://i.imgur.com/fDcuW8E.png)

This is the same font and color styl in an IDE.

bryphe:
Thanks for logging the issue, @rococode ! Looks like it is potentially related to #1083 and #878

Akin909:
@bryphe would the solutions you suggested in #1203 re re rendering the whole canvas or using the `drawImage` method to avoid subpixel rendering wonkiness be relevant to this, personally have 0 clue how to use `drawImage` to render text instead of `fillText` but i'd say this is one of my biggest pain points atm so very motivated to find a solution, as the clipping I believe also impacts nerd fonts and wide unicode chars which  I have littered all over the place in my vim config.

bryphe:
This is a tricky one, @Akin909 ! Unfortunately I don't know what the best solution would be here. But some different possibilities:
- Using `drawImage` instead of `fillRect` might help, if the anti-aliasing behavior is different. We could create a 1-pixel image for the color we want (in a separate canvas), and then use `drawImage` instead of `fillRect` to clear the color. Worth prototyping to see - and it'd be worth trying with both `ctx.imageSmoothEnabled` set to `true` and `false`.
- We could look at a smarter algorithm for selecting when to use `Math.floor` and `Math.ceiling` when rounding the rectangles we use for `fillRect` to paint the background.
- In the case where we clip text, we could re-render the preceding token.

The full redraw would be a fix for sure, but it would also be a huge blow to perf. An alternative option would be to re-render a bigger chunk when the background color changes - if the background color changes, we could dirty all the adjacent tokens. This would still be expensive, but hopefully less expensive.

One other, more exotic possibility I was thinking about - is we skip the canvas for rendering the _alternate_ backgrounds. We use it for the 'common case' of the default background, but for elements with a non-default background (which is I believe should be less common), we render that via DOM elements overlaid. Then, we would never have to use `fillRect` for whitespace, and we could clear text by drawing over it with `fillText` (or `drawImage`). We'd have to essentially diff the previous state and the new state, though.

I wish I knew the best fix for it... but there is definitely more investigation needed. Hopefully that helps give you some ideas to try, though!

Akin909:
[...]

Issue #2047, opened 2018-04-06
Text rendering blurry not retina.


dw9694:
Faced such a problem.
I have no retina, my display is 1920x1080, the operating system is Linux Ubuntu 16.04.
The screen blurs in the center, but at the top and bottom is normal.
Thank you.

![2018-04-06 10-25-26](https://user-images.githubusercontent.com/6016334/38408748-86749b78-3987-11e8-97fc-dc71a4ccacf5.png)

MarcusE1W:
Hi.

I think I have the same problem.
I have a 
- notebook with 1366x767 resolution.
- Archlinux
- SpectrWM
- Oni 0.3.1

![2018-04-06-125712_1366x768_scrot](https://user-images.githubusercontent.com/26278354/38420029-6a5e96ba-399a-11e8-9556-58ccf9ac31a9.png)

My impression is that the left Eplorer part is ok, only the text window is mushy.

What log would you need ?

jordwalke:
What could possibly be causing this?

ArthurZ:
Hi @CrossR it is related to https://github.com/onivim/oni/issues/1252 and I was told that it is due to issues around the subpixel anti-aliasing @jordwalke

MarcusE1W:
My impression is that `"ui.fontSmoothing": "subpixel-antialiased"` improves the font sharpness. Thanks.

dw9694:
I do not see the difference

![xojo](https://user-images.githubusercontent.com/6016334/38819277-a023fd74-41a3-11e8-894b-66fffed23244.png)

bryphe:
Thanks for the updates, @MarcusE1W and @dw9694 !

Kind of makes sense, as looking at the screenshots, @MarcusE1W 's screenshot definitely did not have subpixel anti-aliasing enabled. Unfortunately, @dw9694 , yours _does_ have subpixel anti-aliasing, so it's a different issue.

@jordwalke had some helpful pointers - looks like `xterm.js` had some similiar problems (and fixed them). This issue: https://github.com/xtermjs/xterm.js/issues/985 and this PR: https://github.com/xtermjs/xterm.js/issues/985 are especially relevant.

They ensured that their text are aligned with _pixel_ boundaries. This would be a good next-fix for us (#2090 ensures the text is aligned with the 'y' pixel position, but we still need to align with the 'x' pixel position - this is a bit more work because it impacts ligatures and the way we render tokens), if it turns out the WebGL strategy isn't feasible for some reason.

[...]

PR #2120, opened 2018-04-19
Add experimental WebGL Renderer

Cryza:
~This PR is not ready to be merged yet, I will extend the description once it is. Consider this a sneak peek into my work in progress for now~

~For anyone feeling adventurous until then: The renderer should be working roughly now, with bold, italics and underline missing, and no performance optimizations or nice refactorings made so far.~

This PR adds an experimental new renderer as an alternative to the current canvas-based one. I am doing this with the goal of solving several issues that we currently have with the optimizations done in the canvas renderer. First and foremost #1252 and #1083 should be fixed.

This PR intentionally does not cover all the functionality that we have in the canvas renderer yet, I am planning to add all the missing pieces with several smaller PRs afterwards. See the comments below for details in case you want to jump on the train!


bryphe:
@Cryza , this is awesome! I just had a go with it and it's really great work.

I didn't realize how far along this was based on your description... When I ran it, I thought it would be obvious that it was different, but it looked basically the same on my machine (with no cut-off characters, though!) I had to open the debugger to make sure I was actually running the new code. 😄 

Not only that, the performance was really impressive in my quick testing... I did some simple benchmarks of just navigating to the top/bottom of a file (triggering a full redraw). 

For comparison, this is the old timing with the `CanvasRenderer`:
image

And this was the new timing with the `WebGLRenderer`:
image

Going from _22 seconds_ to _2 seconds_ for that benchmark is an __incredible__ performance improvement. 💯 

Thank you for sharing it out! Hope you don't mind if I post a couple questions, just so I can learn more how it works. 😄

browser/src/Renderer/WebGL/WebGLAtlas.ts
```diff
@@ -0,0 +1,141 @@
...
+        gl.texImage2D(
+            gl.TEXTURE_2D,
+            0,
+            gl.RGBA,
+            this.textureSize,
+            this.textureSize,
+            0,
+            gl.RGBA,
+            gl.UNSIGNED_BYTE,
+            this.glyphCanvas,
```
bryphe: This is cool - if I understand correctly, this is how we can take our 'Canvas' that we render characters / glyphs to, and supply it as a texture that we can read from the shaders?

browser/src/Renderer/WebGL/WebGLRenderer.ts
```diff
@@ -0,0 +1,535 @@
...
+        this._gl.bufferData(this._gl.ARRAY_BUFFER, this.glyphInstances, this._gl.STREAM_DRAW)
+        this._gl.blendFuncSeparate(
+            this._gl.ZERO,
+            this._gl.ONE_MINUS_SRC_COLOR,
+            this._gl.ZERO,
+            this._gl.ONE,
+        )
+        this._gl.drawElementsInstanced(this._gl.TRIANGLES, 6, this._gl.UNSIGNED_BYTE, 0, glyphCount)
+
+        this._gl.useProgram(this.textBlendPass2Program)
```
bryphe: Was curious - are the two passes needed for subpixel anti-aliasing? Or for handling transparency / background?

browser/src/Renderer/WebGL/WebGLAtlas.ts
Cryza: That's exactly what we're doing here :) This way we get the nice font rendering from Chrome while at the same time being able to run our editor through WebGL.

The only trade-off is that we need to manually manage our subpixel offsets: We need to manually manage the kerning.

Without additional effort, there is only one version of a character in the atlas, typically aligned to the full pixel boundary. With non-integer character widths however, this would probably ruin the kerning in many cases. A character might at some point need to start at 0.7 pixels offset within the pixel grid. To approximate the correct position, we have different variants of each rendered character in the atlas, each offset by a different fraction of a full pixel. Then we use the one that is closest to what we need at the given position in the editor panel. The number of different variants is controlled by `subpixelDivisor` in this file.

This idea was completely stolen from xray, BTW! All credit goes to @nathansobo :)

browser/src/Renderer/WebGL/WebGLRenderer.ts
Cryza: The two-pass approach is required to do correct blending of subpixel-antialiased text onto our destination canvas. Basically we are abusing the RGB output channels of the first pass to do some intermediate calculations and blending with the destination for us which we then use to calculate the final result in the second pass.

 I will try to find the article again that explained the equations in a nice way and post it here!

Cryza:
@bryphe thanks for the great feedback! I'm really happy that you like it :)

So I am planning to do a major refactoring of all the low-level WebGL code so that it is more understandable and encapsulated in a better way. I hope that the code will be more maintainable afterwards also for people who didn't do a deep dive into WebGL.

I also still see a few quirks and performance improvements we might want to do and possibly also an adaptation of the `INeovimRenderer`, but I think that most of them should come in a separate PR afterwards, just like the rendering of bold, italic and underlined characters (which might also not be too trivial as it requires enhancements to the atlas mapping).

What I would like to finish before we merge this PR:
- Refactor WebGL code
- Hide the WebGL renderer behind an experimental configuration flag
- Fix vertical alignment of the text with the cursor (currently off by ~1px)
- Include line padding (currently ignored, but still yields the same line height AFAIK? :thinking: )
- Maybe even fix the vertical antialiasing issue from #2047 and bring the fix back to the `CanvasRenderer`

bryphe:
> So I am planning to do a major refactoring of all the low-level WebGL code so that it is more understandable and encapsulated in a better way. I hope that the code will be more maintainable afterwards also for people who didn't do a deep dive into WebGL.

That sounds great! I like the idea of refactoring to the high-level constructs (will help me out too, I spend some time reading through WebGL 2 docs, but still not proficient yet) 

> I also still see a few quirks and performance improvements we might want to do and possibly also an adaptation of the INeovimRenderer, but I think that most of them should come in a separate PR afterwards, just like the rendering of bold, italic and underlined characters (which might also not be too trivial as it requires enhancements to the atlas mapping).

Your plan sounds perfect to me. Adding the experimental configuration flag sounds great - perhaps we could have something like `editor.renderer` with `'canvas' | 'webgl'` as values. For now, `'canvas'` will still be the default, but users can experiment with `'webgl'` - and then eventaully once the remaining work comes in, we'll flip the default. Need to get some champagne ready for that.... that'll be a big day 😀 🎉 

> Include line padding (currently ignored, but still yields the same line height AFAIK? 🤔 )

Interesting - we actually do something kind of crazy and incorporate the line padding into the `fontHeightInPixels` here in `NeovimInstance.ts`: 
```
    public setFont(fontFamily: string, fontSize: string, linePadding: number): void {
        this._fontFamily = fontFamily
        this._fontSize = fontSize

        const { width, height, isBoldAvailable, isItalicAvailable } = measureFont(
            this._fontFamily,
            this._fontSize,
        )

        this._fontWidthInPixels = width
        this._fontHeightInPixels = height + linePadding
```

This is confusing because the `fontHeightInPixels` that is everywhere isn't actually the measured pixel height of the font... its the measured pixel height of the font _plus_ the line padding. This could potentially be causing problems.

Of course, keep me posted if there is anything I can do 👍 Awesome work @Cryza !

nathansobo:
@Cryza thanks for the shout-out. I’d definitely appreciate a ping on PRs related to this code path so I can learn from your progress. Good luck.

package.json
```diff
@@ -750,6 +752,7 @@
         "@types/sinon": "1.16.32",
         "@types/webgl2": "^0.0.3",
         "autoprefixer": "6.4.0",
+        "aws-sdk": "^2.202.0",
```
DeltaEvo: Why is this dependency included ?

DeltaEvo: Aw seems a merge from master, sorry :joy:

bryphe:
Thanks for your help @nathansobo - look forward to seeing where `xray` goes! 👍 It's been exciting the progress.

browser/src/Renderer/WebGL/WebGLRenderer.ts
```diff
@@ -0,0 +1,157 @@
...
+        foregroundColor,
+        backgroundColor,
+    }: IScreen) {
+        const canvasWidth = this._gl.canvas.width
+        const canvasHeight = this._gl.canvas.height
+        const viewportScaleX = 2 / canvasWidth
+        const viewportScaleY = -2 / canvasHeight
+        this._gl.viewport(0, 0, canvasWidth, canvasHeight)
+
+        this._solidRenderer.draw(
```
bryphe: Thanks for refactoring this! Definitely helped make it easier to understand 👍

Cryza: I'm happy that it helped! I think I'll try to make it even more concise but that depends on how much time I will find for doing that. :)

browser/src/Renderer/WebGL/WebGLSolidRenderer.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    private createBuffers() {
+        this._unitQuadVerticesBuffer = createUnitQuadVerticesBuffer(this._gl)
+        this._unitQuadElementIndicesBuffer = createUnitQuadElementIndicesBuffer(this._gl)
+        this._solidInstances = new Float32Array(maxCellInstances * solidInstanceFieldCount)
+        this._solidInstancesBuffer = this._gl.createBuffer()
+    }
+
+    private createVertexArrayObject() {
+        this._vertexArrayObject = this._gl.createVertexArray()
```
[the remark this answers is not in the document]
Cryza: I havent't seen that library before, and it looks really interesting! I actually really dislike the mutative and OO way of programming that the GL APIs enforce, so this hits the spot for me :)

Unfortunately though it seems like there is no WebGL2-compatible version of this library and I currently don't know if our shaders could be rewritten in a WebGL1-compatible way. I had a quick try and changed the typings of the contexts to WebGL1, which resulted in lots of errors. So my gut feeling for this is that it's not really possible, but one would need to do some more investigation into this to be sure.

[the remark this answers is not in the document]
nathansobo: WebGL 1 doesn't support instanced rendering as far as we could tell, so it will be a non-trivial change and slightly less efficient to use.

parkerault:
If there is an experimental flag for this, I would suggest merging it into master as soon as the refactoring is complete. I am very eager to try this out!

browser/src/Renderer/WebGL/WebGLSolidRenderer.ts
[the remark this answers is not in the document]
bryphe: Thanks @Cryza  and @nathansobo ! That's a bummer that regl doesn't support WebGL 2 (I like the functional approach better too) - but definitely a non-starter if we'd have to switch to WebGL 1 👍 Good to know! Definitely saved us time. 😄

adelarsq:
I don't know if its related but after this change shows this messages when using webgl:

```
Received notification for - Unhandled Exception:Error: Texture is too small
Please report this error. Callstack: Error: Texture is too small
    at Wi._rasterizeGlyph (file:///C:/dev/oni/lib/browser/vendor.bundle.js:964:15106)
    at Wi.getGlyph (file:///C:/dev/oni/lib/browser/vendor.bundle.js:964:14430)
    at Xi.populateGlyphInstances (file:///C:/dev/oni/lib/browser/vendor.bundle.js:964:19764)
    at Xi.draw (file:///C:/dev/oni/lib/browser/vendor.bundle.js:964:17781)
    at Ji._draw (file:///C:/dev/oni/lib/browser/vendor.bundle.js:964:22957)
    at Ji.redrawAll (file:///C:/dev/oni/lib/browser/vendor.bundle.js:964:21619)
    at br._onResize (file:///C:/dev/oni/lib/browser/vendor.bundle.js:791:3136)
    at ResizeObserver._boundOnResizeMethod._resizeObserver.window.ResizeObserver.e (file:///C:/dev/oni/lib/browser/vendor.bundle.js:791:2653)
Uncaught Error: Texture is too small
    at Wi._rasterizeGlyph (vendor.bundle.js:964)
    at Wi.getGlyph (vendor.bundle.js:964)
    at Xi.populateGlyphInstances (vendor.bundle.js:964)
    at Xi.draw (vendor.bundle.js:964)
    at Ji._draw (vendor.bundle.js:964)
    at Ji.redrawAll (vendor.bundle.js:964)
    at br._onResize (vendor.bundle.js:791)
    at ResizeObserver._boundOnResizeMethod._resizeObserver.window.ResizeObserver.e (vendor.bundle.js:791)
```

Cryza:
@adelarsq thanks for the report! This is caused by the webgl renderer as you assumed. It currently happens if the texture that contains all the prepared characters runs out of space. Currently we hard-coded the texture size to 512x512 pixels, so for files with lots of different characters, this limit will at some point cause the renderer to throw. Especially so for large font sizes.

My current idea is to implement some logic that catches the error at the level of the text renderer, creates a larger texture and restarts rendering.

To find a reasonable default size in any case, it would be great if you could share some information about your font size, font family, line padding, if you have a retina display (or some other setup with devicePixelRatio > 1) and ideally also the content of the file, if that is possible.

Cheers!

CrossR:
I also hit a similar issue when doing lots of scrolling very rapidly.

I'm using Fira Code as my font at 12px, with the default line padding and a 4k monitor (27" if that matters?) at 150% scaling, with Windows 10 64bit.

The file was my Oni config, which can be found here: https://github.com/CrossR/dotfiles/blob/master/oni/config.tsx

justinmk:
oni 0.3.4, macOS 10.13.4

Setting `"editor.renderer": "webgl"` gives me this error immediately on startup:

```
bundle.js:1 Received notification for - Unhandled Exception:TypeError: Cannot read property 'canvas' of null
Please report this error. Callstack: TypeError: Cannot read property 'canvas' of null
    at Ji._updateCanvasDimensions (file:///Users/justin/opt/Oni.app/Contents/Resources/app/lib/browser/vendor.bundle.js:964:21733)
    at Ji.redrawAll (file:///Users/justin/opt/Oni.app/Contents/Resources/app/lib/browser/vendor.bundle.js:964:21520)
    at eo._render (file:///Users/justin/opt/Oni.app/Contents/Resources/app/lib/browser/vendor.bundle.js:964:44256)
    at _pendingAnimationFrame.window.requestAnimationFrame (file:///Users/justin/opt/Oni.app/Contents/Resources/app/lib/browser/vendor.bundle.js:964:44126)
a @ bundle.js:1
vendor.bundle.js:964 Uncaught TypeError: Cannot read property 'canvas' of null
    at Ji._updateCanvasDimensions (vendor.bundle.js:964)
    at Ji.redrawAll (vendor.bundle.js:964)
    at eo._render (vendor.bundle.js:964)
    at _pendingAnimationFrame.window.requestAnimationFrame (vendor.bundle.js:964)
```

adelarsq:
@bryphe the error above occurs after the change https://github.com/onivim/oni/pull/2183 I think

adelarsq:
@justinmk Was it after the change https://github.com/onivim/oni/pull/2183?

Cryza:
@justinmk thanks for the report! That happens due to the fact that sometimes the drawing call can be triggered before the renderer was initialized. I will try to fix that in my current branch where I am improving the stability of the WebGL renderer. This is the branch: https://github.com/onivim/oni/tree/cryza/bugfix/webgl-renderer-out-of-texture-space

Cryza:
Great to hear, @badosu! Thanks for giving it another try this quickly :)

PR #2121, opened 2018-04-20
Feature/ add delete event so watcher picks up all deletions

Akin909:
This PR adds the `unlink` event to the watcher and a hook for it as it turns out that the change event does not always accurately pick up deletions, the correct event for that seems to be unlink.

Also added a check to only watch workspace changes if the watcher was initialised to watch the workspace, this will allow the watcher to be used for other purposes without always spawning a watcher for the workspace.

Lastly added a check to see if project root is different before triggering a change dir event as it seems the explorer always refreshes and flickers when moving buffer if `navigateToProjectRoot` is set to always as it always triggers a change dir but a check to see if the dir is the same stops that

### DIFF
--- a/browser/src/Services/FileSystemWatcher/index.ts
+++ b/browser/src/Services/FileSystemWatcher/index.ts
@@ -2,8 +2,6 @@ import * as chokidar from "chokidar"
 import { Stats } from "fs"
 import { Event, IEvent } from "oni-types"
 
-import * as Workspace from "./../Workspace"
-
 export type Targets = string | string[]
 
 interface IFSOptions {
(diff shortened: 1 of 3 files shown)


browser/src/Services/FileSystemWatcher/index.ts
```diff
@@ -42,11 +43,13 @@ export class FileSystemWatcher {
         // to avoid a flurry of events when the watcher is initialised
         this._watcher.on("ready", () => {
             this._attachEventListeners()
-        })
 
-        this._workspace.onDirectoryChanged.subscribe(newDirectory => {
-            this.unwatch(this._activeWorkspace)
-            this.watch(newDirectory)
+            if (fileOrFolder === this._activeWorkspace) {
```
bryphe: This actually might be a _bit_ fragile, if we decide to create a `FileSystemWatcher` that is not coupled to the workspace, but happened to be watching the same folder - it might not be obvious that it's going to switch directories underneath if the workspace changes.

We could consider 'inverting' this dependency - meaning, instead of the `FileSystemWatcher` being dependent on the Workspace, we switch gears - and make the Workspace (or explorer) own an instance of `FileSystemWatcher`. Then, in either Workspace / Explorer, we listen to the event there - and call `watch`/`unwatch` from there - that way the `FileSystemWatcher` isn't ['coupled' ](https://en.wikipedia.org/wiki/Coupling_(computer_programming)). 

That makes the `FileSystemWatcher` simpler, since now it doesn't have to care about the workspace at all!

bryphe:
> This PR adds the unlink event to the watcher and a hook for it as it turns out that the change event does not always accurately pick up deletions, the correct event for that seems to be unlink.

Awesome! Thanks for adding this @Akin909 👍 

> Also added a check to only watch workspace changes if the watcher was initialised to watch the workspace, this will allow the watcher to be used for other purposes without always spawning a watcher for the workspace.

I had one concern here - there's an 'implicit' assumption that if a standalone watcher is initialized to the _same directory_ as the workspace, it should also follow workspace changes. That might be a bit risky- it might not be obvious to someone using the `FileSystemWatcher` in a standalone capacity that it has this behavior, so it could be a potential source of bugs. I left a more detailed comment but we could consider factoring that responsibility out into `Explorer` (and have it own the `FileSystemWatcher`) - that way `FileSystemWatcher` doesn't have to care about workspaces at all, just about watching files.

> Lastly added a check to see if project root is different before triggering a change dir event as it seems the explorer always refreshes and flickers when moving buffer if navigateToProjectRoot is set to always as it always triggers a change dir but a check to see if the dir is the same stops that

Nice catch! 👍

Akin909:
@bryphe I've moved references to the workspace out of the watcher and instead instantiate and coordinate the watcher activities in the explorer split so they are decoupled

bryphe:
> @bryphe I've moved references to the workspace out of the watcher and instead instantiate and coordinate the watcher activities in the explorer split so they are decoupled.

Cool, the refactoring looks great! 💯 

> I initially imagined that the majority of the watcher's use cases would relate to the active workspace which is why I was defaulting the instance to watch it but think it began to slip into tighter and tighter coupling when all I intended originally was that if passed no options it would watch the active workspace (so we could default export a version that would apply to most use cases), but guess if that's what someone wants to use it for its best to just specify rather than risk the separation of concerns

Ya, totally, I think that's how coupling tends to find it's way into systems 😄 It's like one of those things that just happens incrementally unless it's actively worked against. I like the way you tweaked it, though, since the `FileSystemWatcher` is very scoped to its particular task of watching a folder - it gives me a lot of confidence that we could use it in different areas of the code. And then the `ExplorerSplit` tailors it to the particular use case in the explorer of managing the active workspace 👍 

Thanks for making that change!

PR #2122, opened 2018-04-20
bugfix/ add check to ensure there is an error event obj and it has a name

Akin909:
Some errors that are save to the `queuedErrors` do not have names and so if an error was thrown without one this check would occasionally itself throw an error, so added a check to ensure that the props being check exist before the actual check.


bryphe:
Thanks for catching this @Akin909 ! Errors in the error handler is no good 😄 

Change looks great!

PR #2124, opened 2018-04-20
Hotfix/fix fs event overload

Akin909:
@bryphe @CrossR important bugfix here, noted since I merged in the FileWatcher that if you switch project roots the filesystem watcher would send an event for all the subdirectories and files of the newly watched directory resulting in 100s of events which bombard oni.

I've added a throttle off `300ms` to the events being fired to stop the bombardment

Discussion Points: Wasn't sure if it would be preferable to throttle the refresh dispatch, and have the watcher respond with all the events it receives?

### DIFF
--- a/browser/src/Services/Explorer/ExplorerStore.ts
+++ b/browser/src/Services/Explorer/ExplorerStore.ts
@@ -633,13 +633,17 @@ export const clearUpdateEpic: ExplorerEpic = (action$, store) =>
         .mergeMap(() => timer(2_000).mapTo(Actions.clearUpdate))
 
 const refreshEpic: ExplorerEpic = (action$, store) =>
-    action$.ofType("REFRESH").mergeMap(() => {
-        const state = store.getState()
+    action$
+        .ofType("REFRESH")
+        .auditTime(300)
+        .mergeMap(() => {
+            console.log("REFRESHING")
+            const state = store.getState()
 
-        return Object.keys(state.expandedFolders).map(p => {
-            return Actions.expandDirectory(p)
+            return Object.keys(state.expandedFolders).map(p => {
+                return Actions.expandDirectory(p)
+            })
         })
-    })
 
 const expandDirectoryEpic: ExplorerEpic = (action$, store, { fileSystem }) =>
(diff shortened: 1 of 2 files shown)


bryphe:
> @bryphe @CrossR important bugfix here, noted since I merged in the FileWatcher that if you switch project roots the filesystem watcher would send an event for all the subdirectories and files of the newly watched directory resulting in 100s of events which bombard oni.

Ah good catch... that would be brutal!

> Discussion Points: Wasn't sure if it would be preferable to throttle the refresh dispatch, and have the watcher respond with all the events it receives?

Yes, perhaps it makes sense actually to move this logic to our `refreshEpic` in `ExplorerStore` - I can't think of a scenario where we'd want to refresh that aggressively.

We could use [`auditTime`](https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/auditTime.ts)  on our epic there, like:
```
    action$.ofType("REFRESH")
   .auditTime(300) 
   .mergeMap(() => {
        const state = store.getState()

        return Object.keys(state.expandedFolders).map(p => {
            return Actions.expandDirectory(p)
        })
    })
```

I think this approach may actually be preferable, because based on the interface of the `FileSystemWatcher` - if we had several 'add' or 'move' events happen in succession, I believe we'd actually drop some of those with the throttle (we'd only get the first/last depending on what the leading/trailing defaults are). So if something else depending on the `FileSystemWatcher`, it might lead to confusing or unintuitive behavior - it wouldn't be clear why sometimes in high-frequency changes we'd drop events.

Akin909:
@bryphe changed this PR to use `auditTime` as you suggested and this seems to work quite well

Akin909:
@bryphe just wanted to mention somewhere since I know you plan on releasing soon that I think this PR might be required before release or alternatively reverting the `FSWatcher` as the bombardment really negatively impacts user experience when loading a project folder (btw I've been running this PR locally and it resolves the issue v. well).

bryphe:
Awesome, thanks @Akin909 ! Change looks great.  Appreciate you calling out that we should include, I'll make sure it's included in 0.3.3 👍

PR #2126, opened 2018-04-20
Add back cache to VSCode scorer.

CrossR:
After adding back the cache, the sorting is certainly quicker but still not at VSCode level.

Need to have a look and see if it makes sense to keep the cache around longer, and also see if we can have the scoring not block the typing which I see occasionally.

I'm testing this on the VSCode git repo, which seems a pretty large repo and seems to show the slow down.

One note though if someone else tries to compare vs VSCode is that they have a pretty large exclude list, so you'd want to stick that in your project specific `.oni/config` to make the comparison a bit fairer. Or disable it in VSCode. It can be found in `.vscode/settings.json`.

### DIFF
--- a/browser/src/Services/QuickOpen/VSCodeFilter.ts
+++ b/browser/src/Services/QuickOpen/VSCodeFilter.ts
@@ -57,7 +57,7 @@ export const vsCodeFilter = (
     // TODO: Is it worth instead persisting this cache?
     // Plus side is repeated searches are fast.
     // Down side is there will be a lot of rubbish being stored too.
-    let cache: ScorerCache = {}
+    const cache: ScorerCache = {}
 
     const filteredOptions = processSearchTerm(vsCodeSearchString, options, cache)
 
--- a/browser/src/Services/QuickOpen/VSCodeFilter.ts
+++ b/browser/src/Services/QuickOpen/VSCodeFilter.ts
@@ -15,6 +15,7 @@ import {
 } from "./Scorer/OniQuickOpenScorer"
 
 import { IMenuOptionWithHighlights, shouldFilterbeCaseSensitive } from "./../Menu"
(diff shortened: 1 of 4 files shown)


bryphe:
Awesome, thanks for bringing this back in @CrossR ! 💯

browser/src/Services/QuickOpen/VSCodeFilter.ts
```diff
@@ -49,7 +50,16 @@ export const vsCodeFilter = (
             ? listOfSearchTerms.slice(1).join("") + listOfSearchTerms[0]
             : listOfSearchTerms[0]
 
-    const filteredOptions = processSearchTerm(vsCodeSearchString, options)
+    // Adds a cache for the scores. This is needed to stop the final score
+    // compare from repeating all the scoring logic again.
+    // Currently, this only persists for the current search, which will speed
+    // up that search only.
+    // TODO: Is it worth instead persisting this cache?
```
bryphe: Good question - do you mean persisting this cache on the filesystem, so that it's available on startup?  Or persisting in memory for the duration of the session? I think the latter may make sense... the former might be overkill for now (but maybe something we could do later)

bryphe: Or we could stick with the current strategy for now - it's at least much faster since we don't repeat the scoring logic for final scoring, and hook up the persistence later (if needed)

CrossR: I was meaning the later, persisting for the session.
It would make subsequent searches pretty much instant, The only concern I had is that every single search is persisted so it could end up getting quite big.

That is for `QuickOpen` the results are stored for `Q`, `Qu`, `Qui`...and so on. I think VSCode does persist it for the whole session though.

Issue #2129, opened 2018-04-21
Feature request: Create a new file from Ctrl+P menu


Kxze:
The Ctrlp extension allows you to type the path of a file and press Ctrl+Y to create it if it doesn't exist. This is a super handy feature, as it lets the user create new files without taking their hands off the keyboard.

tarciozemel:
Check #1839 which was merged some days ago. It's more "vim style".

tarciozemel:
Check #1839 which was merged some days ago. It's more "vim style".

Akin909:
@Kxze I'm currently working on increased explorer functionality as @tarciozemel pointed out with #1839 and also in #2115 the next PR i'll be adding will be using probably `e` binding to add a new file in the explorer, the idea being you move to the selected folder where you want the file using `h,j,k,l` then hit `e` to get an input where you'd type in the file name to create a new file (and have this be undoable with `u`)

badosu:
@Akin909 I have a suggestion, when cursor over a file in a particular folder `o`, opens the dialog to create a new file as well, has the same semantic idea of a new line on vim.

Akin909:
@badosu #2130 adds the ability to press `<c-e>` (`e` seemed too easy to hit by accident, but can easily be rebound) over a file or folder and if a folder is selected the the new file or folder is created within it so I think if i'm understanding it right it should already do what you suggesting? (although the binding is different)

keforbes:
Is there anything pending with this issue?  A lot of ideas were thrown around but all the referenced pull requests have been merged in so this issue feels "done" to me.

keforbes:
I'm just going to close this issue and see if anyone complains. :stuck_out_tongue:

dkns:
What was merged and referenced was sidebar explorer code. If I understand OP correctly he wants to be able to press `<Ctrl+Shift+P>`, type something like "New file" and be able to create new file that way. I think it's quite different from sidebar explorer functionality.

badosu:
Yes, the question is whether this functionality is necessary given the
explorer functionality.

Em Sex, 25 de mai de 2018 16:36, Daniel Kosiński <notifications@github.com>
escreveu:

keforbes:
For what it's worth, @Akin909 described his alternative solution using the explorer and the OP responded with `hooray` emoji so I assumed it was an acceptable alternative.  I guess there's just a question of whether we want to implement a feature as originally described or if we just want to make an OP happy. :slightly_smiling_face:

[...]

PR #2130, opened 2018-04-21
Feature/add new file in explorer

Akin909:
This PR adds the ability to create a new file or folder in the explorer, these are bound to `<c-e>` and `<c-f>` by default both actions are also undoable by pressing `u`, duplicates result in an error notification. Relates #1839 based off of #2115 (should be merged after 2115)

new

### Todo:
- [x] Add Tests


tarciozemel:
@Kxze #2129

Kxze:
This is awesome! Thanks for implementing my request :)

bryphe:
Very cool, @Akin909 ! Can't wait to start using this in my workflow! You've done an amazing job making the explorer useful 💯 

I noticed one minor quirk that was a bit confusing for me at first:
add file

When I'm highlighted on an item, and I press `<C-e>` or `<C-f>`, it shows the insertion UX _before_ the item - that made me think it might put the file in the wrong place. In the example above, I pressed `<c-f>` on `test-rename2`.  

I would've expected the insertion UX is shown _after_ the item - that way, it feels like it will be inserted into `test-rename2`.

It might be even more apparent on the root (which is super cool that it works and I can insert a file/folder there!): 
image

But I would expect that it would show the 'insert' UX _after_ the root, instead of before.

Akin909:
@bryphe I've tweaked the insertion UX to show after the selected item (it was a completely arbitrary decision on my part in the first place and easily fixed 😄)

bryphe:
Sweet, looks great @Akin909 ! 😀 Glad it was a straightforward change, the behavior is much more intuitive to me 👍

bryphe:
@Jimeno0 , you're in luck! These actually made it into our 0.3.4 release: https://github.com/onivim/oni/releases/tag/v0.3.4 😉 Would ❤️ feedback.

Of course, you can always build from the latest `master` to pick up the newest stuff - instructions here: https://github.com/onivim/oni/wiki/Development

We also have the latest `master` binaries available for download (for backers) here: https://www.onivim.io/Insiders

Jimeno0:
Wohaaaa I thought that I was using the 0.3.4 but it turns I wasnt! 😃

(If you go to the [installation guide](https://github.com/onivim/oni/wiki/Installation-Guide#mac) and then search for the download Oni for mac it downloads the 0.3.2)

Im currently using Oni at work so I will try the latest master version and give you all my feedback :)

About the explorer commands I can just say... this is f**ing amazing!😍

Some little thing that disrupts me are that if you delete a file the cursor goes to the top of the explorer and when I try to delete several files I ended deleting files that I didnt want to (its just a mather of time to get use to it but I found it interensting to let you know)

Also whit the new feaure I no longer can use <C-d> to go down for a few lines  (as I use to do in vim), it deletes the cursor file.

Those are just minor stuff, I hope that feedback helps. And again these new features are insane guys! Thanks for such an amazing editor 😍

PR #2143, opened 2018-04-26
#4332 - Fix bug with 'onScrolled' event not firing in all cases

bryphe:
__Issue:__ The `onScroll` event we expose from `NeovimInstance` (and which `NeovimEditor`/`OniEditor` depend on to power the `onBufferScrolled` event for the `Editor` interface) doesn't work reliably - it only dispatches when the cursor moves. However, there are lots of scroll cases where the cursor doesn't move, like:
- Scrolling when the cursor is in the center of the screen
- Using `zz` and scrolling by a single line (`<C-e>`)

It's easy to repro by adding an event listener in the dev tools:
```
Oni.editors.activeEditor.onBufferScrolled.subscribe((evt) => console.log("scroll: " + JSON.stringify(evt)))
```

For those cases above, there will be no output.

__Defect:__ Neovim doesn't expose a `Scroll` autocommand, so we rely on a few things to power that event. The cursor moved events trigger this today, but there is also a redraw event that lets us know when there is an incremental scroll - we weren't using that at the moment.

__Fix:__ Hook up the `onScroll` event to use the scroll redraw message to work reliably in cases where the cursor doesn't move.

Remaining work:
- [x] Create a CiTest to exercise this, so we can trust our `onBufferScrolled` handler build-over-build 😄


browser/src/neovim/NeovimInstance.ts
```diff
@@ -760,6 +765,7 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
                     break
                 case "scroll":
                     this.emit("action", Actions.scroll(a[0][0]))
+                    this._dispatchScrollEvent()
```
Akin909: @bryphe ah this was the line I was referring to which I though represented a scroll event from neovim but that was the what I was logging out which was providing a more accurate response to scrolling didnt realise it was from the redraw event

browser/src/neovim/NeovimInstance.ts
```diff
@@ -701,6 +690,22 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
...
+    private _dispatchScrollEvent(): void {
+        if (this._pendingScrollTimeout || this._isDisposed) {
+            return
+        }
+
+        this._pendingScrollTimeout = window.setTimeout(async () => {
+            if (this._isDisposed) {
+                return
+            }
+
```
Akin909: I was also logging here so I thought this bit being put into the the event loop rather than running synchronously was where things were going missing since if by some quirk of fate if the next scroll event came straight after and the pendingScrollTimeout still existed the function would just return so it might in effect be swallowing scroll events, didn't actually verify that this was happening

browser/src/neovim/NeovimInstance.ts
bryphe: Haha ya, I think the "bug" is that we weren't using it for our `onScroll` event 😄 So we were missing out on this info, which is the most reliable way to check for scroll events when the cursor doesn't move (it misses large scroll events, like `G` on a long file)

browser/src/neovim/NeovimInstance.ts
Akin909: 👍 Makes sense thanks for explaining

PR #2154, opened 2018-04-30
Fix #2149 - Part 1 - Add configuration setting for explorer refresh

### LINKED ISSUE #2149: FileSystemWatcher should ignore folders specified under 'oni.exclude'
In the last version of Oni (v0.3.3-1) it was introduced a file system watcher (#2116 with further fixes in #2121). The file system watcher should exclude folders that is specified under `oni.exclude` configuration, or even added  new configuration.

For projects that generate `tmp` folders with many, many files under it, Oni becomes very unresponsive when opening such projects. After a certain amount of time with Oni open (~2-4 min), it will start to respond as expected.

bryphe:
__Issue:__ There can be a performance degradation when upgrading from Oni 0.3.2 -> Oni 0.3.3 due to the way we watch the file system / update the UI in response to it.

__Fix:__ We'll gate the 'autoRefresh' behavior behind a flag - `explorer.autoRefresh`, which defaults to `false`. We'll enable an explicit refresh gesture with the `explorer.refresh` command, bound to `<C-r>`.

### DIFF
--- a/browser/src/App.ts
+++ b/browser/src/App.ts
@@ -262,7 +262,13 @@ export const start = async (args: string[]): Promise<void> => {
     Sidebar.activate(configuration, workspace)
     const sidebarManager = Sidebar.getInstance()
 
-    Explorer.activate(commandManager, editorManager, Sidebar.getInstance(), workspace)
+    Explorer.activate(
+        commandManager,
+        configuration,
+        editorManager,
+        Sidebar.getInstance(),
+        workspace,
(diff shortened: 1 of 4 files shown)


bryphe:
Note that this doesn't completely address #2149 - we should also hook up the `oni.exclude` files configuration to work with the explorer file watching (and investigate other ways to optimize the perf as @Akin909  mentioned).

PR #2162, opened 2018-05-02
Update delete binding to use the `isExplorerActive` filter.

CrossR:
Updated the explorer delete binding to use the `isExplorerActive` filter, otherwise if you have a folder/file selected in it and hit delete (even whilst in insert mode!) you'll delete that folder.

I was getting random files being deleted whilst working and couldn't work out why. Luckily I have everything in git so it was no hassle, but could have been a lot worse. Since my keyboard has a function layer which sticks `delete` right on the home row, I use it a lot more than maybe other vim users who would need to move their hands to hit it.

What I'm less sure about is why I didn't hit this back when the `delete` was originally added, since it was added in Jan I think. Perhaps it was just because I was using the explorer less than I am now.


bryphe:
Definitely, I agree - I just started a pre-release for `0.3.4` which has this and #2154

PR #2167, opened 2018-05-04
add Change Operator and Visual Mode tutorials

keforbes:
As mentioned in #2035, I felt a "visual mode" tutorial should come next.  However, once I started working on it I realized I would want to demo visual mode with all three operators: `y`ank, `d`elete, and `c`hange.  Since there wasn't a Change Operator tutorial yet, I threw one together.  Personally, I think `change` is most useful when used in conjunction with `t`/`f`/`i`/`a`.  Since we haven't written those tutorials yet, I only included `cw` in the Change Operator tutorial.  While I *could* add steps for `ce`, `cb`, `cj`, and `ck`, I don't feel like those operations are very beneficial and would just confuse a user as to why you'd want to do those things.

After looking back at these tutorials, I feel like I should probably repeat some operations to reinforce the learning.  The current state is function though, so adding reinforcement could probably come in a separate pull request.

### DIFF
--- a/browser/src/Services/Learning/Tutorial/Notes.tsx
+++ b/browser/src/Services/Learning/Tutorial/Notes.tsx
@@ -268,6 +268,27 @@ export const DeleteWordKey = (): JSX.Element => {
     )
 }
 
+export const ChangeOperatorKey = (): JSX.Element => {
+    return (
+        <KeyWithDescription
+            keyCharacter="c"
+            description={
+                <span>
+                    <Bold>+ motion</Bold>: Change text specified by a `motion`. Examples:
+                </span>
+            }
+        />
+    )
+}
+export const ChangeWordKey = (): JSX.Element => {
+    return (
+        <KeyWithDescription
+            keyCharacter="cw"
+            description={<span>Delete to the end of the current word and enter Insert mode.</span>}
+        />
+    )
+}
+
 export const HJKLKeys = (): JSX.Element => {
     return (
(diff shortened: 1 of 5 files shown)


bryphe:
Awesome! Thanks for these, @keforbes !

>  Since there wasn't a Change Operator tutorial yet, I threw one together. 

Nice, this was a big gap in our 'curriculum' - IMO the change operator is very important to get the feel for Vim.

>  Personally, I think change is most useful when used in conjunction with t/f/i/a.

Right, it would be great to have a tutorial that covers `ciw`, `cf<char>`, etc. I think for both 'change' and 'delete' it would be helpful to have a second tutorial, like:
- Delete + Motions (`dj`, `d5k`, `dw`, etc)
- Change + Motions (`ciw`, `cw`, `ct(`, `ci"`, `cit`, etc).

We don't really have a good experience now that brings together the 'operators' and 'motions' - so I think having these next-level change/delete tutorials would be good (and they'd also help address the repetition concern).

>  The current state is function though, so adding reinforcement could probably come in a separate pull request.

For sure! I think what you have here is an excellent start - it's great to see the tutorials fill out and cover more of the functionality. Played through both of these and I like them - great to have coverage over Visual mode! 👍 

Oh, one other idea - perhaps we could have a `blockwise` visual mode tutorial at some point for `<c-v>`... I definitely need to practice that... 😄

bryphe:
Change looks great to me! Thanks for including the new notes in the sidebar, too 👍

Issue #2057, opened 2018-04-06
Bug - Developer Experience: `npm link` woes


bryphe:
On our [Development](https://github.com/onivim/oni/wiki/Development#first-build) wiki page, we currently list the following steps:

- Clone the repository: `git clone --recursive https://github.com/onivim/oni.git`
  - `--recursive` will also update submodules (which are required to build Oni)
- `cd oni` to navigate to the root folder
- `yarn install` to install the project dependencies (If `yarn` is not installed, use `npm install -g yarn`).
- `npm run build` to build the project
- `npm link` to register the `oni` command in the shell, against the local build.

This worked fine for a while, _until we switched to `yarn`_. It seems that `yarn` and `npm` don't play well together... And the `npm link` command at the end is unfortunately not reliable when we use `yarn install` prior.

A few things:
- First, we should standardize on just using `yarn` version of commands to avoid confusion - so we should change `npm run build` to `yarn build`
- Second, we should find a replacement for `npm link`. Unfortunately, the `yarn` version doesn't seem to work in parity. __The only reason we need to the `npm link` call is to create a global shortcut for the development build of Oni__ - an alternative here would be to write a shell script for each platform (or in a cross-platform way, if possible), to achieve the same goal. On OSX / Linux, it seems relatively straightforward to create a symlink in `/usr/local/bin`, but I'm not sure the best way on Windows... perhaps we could write a powershell script to add to the path?

TL:DR - A few things to make the development / set-up experience clearer:
- Standardize to use `yarn` versions of commands vs `npm`
- Replace `npm link` with something else that we can use with `yarn` - this might involve writing custom scripts for each platform, worst case.

Akin909:
@texhnolyze re. the wiki pages you can edit those on github once PR is in you don't need to make a PR for the wiki

texhnolyze:
This can be closed now as it was merged in #2177, forgot to reference the issue in my pull request. I also updated the wiki docs for the [dev setup](https://github.com/onivim/oni/wiki/Development) to include yarn as an installation dependency and to run all previous `npm run` and `npm link` commands with yarn instead.

CrossR:
Yep, looks like the first bullet point here was covered in #2177 so I'll close this out.

We've got a much more indepth ticket for a `link` alternative I think over in #1707, so I'll close this out.

Thanks for the contribution @texhnolyze!

PR #2177, opened 2018-05-07
Use yarn to run npm sripts (#2057)

texhnolyze:
because for dev setup previously yarn was used to install
dependencies while npm was used to run build and setup a
symlink to use local oni from shell resulting in issues


texhnolyze:
Are people actively using the ```install-global``` and ```uninstall-global``` npm scripts? If not I would like to remove them as yarn actively deprecates the ```yarn install -g```  and ```yarn remove -g``` methods as it is not recommended to install packages globally. What do you guys think?

Akin909:
@texhnolyze I have never actually run either command and I'm not actually sure where they are being used @bryphe might know if they are needed, I think personally that if they are being deprecated and they aren't actively being used then removing them sounds fine 👍 

PS: thanks for the PR looks good

bryphe:
Excellent, thanks for getting us all-in on `yarn`, @texhnolyze ! Looks good to me too.

> Are people actively using the install-global and uninstall-global npm scripts? If not I would like to remove them as yarn actively deprecates the yarn install -g and yarn remove -g methods as it is not recommended to install packages globally. What do you guys think?

Nope, these are _very_ old scripts that aren't being used anymore - would be great to remove them. Appreciate you taking a pass through these!

texhnolyze:
@Akin909 I tested the npm scripts to make sure there is no unexpected behaviour after the switch to yarn and have not encountered any issues. 
To the second question, ```yarn install``` installs a packages (either by name or ```package.json```), while ```yarn add``` is used to add new depencies to the ```package.json``` and generate a new lock file. It used to be the case that the two would have the same effect when adding dependencies, which is not the case anymore as add replaced install for that.

@bryphe removed those two npm scripts, so this is merge ready from my side. Also will update the wiki/docs once it is merged.

PR #2179, opened 2018-05-07
Feature/add file drag and drop handler

### LINKED ISSUE #2070: Regression: File-explorer drag and drop
Dragging and dropping in the file explorer, doesn't appear to be working anymore.
I think this might be as a result of other dragging and dropping in oni. While working on #1406, I noted that there's a bit of a delicate balance between the dragging files onto oni and the file explorer drag and drop as higher level events can swallow the event transmission.

I think if `stopPropagation()` is being called on `document.body` then this will prevent events being transmitted down to the file explorer.

In the long run I think all drag and drop functionality needs to be specifically scoped to target elements and events on body should be avoided aka `querySelector(.active .editor).addEventListener("drop...)` as opposed to more general handlers

Akin909:
Fixes #2070 this PR adds (as suggested by @bryphe in #2070 a `FileDropHandler` component to the `NeovimSurface` that handles file drop events allowing the explorer file drag and drop to co-exist with the open by dragging onto oni functionality.

It can be re-used else where if there is a need with the caveat that if the target is passed is too deeply nested and its ancestors have pointer-events enabled they might obscure the event transmission.


browser/src/Editor/NeovimEditor/FileDropHandler.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Takes an element (which can accept drag and drop events) and a file drop event listener callback
+ * N.B. the element cannot be obscured as this will prevent event transmission
+ * @name FileDropHandler
+ * @function
+ *
+ * @extends {React}
+ */
+export default class FileDropHandler extends React.Component<IFileDropHandler> {
```
bryphe: Looks like a great start! Thanks for adding this. 💯

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
```diff
@@ -741,25 +741,20 @@ export class NeovimEditor extends Editor implements IEditor {
             this._neovimInstance.command(`:e! ${path}`)
         })
 
-        // TODO: Factor this out to a react component
-        // enable opening a file via drag-drop
-        document.body.ondragover = ev => {
+        // This is necessary to prevent electron's default behaviour on drag and dropping
+        // which replaces the webContent aka the entire editor with the text, NOT Good
+        // also DO Not Stop Propagation as this breaks other drag drop functionality
+        document.addEventListener("dragover", ev => {
```
bryphe: I was hoping with this change we'd be able to remove this dependency - it's not ideal for `NeovimEditor` to have this global side effect (for example, if we have multiple `NeovimEditor`s). Could we factor this behavior into the `FileDropHandler`?

browser/src/Editor/NeovimEditor/NeovimSurface.tsx
```diff
@@ -67,7 +70,7 @@ class NeovimSurface extends React.Component<INeovimSurfaceProps> {
 
     public render(): JSX.Element {
         return (
-            <div className="container vertical full">
+            <div className="container vertical full" ref={e => (this._editorContainer = e)}>
```
bryphe: What if we made the following changes?

1) Use `<FileDropHandler />` for this line
2) Update `<FileDropHandler>` to render the children (see note above)
3) Have `FileDropHandler` be responsible for getting the ref
4) If `FileDropHandler` owns the ref, it can hook those events on the element, and call `preventDefault` - so now `NeovimEditor` doesn't need to explicitly hook any drag/drop events - it can delegate all the functionality to `FileDropHandler`.

Right now, the `FileDropHandler` feels a bit 'awkward' - it doesn't feel like a first class React component, since it's position in the hierarchy really doesn't matter - all that matters is the target thats passed in. But I think it'd be simpler if we could just put this as the 'container' and have it take care of the events (including the `preventDefaults`).

browser/src/Editor/NeovimEditor/NeovimEditor.tsx
Akin909: @bryphe an alternative option option is to call `stopPropagation` but depending on if we are very clear about how future drag and drop functionality will work it might be fine otherwise I think this could cause unusual long term bugs e.g. if there is another drag or drop bit of functionality that perhaps overlaps with this one then another contributor who is not aware of the inner working of the fileDropHandler wouldn't understand why their drop event was not being dispatched not knowing that somewhere along the line the event's propagation was halted.

I haven't tried it but if we `stopPropagation` in the target element in this case the editor container, it wont move outside to the document. This will mean though that if down the line there is the desire for some functionality maybe drag something into some specific section of the editor for some other reason that won't work

All we really want is to prevent the *document's* global behaviour but otherwise have events bubble up as normal was my thinking hence the global listeners

browser/src/Editor/NeovimEditor/NeovimSurface.tsx
Akin909: @bryphe by using the handler in place of the container the component becomes entirely specific to this one use i've refactored the global handlers into the component and use the [ref callback pattern](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) which I pass down using a render prop aka

```js
<FileDropHandler>
	{({ setRef }) => (
		<div ref={setRef}/>
	}
</FileDropHandler>
```

this has the advantage of allowing the file handler to be reused without needing to render a particular component of a given size, an example of a potential use case of this could be having the screen actually have two drop containers (invisible) and on hovering a file onto oni if on the left then that handler could open in a `vsplit` and vice versa, or potentially dropping an image into a markdown file and having the link extracted and input into the document etc.

PR #2182, opened 2018-05-07
[WIP] Switch explorer rendering to use react-virtualised

Akin909:
An issue I noticed with the explorer was that on reaching the lowest boundaries of the visible files/folders the scroll position would flicker on each move downwards making it hard/impossible to move to files which are out of view.

Having attempted to use/tweak the `scrollIfNeeded` function with no luck I've refactored the explorer to use react virtualize instead which I believe was mentioned at some point was an eventual change we wanted to make.

Todo:
- [ ] fix scroll issues when folder(s) is/are expanded - at present the main hold up with this PR is that if a folder is expanded react virtualised seems maybe to not know how to scroll since ?the dimensions are off?

### DIFF
--- a/browser/src/Services/Explorer/ExplorerView.tsx
+++ b/browser/src/Services/Explorer/ExplorerView.tsx
@@ -12,7 +12,7 @@ import { compose } from "redux"
 
 import { CSSTransition, TransitionGroup } from "react-transition-group"
 
-import { css, styled } from "./../../UI/components/common"
+import { css, enableMouse, styled } from "./../../UI/components/common"
 import { TextInputView } from "./../../UI/components/LightweightText"
 import { SidebarContainerView, SidebarItemView } from "./../../UI/components/SidebarItemView"
 import { Sneakable } from "./../../UI/components/Sneakable"
@@ -39,31 +39,15 @@ export interface INodeViewProps {
     updated?: string[]
     isRenaming: Node
     isCreating: boolean
+    children?: React.ReactNode
 }
 
 export const NodeWrapper = styled.div`
+    cursor: pointer;
     &:hover {
         text-decoration: underline;
     }
 `
-
-interface IOpts {
-    behavior: string
-    block: string
-    inline: string
-}
-
-type ScrollViewOpts = IOpts | boolean
-
-interface IChromeElement extends HTMLElement {
-    scrollIntoViewIfNeeded: (opts?: ScrollViewOpts) => void
-}
-
-const scrollIntoViewIfNeeded = (elem: IChromeElement) => {
-    if (elem && elem.scrollIntoViewIfNeeded) {
-        elem.scrollIntoViewIfNeeded(false)
-    }
-}
 
 const stopPropagation = (fn: () => void) => {
     return (e?: React.MouseEvent<HTMLElement>) => {
@@ -143,14 +127,7 @@ export class NodeView extends React.PureComponent<INodeViewProps, {}> {
         const renameInProgress = isRenaming.name === node.name && isSelected && !isCreating
         const creationInProgress = isCreating && isSelected && !renameInProgress
         return (
-            <NodeWrapper
-                style={{ cursor: "pointer" }}
-                innerRef={
-                    this.props.isSelected
-                        ? (elem: IChromeElement) => scrollIntoViewIfNeeded(elem)
-                        : () => ({})
-                }
-            >
+            <NodeWrapper>
                 {renameInProgress ? (
                     <TextInputView
                         styles={renameStyles}
@@ -293,9 +270,15 @@ import { commandManager } from "./../CommandManager"
 interface ISneakableNode extends IExplorerViewProps {
     node: Node
     selectedId: string
+    list: List
 }
 
 const SneakableNode = ({ node, selectedId, ...props }: ISneakableNode) => {
+    // if (node.type === "folder" && node.expanded) {
+    //     props.list.recomputeRowHeights()
+    //     props.list.forceUpdate()
+    // }
+
     return (
         <Sneakable callback={() => props.onClick(node.id)}>
             <NodeView
@@ -316,7 +299,13 @@ const SneakableNode = ({ node, selectedId, ...props }: ISneakableNode) => {
     )
 }
 
+const ExplorerContainer = styled.div`
+    height: 100%;
+    ${enableMouse};
+`
+
 export class ExplorerView extends React.PureComponent<IExplorerViewProps, {}> {
+    private _list: List
     public render(): JSX.Element {
         const ids = this.props.nodes.map(node => node.id)
 
@@ -332,22 +321,24 @@ export class ExplorerView extends React.PureComponent<IExplorerViewProps, {}> {
         }
 
         return (
-            <TransitionGroup>
+            <TransitionGroup style={{ height: "100%" }}>
                 <VimNavigator
                     ids={ids}
                     onSelectionChanged={this.props.onSelectionChanged}
                     onSelected={id => this.props.onClick(id)}
                     active={this.props.isActive && !this.props.isRenaming && !this.props.isCreating}
+                    style={{ height: "100%" }}
                     render={(selectedId: string) => {
                         const selectedIndex = this.props.nodes.findIndex(n => selectedId === n.id)
                         return (
-                            <div className="explorer enable-mouse">
+                            <ExplorerContainer className="explorer">
                                 <AutoSizer>
                                     {({ height, width }) => (
                                         <List
+                                            ref={e => (this._list = e)}
                                             height={height}
                                             width={width}
-                                            rowHeight={20}
+                                            rowHeight={25}
                                             rowCount={this.props.nodes.length}
                                             scrollToIndex={selectedIndex}
                                             rowRenderer={({ index }) => {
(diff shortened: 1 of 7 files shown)


browser/src/Services/Explorer/ExplorerView.tsx
```diff
@@ -269,11 +265,72 @@ export interface IExplorerViewProps extends IExplorerViewContainerProps {
...
+        const node = this.props.nodes[index]
+        const isSelected = node.id === selectedId
+
+        const renameInProgress =
+            this.props.isRenaming.name === node.name && !this.props.isCreating && isSelected
+        const creationInProgress = this.props.isCreating && isSelected && !renameInProgress
+
+        switch (true) {
+            case renameInProgress:
+                return 35
```
bryphe: Hmm, how do we determine these sizes? I'm wondering if they are dependent on the `ui.fontSize` setting? I'm just worried about a bug where different font sizes cause rendering quirks.

browser/src/Services/Explorer/ExplorerView.tsx
```diff
@@ -289,36 +346,45 @@ export class ExplorerView extends React.PureComponent<IExplorerViewProps, {}> {
...
-                        ))
-
+                        const selectedIndex = this.props.nodes.findIndex(n => selectedId === n.id)
                         return (
-                            <div className="explorer enable-mouse">
-                                <div className="items">{nodes}</div>
-                            </div>
+                            <ExplorerContainer className="explorer">
+                                <AutoSizer>
+                                    {({ height, width }) => (
```
bryphe: Nice! Thanks for wiring this all up 👍

bryphe:
> An issue I noticed with the explorer was that on reaching the lowest boundaries of the visible files/folders the scroll position would flicker on each move downwards making it hard/impossible to move to files which are out of view.

Ah ya, I hit this too - didn't get a chance to log an issue, but it definitely was a problem! Thanks so much for fixing it!

> Having attempted to use/tweak the scrollIfNeeded function with no luck I've refactored the explorer to use react-virtualized instead which I believe was mentioned at some point was an eventual change we wanted to make, which has the added bonus of providing methods that can be used to control scrolling to selected elements.

Awesome!! 💯 Thanks for going to those lengths to fix it. This will be much better long-term too, especially for large directories with tons of files - we only have to render whats on the screen.

> This change also sets up the searchInExplorer functionality I'm planning by allowing for programmatically focusing search of a particular element.

Can't wait! 😄 

Overall the change looks great! I had one small concern around how we calculate the row heights - I'm not sure how it works with different font sizes, so I wanted to see what you thought there.

Thanks @Akin909 !

browser/src/Services/Explorer/ExplorerView.tsx
Akin909: I actually fiddled with the sizes here and tbh the css with `virtualized` involved is more or less opaque to me but I didn't notice a big difference using the values which were trial and error as a result of that process, I've tweaked it however to use `ui.fontSize` to derive the size in case

Akin909:
@bryphe made the changes you suggested re row height 👍 hopefully this is good to go

browser/src/Services/Explorer/ExplorerView.tsx
```diff
@@ -269,11 +266,77 @@ export interface IExplorerViewProps extends IExplorerViewContainerProps {
...
+    }
+
+    public calculateHeight = (selectedId: string) => ({ index }: { index: number }) => {
+        const { nodes, isCreating, isRenaming } = this.props
+        const node = nodes[index]
+        const { id, name } = node
+        const isSelected = id === selectedId
+
+        const PADDING = 10
+        const fontSize = parseInt(this.props.fontSize, 10)
```
bryphe: Hmm, I wonder if this handles the various cases we allow for fontsize? For example, it looks like `11px`, `11pt`, `11em` would all get handled the same way here, but they would actually have very different _pixel_ sizes. I think what we'd want instead is the measured font size, or some more deterministic way to get the size of the cells.

browser/src/Services/Explorer/ExplorerView.tsx
```diff
@@ -269,11 +266,77 @@ export interface IExplorerViewProps extends IExplorerViewContainerProps {
...
+        const rowSize = fontSize + PADDING
+
+        const renameInProgress = isRenaming.name === name && !isCreating && isSelected
+        const creationInProgress = isCreating && isSelected && !renameInProgress
+
+        switch (true) {
+            case renameInProgress:
+                return rowSize * 1.5
```
bryphe: I'm a bit concerned about these magic numbers here - `1.5` and `2.8`. If someone changes the styling of the node or the creation, they might not realize they need to update these here, and get difficult to find bugs. Along the lines of the above comment - is there a more deterministic way we can get these sizes?

browser/src/Services/Explorer/ExplorerView.tsx
Akin909: @bryphe I used the menu component which has a static value of `20` as my initial reference point for implementing this and I haven't noted any issue with the value we pass in there, I also tried various sizes with a static value for `rowHeight` locally and didn't get any issues re. different units I can look at converting em to px as for pt I really hope ppl aren't giving units in pt as that'll be very problematic on multiple fronts.

Akin909:
Switched to using the virtualized `CellMeasurer` which handles row sizing dynamically (not the most performant solution but the *only* one I could find after a great many different attempts)

bryphe:
Wow, thanks for figuring out how to get accurate row sizes, @Akin909 ! Awesome.

bryphe:
Pulled the latest changes to check it out on Windows - I used the `Paint Performance` tool. I wanted to make sure we weren't aggressively re-painting the explorer while typing, which we aren't 👍 

I think the reason this test is failing is because the paint happens just a bit later, so it gets picked up the test. The best bet is to simply turn off the sidebar / explorer in that test's config:
https://github.com/onivim/oni/blob/cd87a1f201ea8b20659c7d76d3a57464ce186cb8/test/ci/PaintPerformanceTest.config.js#L5

(That will reduce the noise - the test is really focused on the _editor surface_). Note that we disable the statusbar, tabs, and animations for the same reason 😄

Akin909:
@bryphe i've updated the test for this and its good to go

CrossR:
Code looks good, and the stuttering is gone, but when scrolling fast I get this slight lag between the selected item and the highlight. 

Bad gif example in the Oni repo just holding down `j`:

issue

That said its still a lot better than the current flickering, do you want to address the lag issue in a separate PR?

Akin909:
Gonna stick a work in progress on this as I've just done some more local testing and theres a weird sizing issue when trying to create a folder in the root dir :S didn't realise 🤦‍♂️

CrossR:
I was using the latest build at the time, so afc992fc4fa27a06cfd8207ac54fe43dc4dd6827.

I'll try again using my own build rather than grabbing the AppVeyor one though and see if its different.

bryphe:
Thanks @Akin909 and @CrossR for the all the testing here! 💯

Akin909:
@CrossR could you give this PR a try especially re creating and renaming and moving files, I've kind of lost perspective on it thing its a big improvement (re. the stutter) and havent noted any bugs today but previously noted a resizing issue which seems to have resolved itself.....

keforbes:
Hey @Akin909, it looks like this PR has a branch conflict now.  Can you take a look?

CrossR:
I'm still getting some weirdness with the scroll sticking to the bottom.

If I scroll past the bottom of the menu, the bar sticks to the bottom for the entire time now. There is also some jumping still where the bar vanishes for a bit and then reappears after letting go of the scroll button.

Another thing I've noticed is that on inserting items sometimes the spacing is off.
Repro: 
 - Open Oni with Oni repo. 
 - `Ctrl-h` to highlight explorer, 
 - `G` to move to bottom, 
 - `Ctrl-e` to make a new file, and set name to `a.txt`.
 - New file is added fine, but weird spacing occurs:

image

Its up to you if you want to sort them in this PR or in a follow up, neither are as bad as the flickering, so I'd understand getting this in and sorting the bugs later(in a smaller more contained PR).

Let me know and if you decide on that, I can approve the PR. No point doing that now since there is a merge conflict.

Akin909:
@CrossR thanks for having another look was reeeeeally stuck with the bug you noticed with the spacing I've switched the strategy to a more brute force one instead of only trying to update the affected nodes when something like a rename etc happens I instead now trigger a redraw/update of all the visible nodes so should avoid random ones here or there not being measured properly.

Re holding the button and the bar disappearing when is that happening? I haven't noted it since I switched to having the bar stick on the bottom although, one of the limitations of using virtualized and the cellMeasurer component from the docs is

> Since this component measures one cell at a time to determine it's width/height, it will likely be slow if a user skips many rows (or columns) at once by scrolling with a scrollbar or via a scroll-to-cell prop. There is (unfortunately) no workaround for this performance limitation at the moment.

So without finding a faster library or a different implementation button mashing could always lead to this although I there might be little ways to optimise this or hide it at least. Although the library authors have been talking about a new version for a while now so we'll see.

Anyway if you get a moment can you try it again to see if the weird spacing bug is still there(locally I havent noted it anymore), I never found a consistent way to reproduce it (your repro doesnt always bring it on nor have I found it to consistently reproduce)

CrossR:
>Re holding the button and the bar disappearing when is that happening? I haven't noted it since I switched to having the bar stick on the bottom although, one of the limitations of using virtualized and the cellMeasurer component from the docs is

Seems to still happen when I scroll rapidly in the Oni repo:

gif

I can't seem to repro the other bug now though, and I tried 10-15 times whereas before it was happening maybe every 2/3 times I tried.

Akin909:
@CrossR thanks for the GIF that helps a lot 💯 , re that issue it relates to the quote above re performance and the strategy were using basically the library cant keep up with the holding the button

As far as the docs say there is no clear way round it I'm hoping future releases improve performance and we get a fix for free if not down the line we might have to switch to a different strategy I have no idea what, the main value of the library is for scrolling v. long lists not quite what were doing which is scrolling to indexes very fast using keyboard input

PR #2183, opened 2018-05-07
Fix #2047 - Blurry canvas rendering

### LINKED ISSUE #2047: Text rendering blurry not retina.

bryphe:
This is totally credited to @Cryza , who did the investigation and identified the issue 👍 

__Defect:__ The `canvas` element is set to width `100%` and height `100%`. This can cause the `canvas` element to have non-integer `height` or `width`. The problem with this is that the `CanvasContext2D` will always have an integer size, and if the size of the `CanvasContext2D` does not match the actual size of the `canvas` element, the browser may do some scaling or interpolation to render the `CanvasContext2D` within the dimensions of the `canvas`. For best rendering results, the `CanvasContext2D` size and the `canvas` element need to match.

__Fix:__ Always set the `canvas` width and height to an integer size. Added a CiTest that checks this constraint.


bryphe:
@Akin909 - ah ya,thanks for trying it out on retina and catching that! It turns out there was a problem with the way I was calculated the element (I shouldn't have applied `window.devicePixelRatio` to the element's `style.width` and style.height`). The latest should work better now!

Akin909:
Just had another look and that is definitely much better 👍 tbh its always looked fairly good on my machine (aka I havent really suffered from easily discernible blurriness, plus my eyesight is far from 20x20, although it does seem crisper, hopefully it resolves the issue for ppl significant issues)

PR #2184, opened 2018-05-07
Feature/add latex preview

TalAmuyal:
Hi,

Added a Latex-preview to the current markdown-preview plugin.
My intention is for that plugin to be a previewer for static data like Markdown, Latex, (maybe) HTML, etc.
Is this acceptable?

Also, there is a huge diff in `package.json` and `yarn.lock` - will try to get rid of the bloatware.

### DIFF
--- a/extensions/oni-plugin-markdown-preview/package.json
+++ b/extensions/oni-plugin-markdown-preview/package.json
@@ -22,6 +22,7 @@
     },
     "tbd-dependencies": {
         "marked": "^0.3.6",
+        "mathjax": "^2.7.4",
         "dompurify": "1.0.2",
         "oni-types": "^0.0.4",
         "oni-api": "^0.0.9"
--- a/extensions/oni-plugin-markdown-preview/src/index.tsx
+++ b/extensions/oni-plugin-markdown-preview/src/index.tsx
@@ -24,7 +24,7 @@ interface IColors {
 }
 
 interface IMarkdownPreviewState {
-    source: string
+    rendered: string
     colors: IColors
 }
 
@@ -46,7 +46,7 @@ class MarkdownPreview extends React.PureComponent<IMarkdownPreviewProps, IMarkdo
             codeForeground: this.props.oni.colors.getColor("foreground"),
             codeBorder: this.props.oni.colors.getColor("toolTip.border"),
         }
-        this.state = { source: "", colors }
+        this.state = { rendered: "", colors }
     }
 
     public componentDidMount() {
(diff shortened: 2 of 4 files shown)


TalAmuyal:
The current implementation renders `.tex` files as Latex and `.md` files as Markdown.
A hybrid is also possible but starts to get complicated (mainly on the UX side).
Maybe as a next step?

badosu:
@TalAmuyal rendering tex files won't work this way for most use cases as they depend on the local latex installation and libs, mathjax is really only useful for embedded text in html or really simple tex files which are of no use for most users.

The approach for this would be to have an embedded latex plugin that renders the tex file as html or as pdf and displays it. I am not sure if there's already a lib for that.

For compiling (rendering) the tex file there are a multitude of complicated setups, the most simple one being running `pdflatex`. But it may be necessary to be able to be configurable.

So if you want to implement a tex previewer the current approach won't work very well. And the ideal approach may be a too much effort for little benefit.

I agree that previewing tex files is a killer feature for science-oriented Oni users, but I am not sure if it's a big audience for the effort. If you decide to proceed with it I would suggest the approach aforementioned.

TalAmuyal:
Just tired with [a "real" tex file](http://physics.clarku.edu/sip/tutorials/TeX/intro.html) and I see what you mean.
It seems that about half of it is not rendered `:,(`

badosu:
@TalAmuyal To be fair, that is a really great goal, as Latex IDEs are something really looked for as you can see by this [list](https://tex.stackexchange.com/questions/339/latex-editors-ides), myself having used Texnic Center and Lyx in the past.

It certainly would be a killer feature for vim users in the scientific realm.

For rendering tex files it seems this is a good wrapper that takes care of performing that: https://www.npmjs.com/package/react-latex. Would be good to have that as a separate plugin, as it is more of a niche.

Same thing for rendering embedded latex in html, as seen in this npm package: https://www.npmjs.com/package/latex.

CrossR:
>The approach for this would be to have an embedded latex plugin that renders the tex file as html or as pdf and displays it. I am not sure if there's already a lib for that.

This the approach I've had a bit of a look into, and how VSCode and similar do it. Then you don't have to deal with the horrible-ness that comes with latex package management, what compiler a user likes etc, you can simply say "What command do you use to make a PDF" and then show the PDF that appears.

I have however seen friends using this in a markdown sense, such that they can quickly stick nice equations in their MD files, but I think that is a lot more work.

PR #2186, opened 2018-05-08
Bugfix/add sneakability back to tabs

Akin909:
NB: Remade this PR due to Git drama ^^

It fixes a regression introduced in #2171 where the argument for the sneak callback was missed out. I've re-added it and added a `ciTest` to prevent future regressions


bryphe:
Ah thanks for catching this, @Akin909 ! Sorry about the regression in #2171  - really appreciate you adding a CiTest to protect this functionality 👍

Akin909:
@bryphe this works locally but i'm guessing that the sneaks are non-deterministic so I'll await you're sneak PR before going on here as the test is unlikely to pass in CI

Akin909:
@bryphe used the changes you made in #2187 thanks 👍 tweaked a check there which looked like a typo and this seems to be working now 😄

bryphe:
> EDIT: kind of sad that adding one argument to a function call and adding a ciTest still result in coverage going ⬇️ if only microscopically

We don't capture code coverage for our CiTests unfortunately - would be great to have that. So our coverage numbers really only track unit tests.

Akin909:
I agree although I was actually thinking of starting to dedicate testing efforts towards our middle layer of sorts as those sections seem very under represented in our tests

PR #2189, opened 2018-05-09
Add `--plugin-develop` flag

bryphe:
This is meant to be used in conjunction with the starter kit I'm working on for Oni plugins here:
https://github.com/onivim/oni-plugin-starter-kit

This change adds a `--plugin-develop` flag, which allows you to specify a plugin directory to load on launch.

This is to facilitate an `F5` scenario from oni-plugin-starter-kit, with the flow being that I have a _workspace command_ defined in `oni-plugin-starter-kit` which will spin up a new Oni instance with the plugin loaded for debugging and testing.

### DIFF
--- a/browser/src/App.ts
+++ b/browser/src/App.ts
@@ -133,6 +133,13 @@ export const start = async (args: string[]): Promise<void> => {
     PluginManager.activate(configuration)
     const pluginManager = PluginManager.getInstance()
 
+    const developmentPlugin = parsedArgs["plugin-develop"]
+    alert(parsedArgs)
+    if (developmentPlugin) {
+        alert(developmentPlugin)
+        pluginManager.addDevelopmentPlugin(developmentPlugin)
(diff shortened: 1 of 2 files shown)


TalAmuyal:
@bryphe, I tried this flag with post `0.3.4` master (taken from the `Insider` page).
Running `oni --plugin-develop` yields an instance of Oni that seems rather stuck.
Have an idea on how to debug it?

Tried on my Ubuntu 18.04 VM.

Definition of `rather stuck`:
    Un-responsive.
    No element it being rendered to the screen (just seeing the background).
    The upper toolbar shows up when pressing Alt, but it seems that the `Help -> Developer Tools` button doesn't open the dev-tools.

TalAmuyal:
Sorry for the spam - just looked at the PR and realized that the `--plugin-develop` should be followed with a path.

Issue #2069, opened 2018-04-07
Close tabs by middle clicking


AUTplayed:
Would love to be able to close editor tabs by clicking mouse3 (middle click, mouse wheel click...) like in many other programs.

texhnolyze:
I am currently having a look at this and did a first shot implementation of this, by adding a mousedown listener to both the tabs name and file icon (similar to the click listener to switch tabs).

Check https://github.com/onivim/oni/compare/master...texhnolyze:middle-click-tab-close?expand=1

I am however not very happy with this solution as it stands and would appreciate some suggestions:
* One option would be to combine the onClick and onMouseDown event handlers and select/close tabs based on the ```React.MouseEvent.button``` value (which I already do for the middle click)
* Another options if something like that is achievable nicely with react (I come more from an angular background) would be to write a general purpose middle click event handler to run any given function on a middle click of an element e.g. 
```Javascript
<div  className="name"
         onClick={this.props.onClickName}
         onMiddleClick={this.props.onMiddleClick}>
```

On another note I noticed, that tabs do not close if you only have a single one which suprised me but makes sense from a neovim perspective. Should I track that in an issue?

Akin909:
@texhnolyze I think from a react perspective option one if *much* more preferable to option two especially as we have a few issues already which are caused by very general event handlers occasionally swallowing or obscuring events. I think having an on click that also listens for a middle click would be a good solution

Tbh never used middle click or listened for it so can't advise much there re the actual event, re the tabs not closing if you only have a single one the reason for that is that an issue was raised a while back that if oni had only one tab remaining and it was closed neovim/oni would close which is how neovim responds to that so what is done instead now is a no name buffer is created on purpose to prevent that behaviour

texhnolyze:
@Akin909 I did an implementation of an ```onMouseDown``` event handler to handle both the click and middle click as sadly ```onClick``` only gets triggered when using left mouse button. So I have to check the events button value for left and middle click to prevent unexpected behaviour on right clicks

Concerning the tabs from what I gather from your response is, that when I have a single tab of let's say ```resume.md``` and I close it, it should be closed but a new empty buffer with the oni tab name ```[No Name]``` should be opened? Did I get that correctly? 
Because what is happening for me is that, when I try to close ```resume.md``` there is just nothing happening, which led me to assume I did not click the close button correctly and clicking it again the first time I encountered it.

[...]

PR #2190, opened 2018-05-10
Enable use of middle click to close tabs

texhnolyze:
Enables closing oni tabs with middle click, much like e.g. firefox/chrome.
Corresponds to #2069. 

**Do not merge yet:**
As mentioned in the issue there is still some clarification needed
concerning the ability to close tabs, when only a single one is opened and how to go 
about testing clicks/middle clicks of tabs to ensure they are correctly selected and closed.


keforbes:
> Do not merge yet:

Please put `[WIP]` at the beginning of the Pull Request's title (Work In Progress) until the Pull Request is ready.

keforbes:
Feel free to edit the title and remove `[WIP]` when you're ready to go. :slightly_smiling_face:

bryphe:
Thank you for the PR, @texhnolyze ! The functionality change looks great to me 👍 

Would it be possible to add a test case to exercise this behavior? I'm wondering if we could use the [simulate](http://airbnb.io/enzyme/docs/api/ReactWrapper/simulate.html) method to exercise the normal click and non-middle button click, and check if the correct `onTabClosed` and `onTabSelect` callbacks from the props are executed as expected?

texhnolyze:
@keforbes thanks, will do in the future
@bryphe ok once I am finished with this I will open a new issue concerning the single tab behaviour.
Concerning the test I will have another try to get that working, however due to the the async nature of `FileIcon` (as described in the issue) the way I see it the only two ways to get that test working is injecting `FileIcon` as a dependency so that I can mock it out in the test easily (which would require quite a bit of change), overwriting the actual imported `FileIcon` in tabs or actually setup the underlying async `IconThemes` first and turn the test more into an integration test. Not sure what the best option would be.

texhnolyze:
@bryphe yeah took my some trial and error, but this is finished from my side for this pull request, if you already had a look at the changes you probably saw what I did with the `classnames` dependency ([commit](https://github.com/onivim/oni/pull/2190/commits/cf6df8d3821a09420b0450bcf3933e5c49b0bc02)).

This works for now, but I would suggest maybe dropping the dependency in favor of an own implementation, because as it currently stands it does not seem like there will be a [new release soon](https://github.com/JedWatson/classnames/issues/152). The implementation of this as a project component should be failrly trivial. What do you think?

Also now that I know a bit more about the test setup and jest and so on, I'll do a pull request, when I find the time, to add a bit more test coverage and to get to know the code base better.

bryphe:
Merged - thank you for the contribution, @texhnolyze ! 💯

Issue #1999, opened 2018-03-31
Tutorial: "Target Practice" - Exercise `targets.vim` plugin that is bundled by default


bryphe:
A cool tutorial would be one that exercises one of our bundled default plugins: 
https://github.com/wellle/targets.vim

As @CrossR suggested, the description would be an excellent place to make a special attribution to the author 👍 

Warm-up with some of the default 'text objects':
- `ci"`
- `ci(`

And then start exercising some of the more in-depth text objects. I only use the basic ones, so it would be fun to build some muscle memory around some of the more involved ones. I'd like to get more proficient at nested pairs and last pair, for example. Part of the task will be designing the exercises here - we can always start small and add more!

A good starting point for this would be looking at some of the existing tutorials. One in particular that would be relevant might be: https://github.com/onivim/oni/blob/master/browser/src/Services/Learning/Tutorial/Tutorials/DeleteOperatorTutorial.tsx

We'd probably be leveraging the `WaitForStateStage` to detect if the modifications are correct and then use `SetBufferStage` to set up the next exercise.

nwneisen:
I'd like to take a look at this.  I'm new to contributing to open source so I might have a few hiccups along the way.

The code for this seems pretty straight forward after looking at other tutorial files, it's organizing all of the information that might take a few tries.  I think following the same basic overview as the plugin's repo and providing a link to the repo for advanced features would work well.  Something along these lines:

Intro
Pair text objects
Quote text objects
Separator text objects
Argument text objects
Explain second character options
Explain third character options
Conclusion

With the intro providing attribution to the author and the conclusion mentioning to go to the repo for advanced usage.  I just learned about targets.vim with this issue but I believe that would give a basic overview of all its features. 

Also, all of the tutorials in onivim are currently under a collapsible arrow labeled "Tutorials".  It seems the current tutorials might belong under a "Basic" label and this new tutorial under a "Plugins" label.  If this is desired then it might be a good idea to put these in their different sections now rather than having to come back and do it later.

nwneisen:
I've opened a pull request for this issue that can be found here:
#2192

keforbes:
#2192 has been merged in, closing issue.

PR #2192, opened 2018-05-11
[RFC] #1999 - Add Targets.vim tutorial

nwneisen:
Add a tutorial with description, useful keys, and interactive examples
to teach users about Target.vim.

### DIFF
--- a/browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
+++ b/browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
@@ -1,5 +1,4 @@
 /**
- * TargetsVimPluginTutorial.tsx
  *
  * Tutorial that exercises the targets.vim plugin
  */
@@ -17,6 +16,7 @@ const Line1c = "Targets.vim adds text objects for additional (operations)."
 const Line2a = "'cin(' selects inside the next pair of (parenthesis)."
 const Line2b = "'can(' selects around the next pair of (parenthesis)."
 const Line2c = "Replacing 'n' with 'l' will move to the previous pair."
+const Line2d = "Omitting either will select the (current pair) or the (next)."
 
 const Line3a = "Quote objects can 'also' be used."
 const Line3b = '\'cIn"\' selects the first characters inside of " quotes ".'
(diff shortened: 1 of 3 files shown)


nwneisen:
Here is my work for issue #1999.  I ran into some trouble when I squashed my commits but I believe everything is straightened out now.  

I'm really happy with how this turned out.  I think the only changes I might look into more are adding a horizontal divider between the note keys and pausing after the user enters a correct position.

Looking forward to your feedback.

browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
```diff
@@ -0,0 +1,142 @@
...
+                41,
+            ),
+            Stages.combine(
+                null,
+                new Stages.FadeInLineStage(null, 1, Line5a),
+                new Stages.FadeInLineStage(null, 2, Line5b),
+                new Stages.SetBufferStage([Line5a, Line5b, Line5c]),
+            ),
+            new Stages.SetCursorPositionStage(0, 0),
+            new Stages.MoveToGoalStage("Use 'dina' to delete the next programming argument", 1, 31),
```
bryphe: I noticed after the delete it moved quickly to the next stage (so I wasn't quite able to see the impact of my change) - is this what you referred to about adding a pause? I think that might be helpful, just to give the user some feedback about what the delete operation did.

browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
```diff
@@ -0,0 +1,142 @@
...
+            new Stages.SetCursorPositionStage(1, 7),
+            new Stages.MoveToGoalStage(
+                "Use 'cIn\"' to move to first character inside double quotes",
+                1,
+                48,
+            ),
+            new Stages.WaitForModeStage("Ready to type.  Go back to 'normal' mode", "normal"),
+            new Stages.SetCursorPositionStage(2, 7),
+            new Stages.MoveToGoalStage("Use 'cAn\"' to select around the double quotes", 2, 26),
+            new Stages.WaitForModeStage("Ready to type.  Go back to 'normal' mode", "normal"),
```
bryphe: I realized at first when I got to this step - I wasn't sure if I was  supposed to do anything - I think the 'Ready to type.' was a bit ambigious (for me). We could consider making it a bit more explicit, like:
- "Press ESC to go back to 'normal' mode."

Or perhaps just:
- "Go back to 'normal' mode"


Or have them actually enter some text:
- "Type 'foo'."
- " Press ESC to go back to normal mode."

bryphe:
Hi @nwneisen , thank you for submitting the PR! Just played through the tutorial, it's great 👍 I like that you attributed the author in the body of the tutorial!

Just left some minor feedback in the comments - let me know what you think.

I also like the shorter version of some the commands:
- `ci"`, `ci'`, `ci(` - I wonder if it is worth calling out those too? (I think technically `ci"` and `ci'` are not part of the plugin, though - so perhaps a different tutorial is the proper place for those).

Overall the change looks good though, really appreciate it @nwneisen !

bryphe:
BTW, I noticed I'm starting to use `cin` more now thanks to this tutorial! 👍

nwneisen:
@bryphe I addressed your comments from the review.  I also added a section to show the difference between `ci` and `cin`.  Let me know what you think!

browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
```diff
@@ -42,25 +42,33 @@ export class TargetsVimPluginTutorial implements ITutorial {
...
             Stages.combine(
                 null,
                 new Stages.FadeInLineStage(null, 1, Line2a),
                 new Stages.FadeInLineStage(null, 2, Line2b),
-                new Stages.SetBufferStage([Line2a, Line2b, Line2c]),
+                new Stages.FadeInLineStage(null, 3, Line2c),
+                new Stages.SetBufferStage([Line2a, Line2b, Line2c, Line2d]),
             ),
             new Stages.SetCursorPositionStage(0, 7),
             new Stages.MoveToGoalStage("Use 'cin(' to move inside the paranthesis", 0, 40),
```
bryphe: Minor, but I just realized `paranthesis` was misspelled a couple of places - should be `parenthesis`.

browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
```diff
@@ -42,25 +42,33 @@ export class TargetsVimPluginTutorial implements ITutorial {
...
+            new Stages.WaitForModeStage("Press ESC to go back to normal mode", "normal"),
             new Stages.MoveToGoalStage(
                 "Use 'cil(' to move back inside the last paranthesis",
                 0,
                 40,
             ),
-            new Stages.WaitForModeStage("Ready to type.  Go back to 'normal' mode", "normal"),
+            new Stages.WaitForModeStage("Press ESC to go back to normal mode", "normal"),
+            new Stages.SetCursorPositionStage(3, 34),
+            new Stages.MoveToGoalStage("Use 'ca(' to select inside the current paranthesis", 3, 32),
```
bryphe: Is this one supposed to be `ci(` instead of `ca(`?

browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
```diff
@@ -42,25 +42,33 @@ export class TargetsVimPluginTutorial implements ITutorial {
...
+            new Stages.WaitForModeStage("Press ESC to go back to normal mode", "normal"),
+            new Stages.MoveToGoalStage("Use 'ca(' to select the next paranthesis", 3, 40),
```
bryphe: I realized for this line, and line 68, the word 'select' is a bit confusing (it's overloaded in Vim, since there is a `select mode`). Perhaps we could replace 'select' with 'change'?

bryphe:
Thanks a lot for making the changes, @nwneisen ! Overall looks good to me. Just had a couple other pieces of minor feedback - let me know what you think.

browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
nwneisen: Good catch.

browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
nwneisen: I had the user continue to use ca( so that there wasn't a pair of parenthesis afterwards.  This will cause the next ci( to select the same pair of parenthesis.  I could change it and add a step to have the user move the cursor to the right and really drive the point of it selecting the current pair.

browser/src/Services/Learning/Tutorial/Tutorials/TargetsVimPluginTutorial.tsx
nwneisen: Good idea.  I will update anywhere using 'ci(' to 'change' and anywhere using 'di(' to delete.  I think associating the words for each shortcut will help to remember them.

nwneisen:
Here are the vocabulary adjustments.  On top of doing 'c' for 'change' and 'd' for 'delete', I also looked at the second character (iaIA) and tried to match the vocabulary used for these in the Targets.vim repo as well.  

I didn't make any change to go from 'ca(' to 'ci('.  I feel like entering two 'ca(' in a row shows the user that its behavior can change without any other input.

PR #2196, opened 2018-05-11
Fix #2191 - Snippet loading issues

### LINKED ISSUE #2191: Slight snippet differences
I was testing some Python snippets before I recommended them to someone, and realised there is 2 small differences between ours and VSCode snippets, or at least it appears to be that way.

1. We require the `body` to be a list.
2. We don't support trailing commas.

https://github.com/Microsoft/vscode-python/blob/master/snippets/python.json is the example I was using. The last three snippets body is a single string rather than a string in a list. Also, the final snippet is terminated with `},`, which seems to both break the snippets and also stop the menu option appearing somehow? Removing that comma fixed it though.

bryphe:
__Issues:__ Oni had issues loading some snippet files in 2 cases:
- Case where the snippet is a `string` instead of `string[]`
- Case where the snippet file is not well-formed JSON

__Defect:__
- We assumed that the body of the snippet is always an array of strings
- We used `JSON.parse` to parse the snippet file, but that is not very forgiving - it doesn't allow for things like trailing commas

__Fix:__
- Handle the case where the body of the snippet is a string, and add test
- Use the `JSON5` parser, which implements a more forgiving JSON syntax (that is also backwards-compatible with JSON).

Fixes #2191


twice:
How can one make snippets modular? I tried global snippets ala vscode (<name>.code-snippets) but they do not seem to work, neither does ultisnips like snippets subdirectories.

CrossR:
I edit most of my snippets via the Command Pallete menu (Ctrl-Shift-P/cmd-shift-p).

There is an option for per filetype snippets in there and global ones, just search for "snippet" I think.
That will drop you into a snippet JSON file in the Oni config folder.

Once in that file you can either manually add, or copy and paste VSCode JSON snippets in to it.

twice:
I probably did not express myself well, the solution I'm looking for was what options are there to modularize snippets per filetype. e.g. in ultisnips I normally have under the javascript subdirectory snippets files for react, redux, es6 e.t.c which are then available for any javascript file.

Thanks for the suggestion you provided and it's what I think am currently using. I believe a cson option ala atom would be a nice addition.

twice:
I have decided to improvise by using subdirectories in the snippets directory and cson and then use a command of the form `cat javascript/*.cson | cson2json > javascript.json` to produce the final json file in the snippets directory that satisfies oni.

PR #2197, opened 2018-05-12
Bugfix: Accumulating error squiggle styles in document head

Akin909:
I have been noticing quite a lot of slow down whilst using oni for 30+ minutes (maybe an hour), it seems that since the refactor of some components to styled components we have run into the styles accumulation issue which occured before.

screen shot 2018-05-12 at 21 12 35

The reason this doesn't seem to have been caught by the `ciTest` -> `PaintPerformance` is that the error occurs *over time* and *with multiple errors*.

I have tweaked the `ErrorSquiggle` and `ErrorMarker` components to use the [`styled.element.attrs`](https://github.com/styled-components/styled-components/issues/134) (link is to an issue that relates). Essentially this method is an escape hatch that just inlines the styles that need to be inlined so styles can still be in a styled component but all the highly dynamic styles have an out.

Also I upgraded styled components as there's been some performance improvements and bug fixes, if interested see [here](https://medium.com/styled-components/v3-1-0-such-perf-wow-many-streams-c45c434dbd03) for perf improvement post.

Also changed token classes for hover to return an empty string if there is a falsy value -> recently discovered that `undefined` isn't totally filtered out and can end up being `preprocessed` aka => `-ms-prefix-undefined`


bryphe:
Excellent! Thanks for catching this, @Akin909 - this is definitely the sort of thing that can cause performance issues. Really appreciate you looking out for this!

Those performance fixes in `styled-components` look promising too - could help our startup time, potentially 👍

PR #2198, opened 2018-05-13
Improve WebGL Renderer stability

Cryza:
This fixes two issues with the WebGL renderer:

### First Issue: Texture space used up
This issue was first reported by @adelarsq
The WebGL texture space that we allocated for prerendering the glyphs was fixed until now. For large screens with high DPI and/or with much scrolling, the renderer ran out of space at some point. This caused it to crash entirely.

#### Fix:
We now use a texture array instead of a single texture which enables us to load multiple layers of our webgl atlas into the GPU. Like that, we can gradually fill each layer while adding new glyphs and switch to the next one whenever the current one is full. We can still run out of space when all layers are full, but the renderer will now recreate a new atlas with twice as many layers once that happens.

The renderer always starts with 2 texture layers at 1024x1024 each. All hardware should support at least 256 texture layers (with many of them going up to 2048 as can be seen [here](https://webglstats.com/webgl2/parameter/MAX_ARRAY_TEXTURE_LAYERS)), so this should be enough.

I chose to go with comparably small texture layers of only 1024x1024 even though most hardware would support at least sizes of 2048 (see [here](https://webglstats.com/webgl2/parameter/MAX_3D_TEXTURE_SIZE)) because the layered approach enables us to do the following optimization:
- Everytime we add a new glyph to the atlas, we have to upload the texture to the GPU again
- Since we only ever change the last layer of the atlas, we would re-upload unchanged data in all other layers if we re-upload everything
- Using `gl.texSubImage3D`, we can thus only re-upload the last layer of the atlas and keep the unchanged data from the previous  uploads

The most expensive part of GPU-based calculation is almost always the communication between CPU and GPU. This approach allows us to minimize the data that needs to be uploaded to the GPU per draw call with new glyphs to 1024x1024x4 (WxHxRGBA) instead of 1024x1024x4xlayerCount. The number 1024 itself only comes from a gut feeling though, so actual experimentation could still lead us to a larger or smaller number.

### Second Issue: `Cannot read property 'canvas' of null`
This error was reported by @justinmk in https://github.com/onivim/oni/pull/2120#issuecomment-388108639
For some reason, the `NeovimRenderer` component can cause `INeovimRenderer.redrawAll()` to be called before `INeovimRenderer.start(editorElement)`, leading to this `TypeError`.

#### Fix:
Since I was not sure how exactly the usage`ResizeObserver` and the React lifecycle methods would have needed to be changed to fix this without breaking something else, I just switched to skipping the entire `redrawAll` body in the `WebGLRenderer`. This is also what we do in the `CanvasRenderer`.

I would however prefer to clean up the interface between the `NeovimRenderer` and the implementations of `INeovimRenderer` instead at some point in time since this seems unintuitive to me; why is there a `start` method if the caller does not guarantee that to be called first? :)

Edit: This change also introduces the fix for blurry rendering from #2183 resp. #2047 to the WebGL Renderer

@nathansobo The sampler array strategy might also be relevant for Xray. Maybe this is could be the first occasion where we pay back some effort to your project :)

### DIFF
--- a/browser/src/Renderer/WebGL/WebGLTextRenderer.ts
+++ b/browser/src/Renderer/WebGL/WebGLTextRenderer.ts
@@ -17,7 +17,7 @@ const vertexShaderAttributes = {
     targetOrigin: 1,
     targetSize: 2,
     textColorRGBA: 3,
-    atlasIndex: 4,
+    atlasLayerIndex: 4,
     atlasOrigin: 5,
     atlasSize: 6,
 }
@@ -29,14 +29,14 @@ const vertexShaderSource = `
     layout (location = 1) in vec2 targetOrigin;
     layout (location = 2) in vec2 targetSize;
     layout (location = 3) in vec4 textColorRGBA;
-    layout (location = 4) in float atlasIndex;
+    layout (location = 4) in float atlasLayerIndex;
     layout (location = 5) in vec2 atlasOrigin;
     layout (location = 6) in vec2 atlasSize;
 
     uniform vec2 viewportScale;
 
     flat out vec4 textColor;
-    flat out int convertedAtlasIndex;
+    flat out int convertedAtlasLayerIndex;
     out vec2 atlasPosition;
 
     void main() {
@@ -44,7 +44,7 @@ const vertexShaderSource = `
         vec2 targetPosition = targetPixelPosition * viewportScale + vec2(-1.0, 1.0);
         gl_Position = vec4(targetPosition, 0.0, 1.0);
         textColor = textColorRGBA;
-        convertedAtlasIndex = int(atlasIndex);
+        convertedAtlasLayerIndex = int(atlasLayerIndex);
         atlasPosition = atlasOrigin + unitQuadVertex * atlasSize;
     }
 `.trim()
@@ -57,13 +57,13 @@ const firstPassFragmentShaderSource = `
 
     layout(location = 0) out vec4 outColor;
     flat in vec4 textColor;
-    flat in int convertedAtlasIndex;
+    flat in int convertedAtlasLayerIndex;
     in vec2 atlasPosition;
 
     uniform sampler2DArray atlasTextures;
 
     void main() {
-      vec4 atlasColor = texture(atlasTextures, vec3(atlasPosition, convertedAtlasIndex));
+      vec4 atlasColor = texture(atlasTextures, vec3(atlasPosition, convertedAtlasLayerIndex));
       outColor = textColor.a * atlasColor;
     }
 `.trim()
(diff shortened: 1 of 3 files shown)


Akin909:
@Cryza really cool to see fixes for these issues 👍

I've been sitting on an issue when trying the renderer since i'm not sure if its known/ I didnt/dont want to pile on but just wanted to mention in case it relates (doesn't sound like it does) which is best illustrated with a screen shot -> 

screen shot 2018-05-13 at 11 37 46

The issue is the small dots/artifacts above random characters, they persist despite scrolling although weirdly they seem to occur over only certain characters like `p`, `)` but these can vary. It seems to occur/be worse depending on fonts. The font I'm using in the screen shot is `input mono regular`, also although it might be random the problem seems to go away/less massively on reloading the editor

Not sure if what this might be down to but seems reminiscent of the canvas crack although since we redraw everything and that was previously the cause it seems unlikely. Can raise a separate issue or leave it be if its known

Cryza:
Thanks for trying this out, @Akin909!

I actually saw some of these artifacts before as well. I'm pretty sure they're related to characters not sticking to their em-rectangle for certain fonts. I expect the dots you see to be the cut-off rest from a descender that was rendered in the line above in your current WebGL atlas.

If you want, you should be able to investigate that by giving the `canvas` element inside your `document.body` a style like `position: absolute` so that you can see it. For the case in your screenshot, there should be a character with a long descender above the `p`. That character exceeds its own rendering area and even reaches into the rectangle of the `p` beneath it, which causes that fraction to be copied into the editor view along with each `p`.

Now in terms of fixing this issue (which I clearly see as a blocker):
Originally, we intentionally left a certain gap between glyphs (`glyphPaddingInPixels`) in the atlas to mitigate this problem, but that doesn't fully solve it in case the padding is not big enough. Maybe we should reintroduce it though, possibly scaled with the font size. I can't really think of a smarter solution at the moment. Maybe you have a better idea, @bryphe?

Akin909:
Not sure I can advise really, but I wonder how we derive the size of each line i.e. the average/max character height as for the canvas renderer we were using the `H` character but I wonder if it'd be possible to have a string of characters like `a-z` and special chars like `{[|` etc. and then use those to calculate the max height of and then use that as the height for the line?

also this is seeming like an issue I should raise separately but whilst running the bold/italic branch I noticed

screen shot 2018-05-14 at 10 15 30

You might need to look a bit more closely to see it but the artifacts seem to occur on the horizontal axis as well once italic is enabled

bryphe:
Awesome work @Cryza !

> Using gl.texSubImage3D, we can thus only re-upload the last layer of the atlas and keep the unchanged data from the previous uploads

Very cool.

> Since I was not sure how exactly the usageResizeObserver and the React lifecycle methods would have needed to be changed to fix this without breaking something else, I just switched to skipping the entire redrawAll body in the WebGLRenderer. This is also what we do in the CanvasRenderer.

> I would however prefer to clean up the interface between the NeovimRenderer and the implementations of INeovimRenderer instead at some point in time since this seems unintuitive to me; why is there a start method if the caller does not guarantee that to be called first? :)

Seems reasonable to me - that interface / interplay is awkward, for sure. Definitely open to improving it! Your solution seems reasonable for the meantime though 👍

browser/src/Renderer/WebGL/WebGLTextRenderer.ts
```diff
@@ -9,16 +9,17 @@ import {
...
 const glyphInstanceSizeInBytes = glyphInstanceFieldCount * Float32Array.BYTES_PER_ELEMENT
 
 const vertexShaderAttributes = {
     unitQuadVertex: 0,
     targetOrigin: 1,
     targetSize: 2,
     textColorRGBA: 3,
-    atlasOrigin: 4,
-    atlasSize: 5,
+    atlasLayerIndex: 4,
```
bryphe: Really cool that we can index into the atlas layer from this vertex data; thanks for plumbing that through! 👍 Neat to see how it works with the `sampler2DArray` and how the lookup works in fragment shader.

bryphe:
Hope you guys don't mind, I pulled the 'render artifacts' issue out to #2202 - it seems like a potentially separate issue from this change (independent of the layers).

Cryza:
Thanks for the feedback! I totally agree on the extraction of the artifacts as a separate issue. Bringing this in now :)

PR #2203, opened 2018-05-14
Add issue template

dkns:
Decided to keep issue template short becuase I'm not sure what kind of information would be valuable to know. Like is yarn version important, etc? Anyway it can be added later on.


Akin909:
Thanks a lot 👍 really great to formalise a structure

PR #2204, opened 2018-05-14
Cleanup tutorials

keforbes:
I went through the tutorials and tried to clean them up a bit.  Originally, I just wanted to add some repetition to reinforce the commands but then I got carried away and kept tweaking things.  I don't think my changes will offend anyone, but who knows.

The changes I made:
- I re-arranged the tutorials so all the "motions" come before editing a file (Insert Mode).  I also re-arranged the motions to follow (what I consider) a more logical flow.  It goes from `h/j/k/l` to `w/e/b` to `0/$/_`, then to `gg/G` and search.  In my mind, this goes from smaller-jumps to larger-jumps and simple movements to more complex movements.
- The `w/e/b` tutorial was using `0/$` when those already have a dedicated tutorial for them.  And after my re-arranging, `w/e/b` came before that tutorial.  I removed all references of `0/$` in `w/e/b` to remove confusion.
- Changed "Motion and Insert" to be "Insert and Undo".  Since all the motion tutorials are non-destructive, I wanted them to be before any edits to the file.  I also wanted to make sure we taught `u`ndo as soon as we taught destructive changes so the user can undo any mistakes when trying to complete later tutorials.  Sometimes you can modify the wrong text and without `u` there's no way to successfully complete the tutorial.
- Added some repetition to the ChangeOperator tutorial.  I had just thrown together the tutorial with a single example so I could reference `c`hange in my VisualMode tutorial.  Now it has a couple more uses of `cw` for reinforcement.
- Added some repetition of `xp` to the CopyPaste tutorial.  Again, I originally only had one example of this and it was the last item in the tutorial so the "you win" animation played before you could even see what `xp` had changed.
- Fixed some issues in the MoveAndInsert/InsertAndUndo tutorial where the marker was claiming success before you entered a `space` after the added word.  This was throwing me off because it claimed I had completed the step but then wasn't matching the text in the next step so I couldn't complete the tutorial.
- Cleaned up the wording in Search tutorial to be more descriptive in case this is the user's first introduction to search mode.

### DIFF
--- a/browser/src/Services/Learning/Tutorial/Notes.tsx
+++ b/browser/src/Services/Learning/Tutorial/Notes.tsx
@@ -112,6 +112,10 @@ export const OKey = (): JSX.Element => {
     )
 }
 
+export const UKey = (): JSX.Element => {
+    return <KeyWithDescription keyCharacter="u" description={<span>Undo a single change</span>} />
+}
+
 export const GGKey = (): JSX.Element => {
     return (
         <KeyWithDescription
--- a/browser/src/Services/Learning/Tutorial/Tutorials/BeginningsAndEndingsTutorial.tsx
+++ b/browser/src/Services/Learning/Tutorial/Tutorials/BeginningsAndEndingsTutorial.tsx
@@ -40,8 +40,8 @@ export class BeginningsAndEndingsTutorial implements ITutorial {
                 2,
                 Line3.length - 1,
             ),
-            new Stages.MoveToGoalStage("Use '0' to move to the BEGINNING of the line", 2, 0),
             new Stages.MoveToGoalStage("Use '_' to move to the FIRST CHARACTER", 2, 4),
+            new Stages.MoveToGoalStage("Use '0' to move to the BEGINNING of the line", 2, 0),
             new Stages.MoveToGoalStage(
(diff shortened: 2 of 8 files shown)


bryphe:
Just played through these w/ the updates, nice! 👍 Thanks for taking a holistic look across the tutorials.

> I re-arranged the tutorials so all the "motions" come before editing a file (Insert Mode). I also re-arranged the motions to follow (what I consider) a more logical flow. It goes from h/j/k/l to w/e/b to 0/$/_, then to gg/G and search. In my mind, this goes from smaller-jumps to larger-jumps and simple movements to more complex movements.

I like the progression, seems reasonable to me!

> Changed "Motion and Insert" to be "Insert and Undo". Since all the motion tutorials are non-destructive, I wanted them to be before any edits to the file. I also wanted to make sure we taught undo as soon as we taught destructive changes so the user can undo any mistakes when trying to complete later tutorials. Sometimes you can modify the wrong text and without u there's no way to successfully complete the tutorial.

Yes, good catch - makes sense to bring up undo here.

Really appreciate you looking at the overall flow. I do like the added repetition for the commands - I think that is important for building up that 'muscle memory'! Nice work @keforbes !

Issue #1822, opened 2018-03-16
High CPU Usage


hkupty:
I've observed for the last release (0.3.1) a big CPU usage on when redrawing the screen.
Idle usage keeps using around 2% CPU (which is strange TBH), but when I do something (say, move the cursor or split the screen, CPU usage for the same process can go up to 80% and then drop again.

Also, it is consistently observed that oni spawns and keeps 50+ processes open at the same time with very few actually doing something.

This can make code editing experience annoying and frustrating.

Akin909:
@bryphe in relation to oni and its processes I've been looking at [`execa`](https://github.com/sindresorhus/execa) which is a library which is a drop in replacement for `child_process` with some really great added advantages one of which is relevant here aka *automatically kills child processes if the parent dies* which `cp` doesnt not do

bryphe:
Thanks for the report @hkupty - there's a few pieces of additional info that would be useful:
- What OS are you using?
- What kind of file are you editing when you see the spike? Does it happen in any sort of file?
- Do you have any customizations to `config.tsx` or `init.vim`?
- What plugins do you use with Oni / vim?

It would also be very helpful to have grab and save a performance profile - https://github.com/onivim/oni/wiki/Performance-Profiling - this can help us determine whats pegging the CPU.

One potential culprit could be the `editor.textMateHighlighting.enabled` flag - this was turned on by default in `0.3.1`, and it can incur some extra CPU for determining the syntax highlighting. Would be worth trying to set that to `false`. (Animations can also cause the idle CPU usage you mentioned - you may want to try setting `ui.animations.enabled` to `false`)

> Also, it is consistently observed that oni spawns and keeps 50+ processes open at the same time with very few actually doing something.

That amount of processes spawned is strange and seems like a bug... and could be related (spawning a process incurs CPU usage). Do you see which processes are being spawned? Are they language servers / etc? Are they being spawned during certain operations? We can get that information from the logs as well if you can attach it.

> execa which is a library which is a drop in replacement for child_process with some really great added advantages one of which is relevant here aka automatically kills child processes if the parent dies which cp doesnt not do

Yes, we definitely need to improve our 'clean up' hygiene when closing Oni. @Akin909 - Do you suspect that some of the hanging processes are child processes of a parent process that was spawned by Oni, and they aren't going away when the parent dies? Curious if you've seen this for a particular process.

Akin909:
[...]

PR #2208, opened 2018-05-15
Performance: Limit chorded keys to evaluate

bryphe:
__Issue:__ This was mentioned in #1822 , and manifests when holding down a key (ie, holding the down arrow key when scrolling). The `keydown` handler takes longer and longer to execute until the CPU is pegged and the responsiveness suffers.

__Defect:__ In our `handleKey` method, we look at recent key presses to see if they match any chorded key bindings (like `input.bind('jk')`. We only look at recent keypresses (~250ms), however, this list grows and grows while the key is being held down.

__Fix:__ Cap the keys to evaluate to four most recent. This improves performance and should be more than sufficient for the chorded key feature. After this change, in my profiling, the `handleKey` method was no longer the bottleneck.

Would love to have some sort of performance regression test for issues like these - does anyone know of a service or infrastructure for this?

### DIFF
--- a/browser/src/Services/InputManager.ts
+++ b/browser/src/Services/InputManager.ts
@@ -20,6 +20,7 @@ export interface KeyBindingMap {
 }
 
 const MAX_DELAY_BETWEEN_KEY_CHORD = 250 /* milliseconds */
+const MAX_CHORD_SIZE = 4
 
 import { KeyboardResolver } from "./../Input/Keyboard/KeyboardResolver"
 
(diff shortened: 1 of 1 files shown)


jeffreydvp:
The performance increase with just this small change is quite astounding!

bryphe:
@jeffreydvp - thanks for confirming the fix!

> The performance increase with just this small change is quite astounding!

Yep, fortunately or unfortunately, there are likely lots of other small optimizations along the pipeline that can improve the performance like this - there is still _a lot_ of room for performance improvement. Thanks a lot for the trace; it's very helpful in tracking these issues down!

PR #2210, opened 2018-05-15
Fix #2205, enable sidebar.width config setting

### LINKED ISSUE #2205: sidebar.width is ignored
In the config, there's a default value of:
```
"sidebar.width": "50px"
```
Any attempt to change this value is ignored.  I tried searching the codebase for its use, and it isn't referenced anywhere.

Can we please add support for this option? That sidebar is too small for my preference.

keforbes:
Fixes #2205.  The `DefaultConfiguration.ts` had an entry for `"sidebar.width": "200px"` but any attempt to modify the value was ignored.  This is because the value was actually hard-coded to be `200px`, ignoring the config completely.  I changed it to pull that value from the config instead.  Also, @badosu suggested we change the units from `px` to `em` to handle high-resolution displays.

The current state of the Pull Request is functional; it behaves exactly as I'd expect.  However, it doesn't automatically update when saving the config.  So you can't really play with values and see which size works best for you.  Instead, you have to close and re-open Oni each time to see the changes.  I'm putting `[WIP]` on the Pull Request because I'd like to fix the dynamic updating before merging in.

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -376,7 +376,7 @@ const BaseConfiguration: IConfigurationValues = {
 
     "sidebar.enabled": true,
     "sidebar.default.open": true,
-    "sidebar.width": "50px",
+    "sidebar.width": "15em",
 
     "sidebar.marks.enabled": false,
     "sidebar.plugins.enabled": false,
--- a/browser/src/Services/Sidebar/SidebarContentSplit.tsx
+++ b/browser/src/Services/Sidebar/SidebarContentSplit.tsx
@@ -8,7 +8,6 @@ import styled, { keyframes } from "styled-components"
 
 import { Event, IDisposable, IEvent } from "oni-types"
 
-import { configuration } from "../Configuration"
 import { enableMouse, withProps } from "./../../UI/components/common"
 
 import { ISidebarEntry, ISidebarState, SidebarManager, SidebarPane } from "./SidebarStore"
@@ -65,6 +64,7 @@ export class SidebarContentSplit {
 
 export interface ISidebarContentViewProps extends ISidebarContentContainerProps {
     activeEntry: ISidebarEntry
+    width: string
 }
 
 export interface ISidebarContentContainerProps {
@@ -174,10 +174,7 @@ export class SidebarContentView extends React.PureComponent<
         const header = activeEntry && activeEntry.pane ? activeEntry.pane.title : null
 
         return (
-            <SidebarContentWrapper
-                className="sidebar-content"
-                width={configuration.getValue("sidebar.width")}
-            >
+            <SidebarContentWrapper className="sidebar-content" width={this.props.width}>
                 <SidebarHeaderView hasFocus={this.state.active} headerName={header} />
                 <SidebarInnerPaneWrapper key={activeEntry.id}>
                     {activeEntry.pane.render()}
@@ -200,6 +197,7 @@ export const mapStateToProps = (
     return {
         ...containerProps,
         activeEntry,
+        width: state.width,
     }
 }
 
--- a/browser/src/Services/Sidebar/SidebarContentSplit.tsx
+++ b/browser/src/Services/Sidebar/SidebarContentSplit.tsx
@@ -8,6 +8,7 @@ import styled, { keyframes } from "styled-components"
 
 import { Event, IDisposable, IEvent } from "oni-types"
 
+import { configuration } from "../Configuration"
 import { enableMouse, withProps } from "./../../UI/components/common"
 
 import { ISidebarEntry, ISidebarState, SidebarManager, SidebarPane } from "./SidebarStore"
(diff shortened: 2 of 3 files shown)


Akin909:
@keforbes sorry I wasn't very clear as to any specifics in the issue, what I'd meant was passing the config option directly will work the reason I advised against it is ideally I think we would rather not pass on api methods directly to `styled-components` which is the `css-in-js` library we use to render that component as any change in the api would immediately break this functionality.

The way it's currently being passed here though also means that any other attempt to programmatically change the size of the sidebar won't work since it will only be taking values from the config.

So something like `Oni.input.bind(<c-+>, "widen.sidebar")` wouldn't be an option,
I was thinking that if the value was passed in from a `redux` store then it could be changed over time by commands.

What this would involve is if the default value of [`SidebarState`](https://github.com/onivim/oni/pull/2210/files#diff-bf052c69d0fc1c237dce47e61607d5a6L192) included a value for width which could be pulled out using the config as you've done then that could later be altered.

example:
```ts
// SidebarStore.ts

const DefaultState = {
	// Stuff thats already there
	width: config.getValue("sidebar.width")
}

// SidebarContentSplit

const mapStateToProps = (state, otherProps) => {
	return {
		// Stuff thats already there
		width: state.width
}

// finally

	<SidebarContentWrapper width={this.props.width}>
```

Let me know if that makes sense and I can expand on any bit (I've left out all types that will need to be assigned), the reason I suggested this approach is that I plan on having a look at programmatic resizing using bindings in a little bit and was thinking being able to change the value of state would be a much simpler and robust way to achieve that if it was already in the store.

Of course when we initially talked about resizing @bryphe you discussed having the windowManager control the resizing so this might go against what you intended.

parkerault:
FWIW I believe it's CSS pixels, which will take screen scaling into account, so 200px at a @2x scaling should be 400 real world pixels. Being able to scale the sidebar width with the font size is not a bad approach though. I think ideally we should just take a string value so the user can decide which units they would prefer to use, so they could use `vw` viewport relative width or something similar instead. Just my two cents!

badosu:
Would be the case if electron supported scaling automatically, which is not
happening for me in a 4k display on Arch.

I have to increase all defaults in px to my values otherwise it's unusable.

I think there's a good case to have a default that works for all cases if
it's so straightforward.

Anyway, just having the ability to configure the width makes the issue
fixable.

Em Qua, 16 de mai de 2018 16:54, parkerault <notifications@github.com>
escreveu:

keforbes:
@Akin909, thanks for that implementation!  That's exactly what I was asking for in the issue, I knew I wasn't understanding you completely.

I've implemented everything you mentioned and while it's technically "working", it's only using the value from `DefaultConfiguration.ts`.  I think the store is being created before the config file has been processed so the config is just returning the default value.  I'm guessing I need to subscribe to the `onConfigurationChanged` event to be notified when the config file is done being processed and then update the store accordingly.  I'm not sure how to do that though.  Here's what I was thinking:
```ts
/// class SidebarManager
   constructor(private _windowManager: WindowManager = null) {
        this._store = createStore()

        configuration.onConfigurationChanged.subscribe(val => {
            if (typeof val["sidebar.width"] === "string") {
                // SOMEHOW UPDATE STORE
            }
        })
```
but so far all of my ignorant attempts to update the store have had no effect.  Or maybe I also need to poke the sidebar into re-rendering with this new value; I don't know how to do that either.

Sorry for the ignorant questions, I'm new to React.

keforbes:
@Akin909, sorry, nevermind.  I found the docs for Redux stores so you don't need to explain that part to me.  I didn't realize `dispatch` was how you set values.  Changing the `config.ts` now dynamically updates Oni but it's still launching with the default value.  So I still need to figure out how to poke it when it first launches but everything else is now behaving correctly (and hopefully implemented correctly too).

keforbes:
Ok, I got everything working.  The issue with the initial value was due to `DefaultSidebarState` being set as a `const` outside of any class so that code was running at class-load time.  I just added a line to explicitly set the width inside the `constructor` and everything behaves correctly now.

@Akin909, I still wouldn't mind one last code review from you since you know what you're doing more than I do. :smile:

keforbes:
Thanks @Akin909 !  You know, one consequence of listening for config changes and updating the store accordingly means you could implement `Oni.input.bind(<c-+>, "widen.sidebar")` by just calling `configuration.setValue()` and the sidebar should detect the change and update.  Of course, we'd probably want to write the current value out to the config file when Oni shuts down so we re-use the last sidebar width during the next Oni launch.

Also, as I was thinking about your key bindings... The `width` property we're using is just a CSS string so it can contain any supported unit.  There's already been some discussion in this pull request about whether it should use `px` or `em` or `vw`.  When you implement your keybinding, I wonder if it'd be best to split the number-value from the unit-value, increase the number by `1`, then re-apply the unit-value.  So if the user was using `em` or `%` or `vw` we'd increase the width by one of those units.  I guess I'm just saying I don't want to hardcode the key binding to automatically switch to `px` to make the math easy.  Just a thought; it might add unnecessary complexity though.

CrossR:
>Of course, we'd probably want to write the current value out to the config file when Oni shuts down so we re-use the last sidebar width during the next Oni launch.

`configuration.setValue()` does actually accept that as a second parameter of if it should save the value. This functionality needs adding to the API, since Oni will complain about getting 2 values not 1, but the functionality is there its just not exposed in the API.

I have something similar in my config for font size changes, with `<c-+>` bound to some simple functions.

https://github.com/CrossR/dotfiles/blob/5f9e5c0169c08b2ef03cd830bb1a093a3e95713a/oni/config.tsx#L184-L216

Just incase anybody would like some similar functionality now!

PR #2212, opened 2018-05-15
Moved Issue template to the correct folder. 

CrossR:
When I was opening a new issue, I realised the new template hadn't been applied, so I've merged it with the existing one (which I had forgotten about since its just a hidden comment!).


bryphe:
Awesome, thanks @CrossR ! I'll bring this in right now (AppVeyor failure is not related)

PR #2216, opened 2018-05-17
Bugfix: oni.input.unbind()

ackey:
If you pass array to oni.input.unbind(), the function hangs up.
This fixes it.

### DIFF
--- a/browser/src/Services/InputManager.ts
+++ b/browser/src/Services/InputManager.ts
@@ -103,7 +103,7 @@ export class InputManager implements Oni.Input.InputManager {
 
     public unbind(keyChord: string | string[]) {
         if (Array.isArray(keyChord)) {
-            keyChord.forEach(key => this.unbind(keyChord))
+            keyChord.forEach(key => this.unbind(key))
             return
         }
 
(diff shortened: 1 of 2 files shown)


bryphe:
Awesome, thanks for catching this and fixing it @ackey !

Would you mind adding a unit test to help protect this case from regressions? We have some unit tests set up here already:
https://github.com/onivim/oni/blob/5f5c2382712893a21e12754818306a9972fac1aa/browser/test/Input/InputManagerTests.ts#L38

So just added a quick one that validates this unbind-array case would be great. Perhaps something like this:
```
it("can unbind an array of keys", () => {
const im = new InputManager()
im.bind(["a", "b"], "test.command")
im.unbind(["a", "b"])

const boundKeys = im.getBoundKeys("test.command")
assert.deepEqual(boundKeys, [], "Validate no bound keys are returned")
})
```

I think that would exercise this issue, and keep it from being regressed in the future

ackey:
@bryphe OK, I will 🙂

bryphe:
Change looks great - thanks for adding the test, @ackey ! 💯

bryphe:
Bringing this in now - thanks @ackey ! 👍

PR #2224, opened 2018-05-19
WebGLRenderer: Lazily (re)initialize the glyph buffer with sufficient size

Cryza:
@CrossR reported an issue with the `webgl` rendering strategy on his huge screen: When opening a file that completely filled the screen with characters, the renderer went blank.

*Issue:*
Up until now, we initialized the buffer where we put the glyph information for the GPU with a fixed size of `1 << 14`, i.e. `2^14` or 16384 glyphs.
With a sufficiently large screen, this limit could be exceeded. I was able to reproduce it with a viewport of 257x66 code cells (16191 cells + some more for the line numbers).

*Fix:*
Before each render, we check if the buffer length is sufficient for containing a screen that is completely filled with characters and re-initialize it with a sufficient length if not.

### DIFF
--- a/browser/src/Renderer/WebGL/WebGLSolidRenderer.ts
+++ b/browser/src/Renderer/WebGL/WebGLSolidRenderer.ts
@@ -6,8 +6,6 @@ import {
     createUnitQuadVerticesBuffer,
 } from "./WebGLUtilities"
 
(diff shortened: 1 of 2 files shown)


CrossR:
This seems to have fixed it for me now, I am able to fill the entire editor and not lose rendering.

PR #2226, opened 2018-05-19
bugfix/ fix parsing issues in ui-test tabs

Akin909:
Incredibly unusually somehow this file in `master` has a parsing error but this seems neither to have tripped up the tests nor failed to build in `ts` in merging this file into #1308, trying to commit fails as prettier suddenly *realises* its unable to parse this file.

It seems to have been failing silently previously which allowed for the semicolons (aka our prettier commit hook would ordinarily have removed them,
Also on `line 188` a parens was missing which should have thrown an error not entirely sure how/why it didn't till I tried merging it in. Also added some types as the `tss` complaining as the jest mocking unfortunately doesnt automatically replace the types for imported modules which are subsequently mocked with `jest.mock`

### DIFF
--- a/ui-tests/Tabs.test.tsx
+++ b/ui-tests/Tabs.test.tsx
@@ -9,9 +9,12 @@ import { Sneakable } from "../browser/src/UI/components/Sneakable"
 
 import { Tab, Tabs } from "../browser/src/UI/components/Tabs"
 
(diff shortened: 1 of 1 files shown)


bryphe:
Strange - thanks for fixing it, @Akin909 !

Issue #2213, opened 2018-05-16
`d` in file tree deletes the whole project


abs-zero:
I accidentally pressed `d` on the container project on the file tree and now the entire project is deleted. Does oni keep deleted files anywhere so I can recover from there?

abs-zero:
I think it would be a good idea to use trash instead of hard delete with module like `https://github.com/sindresorhus/trash`

Akin909:
@abs-zero we actually gone hard delete with the d command, what it actually does is move the folder to the operating systems temp dir under a folder called `oni_backup` and pressing `u` should undo it but if the session is over then undoablity is lost but as for the folder it should be in your tempdir.

I didn't special case  the container folder since I was sure what ppl's use cases would be but given its the root maybe a notification to get confirmation before deleting in that case would be best.

abs-zero:
Oh.. But why does it clear `oni_backup` after I quit oni?

badosu:
@Akin909 Might be a good idea to use `oni_backup_<timestamp>` for the backup folder name as two reversible operations might corrupt the backup, unless you already have a system for that case.

Akin909:
Reading through the docs for `trash` @abs-zero I think that option would unfortunately mean that delete actions can't be undone by oni because of the way the trash dir works cross platform so might include it in this PR with the caveat that it breaks undo functionality and leave the default as the oni strategy so users have access to undo as default.

Re. the timestamps I actually think that possibly why I left that line in there @badosu as brute force way of preventing conflicts the issue with using that strategy would be that we'd have to persist the backup dir with time stamps and check those which I think would result in a much bigger pr as it doesn't do anything close to that at the moment

As of right now its a brute force strategy to overwrite duplicates, I could look at time stamping down the line but I'd like to get a fix for this so we aren't deleting directories irreversibly as that could be catastrophic if not version controlled etc.

Also when you say reversible operations corrupting what did you have in mind there (I'm concerned re. situations I hadn't considered)

dkns:
I think we should add confirmation box before deleting anything. I can see situation where user thinks they navigated out of file browser, presses d and is confused/angry as to what just happened.

For reference vscode asks for confirmation:
![image](https://user-images.githubusercontent.com/4007499/40105515-71e5081a-58f3-11e8-80d6-de430ec1d078.png)

Akin909:
[...]

PR #2227, opened 2018-05-19
Bugfix/ remove call to clear the oni-backup dir on oni startup

Akin909:
A partial fix for #2213 and is a small part of #2214 but is a crucial fix that will stop the backup directory saved to the `tmp` dir from being deleted every time `oni` starts up (which was defeated the point of saving to the tmp dir where a user could potentially recover a any deleted files)


bryphe:
Makes sense to me - thanks @Akin909 !

PR #2228, opened 2018-05-19
Feature/ upgrade prettier and pretty-quick

Akin909:
Upgrade prettier and the pretty quick package which runs it as a pre-commit hook the main reason for this being that prettier incrementally improve their parsing and formatting strategy and crucially for a `typescript` project they add support for new ts syntax that prettier would otherwise choke on without upgrades


bryphe:
Cool, thanks @Akin909 ! 👍

PR #2229, opened 2018-05-19
Feature/upgrade react to 16.3

Akin909:
Whilst looking at packages we could upgrade if safe (tested locally and no detectable change nor warnings that werent already there), `React` v16.3 brings a lot of cool new features like `<React.Fragment>` (my personal favourite) which is amazing for avoiding pointless wrapper divs bogging down the DOM and lifecycle methods are being deprecated in favor of new ones.

Unfortunately the code mod to rename unsafe methods doesn't work in `TS` but this release comes with a strict mode component that can wrap the root component and will log a warning to the console for use of the methods to be deprecated and unexpected side effects in render all of which I think will help us get oni ready for the planned changes. I havent done the wrapping since I imagine its not for everyone but I think it might give us a push whenever we encounter these warnings to refactor something problematic


bryphe:
Looks like the `debounceTime` fix will get merged out since it's already in master, but otherwise this looks good to me 👍

Issue #1240, opened 2018-01-06
The key binding mechanism does not work


saibing:
```js
const activate = (oni) => {
    console.log("config activated")

    // Input 
    //
    // Add input bindings here:
    //
    oni.input.bind("<c-enter>", () => console.log("Control+Enter was pressed"))
    oni.input.bind("<c-s-r>", "language.findAllReferences")

    //
    // Or remove the default bindings here by uncommenting the below line:
    //
    // oni.input.unbind("<c-p>")
}
```

I input <c-enter> , there is no any output on developer console
I input <c-s-r>, there is no any response.

mchalkley:
The way I'm doing this now, per a suggestion from the gittr discussion page, is in the Oni\resources\app\vim\core\oni-core-interop\plugin\init.vim file, using the normal vim remap commands...

You may also need add a line into your config.js file (via File>Preferences) to tell Oni not to use its default config, as shown below:

const activate = (oni) => {
   // access the Oni plugin API here
   
   // for example, unbind the default `<c-p>` action:
   // oni.input.unbind("<c-p>")

   // or bind a new action:
    
   //oni.input.bind("<a-f4>", "oni.quit")
};

module.exports = {
    activate,
    // change configuration values here:
    "oni.useDefaultConfig": false
    //"oni.loadInitVim": true,
    //"editor.fontSize": "14px",
    //"editor.fontFamily": "Monaco",
    //"editor.completions.enabled": true
}

I hope this helps...

Mark

CrossR:
Other thing to check is that you have activate being called in your `module.export`, ie

```
module.exports = {
    activate,
    ...
}

**Comment by saibing** [comment]
@mchalkley 
Thank you. It works well.

@CrossR 
I have already call activate in config.js because that 'config activated' message has been output to the console
```js
const activate = (oni) => {
    console.log("config activated")

    // Input 
    //
    // Add input bindings here:
    //
    oni.input.bind("<c-enter>", () => console.log("Control+Enter was pressed"))
    oni.input.bind("<c-s-r>", "language.findAllReferences")

    //
    // Or remove the default bindings here by uncommenting the below line:
    //
    // oni.input.unbind("<c-p>")
}
```

saibing:
@bryphe 
Could you add a defualt sortcut key for language.findAllReferences command.

It is not included in the [shortcuts](https://github.com/onivim/oni/wiki/Shortcuts) list.

CrossR:
[...]

PR #2236, opened 2018-05-22
Feature/add bindings to change sidebar width

Akin909:
This PR adds the ability to resize oni's sidebar using key binding (defaults are: mac - `<m-s-+>` to increase `<m-s-_>` to decrease, and windows - `<c-s-_>` to decrease and `<c-s-+>` to increase.
If the user input is invalid e.g. `stringpx` in their configuration the default configuration value is returned atm `15em`, if the size is too small aka less than `1em` the last value is returned aka no change in size occurs. If the units a user selects are inappropriate a default unit of `em` is used instead.

Edit: I know there was some discussion about potentially persisting this value so it is used on restart however I've decided to leave that out of this PR to limit its complexity.

### GIF:

resize


Akin909:
Seem to have an issue with finding the cached binaries on AppVeyor hence the failing test restarting doesn't help have you run into this @CrossR?

CrossR:
I've run into it, but a restart allowed it to succeed for me...
I'm hoping next week to have a better look into other solutions/why its happening.

badosu:
@Akin909 I wonder if we'll have to write the configuration if the user drags the sidebar handle when it's implemented.

Intuitively that should be temporary, so not writing the configuration, but then we can't use the same command.

Thoughts?

Akin909:
Running into the following error on the ios travis build -
`The command "npm run check-cached-binaries" exited with 1.`

browser/src/Input/KeyBindings.ts
```diff
@@ -60,6 +62,8 @@ export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configurati
         input.bind("<m-r>", "browser.reload")
     } else {
         input.bind("<a-f4>", "oni.quit")
+        input.bind("<c-s-_>", "sidebar.decreaseWidth")
```
badosu: Just a reminder that in many programs like terminals or other editors `<C-S-_,+>` is used to increase/decrease the font size, so I leave to you to ponder if we should adhere to this convention or not.

Akin909: @badosu I was thinking that control/cmd +/- was what most programs used to increased/decrease font size? if the convention is to use c-s-_/+ I can change these bindings to `c-+`/`c-_` ?

badosu: @Akin909 In my keyboard the key `=` is the main one and `+` is the one in the shift position.

So, `<C-S-+>` would be not possible for me unless it was the same as `<C-S-=>`. Maybe I am misunderstanding anyway.

If someone else does not chime in with an insight please disregard my comment.

CrossR: I hit an unrelated issue with these keybinds.

On Windows, the library is resolving the keys as `<s-c-+>` and `<s-c-_>`, which means the default binds don't work. This is a flaw in the library (I think) and I found it again over in #1240.

Until thats sorted, its probably worth just swapping the bindings from `c-s-` to `s-c-` like the `commands.show` bind etc.

And @badosu you are correct that `=` is the default for that key, but since the bind has `shift` in it, it is resolved as `+`. So you are holding the `Ctrl`, `Shift` and `=/+` keys, but the library will resolve it to `c-s-+`.

badosu:
@Ryan C Thanks! :+1:

2018-06-10 20:59 GMT-03:00 Ryan C <notifications@github.com>:

Akin909:
@CrossR thanks for looking at this 👍, didn't realise about our issue with the keybinds, I've swapped these to be `s-c-_` and `s-c-+` lemme know how that fares on your end

PR #2244, opened 2018-05-24
Improve notification dimensions

### LINKED ISSUE #2233: Improve notifications appearance
Oni Version: 0.3.4
Neovim Version (Linux only): 0.2.2
Operating System: Arch

#### Describe your issue

Notification messages are hard to read, they pop up on the right top corner without any margin and have a square-ish dimension.

See:

screenshot from 2018-05-21 09-35-15

Btw, this is on a vertical split, so the lack of margin is actually a bigger issue than what's represented. This is probably due to a px sized margin instead of a scalable unit in a HiDPI display.

badosu:
Use relative units for positioning and also increase max-size of notifications.

Fixes #2233 

screenshot from 2018-05-24 07-41-06

### DIFF
--- a/browser/src/Services/Notifications/NotificationsView.tsx
+++ b/browser/src/Services/Notifications/NotificationsView.tsx
@@ -35,10 +35,10 @@ const Transition = (props: { children: React.ReactNode }) => {
 
 const NotificationsWrapper = styled.div`
     position: absolute;
-    top: 16px;
-    right: 16px;
+    top: 1em;
+    right: 1em;
     max-height: 90%;
-    max-width: 25rem;
+    max-width: 33vw;
     pointer-events: all;
     overflow: auto;
 


Akin909:
Just placing this comment here from the original issue, I think 50rem is unlikely to work for people with smaller units as thats likely to be enormous on laptops or 17-20inch is displays

> @badosu @keforbes we discussed the sizes of the notification when we initially implemented it and @CrossR had a try with the sizes as I think initially it was too rectangular so I'd be careful there as changes to fix one view port might make it too big on another it might be worth looking at `vw` instead of `rem` or maybe `em`

Akin909:
Just tried `50rem` on a 15inch macbook and it is enormous unfortunately, tried with `33vw` and that seems to scale quite well can you see if that works for you on your screen @badosu ?

badosu:
Thanks @Akin909, let me try that!

badosu:
@Akin909 Updated to 33vw, looks good on my end, I updated the screenshot.

keforbes:
I can't find an error message long enough to test out the current max of `33vw` but the errors I could generate looked fine on a 1920x1200 monitor.

badosu:
@keforbes Try editing `config.{tsx,js}` to have a parse error.

keforbes:
Yeah, that's exactly what I did.  Unfortunately, all the parse errors I generated were still on a single line.  I was hoping to have an error long enough that the notification has to wrap to two lines.  That would tell me how big the new max is on my screen.

CrossR:
You can create a notification through the API if that helps @keforbes, ie in the Dev Console:

```
let test_notification = Oni.notifications.createItem()
test_notification.setContents("Title", "Body")
test_notification.show()
```

Stick a bunch of text in the title/body/both, or use a string template etc and see how it looks I guess.

keforbes:
Perfect, thanks @CrossR!  I gave it a paragraph of text and I'm good with the new max size :+1:

badosu:
I've rebuilt the CI multiple times and it's failing :-(

Akin909:
@badosu we seem to be having some issues with the Ci as its happening across the board

Akin909:
@badosu I cleared the cache for your test and re ran it which seems to have made it mergeable I'll leave the honours to you just to make sure your done with it

Issue #2230, opened 2018-05-19
Autocomplete case insensitivity


bschulte:
I was wondering if there is any ability for autocomplete to be case insensitive. I looked through the documentation/settings and didn't see anything related to that. It would be nice to be able to type something like "usct" and have the option to autocomplete "userController" come up.

CrossR:
Can I check what version of Oni you are running? The more recent versions should already do that.

By default the quick open menus and similar should use a "smart" case sensitivity strategy. That is, if the string is all lower case, the search is case insensitive, but if the string contains uppercase characters, the search is then instead case sensitive.

This can be changed to always to always/never be case senstive by setting `    "menu.caseSensitive": false,` or `true` depending on the required setting. Though you are right, I can't see that option in the config, so we should probably tidy that up...

With a quick test on the latest release (we added a new sorting system in 0.3.4 that fixed a bunch of issues with the search) the situation you described seems to work:

![image](https://user-images.githubusercontent.com/10038688/40271871-ca808954-5b9b-11e8-8b68-5ea136ba4047.png)

(Obviously a fake example that I just made a random file, but still it managed to find it despite it being nested fairly deep in the project)

bschulte:
Thanks for the reply! I'm on Mac 10.13.3 with Oni version 0.3.4. The smart search functionality indeed works for me for the fuzzy file searching, but I'm specifically referring to auto complete within code.

For example this is the behavior I get with and without using a capital letter:

![image](https://user-images.githubusercontent.com/473161/40271925-8b3273bc-5b72-11e8-906e-9587f092c8fb.png)

![image](https://user-images.githubusercontent.com/473161/40271931-a575a8fc-5b72-11e8-9fc7-c03f7ed21eee.png)

CrossR:
Ah! That makes sense then.

Currently no, we don't have a config option for that, but I think it makes sense to have a similar one to the menu, ie `smart` `true` `false`.

mnrce:
Will #2249 be merged in the next stable release?

PR #2249, opened 2018-05-26
Allow for completions filter to be case insensitive

bschulte:
I created an issue about this [#2230](https://github.com/onivim/oni/issues/2230). This change will just allow users to change the completion filter to not be case sensitive.

The setting "editor.completions.caseSensitive" can be set to `true` or `false` (`true` by default) and will tell the completion filter to be case sensitive or not.

### DIFF
--- a/browser/src/Services/Configuration/IConfigurationValues.ts
+++ b/browser/src/Services/Configuration/IConfigurationValues.ts
@@ -150,6 +150,9 @@ export interface IConfigurationValues {
     // a custom init.vim, as that may cause problematic behavior
     "editor.completions.mode": string
 
+    // Decide whether or not the completion matching should be case sensitive
+    "editor.completions.caseSensitive": boolean
+
     // If true (default), ligatures are enabled
     "editor.fontLigatures": boolean
     "editor.fontSize": string
--- a/browser/src/Services/Completion/CompletionSelectors.ts
+++ b/browser/src/Services/Completion/CompletionSelectors.ts
@@ -4,6 +4,7 @@
  * Selectors are functions that take a state and derive a value from it.
  */
 
+import { configuration } from "./../Configuration"
 import { ICompletionState } from "./CompletionState"
 
 import * as types from "vscode-languageserver-types"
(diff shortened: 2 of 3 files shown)


browser/src/Services/Configuration/IConfigurationValues.ts
```diff
@@ -150,6 +150,9 @@ export interface IConfigurationValues {
     // a custom init.vim, as that may cause problematic behavior
     "editor.completions.mode": string
 
+    // Decide whether or not the completion matching should be case sensitive
+    "editor.completions.caseSensitive": boolean
```
CrossR: Thanks for the PR @bschulte!

It may make sense to make this like the other case sensitive option, ie `string | boolean` such that it can be `smart`/`true`/`false`.

Then the check becomes something like this:

https://github.com/onivim/oni/blob/262098fefd7e82bb69b0b9b68ff66ba6af3d18d8/browser/src/Services/Menu/MenuFilter.ts#L25-L39

That should also let you simplify the regex check like we have over here:

https://github.com/onivim/oni/blob/afab2beac8c9374db9996118d1892e2bf074275d/browser/src/Services/QuickOpen/RegExFilter.ts#L34-L38

That lets us use the same regex for all three cases then.

bschulte:
Thanks for the feedback! I've added a commit to allow for the config value of true/false/smart. I also tried to follow the same pattern that you linked to elsewhere in the code.

I did have to also perform the check if the search should be case sensitive when it comes to creating the RegEx, otherwise it wouldn't work. From the code elsewhere in the project, it looks like you didn't have to do that so maybe I'm missing something silly?

bschulte:
OK, I modified things to conform to the already exists patterns a bit more. 

One thing I noticed in testing was that the underlining wasn't correct when the search was case insensitive (which makes sense). I was having a bit of trouble tracking down where that code was to handle that, would you mind pointing me in the right direction?

bryphe:
Thanks for making the changes, @bschulte ! Changes overall look good to me.

It looks like there is a test failure with the `AutoCompletionTest-HTML` integration test:
https://github.com/onivim/oni/blob/d18087914c7aad1f3c65161145ff2150d194f41a/test/ci/AutoCompletionTest-HTML.ts#L18

That test sends these keys:
`<body s`

And waits for completion to appear (it expects that `style` should be in the completion). It looks like perhaps that behavior changed with this - it might be that `style` isn't one of the top-ten completions we show with case-insensitive. A quick fix for this would be to add a configuration setting, like:
https://github.com/onivim/oni/blob/d18087914c7aad1f3c65161145ff2150d194f41a/test/ci/AutoClosingPairsTest.ts#L63

```
export const settings = {
   config: {
        "editor.completions.caseSensitive": true,
    }
}
```
So for now, we could just set that configuration setting specific for that test. Alternatively, we could pick a new item to test to validate.

Speaking of tests - it'd be great to have a unit test documenting this behavior here:
https://github.com/onivim/oni/blob/d18087914c7aad1f3c65161145ff2150d194f41a/browser/test/Services/Completion/CompletionTests.ts#L173

We have a bunch of tests exercising various parts of completion there - it'd be cool to have a case that exercises the new behavior as well. That will help protect it from regressing. Let me know if you have questions about adding a test there.

Thanks again for taking this on @bschulte  - I know users have asked for this before, will be a great addition to  our completion story 👍

PR #2253, opened 2018-05-28
Update contributors list

badosu:


bryphe:
Nice catch! Thanks for updating this list, @badosu 💯

PR #2255, opened 2018-05-28
Use same colors as other status line items for prettier plugin

### LINKED ISSUE #2157: Prettier plugin statusline item highlighted weirdly
Using the [neodark](https://github.com/KeitaNakamura/neodark.vim) theme with dark background.

screenshot from 2018-04-30 20-56-22

It used to work without problems before 0.3.3

badosu:
This is quick hack to fix this glaring issue.

There are some better approaches that can be made, but require some discussion before proceeding.

- Use a general higher order component for status bar items that shares the same style
- Allow each status line component to define it's own style, but define or use a theme color
- Continue as it is

Fixes #2157

### DIFF
--- a/extensions/oni-plugin-prettier/index.js
+++ b/extensions/oni-plugin-prettier/index.js
@@ -147,8 +147,6 @@ const activate = async Oni => {
 function createPrettierComponent(Oni, onClick) {
     const { React } = Oni.dependencies
 
-    const background = Oni.colors.getColor("highlight.mode.normal.background")
-    const foreground = Oni.colors.getColor("highlight.mode.normal.foreground")
     const style = {
         width: "100%",
         height: "100%",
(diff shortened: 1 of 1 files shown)


Akin909:
@badosu how about potentially just removing the color entirely?, I was thinking that over time if you imagine oni has say 10 status item and they are all colored slightly differently or even the exact same but with a bit of padding between each the status bar wouldn't have the best appearance as opposed to `vscode` or `atom` all of which just have the text and an icon and some padding so things aren't too close together

badosu:
@Akin909 Would that be the same as to define the color on the parent component of the statusbar items?

If so, I can try that, the issue with hardcoded color vs theming will persist though.

Akin909:
@badosu just tried a quick test and I think all you would need to do is remove the background-color from the prettier status item aka the one that has a class of `prettier` I know there are a few components in there and you get:
screen shot 2018-06-13 at 16 39 54

Akin909:
@badosu I've actually removed the background color from the plugin so this should resolve the issue this PR was aimed at let me know if the results work for you re. #2157

Issue #2238, opened 2018-05-23
Plugin Development: --plugin-develop flag should given an error message if no plugin specified


bryphe:
Oni provides a `--plugin-develop` flag that expects to be followed by a path to a plugin. If that doesn't happen, or we can't find the plugin, Oni will be unresponsive.

We should improve this experience by:
- If no plugin is specified, show an error notification the effect of "--plugin-develop must be followed by a plugin path", and ignore it.
- If a plugin is specified, but cannot be found, show an error notification to the effect of "Could not find plugin: ${path}"

nwneisen:
I started taking a look at this and ended up making a PR with some changes.
#2266

nwneisen:
Hey @bryphe, i hadn't heard from you so I thought I'd check on this task.

PR #2266, opened 2018-05-30
#2238 - Change handling of --plugin-develop argument

nwneisen:
Add additional handling for the --plugin-develop argument so that no
path and invalid path cases are handled.

### DIFF
--- a/browser/src/App.ts
+++ b/browser/src/App.ts
@@ -5,6 +5,7 @@
  */
 
 import { ipcRenderer } from "electron"
+import * as fs from "fs"
 import * as minimist from "minimist"
 import * as path from "path"
 
@@ -135,9 +136,15 @@ export const start = async (args: string[]): Promise<void> => {
 
     const developmentPlugin = parsedArgs["plugin-develop"]
 
-    if (developmentPlugin) {
+    if (typeof developmentPlugin === "string") {
         Log.info("Registering development plugin: " + developmentPlugin)
-        pluginManager.addDevelopmentPlugin(developmentPlugin)
+        if (fs.existsSync(developmentPlugin)) {
+            pluginManager.addDevelopmentPlugin(developmentPlugin)
+        } else {
+            Log.info("Could not find plugin: " + developmentPlugin)
+        }
+    } else if (typeof developmentPlugin === "boolean") {
+        Log.info("--plugin-develop must be followed by a plugin path")
     }
 
     Performance.startMeasure("Oni.Start.Plugins.Discover")
--- a/browser/src/App.ts
+++ b/browser/src/App.ts
@@ -135,16 +135,25 @@ export const start = async (args: string[]): Promise<void> => {
     const pluginManager = PluginManager.getInstance()
 
     const developmentPlugin = parsedArgs["plugin-develop"]
+    let developmentPluginError: { title: string; errorText: string }
 
     if (typeof developmentPlugin === "string") {
         Log.info("Registering development plugin: " + developmentPlugin)
         if (fs.existsSync(developmentPlugin)) {
             pluginManager.addDevelopmentPlugin(developmentPlugin)
         } else {
-            Log.info("Could not find plugin: " + developmentPlugin)
+            developmentPluginError = {
+                title: "Error parsing arguments",
+                errorText: "Could not find plugin: " + developmentPlugin,
+            }
+            Log.warn(developmentPluginError.errorText)
         }
     } else if (typeof developmentPlugin === "boolean") {
-        Log.info("--plugin-develop must be followed by a plugin path")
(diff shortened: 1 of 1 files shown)


nwneisen:
I had a couple questions while working on this PR:

1. Is checking that a valid path is given the only criteria that should be used for a plugin?  I ran Oni with a valid path that was an invalid plugin and it was still responsive but it seems like more should be checked.
2. Where does Log.info write to and is this the correct location for these messages?  I assumed it was going to a log but I couldn't find one anywhere.  I also wondered if we wanted these to be wrote to the console when Oni is launched.

Thanks fr you feedback!

bryphe:
Hey @nwneisen - thanks a lot for the PR! 👍 Sorry for the delay in my response, been short on time and sporadic on PRs as a result.

> Is checking that a valid path is given the only criteria that should be used for a plugin? I ran Oni with a valid path that was an invalid plugin and it was still responsive but it seems like more should be checked.

I think this is reasonable for now, at least at this layer in the stack. There are other things that could go wrong, but I think it makes sense for `PluginManager` to handle those (like, a `package.json` is present, but can't be parsed). Definitely  much better than what we were doing before, which was nothing 😄 

> Where does Log.info write to and is this the correct location for these messages? I assumed it was going to a log but I couldn't find one anywhere. I also wondered if we wanted these to be wrote to the console when Oni is launched.

`Log.info` just writes to the console in the developer tools. It doesn't persist to a file at the moment (although it might make sense to add that, using a library like `winston`).  Logging these to the console sounds good, I would even bump up the new logs to `Log.warn` so they are more visible in the developer tools.

In addition to the logs, I'd like to pop up a notification in this case. If the user explicitly passed `--plugin-develop`, they probably wanted something to happen - so in either case, if we can't find the plugin, or they didn't specify in argument, it might be helpful to give an explicit heads up that there was an issue. There's an example of popping up an 'error' notification here:
https://github.com/onivim/oni/blob/e46c720fc82f1466347016bc0916a8d744c94793/browser/src/Services/UnhandledErrorMonitor.ts#L89
(And there's actually an example lower down in that file, too: https://github.com/nwneisen/oni/blob/c2ac24a14e30617e81e3339557f26befe1c2aa36/browser/src/App.ts#L210) The only tricky thing is that the `notifications` object gets initialized after we check for errors, so we might have to hold on to the errors and pop them up after the `notifications` gets created in line 206.

Thanks again @nwneisen for your help with this, really appreciate it! I'm looking forward to having a more robust plugin development story, and having this flow work is critical. 

Once we have this flag all set, I want to wire up https://github.com/onivim/oni-plugin-starter-kit so that we have a key binding (like `F5`) that will automatically start a new instance of Oni, and pass in the `--plugin-develop` flag for the plugin you have open. (Not needed for this PR, but just wanted to let you know what I was thinking as next steps)

nwneisen:
Thanks for the feedback @bryphe.  I figured you were keeping busy and tried to wait a while before commenting.  If I had more Typescript experience I'd be trying to help out with PRs but my background is in C/C++ and so this is still new territory to me.

I went ahead and changed the Log.info messages to Log.warn.  I'm not sure how I missed these previously but I'm seeing the logs now.  

I also added a popup to get the user's attention when --plugin-develop has an error.  I debated about keeping a more generic list of startup errors besides just plugins but I wasn't sure if there were any and it might be better in a separate PR.

> Once we have this flag all set, I want to wire up https://github.com/onivim/oni-plugin-starter-kit so that we have a key binding (like F5) that will automatically start a new instance of Oni, and pass in the --plugin-develop flag for the plugin you have open. (Not needed for this PR, but just wanted to let you know what I was thinking as next steps)
Good info to have.  I saw that plugins were the next big push in the development process and there are a number of issues for them.  I will be taking a look at them to see if any are up my alley.

bryphe:
Just tried this out @nwneisen , looks great! Thanks for incorporating the notifications 💯 

> If I had more Typescript experience I'd be trying to help out with PRs but my background is in C/C++ and so this is still new territory to me.

Ah ya a lot of differences from C++... seems like you picked it up fast though! Maybe down the road we could leverage your C++ skillset more - I'd like to at some point look at either native modules for some of the performance-critical paths, or webassembly - bit further out though..

> I also added a popup to get the user's attention when --plugin-develop has an error. I debated about keeping a more generic list of startup errors besides just plugins but I wasn't sure if there were any and it might be better in a separate PR.

Makes sense, as we start adding more errors we could pursue this.

> Good info to have. I saw that plugins were the next big push in the development process and there are a number of issues for them. I will be taking a look at them to see if any are up my alley.

Awesome, really appreciate the help! 👍

bryphe:
Will bring this in now - thanks again @nwneisen ! 💯

PR #2270, opened 2018-05-30
Fix #2185 - Remove call to set scrollbind

### LINKED ISSUE #2185: Multiple windows of same buffer shares scrolling position
If you start Vi/Vim/Neovim and open a buffer and then split the window `C-w` `v` or `C-w` `s` then both windows have separate scrolling positions, you can for instance have the top of a large file in one window and the bottom part in the other. But in Neovim if you scroll one of the windows, the other is scrolled as well. Is this intentional behaviour?

Akin909:
This PR removes the scroll binding of the reference buffer, fixes #2185. This is the only point in the app where we call `scrollbind` so removing it should resolve this issue.


Akin909:
@CrossR can you have a look at this its a very small change that I think is a real annoyance for a lot of users

Issue #2251, opened 2018-05-27
Oni does not correctly locate executables


jordan-arenstein:
Oni Version: 0.3.4
Neovim Version (Linux only):
Operating System: macOS 10.13.4

#### Describe your issue
The vim command `:executable` in oni does not match neovim's behaviour when using LaTeX commands latexmk and bibtex, in '/Library/TeX/texbin/'

#### Expected behaviour
Using neovim, `executable('latexmk')` returns 1

#### Actual behaviour
In Oni, `executable('latexmk')` returns 0. This breaks the package lervag/vimtex

#### Steps to reproduce
In Oni, type `:echo executable('latexmk')`

jordan-arenstein:
I have restored functionality by running `ln -s /Library/TeX/texbin/latexmk /usr/local/bin` and `ln -s /Library/TeX/texbin/bibtex /usr/local/bin`. Seems to be an issue with Oni's $PATH variable.

Akin909:
@jordan-arenstein if you run `:echo $PATH` in terminal nvim vs. in Oni (keep in mind that oni's path isn't quite the same) do you find that the Library dir is in your $PATH, as I know we add `/usr/local/bin` to the path oni uses but not the Library?

jordan-arenstein:
A few directories differ from the nvim and oni `$PATH`s, including the Library paths.

nvim $PATH: `/Library/Frameworks/Python.framework/Versions/3.5/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/TeX/texbin:/opt/X11/bin`

oni $PATH: `usr/bin:/usr/sbin/sbin:/usr/bin:/usr/local/bin`

PR #2274, opened 2018-05-31
Bugfix/ get true PATH variable

Akin909:
This PR should address #2251 and covers a little though not a big change in itself.

* It seems running `shellEnv.sync()` has been failing silently as `import('shell-env')` returns an object like `{ default: Function }` so calling sync on it fails but the way the try catch was structured led to a silent failure - ~I've switched the call to a simple require as `shell-env` is a tiny library and not sure we gain much by async importing it (guesstimation) also the behaviour between, webpack and typescript of async importing seems to be in flux a little so puts this import at risk.~ I've imported this as `shellSynv.default.sync` which is possibly brittle or maybe what is now the standard way that module.exports map to dynamic imports 🤷‍♂️. 

~* secondly I use a similar pattern as https://github.com/sindresorhus/fix-path to correct the path inconsistency on mac, not sure if this is definitely mac only but it is definitely an issue on mac - so on mac we set the path var to shell envs path
*NOTE:* since we had the heisenbug due to process constantly being reassigned the way that this works is as a **one off** call in the constructor of the shellEnvFetcher - this means that the packaged version of Oni has access to the users actual path variable.~

This fixes issues I noticed for example trying to run a language server that depends of environment variables like the PATH location of a users `goroot` or `cargo/bin` for go and rust respectively.

### Loose Ends
* A question I have is that we seem to dynamically pass in the path but the variable seems like something that can be static unless we are accounting for the edge case where a user modifies the additionPaths var in their config, as even in terminal if I update my `PATH` i would have to restart the shell.


browser/src/Plugins/Api/Process.ts
```diff
@@ -1,27 +1,46 @@
...
 
-    public async getEnvironmentVariables(): Promise<any> {
-        if (!this._shellEnv) {
-            this._shellEnv = await this._shellEnvPromise
-            return this._shellEnv.sync()
+    private _fixPath() {
+        // Set the PATH to a derived path on MacOS as this is not set correctly in electron
+        // the idea is based on https://github.com/sindresorhus/fix-path
+        if (Platform.isMac() && this._env.PATH) {
+            process.env.PATH = this._env.PATH
```
bryphe: Hmm, I don't think it makes sense for us to ever change `process.env.PATH` ourselves. From my understanding, all the cases where we need the full `path` are in _spawned processes_, and in those cases, we can send this 'augmented' path to them - which is what `mergeSpawnOptions` does today. 

To me, it seems like the root problem is that, in some cases, we aren't sending the `PATH` properly to the spawned processes. Setting `process.env.PATH` directly here is a janky workaround that is really risky and susceptible to heisenbugs, so I'd rather us understand the root cause and fix it.

In other words - why would we need to do this if Node gives us the ability to send `PATH` to spawned processes? It makes sense that we need to get the proper shell path, but for things like #2251, it should be sufficient to send it to the `nvim` binary. If it's not making it there, we should understand why that pipeline isn't working.

bryphe:
One thing that would be helpful here is to create a test case to document this behavior - it seems like we've had issues with this for some time. So having a test case can help us make sure we keep this behavior working, since respecting the `path` is important.

We already have the capability to pass in custom stuff with the `env` option in the test settings:
https://github.com/onivim/oni/blob/42840530d89f498f5bd55046573eaa448dfc16d5/test/common/runInProcTest.ts#L40

So what I'm thinking would be useful is a test case where we override `PATH`, and then run `echo $PATH` in neovim, and verify the output is as expected. This would be great to help make sure the functionality stays safe, regardless of the fix we take (and as a side benefit, it might help clarify the exact cases where we run into problems - so we can evaluate where in the pipe the fix needs to be)

bryphe:
> A question I have is that we seem to dynamically pass in the path but the variable seems like something that can be static unless we are accounting for the edge case where a user modifies the additionPaths var in their config, as even in terminal if I update my PATH i would have to restart the shell.

Yes, most applications don't expect / handle the environment variable to change. They get a set of environment variables when the process starts, and use those. But ya, we don't expect Oni to have that full context (which is why we need to ask the shell for the path), so then we can pass the full-shell path to the process when it starts. 

Oni itself shouldn't care about the PATH so much, since it's the language servers / neovim / etc that care about it - so we should make sure to send them the proper, expected path, but it shouldn't be necessary to change Oni's `process.env.PATH` at runtime - sending the `PATH` to the spawned processes should have the same effect.

browser/src/Plugins/Api/Process.ts
Akin909: @bryphe happy to leave this be and leave this change as a simple rectification of the the import error (which is unfortunately definitely failing in master), and remove the above changes.

Akin909:
@bryphe I've removed the changes as requested (leaving in the fix to the import which is definitely broken), regarding the test if I were to override the path in the test wouldn't that impact neovim's startup i.e. not sure if any particular variables are required that my overwriting might remove, also if any variables are added to the path maybe by neovim on startup the test could end up being flaky (I'm vaguely aware of neovim setting some env vars on startup)

PR #2277, opened 2018-06-01
Feature/ convert quickinfo to styled components

Akin909:
This targets the QuickInfo components and converts them to `styled-components` removes the `.less` file and converts where possible class components to `stateless functional components` also uses `fragments` where possible to return components as variables rather than arrays. Added tests and snapshots for the new components.

N.B. I forgot to update `react-dom` types when I made a PR for react I've done so here and the new types register some of the styling options we had in place as invalid e.g. `webkitAppRegion` and `webkitPaddingStart` I've used these without issue in the styled components where possible otherwise chosen close alternatives


bryphe:
Awesome, thanks @Akin909 ! Glad to see those `less` files dwindled 😄 Looks good to me!

PR #2285, opened 2018-06-04
Bugfix/ Fix staggered cursor positioner rendering

Akin909:
This PR fixes what is arguably a small but potentially (at least for me *incredibly frustrating*) bug which is that when the hover renderer appears depending on where it is it seems to shuffle into place aka renders in one position then repositions v. quickly before settling into place 💢. Hopefully this PR helps resolve this frustration for anyone else who is as finicky as me re. such things.

Before:

broken_hover

After:

fixed_hover


bryphe:
> This PR fixes what is arguably a small but potentially (at least for me incredibly frustrating) bug which is that when the hover renderer appears depending on where it is it seems to shuffle into place aka renders in one position then repositions v. quickly before settling into place

Ah yes, these are _exactly_ the sort of issues we need to tackle to get to a 'polished' level - all these sort of small frustrations can add up to create a poor UX. Very cool to have this addressed!

Thank you so much for investigating it, @Akin909 👍

Akin909:
@bryphe this is easily the most cathartic bugfix I've contributed, sooooo glad to have resolved it I definitely agree that little niggles like that can really detract from the overall experience

adelarsq:
@Akin909 I notice that the hover on my machine appears two times (appears, disappears and appears again). Is it related?

Akin909:
@adelarsq the change I made is the staggering in the GIF above not sure if thats what you mean by appearing and disappearing? if not, Im not sure about that issue but maybe raise it separately if it still happens with this PR

adelarsq:
@Akin909 I have confirmed is another issue. I will open a new issue tomorrow about it and try to solve when possible.

PR #2287, opened 2018-06-04
Update all submodules.

CrossR:
Just a quick maintenance PR to update the Git Submodules we have to their latest versions.


bryphe:
Awesome, thanks @CrossR ! 💯

PR #2288, opened 2018-06-06
Security - Add Content Security Policy

bryphe:
Thanks @CrossR for catching this! There's a CSP error that the latest verison of Electron highlights. We should be setting a content security policy for Electron, to help reduce the risk of malicious script execution.

This change adds a CSP for `index.html`, and provides an alternate entry point for the webpack-dev-server build (`npm run start`), since that needs `eval`.

There's still one problem that needs to be fixed:
image

With the new CSP, we can't load configuration files - they get blocked. We should move to the same strategy we use to load plugins - creating a new vmcontext to run them, so that they are isolated.


CrossR:
Ideally, will we want this in the next release?

bryphe:
Ideally! But unfortunately it's cutting it a bit close (and it's easy for this to regress some corner cases, especially around configuration), so I may not have time to complete it by Wednesday.

It's not a 'new' issue per-se, it's been around for a while - Electron is just better about giving us a heads up, so I don't think it necessarily needs to block this release. But we definitely should fix it!

Issue #1978, opened 2018-03-29
Tutorial Proposal: Finding within a line - `f`, `F`, `;`, and `,`


bryphe:
A helpful tutorial to have would be in linewise-character finding, exercise the following commands:
`f` - find a character
`F` - find a character (reverse)
`;` - next match
`,` - previous match

A good example for this would be the `BasicMovementTutorial`: https://github.com/onivim/oni/blob/master/browser/src/Services/Learning/Tutorial/Tutorials/BasicMovementTutorial.tsx

We'd just need to figure out the set of text that makes sense, and add the appropriate `Stages.MoveToGoalStage` to exercise those motions.

The tutorials are relatively straightforward to add, so this would be a fun and visible `first issue` if anyone is up for it 👍

robin-pham:
I'd be pretty happy contributing to this project. What does the `insider` tag mean though?

bryphe:
@robin-pham - that would be great! Would ❤️ help here, and contributions are most welcome.

The bot tags issues with `insider` for people who've contributed to the project via Patreon/Bountysource/OpenCollective. Helps me as a maintainer figure out how to prioritize issues (I try to prioritize first). But there are lots of ways to contribute to the project, and a PR would be awesome.

If you need help getting started, let me know!

keforbes:
@robin-pham, are you still working on this feature?  Do you need any help?

keforbes:
There's no rush at all, I just wanted to see if you were actually working on it or if this was just a random flyby comment. :slightly_smiling_face:

brian-yu:
Hey @robin-pham, are you still working on this issue? If so, no worries and good luck! If not, I'd like to pick this up.

robin-pham:
@brian-yu ahh I completely forgot. Yeah I've been pretty bad at working on this, you can definitely take pick this up.

keforbes:
#1978 has been merged in.

PR #2294, opened 2018-06-07
#1978 Tutorial: f, F, t, T, ;, ','

brian-yu:
Added tutorial for f, F, t, T, ';', and ',' motions, as discussed in #1978 
Please let me know if anything needs to be changed!

Didn't know whether or not to include motions with [count] in the tutorial, or include examples of using these motions with 'd' or 'c'.

### DIFF
--- a/browser/src/Services/Learning/Tutorial/Tutorials/InlineFindingTutorial.tsx
+++ b/browser/src/Services/Learning/Tutorial/Tutorials/InlineFindingTutorial.tsx
@@ -10,12 +10,12 @@ import { ITutorial, ITutorialMetadata, ITutorialStage } from "./../ITutorial"
 import * as Notes from "./../Notes"
 import * as Stages from "./../Stages"
 
-const Line1 = "Use 'f' to move to a character to the RIGHT of the cursor."
-const Line2 = "And use 'F' to move to a character to the LEFT."
+const Line1 = "Use 'f' to move to the next occurrence of a character within the same line."
+const Line2 = "And use 'F' to move to the previous occurrence of a character."
 const Line3 = "'t' is like 'f' except it moves to one spot before the character."
 const Line4 = "And 'T' is like 'F' except it moves one spot after."
 const Line5 = "Awesome! You can also use ';' to repeat the last f, t, F, or T."
-const Line6 = "And use ',' to repeat the last f, t, F, or T in the opposite direction."
+const Line6 = "Now use ',' to repeat the last f, t, F, or T in the opposite direction."
 
 export class InlineFindingTutorial implements ITutorial {
     private _stages: ITutorialStage[]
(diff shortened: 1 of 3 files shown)


glhrmv:
I'm wondering if it would be better to use "relative" directions like "next instance of" and "previous instance of" instead of "left" and "right, etc..

A popular [Vim cheat sheet](https://vim.rtorr.com/) describes the command `[count]f[character]` as "`jump to next occurrence of [character]`", which seems more elegant and concise than saying left or right. 

I think it's also important to mention that `f` & `t` (and backwards variants) only work in the line that the cursor is in, if memory serves correctly.

keforbes:
> Didn't know whether or not to include motions with [count] in the tutorial, or include examples of using these motions with 'd' or 'c'.

Personally, I think you shouldn't mention `[count]` at all.  We can add another tutorial specific to counts later where we can show counts associated with multiple operations.

Given that you made this the fourth tutorial (along with all the other "motion" tutorials) you shouldn't mention `d` or `c`.  The `d` and `c` tutorials come later and you wouldn't want to reference those features before we teach them.

browser/src/Services/Learning/Tutorial/Tutorials/InlineFindingTutorial.tsx
```diff
@@ -0,0 +1,75 @@
...
+    constructor() {
+        this._stages = [
+            new Stages.SetBufferStage([Line1]),
+            new Stages.MoveToGoalStage("Move to the 'R' in 'RIGHT' using 'fR'", 0, 38),
+            new Stages.SetBufferStage([Line1, Line2]),
+            new Stages.MoveToGoalStage("Use 'j' to move down a line", 1, 38),
+            new Stages.MoveToGoalStage("Use 'F' to move LEFT to the goal", 1, 9),
+            new Stages.SetBufferStage([Line1, Line2, Line3]),
+            new Stages.MoveToGoalStage("Use 'j' to move down a line", 2, 9),
+            new Stages.MoveToGoalStage("Use 't' to move to the goal", 2, 43),
```
keforbes: Since this is the first use of `t` I think the step should be more clear, like:
> Use 't' to move to the character before the letter 'b'

browser/src/Services/Learning/Tutorial/Tutorials/InlineFindingTutorial.tsx
```diff
@@ -0,0 +1,75 @@
...
+export class InlineFindingTutorial implements ITutorial {
+    private _stages: ITutorialStage[]
+
+    constructor() {
+        this._stages = [
+            new Stages.SetBufferStage([Line1]),
+            new Stages.MoveToGoalStage("Move to the 'R' in 'RIGHT' using 'fR'", 0, 38),
+            new Stages.SetBufferStage([Line1, Line2]),
+            new Stages.MoveToGoalStage("Use 'j' to move down a line", 1, 38),
+            new Stages.MoveToGoalStage("Use 'F' to move LEFT to the goal", 1, 9),
```
keforbes: When running the tutorial, it's kind of awkward hitting `F` to move to the character `F`.  Can we move to a different character in that sentence?

brian-yu:
@keforbes thanks for your input, I agree that counts should be introduced in a later tutorial. @glhrmv totally agreed, I was having trouble with wording it in a way that was satisfactory. I'll make those changes and commit soon.

bryphe:
Build looks good! Bringing this in now - thanks for the contribution, @brian-yu !

brian-yu:
@bryphe Thanks! I'm looking forward to contributing more to oni in the future 😃

bryphe:
Awesome! 💯

PR #2298, opened 2018-06-08
Fix clipboard content trimming

TalAmuyal:
Tested on Ubuntu, will appreciate another tester or scenarios.

**Cases tested:**
 
1. Copy a line and paste a few lines below (using `yy` and `p`).
2. Copy a line without the line termination and paste it between other characters (`v$hy`)

Also tried copy-pasting back and forth between Oni and a non-Vim-ish text editor.

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -4,6 +4,7 @@
  * IEditor implementation for Neovim
  */
 
+import * as os from "os"
 import * as React from "react"
 
 import "rxjs/add/observable/defer"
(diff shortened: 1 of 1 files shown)


CrossR:
Gave this a quick test on Windows and `yyP` and `v$hp` both work as expected, and the same as terminal neovim on Windows, as well as `Ctrl-v` to paste and `y` syncing with the clipboard.

The only difference I can spot between terminal and Oni is when copying with `v$p`. Ie the following (the new lines are important).

```
test

```

Should become 

```
test
test

```

But instead becomes 

```
ttest
est

```

This is the same on both master and this branch, so this hasn't broken anything at least.

bryphe:
Sweet! Thanks for the fix, @TalAmuyal and for testing it @CrossR ! 💯 Looks great to me.

TalAmuyal:
@CrossR, thanks for the feedback!
Will merge this for now and will have a go on that case next.

PR #2301, opened 2018-06-08
Make FileSystem.readdir more robust

rhendric:
This specific change is motivated by a broken symlink on Linux, which
will show up in an fs.readdir but will cause an error with fs.stat. If
there are other situations where fs.stat will produce an error on a
result returned from fs.readdir, this is probably the right thing to do
in those as well.


bryphe:
Awesome! Thanks for catching this @rhendric and for the fix.  Makes sense to make the `ExplorerFileSystem` robust in the case where `stat` fails.

The test is much appreciated, too - thanks for figuring out a way to exercise this behavior 💯

bryphe:
Looks like the build passed - bringing this in now. Thanks @rhendric for the contribution!

PR #2303, opened 2018-06-09
[WIP] Skip CI for commits that only contain markdown changes.

CrossR:
This should speed up the AppVeyor build queue, since currently each time a backer is added, it uses 40 mins of build time, for no changes to the code!

Unfortunately, it looks like the same functionality isn't available on Travis: https://github.com/travis-ci/travis-ci/issues/6301

I've left this PR as [WIP] for now, as I might make a few other changes to the CI and its easier than creating lots of little PRs, especially if I end up undoing it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,6 +9,12 @@ branches:
     - master
     - /^release.*/
 
+# Skip CI build if the changes match these rules exactly.
+# Ie, if the BACKERS.md file is changed, we don't need to build.
+skip_commits:
+  files:
+    - '**/*.md'
+
 cache:
     - .oni_build_cache -> package.json 
 


CrossR:
For Travis (or both for consistency) we could add a `before_install` step instead to check the file types.

Saves less time (tiny build vs no build) but would work on both.

bryphe:
Sweet! Nice find @CrossR . Makes sense to have this filter. Will especially help the case where master is eating up builds on just markdown changes, and starving out the PR builds.

> For Travis (or both for consistency) we could add a before_install step instead to check the file types.

I think it's ok for this to be a bit inconsistent. Unfortunately the build providers have different capabilities so we have to fork anyway. Up to you though! Thanks again for the investigation here 👍

CrossR:
I'm going to keep looking over the CI stuff, so I can always add this to the Travis build when I make any other changes.

Issue #2302, opened 2018-06-08
Feature: Add support for Neovim Buffer Update Deltas


bryphe:
Today, Oni syncs the _entire buffer_ when not in insert mode (so that it has the latest buffer state for language services). Until #7917, that was the best we could do.

However, with #7917, we can now hook up to the `nvim_buf_lines_event` and `nvim_buf_changedtick_event`, and _only_ get the lines that have changed. This will be very helpful for performance, especially when manipulating large files. And it means we can potentially simplify our update management - we fork our buffer update handling depending on whether we are in insert mode or not (in insert mode, we push up a single line, whereas in normal mode, we push everything) - with this change, we could simply send up whatever the delta is.

On the vim side, a good starting point for this to look at the current synchronization logic:
https://github.com/onivim/oni/blob/68d1f1f816dab516709626499ab62f39fb523135/vim/core/oni-core-interop/plugin/init.vim#L37

We should be able to replace that `OniNotifyBufferUpdate` with the new update-delta events. In addition, we should be able to scrap this `augroup` here:
https://github.com/onivim/oni/blob/68d1f1f816dab516709626499ab62f39fb523135/vim/core/oni-core-interop/plugin/init.vim#L144

On the Oni side, we handle these updates here:
https://github.com/onivim/oni/blob/68d1f1f816dab516709626499ab62f39fb523135/browser/src/neovim/NeovimBufferUpdateManager.ts#L36

We should be able to simplify this - instead of splitting between `notifyFullBufferUpdate` and `notifyIncrementalBufferUpdate`, we can simply do a `notifyDeltaBufferUpdate`.

It'd be great to have some tests here that cover this surface area (when we get a delta update from Neovim, we incorporate that into our model we send to LSPs). And of course, we need to make sure it continues working with our set of LSPs.

Another good point of reference as an example would be the commit that @lunixbochs incorporated into ActualVim: https://github.com/lunixbochs/ActualVim/commit/414e8d4c3feb4c798a611c14ef47741e8b952146

bryphe:
It might be a bit before I get time to tackle this, so if anyone would like to take this on in the meantime, that'd be amazing! ❤️  💯 

A pre-requisite too is upgrading to the latest version of Neovim, but to test locally, we could use the `debug.neovimPath` configuration setting to grab a newer version of Neovim and override the bundled one Oni uses.

CrossR:
[...]

PR #2304, opened 2018-06-11
Bump to Neovim 0.3.0

CrossR:
Depends on my open PR in the `oni-neovim-binaries` being merged first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -861,7 +861,7 @@
         "msgpack-lite": "0.1.26",
         "ocaml-language-server": "^1.0.27",
         "oni-api": "^0.0.45",
-        "oni-neovim-binaries": "0.1.1",
+        "oni-neovim-binaries": "0.1.2",
         "oni-ripgrep": "0.0.4",
         "oni-types": "^0.0.8",
         "react": "^16.3.2",


badosu:
Do you know what this package does for each OS? Just wondering how this could impact Linux Distros.

CrossR:
Yep, the code is pretty simple:

https://github.com/onivim/oni-neovim-binaries/blob/master/lib/postinstall.js

```
    await releaseDownloader.downloadGithubRelease(path.join(__dirname, ".."),  {
        "user": "neovim",
        "repo": "neovim",
        "tag": "v0.3.0",
        "platforms": {
            "win32": {
                "name": "nvim-win32.zip"
            },
            "win64": {
                "name": "nvim-win64.zip"
            },
            "darwin": {
                "name": "nvim-macos.tar.gz"
            }
        }
    })
```

So this will not change Linux behaviour at all. If a Linux user wants Nvim 0.3.0 they'd simply have to update their nvim binary on their machine in whatever way they normally do (download an appimage, `apt`/`pacman` etcetc)

Since we only bundle Neovim for Window and Mac, we don't need to make any changes to the Linux side of things, except for maybe Arch AUR build scripts and similar, though they aren't kept in the repo.

badosu:
@CrossR Thanks for the detailed explanation.

Do we have a place where we state the minimum neovim version Oni supports? If this change gets merged, does that change to 0.3.0?

CrossR:
The Wiki has some stuff here, and perhaps in other places : https://github.com/onivim/oni/wiki/Installation-Guide#linux

That said, we aren't using any features of 0.3.0 yet, so its not going to break anything. However, once #2302 is added, we will need to bump the required version number.

badosu:
@CrossR Great, thanks! I'll keep that in mind when #2302 is merged and a new version is released!

bryphe:
Awesome, thanks @CrossR ! I just published the new version of `oni-neovim-binaries@0.1.2`, and verified it picks up the latest `nvim`. I'll restart the builds here.

bryphe:
Woohoo! Build is good now - thanks @CrossR for getting us upgraded to the latest Neovim! I'll bring this in now in preparation of our next release 👍

PR #2309, opened 2018-06-11
Update ISSUE_TEMPLATE.md

TalAmuyal:
Fixed a typo and tried to polish a bit.


TalAmuyal:
Oops, didn't know it has different spelling...
I just trusted Chrome ><
So... do we care about spelling conventions?...
And aside from that, is the rest of the change acceptable?

keforbes:
I agree, I'm good with this too.  It doesn't bother me if we use US or British spelling for words.

Issue #2136, opened 2018-04-24
L binding prevents inputting l in browser window


Akin909:
Opening this to track a bug I accidentally introduced adding bindings to the browser window, seems that when inputing text if the l character is hit the `l` isn't being input as text but the command is being triggered. A fix for this would be to use a similar check as the other bindings to ensure the mapping only applies when not inserting text

bryphe:
Ah ya, I've hit this too - it seems like it breaks sneak mode and entering text in the address bar. Thanks for logging it!

Most of those commands seem to use this `enabled` filter:
```
    const isBrowserLayerActive = () =>
        !!activeLayers[editorManager.activeEditor.activeBuffer.id] &&
        browserEnabledSetting.getValue()
```

We might need to refine that a bit - we really only want those commands enabled when the focus is in the browser, but not in any of the textareas.

Govinda-Fichtner:
When having a browser tab open "l" character seems to be swallowed. Also in the command palette or the command mode of vim...
Is that to be expected with what you described @Akin909 ?

CrossR:
This has stopped me using the browser a bit, any pointers on how we can update the bind?

Akin909:
@CrossR not sure of a *simple* way tbh the easiest fix would be to change the bindings to have `<ctrl>` prefixes, the best (imo) way though I can think of would be to pass a callback to the `BrowserView` component from the `BrowserLayout` component which updates the `Layer` with if user is inputting i.e. `onChange` and `onBlur` of the input element update a boolean then that would hopefully allow for `BrowserLayer.inputInProgress: boolean` check in the function above.

CrossR:
Sounds reasonable, perhaps a short-term workaround would be swapping the binds for now?

The second does sound much better overall, so we maintain vim keybinds like people expect.

PR #2312, opened 2018-06-12
[WIP] Bugfix/browser bindings

Akin909:
cc @CrossR this implements a solution as discussed in #2136, still WIP but currently allows the binding being toggled off and on depending on if user is inputting.

### Todo

- [x] state transition of `inputInProgress`  (i.e. track changes in a centralised place)- potentially impossible due to the how the component works.
~- [ ] Add keyup listener~
- [x] ensure sneak functionality still works

### DIFF
--- a/browser/src/Services/Browser/BrowserView.tsx
+++ b/browser/src/Services/Browser/BrowserView.tsx
@@ -60,6 +60,7 @@ export interface IBrowserViewProps {
     initialUrl: string
 
     configuration: Configuration
+    setInputStatus: (inputting: boolean) => void
 
     debug: IEvent<void>
     goBack: IEvent<void>
@@ -157,7 +158,19 @@ export class BrowserView extends React.PureComponent<IBrowserViewProps, IBrowser
             scrollLeft,
             scrollRight,
         ])
+
         this._initializeElement(this._elem)
+    }
+
+    public toggleInputStatus = () => {
+        this._webviewElement.executeJavaScript(
+            `window.document.activeElement.tagName`,
+            true,
+            (element: string) => {
+                const isInputting = element.toLowerCase() === "input"
+                this.props.setInputStatus(isInputting)
+            },
+        )
     }
 
     public _triggerSneak(id: string): void {
@@ -306,6 +319,8 @@ export class BrowserView extends React.PureComponent<IBrowserViewProps, IBrowser
             elem.appendChild(webviewElement)
             this._webviewElement = webviewElement
             this._navigate(this.props.initialUrl)
+
+            this._webviewElement.addEventListener("keydown", this.toggleInputStatus)
 
             this._webviewElement.addEventListener("dom-ready", () => {
                 this._webviewElement.setZoomFactor(this._getZoomFactor())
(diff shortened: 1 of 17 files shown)


Akin909:
@CrossR when you have a chance can you give this a try to see if there are cases I haven't accounted for, this currently works by checking on `keydown` if the `document.activeElement` is an input type, if it is it doesn't apply the bindings. Hopefully it covers the issue

CrossR:
One thing I have noticed is that the binding seems to stick around for 1 keypress.

That is, if I stick my cursor in the middle of some inputted text and hit `h` or `l` I see the cursor move to the left/right once, and no letter is generated. `j` and `k` also produce no input for a single keypress, and move the cursor to the start/end of the line.

Also just to check the scope of the PR, is this intended to fix it for just the browser view? Just wanted to check if I should be messing with the Oni address bar too.

Akin909:
@CrossR tbh i'd been focusing on the content of the page but isn't too much of a change to consider changes to the address bar also thanks for testing it out, I always use oni a specific way so sometimes forget all the other ways someone else might interact with it.
 I'm currently not covering mouse events which is what leads to the lag basically I need to find a general event for when a user is interacting with the page or a variety of different events not sure of what a general one would be but for now ill add a mousedown listener as well (was it on clicking into the field that you noticed this lag?)

bryphe:
Shoot, unfortunately I duplicated some work here @Akin909 - sorry about that! The last couple times I was on a plane I was trying to do some react development with the embedded browser, and hit these issues, so I worked on this PR: #2239  LMK what you think. It was a 'bug bear' for me so definitely something I want to get fixed in this next release - IMO the browser isn't usable at the moment.

Again, sorry about the duplicate effort!

browser/src/Services/Browser/BrowserView.tsx
```diff
@@ -157,9 +158,21 @@ export class BrowserView extends React.PureComponent<IBrowserViewProps, IBrowser
             scrollLeft,
             scrollRight,
         ])
+
         this._initializeElement(this._elem)
     }
 
+    public toggleInputStatus = () => {
+        this._webviewElement.executeJavaScript(
```
bryphe: Ideally we could 'push' this information up, instead of having to check everytime a key is pressed (which will hurt the responsiveness of key presses). In #2239 I set up this 'push' pipeline where the `webview` tag will let us know whenever the focused tag changes - that way we don't have to do stuff on keypress.

browser/src/Services/Browser/index.tsx
```diff
@@ -164,7 +178,7 @@ export const activate = (
         const activeBuffer = editorManager.activeEditor.activeBuffer
 
         const browserLayer = activeLayers[activeBuffer.id]
-        if (browserLayer) {
+        if (browserLayer && !browserLayer.inputInProgress) {
```
bryphe: I think we might want to refine this logic a bit. Semantically, it would probably make more sense to be in the `enabled` filter, since this is gating whether or not the command is needed (vs the `execute`, which  should really just be responsible for _executing_ the command).

I don't think this would handle both cases of having a textarea / input in focus and the address bar. I think a slightly more robust condition would be to only allow our scroll commands if the webview is focused _and_ we're not in an input / textarea element (this is what I implemented in #2312 ) - let me know what you think though!

Akin909:
@bryphe no worries just happy the problem is solved and its one less PR for me to work on, had a look at your branch and its a much more robust solution not particularly attached to this PR so will close this in favour of yours

PR #2315, opened 2018-06-13
Add a 'find in files' quick-menu

TalAmuyal:
**Disclaimer:**

I (shamelessly) copied the code from the sidebar's search and injected it into a copy of `QuickOpen`.
Personally, I find it more useful that way.

The "new" feature is called `QuickFind` :P


CrossR:
Gave it a test and works fine for me!

One thing that struck me was if the search is in a menu bar, does it not make sense for the results to be in there as well? Though that could potentially warrant an issue to get some comments. I know there was talks of a combo search box.

Ie Ctrl-p but if you prefix the search with ? its file wise, > is commands, nothing is fuzzy finder etcetc.

Akin909:
I haven't had the time to try it myself but I agree with @CrossR re. the results I think one of my main blockers for using the search feature is that the results get placed in another window I have to navigate to would be nice to have an FZF like Rg command that the results were what the menu is poplulated with and you can just select the one you want

TalAmuyal:
How about this:
The results will appear in the drop down box of the menu (like c-p).
Enter will set the results to quick-fix.
C-{Enter/v/h} will act like in c-p menu.

Akin909:
@TalAmuyal that sounds good think its a good idea to have a way of saving the results anyway, re. Enter as saving to the qf menu (my own 2cents here) I think the definitive action should be opening the result and the optional action should be opening in the qf e.g. `c-enter`, my thinking being if you find what you want in the menu then hitting enter and finding that it opens but also creates another window might not be the desired effect going off of other editor experiences I think its more likely that user would want to make a conscious effort to keep the results and would probably default expect that enter just opened the result

TalAmuyal:
Note: I pulled out the whole QuickOpen folder as a stand-alone plugin and merged the find-in-file functionality as another use-case.

The code is refactored since it was a bit tangled and was hard for me to just inject the additional logic.
I tried not to change the previous business logic.

extensions/oni-plugin-quickopen/src/QuickOpen.ts
```diff
@@ -0,0 +1,435 @@
...
+
+    public openFileWithAltAction(): void {
+        const mode: Oni.FileOpenMode = this._oni.configuration.getValue(
+            "editor.quickOpen.alternativeOpenMode",
+        )
+        this.open(mode)
+    }
+
+    public async searchFileByContent() {
+        const filterName = "none"
```
CrossR: Should this be a config option?

My first instinct was to try and use smart search (ie case insenstive unless you use uppercase characters), which then failed.

CrossR:
Couple of general thoughts:

When selecting something using the `quickOpen.searchFileByContent`, I would expect the action of selecting an entry to move me to that line, rather than that top of that file.

For larger Repos, the `quickOpen.searchFileByContent`, seems to stop working?
That could just be my VM choking on larger repos (I tested the Oni + VSCode repos), so would be useful for someone else to test too.

Similarly, swapping between big repos seems to cause issues where the previous repos files are still shown. I think the process for ripgrep may not be being closed or something, since I saw the menu fighting between repos and swapping back and forth between results from  two different repos. I'm not too sure what the fix here is, but I did get a "`Stopped before out-of-memory crash`" in the debugger for this line:

```
const menuItems = result.items.map(e => e.toMenuItem(this._oni, this._seenItems))
```

So perhaps that is the line taking up most of the time? At the point of the crash I had just short of 4GB used, so something must be up. Also I had duplicated items somehow:

image


When I was testing on smaller repos, worked  as expected and looks good tho! Some of the memory stuff may just be us not controlling ripgrep correctly or its results, since it should be fine.

extensions/oni-plugin-quickopen/src/QuickOpen.ts
TalAmuyal: This `"none"` is intentional.
This `filter` field is a second filtering done after `rg`.
Since `rg` can accept a regex, a search string like `ab.*cd` will be computed as `rg ab.*cd | egrep "ab\.\*cd"` which will very likely cause similar searches the yield no results.

However:
1. Found the reason that smart-case didn't work and fixed it.
2. Added a TODO to use a filter for highlighting w/ support for regex.

I tried the `regexFilter` for the menu, but it throws some exceptions.
It might be that a slight update/variation can fix it. (But I prefer to do less changes that will keep this PR further for merging)

(Will push soon)

TalAmuyal: P.S: fixed the latency/hanging issue.

TalAmuyal:
@CrossR, thanks for the fast review!

Regarding the performance issue, it should be better now and I have an idea for another improvement, though I prefer to have it in a future PR since it requires another API change.

Regarding opening in the right place - fixed! :P

Also added tests :)

TalAmuyal:
@CrossR, @bryphe,

I've added some tests and made them pass locally but on some of the servers it just fails.
Specifically it fails on MacOS & Windows, though it passes on the Linux server and on my MacOS laptop.

For now the tests I've added are commented out.
Could we pull this in that way?

CrossR:
I can give the tests a run on my Windows machine if that would help?

I think I can see part of the issue, so I'll stick a review comment on.

test/ci/QuickOpenTest.ts
```diff
@@ -1,21 +1,50 @@
...
+    const editor = oni.editors.activeEditor
+    await oni.automation.waitFor(() => editor.activeBuffer.filePath.endsWith("/SearchProvider.ts"))
+    // End
+
+    // Start
+    await waitForCommand("quickOpen.searchFileByContent", oni)
+    oni.commands.executeCommand("quickOpen.searchFileByContent")
+    await oni.automation.waitFor(() => oni.menu.isMenuOpen())
+    await oni.automation.sleep(100)
+    anyOni.automation.sendKeysV2("^exp.*?class.*?Qu.*?Op.*?Item")
```
CrossR: I'm not fully sure what this search string is doing, but it kills my dev instance, so I assume that it either does the same to the tests, or its taking that long the correct file isn't opened (a few times on my machine I saw `ACCOUNTING.md` opened instead).

Perhaps its worth changing this, or swapping the working directory to a different folder in the Oni repo, so the search is easier?

test/ci/QuickOpenTest.ts
```diff
@@ -1,21 +1,50 @@
...
+    await oni.automation.sleep(100)
+    anyOni.automation.sendKeysV2("<CR>")
+    await oni.automation.sleep(500)
```
CrossR: On my machine, I needed slightly longer sleeps, so I swapped them all to `2000`.

I also swapped the `sendKeysV2("<CR>")` to use `await`.

test/ci/QuickOpenTest.ts
```diff
@@ -1,21 +1,50 @@
...
+    const filePath = oni.editors.activeEditor.activeBuffer.filePath
+    assert.assert(filePath.endsWith("QuickOpenItem.ts"), `filePath`)
```
CrossR: I was hitting issues with this assert where the test was finding itself! So it would search for some string and the first occurrence would be the test file itself. I think if  you swap the search string to something unique to the test (`MyUniqueTestString`) and then assert the file is the test file, you should get better results.

TalAmuyal:
@CrossR, thanks a bunch for that last batch or comments - I am now confident that I can make it work on the CI servers!

bryphe:
Sweet! Thanks for all the help @CrossR ! 💯

TalAmuyal:
@CrossR, when you have the time, please provide feedback or approve this PR.

Akin909:
It sounds like a fairly small bug that can be fixed in a follow up PR I'd love to see this functionality personally 👍

bryphe:
> That said....its a small bug, so I think its worth getting this merged so it can be used. The bug can be fixed later, and it lets us get more feedback. 

Sounds good to me!

TalAmuyal:
I'm so happy it is finally merged that I'm almost tearful...
Thanks guys, will move on the the next PR.
(BTW, will fix that hidden folder issue)

bryphe:
Thanks for all the hard work on it, @TalAmuyal ! Excited it made it in too! 🎉

PR #2319, opened 2018-06-15
create Text Objects tutorial

keforbes:
I felt that the Targets.vim tutorial was too much of a jump in learning since it covered extensions to text objects before regular text objects had been discussed.  So I created a Text Objects tutorial for the user to complete before moving on to the Targets.vim tutorial.

### DIFF
--- a/browser/src/Services/Learning/Tutorial/Notes.tsx
+++ b/browser/src/Services/Learning/Tutorial/Notes.tsx
@@ -508,3 +508,21 @@ export const RepeatOppositeKey = (): JSX.Element => {
         />
     )
 }
+
+export const innerTextObjectKey = (): JSX.Element => {
+    return (
+        <KeyWithDescription
+            keyCharacter="i"
+            description={<span>Select a Text Object with delimiter characters</span>}
+        />
+    )
+}
+
+export const aTextObjectKey = (): JSX.Element => {
+    return (
+        <KeyWithDescription
+            keyCharacter="a"
+            description={<span>Select a Text Object and its delimiter characters</span>}
+        />
+    )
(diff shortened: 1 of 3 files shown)


keforbes:
Added some requested reviewers.  I know you guys don't really care about tutorials but I can't approve my own PRs. :smile:

CrossR:
I should be able to review this tomorrow if nobody else has!

And I'm pretty excited for the ongoing tutorials, we are quickly moving into areas I know, but would love some thing to practise on and actually fully learn rather than the fragments I know!

CrossR:
Gave the tutorial a test and checked the code and looks good!

Only comment I have is that the moving between stages is a little quick for some parts. Would it be possible to add a slight delay, say between the end of the HTML tags bit and swapping to the next so its clearer to see the result of deleting the tags etc?

keforbes:
> Would it be possible to add a slight delay, say between the end of the HTML tags bit and swapping to the next so its clearer to see the result of deleting the tags etc?

Unfortunately, no.  The best I could do would be to add another step just so the text stays on screen longer.  When showing `va{` I added a "hit <esc> to exit visual mode" step just so the selection stays on screen until the user chooses to hit `<esc>`.  For the HTML tags section, I guess I could add a simple "move to the top of the file" motion just to give the user something to do.  I'm not sure if it's necessary though.

bryphe:
Thanks @keforbes for putting this together and @CrossR for testing it out! 👍 If there isn't an easy fix, I'm cool bringing this in for the next release - interested in getting feedback on our tutorial experience, and it's fun to have new tutorials to include 👍

PR #2321, opened 2018-06-15
[GrammarLoader] Properly check if key is in cache, #1940 fix

### LINKED ISSUE #1940: Weird infinite-loop 100% cpu usage interaction between oni and deoplete and HTML files
Reproduction steps:
1. Configure oni to use init.vim
2. Setup the init.vim with [deoplete](https://github.com/Shougo/deoplete.nvim#install) 
4. Edit an html file, try to prompt suggestions by going into insert mode and start to type inside an html tag, e.g.  `class="type here"`.

screenshot_2018-03-26_22-47-20
screenshot_2018-03-26_22-47-54
^ browser console keeps on complaining about 'No grammar found for language: html'

This is a pretty bizarre bug, just writing it down here in case other people encounter the same issue.

P.S: Oni is realllly coool.

GeorgeTG:
The current check will invalidate the grammar cache if the previous run returned null.
This is because the key is in the cache but the value is null.
I think this was not intended since it forces a sync expensive function to be re-run every time `getGrammarForLanguage` is called resulting in really bad performance and making the app unusable.
_For example when scrolling this method is called hundreds of times._

Also I cannot think of a situation the grammar is going to be found in the next call if the previous failed.
This also fixes #1940

### DIFF
--- a/browser/src/Services/SyntaxHighlighting/GrammarLoader.ts
+++ b/browser/src/Services/SyntaxHighlighting/GrammarLoader.ts
@@ -38,7 +38,7 @@ export class GrammarLoader implements IGrammarLoader {
             return null
         }
 
-        if (this._grammarCache[language]) {
+        if (language in this._grammarCache) {
             return this._grammarCache[language]
         }
 


TalAmuyal:
@GeorgeTG, can you `Squash and merge` this?
If you want to but can't (because of permissions) just say so and I'll do it.

BTW, thanks for the fix!

Akin909:
@TalAmuyal I don't think @GeorgeTG can merge this PR I was going to but thought I'd give other collaborators a chance to have a look first

GeorgeTG:
Oh yeah I couldn't merge of course

PR #2328, opened 2018-06-16
Update copyright year

JosefKuchar:


keforbes:
Shouldn't it be `2017-2018` and not just `2018`?  I'm not sure how these things work.

keforbes:
Fine with me!  I just never know the rules around copyright years.

PR #2330, opened 2018-06-18
[WIP] #2302 - Integrate with the buffer updates API

bryphe:
This PR integrates with the buffer updates API. It deprecates / removes the old sync strategy, which wasn't 100% reliable.

This is a __BREAKING CHANGE__ because it will necessitate neovim 0.3.0 - we won't keep the old strategy around.

Items:
- [ ] Remove old sync strategy from `oni-core-interop/init.vim`
- [ ] Update `NeovimBufferUpdateManager` to call `nvim_buf_attach` when entering a buffer
- [ ] Handle `nvim_buf_detach` event
- [ ] Properly coerce the live-update input into ranges that the LSP expects

Current bugs:
- [ ] Syntax highlighting is not working. Might be an issue with how the enter / update flow has changed with these chagnes
- [ ] Incremental updates aren't working correctly with TypeScript. Need to debug in our LSP implementation further.

### DIFF
--- a/browser/src/neovim/NeovimBufferUpdateManager.ts
+++ b/browser/src/neovim/NeovimBufferUpdateManager.ts
@@ -19,124 +19,62 @@ export interface INeovimBufferUpdate {
     contentChanges: types.TextDocumentContentChangeEvent[]
 }
 
+export type NeovimBufferLineEvent = [number, number, number, number, string[], boolean]
+
 export class NeovimBufferUpdateManager {
     private _onBufferUpdateEvent = new Event<INeovimBufferUpdate>()
-    private _lastEventContext: EventContext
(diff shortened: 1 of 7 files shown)


Akin909:
@bryphe, *preface: definitely not intended to rush or poke* was wondering when/if you were still working on this, reason being I'm curious about trying to implement a [`treesitter`](https://github.com/tree-sitter/node-tree-sitter) syntax highlighter/higlighting opt. at some point (probably not v. soon) and would like to base it off an updated version of this because of the way tree-sitter works it basically seems like it would go hand in hand with incremental updates or rather it works on incremental parsing theres some research I'd need to do but either way since syntax highlighting's structure would change with this I'd rather wait

PR #2331, opened 2018-06-18
[WIP] Feature/set press and hold on ios

### LINKED ISSUE #1888: Set ApplePressAndHoldEnabled on macOS
According to the [gist](https://gist.github.com/kconragan/2510186), on macOS, it would not treat press and hold (especially h/j/k/l) as repeat key press by default.
Should oni change the default settings of this [defaults](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/defaults.1.html) out-of-box?

Akin909:
Set a `mac` users press and hold setting to facilitate ease of use of oni for new users fixes #1888, @CrossR thanks for locating this setting 👍 this PR adds a call to `main.ts` to see if the setting is false if so it applies it.

Todo
- [ ] Test locally

### DIFF
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -1,7 +1,7 @@
 import * as minimist from "minimist"
 import * as path from "path"
 
-import { app, BrowserWindow, ipcMain, Menu } from "electron"
+import { app, BrowserWindow, ipcMain, Menu, systemPreferences } from "electron"
 
 import * as PersistentSettings from "electron-settings"
 
@@ -181,6 +181,10 @@ export function createWindow(
         height: windowState.bounds.height,
         width: windowState.bounds.width,
     })
+
+    if (process.platform === "darwin") {
+        systemPreferences.setUserDefault("ApplePressAndHoldEnabled", "boolean", "true")
+    }
 
     if (windowState.isMaximized) {
         currentWindow.maximize()
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -183,7 +183,14 @@ export function createWindow(
     })
 
     if (process.platform === "darwin") {
-        systemPreferences.setUserDefault("ApplePressAndHoldEnabled", "boolean", "true")
+        const pressAndHold: boolean = systemPreferences.getUserDefault(
+            "ApplePressAndHoldEnabled",
+            "boolean",
+        )
+
+        if (!pressAndHold) {
+            systemPreferences.setUserDefault("ApplePressAndHoldEnabled", "boolean", "true")
+        }
     }
 
     if (windowState.isMaximized) {
--- a/BACKERS.md
+++ b/BACKERS.md
@@ -128,6 +128,8 @@ Thanks you to all our backers for making Oni possible!
 *   James Herdman
 *   Wayan Jimmy
 *   Alex
+*   Phil Plückthun
+*   Norikazu Hayashi
 
 <a href="https://opencollective.com/oni/tiers/backer/0/website" target="_blank"><img src="https://opencollective.com/oni/tiers/backer/0/avatar.png"></a>
 <a href="https://opencollective.com/oni/tiers/backer/1/website" target="_blank"><img src="https://opencollective.com/oni/tiers/backer/1/avatar.png"></a>
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -4,6 +4,7 @@
  * IEditor implementation for Neovim
  */
 
+import * as os from "os"
 import * as React from "react"
 
 import "rxjs/add/observable/defer"
@@ -414,7 +415,9 @@ export class NeovimEditor extends Editor implements IEditor {
                     const isAllowed = isYankAndAllowed || isDeleteAndAllowed
 
                     if (isAllowed) {
-                        clipboard.writeText(yankInfo.regcontents.join(require("os").EOL))
+                        const content = yankInfo.regcontents.join(os.EOL)
+                        const postfix = yankInfo.regtype === "V" ? os.EOL : ""
+                        clipboard.writeText(content + postfix)
                     }
                 }
             }),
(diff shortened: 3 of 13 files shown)


main/src/main.ts
```diff
@@ -182,6 +182,17 @@ export function createWindow(
...
     })
 
+    if (process.platform === "darwin") {
+        const pressAndHold: boolean = systemPreferences.getUserDefault(
+            "ApplePressAndHoldEnabled",
+            "boolean",
+        )
+
+        if (!pressAndHold) {
+            systemPreferences.setUserDefault("ApplePressAndHoldEnabled", "boolean", "true")
```
bryphe: Wow, super cool! Didn't know we could do this. This will alleviate a common pain point on OSX. Nice find @Akin909 💯

bryphe:
Wow, very cool! Didn't know about the `systemPreferences` API here, that's awesome. Code change looks great 👍 

One thing I did notice though is that it looks like your submodules are out-of-date - you might need to run `git submodule update` to bring in the latest, so they don't show as diffs here.

Akin909:
@bryphe thanks for the heads up re the submodules unfortunately that command seems not to change their status, 🤔 I dont work with submodules ever so doing a bit of fiddling to get them inline

breja:
Does this setting get changed globally for `ApplePressAndHoldEnabled`? That would be weird.

How are we reading input in Oni? Is it from this file https://github.com/onivim/oni/blob/master/browser/src/Input/KeyboardInput.tsx which uses an `<input>` field? I think `<input>` elements process key inputs differently than a raw/global event handler.

If we attach key event handlers on window, there should be no rate-limiting on keys being held.

I tried this in the Oni devtools console (on mac os) and it correctly spams keys being held down: `window.addEventListener('keydown', e => console.log(e.key))`

Are we using `window.addEventListener` or can we use it for input events? Is there a reason not to use it? I use this method in my electron app on multiple macs, and I never had to mess with the `ApplePressAndHoldEnabled` flag. But maybe I'm wrong/missing something.

Akin909:
@breja the documentation regarding that tbh is a bit ambiguous I still need to test the behaviour of this locally, but from what I have seen even if it were the case that the action were global which I agree would be unexpectedly weird, there are additional methods available to restore a users setting before the window is closed.

I personally would cc @bryphe re. the input handling side of things its not a part of the code base I've looked at much

bryphe:
Ah ya, it does seem like the behavior is different when using `window.addEventListener` vs `<input />` events.

Originally we only used `window.addEventListener` (which is simpler), but we switched to the `<input />` in order to support a few international scenarios:

Dead keys:

IME:

bryphe:
And good point about the setting - we definitely shouldn't be setting it globally! We should validate that it only gets set for our app.

bryphe:
@breja - nice! Sure thing, it's great to see you around here. We're big fans of [veonim](https://github.com/veonim/veonim/) too, really impressive work!

CrossR:
Gave this a try on my machine, this is how it sets it: 

image

So its local just to Oni!

main/src/main.ts
```diff
@@ -182,6 +182,17 @@ export function createWindow(
...
     })
 
+    if (process.platform.includes("darwin")) {
+        const pressAndHold: boolean = systemPreferences.getUserDefault(
+            "ApplePressAndHoldEnabled",
+            "boolean",
+        )
+
+        if (!pressAndHold) {
+            systemPreferences.setUserDefault("ApplePressAndHoldEnabled", "boolean", "true")
```
CrossR: This should be `false` not `"true"`, since it needs to be a `bool` and it should be setting this value to false.

main/src/main.ts
```diff
@@ -182,6 +182,17 @@ export function createWindow(
         width: windowState.bounds.width,
     })
 
+    if (process.platform.includes("darwin")) {
+        const pressAndHold: boolean = systemPreferences.getUserDefault(
+            "ApplePressAndHoldEnabled",
+            "boolean",
+        )
+
+        if (!pressAndHold) {
```
CrossR: This should be `pressAndHold` (no `!`), since if `ApplePressAndHoldEnabled` is enabled, we want to disable it.

CrossR:
For completeness sake, now I've looked into it there is also `registerDefaults` in Electron 2+ which takes key:value pairs and registers them (ie temporary), rather than setting them.

For something so critical, I think this being a set makes sense, since otherwise we'd have to do it every time.

Akin909:
@CrossR its cool to have it confirmed thanks for having a look 👍 , is this working for you locally personally it throws an error for me re. the third argument not being able to be converted to a boolean, couldn't find any reason to explain why ended up trawling through the code to find out why but never got to the bottom of it

CrossR:
@Akin909 I originally got that error, but have added the changes I made as review comments, which then made this work.

main/src/main.ts
Akin909: yeah that bit is quite confusing as the type for it from electron and the docs state the argument is a string got stuck on that for a while, seems like an issue for electron and their types for this function

Akin909:
@CrossR thanks for checking in on this I kind of let this one fall by the wayside anyway all up to date just built it locally with no issues as well 👍

kopischke:
Great to see platform adequate handling for this! One nit: for future reference and search, this PR should not reference “iOS”: that is the iPhone / iPad OS. The desktop OS was called, until recently, “OS X” and is now called “macOS”. Mac users are _highly_ unlikely to search for a different OS when looking up issues for their platform,.

keforbes:
This PR has Changes Requested by @CrossR but it looks like they were resolved.  Is this PR waiting on anything?

CrossR:
I think its waiting on one of us testing it again now with the changes.

I was unable to get it to work, but I'm not convinced that its the codes problem and more me deleting the old key caused this way to not work.

Akin909:
@CrossR tbh the same goes for me I'm not entirely sure what a good way to test this specific functionality would be I have globally on my mac already turned `pressAndHold` off, if I turn it on I think it globally overrides all my apps' behaviours and not sure if this command should be able to override it if ive set it via the cli, I need to try turning it off for `oni` alone and seeing if that ignores it but I'm not sure either if thats a valid test if it works.

PR #2332, opened 2018-06-19
Markdown Syntax Highlights

### LINKED ISSUE #1653: Images not shown in Markdown preview
The images should be sourced from the directory where the document is saved to:

image

CrossR:
Second set of Markdown changes that fixes #1653 and adds Syntax highlights to code blocks.

image

Things to check:

* [x] - Performance as highlight js is synchronous. Maarked does have an option for async, but I didn't really find much documentation on it, nor any well maintained async libraries compared to highlightjs.
* [x] - Add some testing for both this, and my previous PR. 
  * Check syntax is loaded, check applied to elements, check not applied when relevant language is given, check swapping back and forth from md buffers to close etc.
  * Not sure if its worth adding this in a separate PR to get this merged quicker and with further testing?
* [x] - Currently on the syntax highlights, should it change the code block background colour too?

Something I realised from my previous PR is that when opening a browser link, the markdown preview remains open for some reason, though swapping to any other non-markdown buffer works fine. It looks like the leave command is called once to close, but the enter is called multiple times after and re-opens the markdown preview as the buffer is still a markdown file.

I could update the `onBufferEnter` command to close the preview too, but haven't tested that yet.

Similarly, opening a MD file directly doesn't spawn the preview, but opening one once Oni is open works fine.


CrossR:
Pushing up some tests for this and the previous PR now.

I think the background colour can stay as is for now, it makes sense to use the Oni defined colours to match the theme its rendered in.
Still not sure about the async vs sync plugins though.

CrossR:
Also looks like we've got CodeCov to only look at the main repo, so despite most of the PR being updated tests, the coverage has dropped.

browser/src/Services/Browser/index.tsx
```diff
@@ -290,10 +290,6 @@ export const activate = (
         detail: "",
         enabled: isBrowserScrollCommandEnabled,
     })
-
-    ipcRenderer.on("open-oni-browser", (event: string, args: string) => {
-        openUrl(args)
```
CrossR: Left over from previous PR, tidied up in this one.

CrossR:
I've added a config option for this now, incase it is needed.

I'm still not too sure what to do about the sync vs async nature, but I think the only way I'll really be able to tell is if more people start using it and give some feedback.

I've defaulted this to on, since having an experimental feature in an experimental feature seemed like too many gates to jump through.

CrossR:
Could I get somebody to restart my AppVeyor build/merge this?

Akin909:
@CrossR seems our AppVeyor links or mine at least now leads me to a project not found or no access but if I sign in manually I get access to the oni project but then finding the related build is very manual checking through all the previous builds so not sure how to reset this tbh except manually searching which I can try later

CrossR:
Yep since we moved, the build is still pointing to the old one. I'll push a change over the weekend and that should fix it.

bryphe:
Just tried out! Looks great:
image

And the code change looks good to me 👍 Thanks for the test coverage! 💯

Issue #2329, opened 2018-06-17
Feedback about `:` command pallet.


jordwalke:
Oni Version:0.3.4 OSX

I noticed that the `:` command pallet is kind of slow. The other command pallets (Like the recent files pallet) are faster.

Also, the `:` command pallet deactivates the currently active tab's highlight which is kind of distracting.

bryphe:
Thanks for the feedback, @jordwalke ! 

cc @Akin909 - Akin did the majority of the externalization / React work here so he might have some ideas.

The `:` menu actually has a different architecture than the other menus (like quick open / command palette) - all the UI is round-tripped to Neovim (and there isn't any typing prediction today IIRC) - so it'd be worth profiling the latency from keypress -> rendering. 

There are several potential bottlenecks along that pipeline. Another potential bottleneck could be the `wildmenu` results we get from Neovim - might be some cases where we are rendering a lot of items.

Akin909:
@bryphe @jordwalke my first thought would be whether or not you have any `cabbrev` mappings or commandline mappings as those can have a timeout I think. If the slowness is due to the speed at which the ui appears then theres an `80ms` wait which is there to prevent flickering if for example navigating between buffers using a mapping e.g. `bnext` in quick succession, without it the ui would visibly appear and disappear, which is what actually happens in neovim but due to placement of the commandline is less noticeable.

As for the round trip we don't actively make any calls to neovim, we just listen for commandline events (along with all the other neovim ones) and put the results in redux then read them out, there's no processing or bottle neck along that path other than perhaps the time it takes neovim to register the action and dispatch the event we listen for. I think a profile would be really helpful I tried a quick comparison between my terminal nvim setup and oni, and don't get a discernible difference but could just be fudging it on my end.

As for the highlighting thats definitely a bug looking at the reducer theres no obvious relationship between them but we do depend on an active bufferId when highlighting the buffer's tab and its possible that that is set to null when the commandline is open 🤔

jordwalke:
I took a look at the timing chart and I think it could be a combination of:
1. The fact that there's an 80ms ease-in animation.
2. The dialog is created and destroyed every time you hit `:`. Would it be possible to avoid creating/destroying those DOM nodes and simply hide it? (That might be beneficial for all of the modal UI dialogs). That way, there would be fewer things in between the keystrokes and showing the UI.

justinmk:
[...]

PR #2334, opened 2018-06-19
Feature/remove commandline timeout

Akin909:
Following on from #2329 this PR removes the commandline menu's timeout and adds a special case to the tab color to handle commandline mode (aka same color as normal mode is used).

As I mentioned in #2329 the plan would be to add some documentation about using `<silent>` or `<Cmd>` to prevent any flickering. I think its arguably safe to say users using the `commandline` with mappings aka those who hit this bug are likely to be users who would be comfortable making the required changes to their mappings...

I've left off the changes re. keeping the elements vs. destroying them as at a glance this will require a much bigger/ less straightforward change than I currently can spare the time for


Akin909:
@bryphe this PR should address the performance issues raised in #2329 as well as the tab highlighting, running it locally with no issues.

Re. codecov I think it's not quite picking up salient changes as other than some styling tweaks, shuffling the same code to different and some deletions I haven't really made any changes that should drop the coverage (though it seems to..) can always remove the styling changes (they just mean the component would hide if its parent were hidden rather than be re-created every time)

Akin909:
Is this alright to merge @bryphe I'm planning to add a section to the docs to explain using a `<cmd>` binding to avoid flickering

bryphe:
Go for it! If both you and @CrossR are cool with it, I'm cool with it too 😄  👍 

> I'm planning to add a section to the docs to explain using a <cmd> binding to avoid flickering

A section in the docs would be great! I'll link to it in the release notes for the next release.

PR #2337, opened 2018-06-20
Add hidden option to Menu bar.

### LINKED ISSUE #2336: Feature request : alt doesnt show menubar (this is annoying)
Vscode has four options for this setting
```tsx
"window.menuBarVisibility": "default", // menu only hidden in fullscreen
"window.menuBarVisibility": "hidden", // (what Im requesting we have) menu is always hidden even if alt is pressed
"window.menuBarVisibility": "toggle", // current hidden behaviour
"window.menuBarVisibility":  "visible" // current visible behaviour 
```

CrossR:
This adds an option to hide the menu bar entirely, so alt does not trigger it at all.

Fixes #2336 and I seem to think possibly other issues.

One point I am not quite sure on is if there should be/is a need to have some other way of toggling the menu bar, like a command? That way a user could still toggle it with a command in the `Ctrl-Shift-P` menu etc.

If that is the case, then it would instead need to be hidden and have alt unbound from it, since the currently strategy  removes the menu entirely.

### DIFF
--- a/browser/src/Services/BrowserWindowConfigurationSynchronizer.ts
+++ b/browser/src/Services/BrowserWindowConfigurationSynchronizer.ts
@@ -45,9 +45,13 @@ export const activate = (configuration: Configuration, colors: Colors) => {
             document.body.style["-webkit-font-smoothing"] = fontSmoothing
         }
 
-        const hideMenu: boolean = configuration.getValue("oni.hideMenu")
-        browserWindow.setAutoHideMenuBar(hideMenu)
-        browserWindow.setMenuBarVisibility(!hideMenu)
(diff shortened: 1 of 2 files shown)


bryphe:
Awesome! Thanks for the quick fix @CrossR 💯 

> One point I am not quite sure on is if there should be/is a need to have some other way of toggling the menu bar, like a command? That way a user could still toggle it with a command in the Ctrl-Shift-P menu etc.

Potentially, but I'd be OK tracking that as a separate issue. It seems like niche behavior - if we find that it is something users want, we could implement it (it'd be a bit more complex as you mentioned). IMO the main cases for the menu are: always off, off-but-toggle-with-alt, always on, which would be covered now with your PR 👍

PR #2338, opened 2018-06-20
Fix IME with autoClosingPairs.enabled

ackey:
1. When autoClosingPairs.enabled is true (default)
2. Input some words with IME
3. Hit enter key to submit the input
4. A wrong return is inserted before the input.

This PR fixes it.


bryphe:
Awesome! I just tested out the IME yesterday and saw this issue - thanks for the fix @ackey . Tried it out on OSX and the behavior is much improved. 👍 

Bringing this in now-

PR #2341, opened 2018-06-21
Fix a first character problem in IME with autoClosingPairs.enabled

ackey:
1. With autoClosingPairs.enabled (default)
2. In IME, enter "「" (for instance)
3. Wrong braces appear around the input.

#2338 doesn't fix this case. This PR fixes it.


bryphe:
Just tried it out - looks good to me. Thanks @ackey for all your help fixing up the IME - really appreciate it! 💯  Bringing this in now.

PR #2344, opened 2018-06-21
Feature/experimental indent guide lines

Akin909:
Very early POC for an Oni gui indent lines buffer layer based on @bryphe's suggestions in #2188 

Improved (thanks @bryphe)
screen shot 2018-06-21 at 23 40 09
Line Wrapping:
screen shot 2018-06-24 at 13 56 13

Issues:
- [x] Line positioning!!
- [x] render discrepancy with guide line width
- [x] Line wrapping - not sure how we deal with this but it isn't communicated to the context layer 😿- ~* to be addressed in follow up PRs*~
- scrolling can appear janky - especially on jumping multiple lines at a time repeatedly aka r. fast scrolling (perf metrics on the functions and they take a millisecond each to run, I'm thinking possibly the lines are being rendered with slightly incorrect data midscroll - *be Addressed in follow up PRs*
- [x] JsDoc and possibly more have odd slightly off indentation ?? 🤷‍♂️


bryphe:
This is awesome, @Akin909 ! Great to see an example of using the buffer layer API! 💯

Akin909:
@bryphe managed to get the CI test working

Re. the wrapping I tried the strategy you mentioned on discord but it spirals a little in complexity, the dimensions and length method also seems to work but the dimensions that the context passes includes I think the `numbercolumn` so I still need to figure out how to calculate the offset

Anyway I plan on doing that and addressing the scroll lag separately (as I mentioned)

Akin909:
Actually managed to get wrapping working except for two outstanding edge cases
1. If the first visible line is wrapping this throws everything off not clear on why
2. Sometimes neovim wraps even if there is no content wrapping possibly the line break is wrapping 😕, the method you suggested doesn't seem to detect this though aka `bufferToScreen` doesn't report a wrap in those cases.

Either way its a big improvement so If you're amenable and tests pass I'd like to close this out and tweak further over time.

@bryphe one question though is that having to re-implement the wrapping would be a real pain for any other layer functionality maybe it can be moved to the context itself and that could provide a `wrappedLines` field, also I noticed I largely just used it to adjust the height and position (vertical) of the lines so possibly the method could be used to update the results that bufferToPixel returns to actually just provide the dimensions with the wrapping accounted for.

bryphe:
Wow, just tried out the latest build - looks great @Akin909 ! The scrolling is actually smoother than I expected... really impressive work!

> Either way its a big improvement so If you're amenable and tests pass I'd like to close this out and tweak further over time.

Sounds good, I'm sold on it 👍  Thanks for all the work you did to add tests on it, as well!

> @bryphe one question though is that having to re-implement the wrapping would be a real pain for any other layer functionality maybe it can be moved to the context itself and that could provide a wrappedLines field, also I noticed I largely just used it to adjust the height and position (vertical) of the lines so possibly the method could be used to update the results that bufferToPixel returns to actually just provide the dimensions with the wrapping accounted for.

Definitely - I think we can take some of the lesson here to improve the API. One thing I wanted to do was have a function that would take a rectangle (startLine, startColumn, endLine, endColumn), and return an array of pixel positions back. Sometimes, for example, an error underline might get 'wrapped' - so we'd actually need _two_ sets of rectangles - one in the start line and one in the end line - maybe more if the range spans multiple lines. Open to ideas here in terms of how we can improve the layer API to make these scenarios easier to build!

Awesome work @Akin909!

badosu:
@Akin909 Is this documented somewhere?

Akin909:
sorry for the late response @badosu yes this is document in the configuration section in the experimental subsection

PR #2346, opened 2018-06-21
CI Server Updates

CrossR:
Updated ensureProcessNotRunning to run on both Nvim and Oni processes.

May help with the CI server issues on Mac (or at least killing the processes, won't change whatever bug/weirdness is causing it).

EDIT:

- Additionally, I added a skip to travis for MD only files.
- I also updated the CheckBinariesForBuild script to check for nvim binaries on Windows.
- Skip screenshots for PRs (Its failed a few times and seems unnecessary to me for a PR)
- Use Neovim 0.3.0 on Linux.
- Add mac OS 10.11 build (currently fails)

### DIFF
--- a/build/script/travis-test.sh
+++ b/build/script/travis-test.sh
@@ -23,7 +23,7 @@ fi
 
 echo Using neovim path: "$ONI_NEOVIM_PATH"
 
-if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+if [[ "$TRAVIS_OS_NAME" == "osx" && "$TRAVIS_PULL_REQUEST" != "false" ]]; then
    npm run demo:screenshot
 fi
 
--- a/build/script/travis-test.sh
+++ b/build/script/travis-test.sh
@@ -13,7 +13,7 @@ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
   sleep 3
 
   # Install Neovim
-  curl -LO https://github.com/neovim/neovim/releases/download/v0.2.2/nvim.appimage
+  curl -LO https://github.com/neovim/neovim/releases/download/v0.3.0/nvim.appimage
   chmod u+x nvim.appimage
   ./nvim.appimage --version
   ONI_NEOVIM_PATH="$(cd "$(dirname "$1")"; pwd)/nvim.appimage"
--- a/test/common/Oni.ts
+++ b/test/common/Oni.ts
@@ -151,6 +151,7 @@ export class Oni {
                 if (!didStop) {
                     log("- Attemping to force close processes:")
                     await ensureProcessNotRunning("nvim")
+                    await ensureProcessNotRunning("oni")
                     log("- Force close complete")
                 }
 
--- a/test/common/Oni.ts
+++ b/test/common/Oni.ts
@@ -19,7 +19,7 @@ const isCiBuild = () => {
     return ciBuild
 }
 
-const getWindowsExcutablePathOnCiMachine = () => {
+const getWindowsExecutablePathOnCiMachine = () => {
     switch (process.env.PLATFORM) {
         case "x86":
             return path.join(__dirname, "..", "..", "..", "dist", "win-ia32-unpacked", "Oni.exe")
@@ -33,7 +33,7 @@ const getWindowsExcutablePathOnCiMachine = () => {
 const getExecutablePathOnCiMachine = () => {
     switch (process.platform) {
         case "win32":
-            return getWindowsExcutablePathOnCiMachine()
+            return getWindowsExecutablePathOnCiMachine()
         case "darwin":
             return path.join(
                 __dirname,
@@ -149,7 +149,7 @@ export class Oni {
                 log("- Race complete. didStop: " + didStop)
 
                 if (!didStop) {
-                    log("- Attemping to force close processes:")
+                    log("- Attempting to force close processes:")
                     await ensureProcessNotRunning("nvim")
                     await ensureProcessNotRunning("oni")
                     log("- Force close complete")
--- a/test/common/Oni.ts
+++ b/test/common/Oni.ts
@@ -152,6 +152,7 @@ export class Oni {
                     log("- Attempting to force close processes:")
                     await ensureProcessNotRunning("nvim")
                     await ensureProcessNotRunning("oni")
+                    await ensureProcessNotRunning("chromedriver")
                     log("- Force close complete")
                 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ matrix:
   - os: osx # OSX 10.11
     node_js: 8
     osx_image: xcode8
+  allow_failures:
+    - osx_image: xcode8
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,11 @@ matrix:
     sudo: required
     dist: trusty
     node_js: 8
-  - os: osx
+  - os: osx # OSX 10.12
     node_js: 8
+  - os: osx # OSX 10.11
+    node_js: 8
+    osx_image: xcode8
 
 addons:
   apt:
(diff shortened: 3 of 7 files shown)


CrossR:
Guess it would be too easy for the Mac build to fail and let me see that the code changes are working...

It at least didn't break the builds!

Might do a few other tiny changes in this branch, to see if I can get the code changes to trigger.

CrossR:
@bryphe, I've updated the check in CheckBinariesForBuild to include Windows, which I think is the intended result. There was occasionally errors where the cache files would be restored fine, but Windows was only restoring the ripgrep binaries, so it would fail to launch nvim.

If I'm missing something and that isn't the intended behaviour, I can restore it.

CrossR:
I've also moved the Demo screenshot out of PR builds. Since it was labelled `demo` I assume thats fine (and I don't see much benefit in us building a screenshot  for PRs), as it was failing occasionally and stopping the build.

If its actually performing something useful for testing, I can add it back.

CrossR:
Finally got the Mac build to call the changed code, and it looks like the process stopping works fine, but now that Oni is being killed it causes an issue with the following code (I'm assuming it is this code as the other promise this races against is just a sleep)

```typescript
const promise1 = this._app.stop().then(
                    () => {
                        log("_app.stop promise completed!")
                        didStop = true
                    },
                    err => {
                        // tslint:disable-next-line
                        console.error(err)
                    },
                )
```

```
- Force close complete
- Calling _app.stop
- Racing with 15s timer...
{ Error: Request timed out after the element was still found on the page.
    at windowHandles() - application.js:245:17
  details: undefined,
  message: 'chrome not reachable',
  type: 'RuntimeError',
  seleniumStack: 
   { type: 'SelectorTimeoutError',
     message: 'Request timed out after the element was still found on the page.',
     orgStatusMessage: 'chrome not reachable\n  (Session info: content shell=)\n  (Driver info: chromedriver=2.29 (8e8216e581c512667203931f81c1a1ead47222e5),platform=Mac OS X 10.12.6 x86_64)' } }
- Race complete. didStop: false
```

What is interesting is that if you look at the final try to quit out (https://travis-ci.org/onivim/oni/jobs/395803384#L3831) we can see that it fails one last time, and moves to the next test, then in that next test, the promise seems to complete:

```
[379.069] - Calling oni.start
isCiBuild: true
Using executable path: /Users/travis/build/onivim/oni/dist/mac/Oni.app/Contents/MacOS/Oni
Using executable args: 
Start options: {"env":{"ONI_CONFIG_FILE":"/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/config_1529756087313.js"}}
Oni starting...
_app.stop promise completed!
```

Not fully sure how that promise completed that much further into the future, when the next test had started.

CrossR:
@bryphe do you have any input on what a suitable timeout for the CI tests is?

Currently we have two values I think, which causes issues like in https://travis-ci.org/onivim/oni/jobs/395911164#L3702.

runInProcTest.ts runs the tests with a 300000ms timeout (https://github.com/onivim/oni/blob/master/test/common/runInProcTest.ts#L148) but the actual node call defined in the package.json uses a 120000ms timeout. Ideally it would never be reached....but its probably best to have the in code one be the shorter one, so a retry can be performed.

bryphe:
Thanks for all your help investigating here @CrossR ! Really appreciate it 💯 Sorry I've been slow to reply.

> I've also moved the Demo screenshot out of PR builds. Since it was labelled demo I assume thats fine (and I don't see much benefit in us building a screenshot for PRs), as it was failing occasionally and stopping the build.

That's reasonable! I was hoping it'd be more stable, but yes, let's move it out - don't want that to hurt stability of the builds. I think @badosu logged an issue tracking this, too.

> Not fully sure how that promise completed that much further into the future, when the next test had started.

Hmm, I wonder if Spectron is doing something extra to close or something that causes the promise to resolve?

> runInProcTest.ts runs the tests with a 300000ms timeout but the actual node call defined in the package.json uses a 120000ms timeout. Ideally it would never be reached....but its probably best to have the in code one be the shorter one, so a retry can be performed.

Regarding the timeouts - I think we only hit the timeout in the case where there is a hanging process (I could be wrong, though). I'd be okay adjusting these, though. We could bump the code one down to 120000 to match.

test/common/Oni.ts
```diff
@@ -149,8 +149,9 @@ export class Oni {
                 log("- Race complete. didStop: " + didStop)
 
                 if (!didStop) {
-                    log("- Attemping to force close processes:")
+                    log("- Attempting to force close processes:")
```
bryphe: Thanks for fixing this 😄

build/script/travis-test.sh
```diff
@@ -13,7 +13,7 @@ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
   sleep 3
 
   # Install Neovim
-  curl -LO https://github.com/neovim/neovim/releases/download/v0.2.2/nvim.appimage
+  curl -LO https://github.com/neovim/neovim/releases/download/v0.3.0/nvim.appimage
```
bryphe: Nice catch!

bryphe:
And ya, I hadn't seen this error before on OSX before, either:
```
- Racing with 15s timer...
{ Error: Request timed out after the element was still found on the page.
    at windowHandles() - application.js:245:17
  details: undefined,
  message: 'chrome not reachable',
  type: 'RuntimeError',
  seleniumStack: 
   { type: 'SelectorTimeoutError',
     message: 'Request timed out after the element was still found on the page.',
     orgStatusMessage: 'chrome not reachable\n  (Session info: content shell=)\n  (Driver info: chromedriver=2.29 (8e8216e581c512667203931f81c1a1ead47222e5),platform=Mac OS X 10.12.6 x86_64)' } }
- Race complete. didStop: false
```

But I think it's due to the fact that we kill the `Oni` process while the `chromedriver` process is still doing something. The `spectron` framework we use for testing uses `chromedriver` to connect to Oni - and whenever we make a call to our automation it goes through `chromedriver`'s protocol. I wonder what happens if we try stopping the `chromedriver` process as well?

CrossR:
Not sure how much more I can do on this....

I've added the killing of `chromedriver`, but haven't managed to get an error since to actually test it. I also added the macOS 10.11 builds as optional for now.

As far as I can tell, the main outstanding issue is that the AppVeyor builds sometimes just don't have the cached binaries since the release isn't downloaded, nor the cache restored. If I'm remembering correctly, the releases sometimes aren't downloaded due to GitHub API rate limiting? But we evidently don't get that on the Linux builds, which also download things.

I could be missing something, but what is stopping us adding a `curl`/`Invoke-WebRequest` to the release downloader to get the things we need? Or at least to add optional "if CI use `curl` else use `github-releases`". Just missing what makes the Linux downloads and the Mac/Windows downloads different.

The other solution is the way Veonim does it, which is basically how we had it before.  The entirety of the binaries is stored in the repo, just in a seperate repo. Then they publish individual npm modules for each platform. Lot less clean...but also doesn't have the issues we are having on Windows CI. https://github.com/veonim/neovim for reference. Failing that, its storing it all somewhere, which I think we'd rather avoid.

bryphe:
Thanks for all your help with this, @CrossR ! 💯 

> I could be missing something, but what is stopping us adding a curl/Invoke-WebRequest to the release downloader to get the things we need? Or at least to add optional "if CI use curl else use github-releases". Just missing what makes the Linux downloads and the Mac/Windows downloads different.

Sounds good.  I thought that the `curl` / `Invoke-WebRequest` might be throttled in the same way, but that might not be case - definitely worth a shot 👍 

> The other solution is the way Veonim does it, which is basically how we had it before. The entirety of the binaries is stored in the repo, just in a seperate repo.

Yes, given all the issues we've had, I'm OK going back to that approach. It's not ideal, but it would definitely be more reliable. We could just stash the binaries in a `bin` folder in `oni-neovim-binaries` and `oni-ripgrep-binaries` and add them as a submodule. It completely gets us out of the caching shenanigans - so I'm cool with moving back to that approach!

PR #2355, opened 2018-06-25
Pass event names to all Event constructors

### LINKED ISSUE #2323: Two NeovimEditor instances, two nvim processes results in events being missed
Oni Version: This must be happening for many versions now
Neovim Version (Linux only): 0.3.4 and 0.3.5 from latest git
Operating System: Arch-Linux 

#### Describe your issue
When navigating with j,k a lot of errors are thrown in the console window because `editorManager.activeEditor.activeBuffer.` is null.
This happens because NeovimEditor's lastBufferId is null which in turn happens because
onBufEnter event is _seemingly_ never called.

Now that is the part things get complicated.

Event handling happens through oni-types Event class.
So the oni-types Event class instantiates a new EventEmitter for each instance of the Event class.
Because of this for subscribe and dispatch to work, the same instance of the Event class must be used by the subscriber(s) and the dispatcher(s) basically meaning that **all events must be singletons**, although no measures are taken to enforce this.
(This behaviour also eliminates the need for the name parameter, since it is ignored and not used anyway.)
_Of course if the name argument was provided for all events a single EventEmitter would be enough and the whole problem would be "patched", but the core of the problem is not the Event system by itself._

In the current state of the code, **some events have two instances**. One instance is dispatching, while the Editor is subscribed to the other. As of now I'm pretty sure that this happens with atleast all NeovimAutoCommands/Editor related Events.
A NeovimAutoCommands instance is created only in NeovimInstance.

_Now things get even more complicated and confusing._
 *Two* NeovimInstance instances are created that result in two NeovimProcessSpawner::startNeovim calls which basically runs nvim two times and I'm not sure if this is intended. Again it looks like NeovimInstance should be a singleton, but it isn't.

The first NeovimInstance is created in NeovimEditor which is a SuperClass of OniEditor and is instantiated in startEditors.
The second one is created in SharedNeovimInstance which is a NeovimInstance wrapper.
Now both startEditors and SharedNeovimInstance reside in App.ts so I guess the intended behaviour would be to pass the SharedNeovimInstance's NeovimInstance to startEditors or something similar.

I would make a PR but I'm not sure how to proceed since the intended behaviour is not that clear with this.

#### Expected behaviour
Events handled and program working smoothly.

#### Actual behaviour
Hundreds of errors spamming the console and at least all NeovimAutoCommand events are missed.

#### Steps to reproduce
Open any file, navigate with j,k in visual mode, with developer tools open.

GeorgeTG:
Pass names to all event constructors, first step to fix #2323
Also properly dispose some Event Handlers in tests.

### DIFF
--- a/browser/src/Editor/Editor.ts
+++ b/browser/src/Editor/Editor.ts
@@ -24,13 +24,17 @@ export interface IEditor extends Oni.Editor {
  */
 export class Editor extends Disposable implements Oni.Editor {
     private _currentMode: string
-    private _onBufferEnterEvent = new Event<Oni.EditorBufferEventArgs>()
-    private _onBufferLeaveEvent = new Event<Oni.EditorBufferEventArgs>()
-    private _onBufferChangedEvent = new Event<Oni.EditorBufferChangedEventArgs>()
-    private _onBufferSavedEvent = new Event<Oni.EditorBufferEventArgs>()
-    private _onBufferScrolledEvent = new Event<Oni.EditorBufferScrolledEventArgs>()
-    private _onCursorMoved = new Event<Oni.Cursor>()
-    private _onModeChangedEvent = new Event<Oni.Vim.Mode>()
+    private _onBufferEnterEvent = new Event<Oni.EditorBufferEventArgs>("Editor::onBufferEnterEvent")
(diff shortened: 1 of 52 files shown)


Akin909:
@bryphe when you get a chance would be great to get your input on this issue as it affects quite a bit how events are used in oni as well as `oni-types`

PR #2360, opened 2018-06-26
Feature/use local prettier

### LINKED ISSUE #2348: Prettier plugin should prioritize project or global prettier installation over the plugins
I've recently noticed that depending on the project I'm working on I get different prettier behaviour which I've narrowed down to prettier version mismatches.

The plugin should instead of only ever using its own version of prettier which may be ahead or behind of a users, it should search for a project prettier if there isnt one it should then search for a global prettier and lastly fallback to the one bundled in oni.

Akin909:
This PR fixes #2348, rather than always using the default prettier bundled with the plugin it uses the local prettier version and falls back to the bundled one if a local installation isn't present also. 

All credit to the `prettier-vscode` plugin I basically fully borrowed their implementation.


bryphe:
Awesome! Makes sense to use local prettier if it's available - looks good to me 👍

badosu:
@Akin909 Does this fix #2255 ?

Akin909:
@badosu I basically took the opportunity to just remove the background color prop on the prettier status item and the foreground should just be white I believe so hopefully this solves that

Akin909:
@CrossR I imagine its just a project without prettier, I haven't seen it before but its likely due to this I'll have a look after work I've used a build with this PR in in a project without prettier and haven't encountered it could you share the package json? probably the function I borrowed throws an error at some point

CrossR:
image

It happens for me in Oni as well, at least on a machine that doesn't have Oni's node_modules downloaded etc.

Akin909:
Ahh in which case is prettier specified in the package.json but not available in the node modules?

CrossR:
Ah! Yep, my dev machine with Oni on and node_modules downloaded this works fine, but for my local machine prettier is specified but not available.

PR #2366, opened 2018-06-27
Fix error when using Open With option in OSX finder

### LINKED ISSUE #2041: "A JavaScript error occurred in the main process" - Opening a file before the app is opened.
I get the following message:

```
A JavaScript error occurred in the main process

Uncaught Exception:
Error: Cannot create BrowserWindow before app is ready
    at createWindow (/Applications/Oni.app/Contents/Resources/app/lib/main/src/main.js:124:25)
    at App.electron_1.app.on (/Applications/Oni.app/Contents/Resources/app/lib/main/src/main.js:189:9)
    at emitTwo (events.js:125:13)
    at App.emit (events.js:213:7)

```

Steps to reproduce:

1. Assign a file extension to "Always open with..." e.g. `.js`
2. Double click a file set to be opened by Oni by default (I'm using .js in this example but seems to apply to any file).
3. Error above displays.

Details:

- System: macOS 10.13.3, MacBook Pro Mid 2015
- Oni: v0.3.2
- Neovim: v0.2.2 (installed via home-brew)

psxpaul:
Fixes #2041. I've never done anything with Electron or Typescript, so let me know if I'm doing anything crazy here 😃.

### DIFF
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -260,8 +260,15 @@ app.on("open-file", (event, filePath) => {
     if (activeWindow()) {
         activeWindow().webContents.send("open-file", filePath)
     } else if (process.platform.includes("darwin")) {
-        const argsToUse = [...process.argv, filePath]
-        createWindow(argsToUse, process.cwd())
+        const argsToUse = [...process.argv.slice(1), filePath]
+
+        if (app.isReady()) {
+            createWindow(argsToUse, process.cwd())
+        } else {
+            app.on("ready", async () => {
+                createWindow(argsToUse, process.cwd())
+            })
+        }
     }
 })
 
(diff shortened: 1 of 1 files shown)


main/src/main.ts
```diff
@@ -260,8 +260,15 @@ app.on("open-file", (event, filePath) => {
...
         activeWindow().webContents.send("open-file", filePath)
     } else if (process.platform.includes("darwin")) {
-        const argsToUse = [...process.argv, filePath]
-        createWindow(argsToUse, process.cwd())
+        const argsToUse = [...process.argv.slice(1), filePath]
+
+        if (app.isReady()) {
+            createWindow(argsToUse, process.cwd())
+        } else {
+            app.on("ready", async () => {
```
Akin909: @psxpaul thanks so much for the fix 👍 , One quick note is that the `app.on("ready")` callback is an async function in your change but we dont await `createWindow` which I believe is sync, its I think, a minor point but is it intentional?

psxpaul: Thanks for the feedback! I removed the async keyword and re-tested.

PR #2371, opened 2018-06-29
Feature/upgrade ocaml lsp

Akin909:
Another attempt to upgrade the `ocaml-lsp` hopefully there is no interaction with the css lang server this time although not sure that was spurious last time ^^


bryphe:
Thanks for keeping this up-to-date here @Akin909 ! Really appreciate it 👍 
 
I'm interested to see if we could potentially pick this up for OCaml & Reason support? https://github.com/jaredly/reason-language-server

Some cool things about it:
- LSP is natively implemented in Reason, so no change of mismatched dependencies
- Works on Windows! (Big plus for me 😄 )

Akin909:
@bryphe I was actually looking at exactly the same thing just before I made this PR as it seems like a really cool project, what I saw though was tbh other than as a vscode extension I couldn't tell from the project docs how we could integrate i.e. is there an npm package what flags need to be passed etc. also from the issues on the project it definitely seemed like more of a work in progress than our current solution and tbh I'm not sure the author intends to match the same feature set e.g. https://github.com/jaredly/reason-language-server/issues/21

I definitely think it would be a really cool project to keep an eye on as an alternative ? preferable option but I think beyond not knowing how to integrate I also think anyone who writes reason or ocaml in oni would loose a few bits of functionality

Akin909:
That being said 🤣 we still seem to be getting an error with the `ocaml-lsp` which I think just the css test is picking up but checking it locally an error does seem to be thrown trying to resolve a sub dir of the vscode-language-server package 🤷‍♂️ in which case waiting to see how the `reasonml-language-server` progresses and if we can intergrate it, so will close this

bryphe:
Ya unfortunately I think there is still a conflicting dependency here 😦 Thanks for looking at this @Akin909 !

For the `reason-language-server` - when you run `npm run build`, it actually drops an `exe` that follows the language server protocol. You get it it in `lib\bs\native\bin.native.exe`. So basically we'd use that as the language server.

jaredly:
It is definitely a work in progress :) but if you're interested in trying it, I publish built binaries for linux/windows/macos on the releases page https://github.com/jaredly/reason-language-server/releases

PR #2372, opened 2018-06-29
Sort out CLI for Oni

CrossR:
This PR aims to address some of the points outlined over in #1707.

The first thing I'm aiming to fix is the ability to launch in the background, by fixing/editing the existing launch script (well currently I've deleted the old one, but this is very WIP so I'm just seeing what works).

I've based it on the VSCode script that is outlined in the linked issue.

Todo:

* [x] Windows Script.
  * [ ] Look into double click issue (ie closing powershell takes two clicks of the close X?
* [x] Linux Script.
* [x] macOS Script (we do have one already, so I'll see about editing this etc)
* [x] Check existing functionality isn't changed:
  - Normal opening (Invoking Oni directly with a click)
  - Opening a file that Oni is registered for.
  - Opening a file via `Edit with Oni` (Windows only I think)
  - Launching via Dev.
* [x] Sort out the meshing of  the current CLI functionality and the new changes, ensure swapping folders, opening files and such is okay with the change.

I'm considering the following but will probably just leave them to a second PR:
- The updated workspace detection (may be simple enough when I'm checking/sorting the existing logic, so I'll see).
- Add the option to pass neovim flags.
- Instance management.

There is a very basic implementation I've done already that works on Windows, so seems like a good start, so hopefully it works across the other platforms.


CrossR:
Think the basics are done for the 3 platforms (at least from the very quick testing I did).

CrossR:
I've done a lot of testing with this on Windows, and bits on Mac/Linux, but I'd appreciate if anybody else could give it a test.

I'm going to ignore the powershell issue for now...since I couldn't find any mention of it elsewhere online.

@badosu probably worth you being aware of this since I know you maintain the AUR version of Oni, and this changes what is used to launch Oni to `cli/linux/oni.sh`.

bryphe:
Change looks great, @CrossR ! Thank you for all your work on this! 💯

CrossR:
@bryphe, I've actually just realised that the `Add to PATH` functionality on macOS isn't hooked up.
Its probably best to address that here in a basic way so it can be tested more easily.

I'm aiming to just check the symbolic link and make sure it points to `cli/mac/oni.sh`. For something more stable, we might want a check like @Akin909 `init.vim` check, so when people launch they know  to re-run `Add To Path`.

bryphe:
Thanks for catching that! That plan sounds reasonable to me. I'll tag this so I know to call it out in the release notes, as well 👍

CrossR:
Had a stupid logic error, but fixed that now and it works to add to the path on mac.

I'll leave potentially adding an on launch check to a later PR, I'd rather stick with the CLI stuff whilst I'm in the area.

bryphe:
> Had a stupid logic error, but fixed that now and it works to add to the path on mac.

Ah I missed that too when I looked at the code!  Glad you caught it.

> I'll leave potentially adding an on launch check to a later PR, I'd rather stick with the CLI stuff whilst I'm in the area.

Sure, sounds reasonable as a separate PR - this change looks great 👍

badosu:
@CrossR I'll check th CLI for any problems with the AUR wrapper script.

De we have a standard interface that I can check list? Nothing high level, just the most important bits, e.g. *What happen when oni `...` is run`.

Thanks!

CrossR:
>De we have a standard interface that I can check list? Nothing high level, just the most important bits, e.g. *What happen when oni ... is run`.

Not really right now, and to be honest it will hopefully be changing in the next few PRs to match what #1707 has outlined, so it probably makes more sense for the documentation to wait until I'm done.

All this PR should change is that you can close the terminal without closing Oni now. Everything else should be the same as the previous versions of Oni from the CLI. Non-CLI launching shouldn't have been changed at all.

PR #2374, opened 2018-06-29
add slash back for cmdline search

Piping:
Hi,

vim has `//` syntax for last search. However, the Oni command line search will not display the first character like `/` or `?` .

This pull request is to restore the behavior.

### DIFF
--- a/browser/src/UI/components/CommandLine.tsx
+++ b/browser/src/UI/components/CommandLine.tsx
@@ -108,11 +108,13 @@ export class CommandLine extends React.PureComponent<ICommandLineRendererProps, 
                         iconName="arrow-right"
                         key={`${character}-arrow-right`}
                     />,
+                    character,
                 ]
             case "?":
                 return [
                     <CommandLineIcon iconName="search" key={`${character}-search`} />,
-                    <CommandLineIcon key={`${character}-arrow-left`} arrow iconName="arrow-left" />,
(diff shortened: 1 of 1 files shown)


Akin909:
@Piping there is an option which possibly isn't sufficiently well documented with the commandline externalisation to actually disable the use of the icons if you would prefer to see the original vim characters.

The intention was that users can opt in to gui icons if they like otherwise they can stick with the vim characters the option I believe is `"commandline.icons": false`

Piping:
@Akin909  I like the icon! And I want to keep it. `/` or `?` is not just a character that active the search, but for vim users, it can also mean last searched text. E.g. `//` or `??`

Akin909:
@Piping the though process behind the icons was that it would be the one or the other aka if new users happened upon the command line they would get a gui icon they could guess at functionality for from the icon and power vim users would just turn it off I think a hybrid of the two sort of defeats the intention, the icons are supposed to be a gui stand-in for `/` and `?`

Piping:
@Akin909 gui icon add more info for users to understand whats going on. But icon definitely not replace the `/`. I was very surprised when I pressed `/` and did not see it on the screen. I have to guess what's happened. 

consider the case, when the user coming from stack overflow, and learn how to use search in vim.
and then they open oni and press `/`, they probably used to the cmdline indicating `/` is pressed.

the search pattern always looks like `/abc` or `//` etc.

PR #2378, opened 2018-06-30
Automation: Install tests (companion to #2372)

bryphe:
This is a draft of some quick tests that validate the setup install / uninstall behavior (creating a directory, dropping an executable, running the uninstaller). It also queries the `PATH` registry key to check if it looks as expected. Hopefully it helps with validation for #2372 ! 💯


CrossR:
The caches on AppVeyor seem to keep failing.... no idea why, they just aren't being restored at all.

That said, should be able to uncomment the final test now my PR is in.

PR #2390, opened 2018-07-02
Workaround for devtools keyboard shortcuts not working on OSX

### LINKED ISSUE #2347: devtools clipboard copy does not work
**Oni Version:** 0.3.5 (and earlier)
**Neovim Version (Linux only):** 0.3.0
**Operating System:** macOS

**Issue:**

Copying text to clipboard doesn't work in devtools.

**Steps to reproduce:**

1. open devtools
2. select the Console tab
2. select some log text in the Console log
3. type cmd-c or ctrl-c, or select Edit ... Copy from the system menu. 
    - Also, right-click menu doesn't show Copy/Paste when text is selected.
4. Try to paste the text using the system clipboard (cmd-v or ctrl-v).

psxpaul:
Fixes #2347 using the workaround suggested in [the Electron bug](https://github.com/electron/electron/issues/11998)

### DIFF
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -174,6 +174,26 @@ export function createWindow(
         width: windowState.bounds.width,
     })
 
+    // Workaround for cut/copy/paste/close keybindings not working in devtools window on OSX
+    if (process.platform === "darwin") {
+        currentWindow.webContents.on("devtools-opened", () => {
+            currentWindow.webContents.devToolsWebContents.executeJavaScript(`
+                window.addEventListener('keydown', function (e) {
+                    if (e.keyCode === 65 && e.metaKey) {
+                        document.execCommand('Select All');
+                    } else if (e.keyCode === 67 && e.metaKey) {
+                        document.execCommand('copy');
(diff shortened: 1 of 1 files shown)


main/src/main.ts
```diff
@@ -174,6 +174,26 @@ export function createWindow(
         width: windowState.bounds.width,
     })
 
+    // Workaround for cut/copy/paste/close keybindings not working in devtools window on OSX
+    if (process.platform === "darwin") {
+        currentWindow.webContents.on("devtools-opened", () => {
+            currentWindow.webContents.devToolsWebContents.executeJavaScript(`
```
bryphe: This is really cool, didn't know this was possible! 👍

bryphe:
Just tested it out - works great @psxpaul ! Nice find on the workaround - major quality of life improvement  💯 

Merging the PR now - thanks for the contribution!

main/src/main.ts
Magistern: Cool, but it doesn't work for "quit" CMD+Q will not close window if focus is on the devtools. It used to in previous version of electron.

PR #2392, opened 2018-07-03
Reduce app folder size.

CrossR:
Currently, we are packing some folders that I don't think we need to into the final installers.

This includes `ui-tests`, the `main` src folder, `.github` folder, but the worst offender is the `.oni_build_cache` folder. On my machine with just `ripgrep` in it, its 2MB, but I think on Mac where `nvim` is also cached, it could be almost 50MB, which is a big chunk of the total size of that folder. I'm basing that on the fact `nvim` for Windows is about 45MB.

Assuming this doesn't break everything, is there anything else we can remove? The `README.md` etc is also included and so forth...is it worth swapping from an "Everything but ..." to specifying the individual folders we need? We currently include the `.map` files, which I'm not sure how much value they add.

Setting up this PR to test the CI builds, since that cache is only added for them. Its potentially also worth us adding some size constraints? Or maybe warnings if the installer size deviates from some set value?


bryphe:
Great catch @CrossR ! Thanks for cleaning this up 😄

> but the worst offender is the .oni_build_cache folder. On my machine with just ripgrep in it, its 2MB, but I think on Mac where nvim is also cached, it could be almost 50MB, which is a big chunk of the total size of that folder. I'm basing that on the fact nvim for Windows is about 45MB.

Wow, that's massive - really glad you caught that! 💯 I noticed the Mac `dmg` was significantly larger than the other platforms last release... this explains it.

> The README.md etc is also included and so forth...is it worth swapping from an "Everything but ..." to specifying the individual folders we need? We currently include the .map files, which I'm not sure how much value they add.

I like this idea - it's definitely the right strategy - the 'everything but...' strategy is clunky and unreliable (and makes it easy for some of these issues to sneak in!). 

We shouldn't need the `.map` files for release - they're only used for debugging. The only challenge I see with the 'whitelist' is specifying the set of `node_modules ` we need - but I think we can just use the `dependencies` in our `package.json`. Would be awesome to switch over to explicitly specifying what we pack for our release packages.

> Setting up this PR to test the CI builds, since that cache is only added for them. Its potentially also worth us adding some size constraints? Or maybe warnings if the installer size deviates from some set value?

That's a great idea! I like the idea of having a test that validates that the installer isn't greater than a certain size - it builds up our safety net to keeps issues like this from popping in later, unexpectedly.

build/BuildSetupTemplate.js
```diff
@@ -20,7 +20,7 @@ const packageJsonContents = fs.readFileSync(path.join(__dirname, "..", "package.
 const packageMeta = JSON.parse(packageJsonContents)
 const { version, name } = packageMeta
 const prodName = packageMeta.build.productName
-const pathVariable = "{app}\\resources\\app\\cli\\windows\\"
+const pathVariable = "{app}\\resources\\app\\cli\\win\\"
```
CrossR: I swapped this to take advantage of the `${os}` variable in `electron_builder`, such that we only include the CLI script for the needed platform.

bryphe:
Just checked it out on Mac, @CrossR - looks great! Thank you very much for your work on this 👍 

I like that we have the whitelist/include strategy now - that  make something like #350 easier, as we can build the `asar` ourselves and whitelist it in without worrying about double-packing it.

PR #2394, opened 2018-07-03
Add indents to every line in the snippet.

### LINKED ISSUE #2242: Snippets don't auto indent
Inserting a snippet on anything but col 1 doesn't preserve indentation beyond first line.
Example:
Snippet code:
image

From col 1:
image

From col 5:
image

CrossR:
Fixes #2242.

Adds the correct indent level to every line in a snippet.

Is there any edge cases I'm missing where this could fall apart?

### DIFF
--- a/browser/test/Services/Snippets/SnippetSessionTests.ts
+++ b/browser/test/Services/Snippets/SnippetSessionTests.ts
@@ -50,7 +50,7 @@ describe("SnippetSession", () => {
             assert.strictEqual(firstLine, "somefooline")
         })
 
-        it("matches existing whitespace - 2 spaces", async () => {
+        it("matches existing whitespace for first line - 2 spaces", async () => {
             snippetSession = new SnippetSession(mockEditor as any, "\t\tfoo")
 
             const indentationInfo = {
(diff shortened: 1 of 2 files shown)


Akin909:
@CrossR I think if the snippets are subdivided by line then adding an indent should be fine worst case a line might wrap because of indentation but I dont think anything would blow up, also from the indent lines plugin Im pretty sure detect indent handles empty lines fine as well

CrossR:
Just gave this an actual test (vs just a visual test of the lines) and realised the way I was doing it wasn't updating the placeholders, so tabbing through left you in the old positions.

I'm pushing up a change for that now, that fixes it on my end.

CrossR:
Unit tests added now, is it worth adding some for the tabbing through placeholders? Technically I didn't actually change any of that, since the stuff I've added is all done before that.

browser/test/Services/Snippets/SnippetSessionTests.ts
```diff
@@ -119,6 +119,54 @@ describe("SnippetSession", () => {
...
+
+        // Add a line, and move cursor to line
+        mockBuffer.setLinesSync(["\t"])
+        mockBuffer.setCursorPosition(0, 1)
+
+        await snippetSession.start()
+
+        const [firstLine, secondLine, thirdLine] = await mockBuffer.getLines(0, 3)
+
+        assert.strictEqual(firstLine, "\tfor {")
```
bryphe: Nice! Thanks for adding these tests! 💯

bryphe:
> Unit tests added now, is it worth adding some for the tabbing through placeholders? Technically I didn't actually change any of that, since the stuff I've added is all done before that.

This is great! Just glad we have the indentation behavior covered. I think if we find bugs with the placeholder behavior we could add some additional ones to cover it, at that time.

PR #2402, opened 2018-07-05
Fix badges.

CrossR:
The badges were broken, and I think I've now got it pointing to the correct place.
Was surprisingly confusing, there is a lot of brackets in there...

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 ### Modern Modal Editing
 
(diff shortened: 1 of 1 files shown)


CrossR:
Thats a weird result of the AppVeyor skip....it doesn't seem to complete, meaning I wouldn't be able to merge this. Very odd, wonder if that is intentional, or not setup correctly.

bryphe:
Shoot, thanks for fixing that @CrossR 💯 

Yes - that reporting seems busted... It must be something broken with the migration, since this case was working before AFAIK.

I opened a support ticket here: https://help.appveyor.com/discussions/problems/14969-ci-status-reporting-as-waiting-for-commits-that-are-skipped

IlyaFinkelshteyn:
Related discussion https://github.com/appveyor/ci/issues/1848

bryphe:
Thank you @IlyaFinkelshteyn !

For our purposes - we'd want the skipped commits to show as `passed`. Would it be possible to have a configurable option for that, like in `appveyor.yml`?

For example, having a flag like `pass_on_skip: true`: 
```
# Skip CI build if the changes match these rules exactly.
# Ie, if the BACKERS.md file is changed, we don't need to build.
skip_commits:
   pass_on_skip: true
   files:
    - '**/*.md'
```

bryphe:
I'll merge this in, but I'll keep tracking in issue #2403 on our side. Thanks @CrossR !

IlyaFinkelshteyn:
@bryphe  thinking about UI-only option so people cannot intentionally commit "merge-able" trash.

PR #2404, opened 2018-07-06
Fix: quit the application with Cmd + Q

ackey:
On mac we should quit the application when the user press Cmd + Q.
Currently we need to use Cmd + Q twice to quit the application.
This PR fixes it.


bryphe:
Thanks @ackey and @CrossR ! Really appreciate the contribution + review 💯

 Looks like the builds passed now - bringing this in 👍

PR #2408, opened 2018-07-06
added dracula theme

AlexanderElias:


CrossR:
Since its nice to have an image for themes:

image

Looks good to me though! Only point I have is maybe having the tabs stick out from the background a bit more? As it is, the tabs are hard to distinguish from the background colour. I set `        "background": "#191A21",` to get it a little darker.

Also, is there a reason for using Dracula 1.2.7 vs the latest version (which I think is 1.5.0) : https://github.com/dracula/vim/blob/master/colors/dracula.vim. Thats where I got the darker background colour from as well.

AlexanderElias:
Sounds good. I updated to the darker color and fixed the old vim color version. 

I must have downloaded an old commit of the repo.

bryphe:
Just tested out the update - looks great @AlexanderElias ! 💯 

image

I like the mode highlight colors a lot (and I like the look of the background with the darker color).

Thanks for the contribution @AlexanderElias - bringing it in now!

PR #2413, opened 2018-07-08
Feature/Color layer

Akin909:
Having had a go at implementing a layer in #2344, I've also implemented a layer to highlight css colors in certain files. The way that this works is that if a file extension is in the whitelist any hex/rgba/hsla color is colored.

screen shot 2018-07-08 at 23 10 12

Todos:
- [x] Experimental flag and user config options for file extensions
- [x] Doesn't match the css color names aka `rebeccapurple, goldenrod` etc.
- [x] Ci Test

Outstanding Issues:
~- The layer renders on top of the renderer so the color highlight lies on top of the text obscuring the syntax highlight, currently I reduce the colors opacity but this actually changes the color so its not quite accurate
	- Potentially I could put the matching color into the element but set the `color` prop to transparent so it makes a word shaped window 😆
	- Another idea would be just to render the text inside the highlight using the user's font~

- The regex tries to fuzzy match inside a string this can lead to ?rare situations where things are incorrectly highlighted - for example html entities can contain what the regex would recognise as a hex string.

This is a tough one as its beyond my abilities with regex to not break the regex but have it fail if it matches inside another *word* I think possibly using some sort of word boundary `\\b` check might work but would rather not try and fix this right off the bat as its a potential rabbit hole.

screen shot 2018-07-08 at 20 57 32

- Currently wouldn't handle wrapping at all as the logic used in #2344 is not generalised so would have to be re-implemented here


CrossR:
>The regex tries to fuzzy match inside a string this can lead to ?rare situations where things are incorrectly highlighted - for example html entities can contain what the regex would recognise as a hex string.

I think leaving this for a second PR makes sense, but just in case its useful, have you considered trying to use the text mate grammar scopes for this? You could potentially restrict it to just strings etc.

I've just given it a try, and looks good! Only two comments I have are potentially adding some error checking, and how the layers are applied to the editor, though I expect that is an API question.

I added `.vim` to my list of filetypes for this, and upon opening the `onedark.vim` file, the editor crashed to the "Oh no!" screen. I probably shouldn't have tried opening a different filetype, but I also wouldn't have expected the editor to die.

The second point is more about the API I think, but due to this being a layer that sits on top, the cursor is hidden when it enters the string, meaning you can't actually see where you  are in the string. (I can just about see the cursor since the highlight box doesn't seem to be exactly aligned with the cursor, but still).

Akin909:
@CrossR thanks for trying this out 👍, tbh for this first thought was using a grammar but tbh I think it will make this process way more expensive I created what should be fairly multi purpose `tokenForLine` fn when doing the hover syntax highlighting and tbh the process is async calling in a grammar for the language and then it would involve parsing the entire string then looking for a color token which tbh could have a variety of tokens assigned to it depending on the language.

I think this regex is fairly sound but just needs a tweak for the edge case, it could turn out to not be the case in the long term but for now the process is way more straightforward, I can have a go later on to see how it can be tweaked.

Re. the cursor thats a very good point I didn't realise that. I didn't put a `z-index` on the highlights but I imagine the cursor sits over things in a very similar manner so can be covered, maybe it could hopefully be as simple as assigning it a high z-index

Re. the vim filetype error can I ask what the error was I'm not sure what case would have caused the error there the function should only run if the file hits any color matches

CrossR:
I added `.vim` to the list of extensions and opened `onedark.vim` in the repo and get the following:

image

I can get the output from the un-minified build in a bit if you aren't able to reproduce it.

Weirdly, `solarized.vim` worked fine, so it must be some colours they have defined just in one dark.

Akin909:
Pushed a fix and tested it out in `onedark.vim` seems to have resolved it was down to the color names also just added a try catch to the whole thing so it doesn't throw errors all the way up

Akin909:
Small explainer of added changes slightly unrelated changes, because of the cursor hiding issue I added a `StackLayer` styled component that takes a `z-index` as an optional argument, i've set the cursor to have a z-index of 2 and the overlay to have one of 3, the overlay is higher because the wild menu etc should cover the cursor but without a higher zIndex the cursor pokes through

PR #2423, opened 2018-07-13
Remove 'Split -> File Explorer Split' menu item

keforbes:
Oni's menu items were originally created to mimic GVim's as closely as possible.  One of those items was "File Explorer Split" which just ran the `:Lexplore` command.  Now that we have a legitimate File Explorer there's no reason for this menu item to exist.  Besides, if there is some Vim expert who prefers `:Lexplore` over our File Explorer they wouldn't have used the "File Explorer Split" menu item anyway.

### DIFF
--- a/main/src/menu.ts
+++ b/main/src/menu.ts
@@ -634,12 +634,6 @@ export const buildMenu = (mainWindow, loadInit) => {
                 },
             },
             {
-                label: "File Explorer Split",
-                click(item, focusedWindow) {
-                    executeVimCommand(focusedWindow, ":Lexplore | vertical resize 30")
-                },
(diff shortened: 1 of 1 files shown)


keforbes:
Yes, we can call `executeOniCommand` but personally, I don't think that's the right way to go about it.  If we really wanted a menu item for the explorer I'd prefer adding items like "Show Explorer", "Show Search", "Show Tutorials" to jump directly to those panels.  But I don't know what benefit a menu item would have over just clicking the icons.  Also, clicking the icons already toggles those panels.

Akin909:
@keforbes that makes sense 👍

Issue #1625, opened 2018-02-24
Skipping of closed brackets fails when some characters are present


succcubbus:
When `autoClosingPairs.enabled` is true any closing bracket can by typed over (e.g. for `(|)` typing `)` results in `()|`). This behaviour does not work when some characters are present in the same line.

How to reproduce:
- insert an `ä` in any way
- type `()` behind it
- expected: `ä()`
- actual: `ä())` with cursor between the two closing brackets

This only happens if the problematic character is preceding the closing bracket in the same line. It does not matter if the character was typed or was already present. The character must not but can be immediately preceding the brackets for this problem to occur.
Some non-working characters: `´`, `°`, `§`, `ß`, `ä`.
I'm using a german keyboard, and these are all available characters. I believe all characters available on the us keyboard are working fine.

oni v0.3.0 on linux

succcubbus:
closing due to inactivity and oni2

PR #2427, opened 2018-07-15
Fix cursor position when multi-byte characters are in the line

succcubbus:
The vimscript `col()` and `getpos()` functions return the byte the cursor is at, as opposed to the character index.
Inside Oni this cursor value is often used to index into the line content.
That failed whenever any multi-byte character was present in the line.
This value need to be in byte-form for setting the cursor-position though.

As vim does not have a nice way to access the character index, I used the following workaround:
`strchars((getline('.') . '.')[0:col('.') - 1])`

`(getline('.') . '.')`
get line content and append a dot to lengthen it by one (because columns are one indexed in the std `col()` function)
`[0:col('.') - 1]`
slice the line from beginning to selected column (-1 because the slice is inclusive)
`strchars(...)`
count the characters

When I tested this with very long lines I couldn't notice a significant speed difference in moving the cursor.
I'm open to a nicer solution, but couldn't find a vimscript function to get the character index the cursor is on directly.

This affected all Oni code which needed to analyse text under the cursor (e.g. completion, highlighting, hovers, rename, ...)
Example of badly matched code:
image
Highlighting is still broken even with this pr, because I couldn't figure out how it works and it is not dependent on the cursor.

Fixes https://github.com/onivim/oni/issues/1625

### DIFF
--- a/browser/src/Editor/BufferManager.ts
+++ b/browser/src/Editor/BufferManager.ts
@@ -213,9 +213,13 @@ export class Buffer implements IBuffer {
     }
 
     public async getCursorPosition(): Promise<types.Position> {
+        return this.getPosition(".")
+    }
+
+    public async getPosition(element: string): Promise<types.Position> {
         const row: number = await this._neovimInstance.callFunction("line", ["."])
         const column: number = await this._neovimInstance.eval<number>(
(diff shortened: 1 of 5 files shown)


succcubbus:
closing due to inactivity/oni2

PR #2429, opened 2018-07-15
Explorer: Locate Current Buffer

feltech:
Add command palette option "Explorer: Locate Current Buffer" that will
expand directories down to the current buffer, then select the
corresponding file/directory.


feltech:
As I'm entirely new to the Oni codebase I thought I would submit this before I tackle writing tests, in the hopes someone can take a look and let me know if my approach is reasonable.

feltech:
@Akin909 Thanks for the quick feedback!  

I've changed the SUCCESS action logic so that it doesn't trigger unnecessarily.  It seems odd that dispatching such a simple action (which isn't part of the epic) would slow things down.  When I tried I could maybe feel some slight lag, but not particularly noticeable.  Perhaps my machine is just able to absorb the extra effort.

I do find the Explorer pane is a bit buggy when scrolling down beyond the last visible file. The pane seems to scroll jump up and down.  That bug is present in the current stable release, which I'm using for development, so I'm not too concerned that my changes broke it.  I'll try to remember to check if there's an existing bug report and if not I'll make one.

Activating the Explorer if it's closed is a nice idea, I'll add that asap.

browser/src/Services/Explorer/ExplorerSelectors.ts
```diff
@@ -50,6 +50,20 @@ export const isPathExpanded = (state: IExplorerState, pathToCheck: string): bool
     return !!state.expandedFolders[pathToCheck]
 }
 
+/**
```
Akin909: @feltech a version of this same functionality already exists in [`ExplorerStore.ts`] (https://github.com/Akin909/oni/blob/c8a6051e390449873a1ed80e385643ad11adfe8b/browser/src/Services/Explorer/ExplorerStore.ts#L392) maybe you could import that and use it instead

Akin909:
Re. the flicker in the explorer that isn't because of your changes its a known bug i've opened #2182 in order to fix that

feltech:
@Akin909 - well spotted on the duplicated `nodePath` function, thanks for that.

Now the Explorer will open if it was closed when locating a file.

Akin909:
@feltech just tried it again works really well and now opens the explorer, thanks for adding that 👍. re. testing I don't think you need too many here my recommendation would be just testing the reducer and epic, the other tests are [here](https://github.com/Akin909/oni/blob/local-build/browser/test/Services/Explorer/ExplorerStoreTests.ts) in terms of examples of how those work

feltech:
Still more tests to write.  That weird `redux-mock-store` bug was an odd one.

Akin909:
@feltech thanks for adding so all those tests 🎉

Akin909:
@feltech is this complete?

feltech:
Nearly, working on the final test now (though it's late here so might take another day).

feltech:
OK, assuming checks/review pass, we're good to go.  

I had a problem running the entire integration test suite locally - it stops on `Configuration.TypeScriptEditor.CompletionTest.ts`, where the autocomplete never shows.

I don't think anything I did affects that, so I'm hoping it's just my local machine is misconfigured (I suspect the vim-session plugin is causing problems).

feltech:
Hmm, this OSX integration test failure is interesting.  The buffer's path (`oni.editors.activeEditor.activeBuffer.filePath`) begins with `/var/`.  Whereas the workspace root (`oni.workspace.activeWorkspace`) is `/private/var`.  If I remember the OSX file system correctly, `/var` symlinks to `/private/var` (maybe the other way around).  So a naive check for the buffer's path being a descendent of the root fails.

feltech:
These integration tests are annoyingly unstable.  In this most recent case a test passed on OSX, then ran again and failed, failing the job.  Plus the Linux test didn't even get past installing packages.

I think it would be worth removing the need to run the test twice if it succeeds, and only run multiple times on failure.

I also think it would be nice to remove the `--bail` parameter, so that all tests run even if they fail, to get a better idea of what needs fixing.

Akin909:
@feltech 

> These integration tests are annoyingly unstable. In this most recent case a test passed on OSX, then ran again and failed, failing the job. Plus the Linux test didn't even get past installing packages.

Unfortunately they are still quite flaky I think because of all of the various dependencies and having to open an editor cross platform it results in a lot of false negatives. I agree it would be nice to have them all run I think the idea though was to save time rather than having a full long test suite run this way we find out if theres an error as soon as possible would be interested in @bryphe opinion

Re. this PR ill re-run the tests a couple of times to see if the error resolves thanks for bearing with it will be great to have this in

feltech:
Yeah it works locally for me, but I'm on Linux.  It consistently fails on OSX.  I'll try to get some time with an OSX box tomorrow to check it out.

Akin909:
I use a mac so can have a look at the test later as well when I get a chance

feltech:
I tried at lunch today on OSX (10.12 on a Mac Pro).  I still couldn't get PrettierPluginTest to fail (though AutoCompletionTest-Reason consistently failed with a timeout). I tried the test in isolation, running test:integration, even running the exact scripts that the travis build runs.  

I didn't have much time to play, so perhaps I didn't set it up properly or something.  I'll keep hacking at it..

Akin909:
@feltech since the changes involving the CI test are being quite flaky, how about separating this branch at commit 96ddb79 and submitting that which was the last passing build and opening a separate PR for the work on the CI test I think the other tests are enough to meet the diff and coverage requirements and cover the functionality well and if you would still like to finish work on the CI test you can submit that separately so you arent blocked by this

feltech:
That's a good plan.  My one concern is that the OSX integration test for this feature did highlight a genuine issue to do with symlinks, so it would be good to include the fix for that.  I suppose that could be part of a separate PR to improve the feature (along with the fixes to the CI tests).

feltech:
I've now split at the last successful build as #2456.  However, I had to merge in master so the tests are going to run again...

I still hold out some hope of fixing all this up. So it would be good to keep this PR around so I can abuse Travis a bit longer.

feltech:
No freaking way.  Finally Travis stops complaining!  There's a bunch of debug logs to remove and other tidying.  Hopefully the build still passes after that.

feltech:
OK then, pending review I think this is good to go.

package.json
```diff
@@ -628,7 +628,7 @@
         "test:setup":
             "yarn run build:test:integration && mocha -t 120000 --recursive lib_test/test/setup",
         "test:integration":
-            "yarn run build:test:integration && mocha -t 120000 lib_test/test/CiTests.js --bail",
+            "yarn run build:test:integration && mocha -t 120000 lib_test/test/CiTests.js --bail --runInBand --no-cache",
```
Akin909: Just for my info what does this change `runInBand`

Akin909:
@feltech 🎉 glad you managed to get the CI tests working 👍 looks good to me. 

Re. the CI test changes tbh I havent spent a great deal of time looking at how its all setup what do the changes you've made do? also might be worth if (hes available getting @bryphe or) @CrossR to review

CrossR:
From a quick look, it looks like the main changes are swapping to using a temporary workspace in some tests, since currently they run in the full Oni repo. That should help a few of the tests, but we may want to revert it if in the future we do some performance tests since the Oni repo is nice to test against as its a decent size.

The other changes seem to be removing the "before" and "after" parts of the CI tests, and collapsing it down into the test itself. It doesn't look to change the tests themselves, but if it helps reliability then sounds good! Only part I wasn't quite sure on was the catching and then doing nothing with the error in the final try-finally of `runInProcTest`.

As for the `runInBand` and `no-cache`, from a quick Google they looked like `jest` flags not `mocha` flags, but I could be wrong.

feltech:
Ah good point.  A summary of the changes:
* A few integration tests now use a temporary workspace, rather than relying on whatever default workspace Oni opens.
* The `PrettierPluginTest` now checks for the presence of the statusbar icon in a `waitFor`.  Looking at the flux actions in the log output there were a lot of `STATUSBAR_SHOW` (or some such) around the same time as the assertion, so I concluded it was a race condition.
* I removed the `beforeEach` and `afterEach` in the main `runInProcTest`, and instead put their logic inside the (single) `it`, with an appropriate `try`/`catch` clause.  This means if the setup/teardown fails, the configured retries will still take place.  I.e. if there is an error in a `before/afterEach` then the test does not retry, which was annoying.
* During teardown of Oni, I noticed that it would sometimes hang trying to get the `windowCount`, which only appears to be used for logging, so I made that time out non-fatally.
* In package.json I added a couple of switches to the `test:integration` call.... OK just read your comment @CrossR - I may well have been mistaken there and those switches are pointless.

package.json
feltech: Yeah, apparently I got my wires crossed.  Looks like it has no effect.  It should ensure that all tests run in the same process.  But that looks to be a jest config option, not mocha.  From [the jest docs](https://jestjs.io/docs/en/cli.html#runinband)
> Run all tests serially in the current process, rather than creating a worker pool of child processes that run tests. This can be useful for debugging.

feltech: I'm pushing a commit to remove those pointless flags.  Fingers crossed the tests still pass!

feltech:
@CrossR 
> Only part I wasn't quite sure on was the catching and then doing nothing with the error in the final try-finally of runInProcTest.
My thinking was that if Oni had failed to start or was in some other bad state, then presumably we could get an exception trying to stop it, which we don't really care about.

I suppose the downside is that if Oni fails to shut down, then that specific test won't fail (should it?).  

However, I suspect in that case subsequent tests will fail, because there would be two instances of Oni running, which is bad news for these integration tests.

feltech:
@CrossR Re. swapping to use a temporary workspace in some tests:  

I did this originally in the integration test for this feature, so that I wasn't relying on a specific workspace and some arbitrary file existing in it.  

Then I found that subsequent tests will run in that same temporary workspace, and at least one of those tests _does_ rely on an arbitrary file in the Oni repo.  I judged that to be bad practice, so went with refactoring that test to also use a temporary workspace.

I do get your point about monitoring performance.  Though integration tests that break because of a side-effect of a previous integration test are a pain to debug.

I'm happy to change the test I wrote to rely on running in the Oni repo as it's workspace, if that's the standard.

CrossR:
>I'm happy to change the test I wrote to rely on running in the Oni repo as it's workspace, if that's the standard.

Nah, its fine! Your new implementation is great, its better to setup a proper workspace rather than depend on a file which could maybe move etc. It was more just a mental note that if in the future we swap to checking performance, we'd want to update it so we are running in a bigger repo for more realistic performance.

CrossR:
Just wanted to check this PR can be merged, since its got WIP in the title still! Otherwise looks good.

I assume the split-off PR can be closed?

feltech:
Hey, yep it's ready to go.  I'll close the other PR.

Issue #251, opened 2017-02-23
Make scrollbar clickable


keforbes:
This is one of those annoyances that I hit multiple times per day.  I keep thinking I can click and drag on the scrollbar to move it but instead this is registered as a mouse click in NeoVim and moves into Visual mode.  I know this won't be fixed anytime soon but I just wanted to file it.

![scrollbar](https://cloud.githubusercontent.com/assets/817509/23273155/27c998b0-f9bb-11e6-95a1-77ecaee8e84b.gif)

gildean:
The missing grip stops me from even trying out this program.

bryphe:
Looks like @aefimenko might be looking at this 👍

bryphe:
@jordan-arenstein , thank you so much for fixing this! 

FYI, there was a bounty associated with this item - make sure to collect it here: https://www.bountysource.com/issues/42373618-make-scrollbar-clickable 

Click 'Solve Issue' on that screen to submit the fix, and we'll approve it! 😄

Akin909:
Can this be closed since its now implemented @keforbes @jordan-arenstein ?

keforbes:
Yep, that's fine.  Go ahead and give @jordan-arenstein that bounty!

PR #2431, opened 2018-07-16
clickable and draggable scrollbar

jordan-arenstein:
clicking or dragging the scrollbar now moves the cursor to the correct location. Should deal with #251

### DIFF
--- a/browser/src/UI/components/BufferScrollBar.tsx
+++ b/browser/src/UI/components/BufferScrollBar.tsx
@@ -17,6 +17,10 @@ export interface IBufferScrollBarProps {
     visible: boolean
 }
 
+export interface IBufferScrollBarState {
+    scrollBarTop: number
+}
+
 export interface IScrollBarMarker {
     line: number
     height: number
@@ -43,9 +47,46 @@ const ScrollBarWindow = styled.div`
     pointer-events: none;
 `
 
-export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps, {}> {
+export class BufferScrollBar extends React.PureComponent<
+    IBufferScrollBarProps,
+    IBufferScrollBarState
+> {
     constructor(props: any) {
         super(props)
+        this.state = { scrollBarTop: 0 }
+        // allow scroll events to be added and removed as event handlers
+        this.trackScroll = this.trackScroll.bind(this)
+        this.endScroll = this.endScroll.bind(this)
+    }
+
+    setLine(y: number) {
+        const lineFraction = Math.min(
+            Math.max((y - this.state.scrollBarTop) / this.props.height, 0),
+            1,
+        )
+        const newLine = Math.ceil(editorManager.activeEditor.activeBuffer.lineCount * lineFraction)
+        editorManager.activeEditor.activeBuffer.setCursorPosition(newLine, 0)
+    }
+
+    beginScroll(e: React.MouseEvent<HTMLDivElement>) {
+        e.preventDefault()
+        // offsetY is definitely on the scrollbar in the beginning of the click
+        this.setState({ scrollBarTop: e.nativeEvent.clientY - e.nativeEvent.offsetY })
+        this.setLine(e.nativeEvent.clientY)
+        document.addEventListener("mousemove", this.trackScroll, true)
+        document.addEventListener("mouseup", this.endScroll, true)
+    }
+
+    trackScroll(e: MouseEvent) {
+        e.preventDefault()
+        this.setLine(e.clientY)
+    }
+
+    endScroll(e: MouseEvent) {
+        e.preventDefault()
+        this.setLine(e.clientY)
+        document.removeEventListener("mousemove", this.trackScroll, true)
+        document.removeEventListener("mouseup", this.endScroll, true)
     }
 
     public render(): JSX.Element {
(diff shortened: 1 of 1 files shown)


browser/src/UI/components/BufferScrollBar.tsx
```diff
@@ -76,13 +78,44 @@ export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps,
                 height: size,
                 backgroundColor: m.color,
                 width: "100%",
+                pointerEvents: "none",
             }
 
             return <div style={markerStyle} key={`${this.props.windowId}_${m.color}_${m.line}`} />
         })
 
+        let scrollBarTop: number
```
Akin909: @jordan-arenstein thanks for the PR 😍, one slightly cosmetic-ish thing could the functions in the render be methods on the class and the `scrollBarTop` be in state so that its more explicit that the value is being changed?

jordan-arenstein: Thanks for the review @Akin909. I'm new to Flux and React, so I'm not 100% on the changes. I've committed, let me know if it's okay 😄 👍

Akin909: @jordan-arenstein no worries the changes I'm suggesting are that `setLine` `beginScroll` and `trackScroll` can be used as methods for the react component rather than re-created every time the component re-renders which is currently the case e.g.

```ts
interface IState {
	scrollBarTop: number
}
export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps, IState>
	public state = {
		scrollBarTop: null,
	}
	componentDidMount(){
		e.preventDefault()
        // offsetY is definitely on the scrollbar in the beginning of the click
         scrollBarTop = e.nativeEvent.clientY - e.nativeEvent.offsetY
         setLine(e.nativeEvent.clientY)
         document.addEventListener("mousemove", this.trackScroll, true)
         document.addEventListener("mouseup", this.endScroll, true)
	}
	
	componentWillUnmount() {
		document.removeEventListener("mousemove", this.trackScroll, true)
        document.removeEventListener("mouseup", this.endScroll, true)
	}

	// Above the render function
	public setLine = (y: number) => {
		const lineFraction = Math.min(Math.max((y - this.state.scrollBarTop) / this.props.height, 0), 1)
            const newLine = Math.ceil(
                editorManager.activeEditor.activeBuffer.lineCount * lineFraction,
            )
            editorManager.activeEditor.activeBuffer.setCursorPosition(newLine, 0)
	}

	public beginScroll = (e: React.MouseEvent<HTMLDivElement>) => {
            e.preventDefault()
            // offsetY is definitely on the scrollbar in the beginning of the click
            this.setState({ scrollBarTop: e.nativeEvent.clientY - e.nativeEvent.offsetY })
            setLine(e.nativeEvent.clientY)
            document.addEventListener("mousemove", trackScroll, true)
            document.addEventListener("mouseup", endScroll, true)
        }
```

Currently its implicit that the scrollBarTop bit of information is being changed, and is a dependency for other functions react aims to make state explicit and provides an api for changing which is `setState` by setting state the component updates correctly based on the new state and the state changes can be accessed inside of react `lifeCycleMethods` like `componentDidUpdate`.

Currently the functionality should be largely unchanged by switching to this but it gives us a good jumping of point for any future changes, lemme know if any of that doesn't make sense

jordan-arenstein: Okay I was on the right track. I've pushed my changes over an hour ago, but they aren't showing here?? They're visible on my fork though. I'm trying to figure it out.

browser/src/UI/components/BufferScrollBar.tsx
```diff
@@ -39,11 +44,49 @@ const ScrollBarWindow = styled.div`
...
-export class BufferScrollBar extends React.PureComponent<IBufferScrollBarProps, {}> {
+export class BufferScrollBar extends React.PureComponent<
+    IBufferScrollBarProps,
+    IBufferScrollBarState
+> {
     constructor(props: any) {
         super(props)
+        this.state = { scrollBarTop: 0 }
+        // allow scroll events to be added and removed as event handlers
+        this.trackScroll = this.trackScroll.bind(this)
```
Akin909: @jordan-arenstein thanks so much for making the changes 👍 I'm wondering if the event handlers cant be set in componentDidMount and removed in willUnmount since atm they are created and removed all they time over and over as a user scrolls, so I think the above event handlers in the didMount and willMount would probably do the same job but will only create one set of handlers

jordan-arenstein:
I don't know what changed?? 🤷‍♂️ but it's up now

browser/src/UI/components/BufferScrollBar.tsx
jordan-arenstein: Can you toggle a handler on and off without destroying it?

jordan-arenstein: If not, I don't see any alternative: the event handler is created when the scroll begins and is removed when it ends.

Akin909: ahhh 😆  I didn't quite understand what those event listeners were doing are they only supposed to exist whilst a user is scrolling? in which case the mounting and unmounting wouldnt be the right solution 👍

jordan-arenstein: They allow the drag to continue outside of the scrollbar - I figured it was too narrow a target, so the scroll continues as the mouse moves anywhere on the screen

jordan-arenstein: I'm taking a look at the arrow functions now.

jordan-arenstein: Based on [this MDN link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) `this` is not supported in arrow functions, and so I  can't use state within them

jordan-arenstein: or do you mean using arrow functions instead of `method.bind(this)`?

Akin909: yes exactly so instead of

```ts
constructor(props: any) {
        super(props)
        this.state = { scrollBarTop: 0 }
        // allow scroll events to be added and removed as event handlers
        this.trackScroll = this.trackScroll.bind(this)
```

you can do away with the constructor and do

```ts
	public state = {
		scrollBarTop: 0
	}
	
	public endScroll = (e: MouseEvent) => {
	// add content here
	// essentially the this binding will be permanently set to this component so the
	// this cannot be reassigned aka it implicitly works the same as a call to .bind(this)
	}
```

jordan-arenstein: fantastic! thank you! I'm testing now, will commit soon after 😄

jordan-arenstein:
@Akin909 full steam ahead 👍

PR #2433, opened 2018-07-16
Feature/render indent lines at correct levels

Akin909:
@CrossR following our discussion a little while ago on discord I've tweaked the indent lines so they don't render a guide for the first level of indentation and the last resulting in:

screen shot 2018-07-16 at 23 32 07

Not sure if the image conveys the change correctly but the aim was to have indents render more in line with the way that sublime or vscode renders them as they seem not to render the first and the deepest level of indentation per line


Akin909:
@CrossR thanks for the review re. the 0 column tbh I was thinking that that would have to be configurable at some point as I can very easily see it being raised as an issue down the line since that behaviour isn't going to be universally expected I think users will probably want the option of choosing.

Re. the file types I started off working off of how vim-indent-guides works and it gives a user the option to select per filetype tbh that list can probably be greatly expanded although I think its still very useful to be able to select where you want them to show as for example you might not want them in markdown (I know I wouldn't) or some other file for some other reason, without an option a user is stuck either turning them on or off with no more granular control

PR #2440, opened 2018-07-18
Fix for #2251 Oni does not correctly locate executables

psxpaul:
Issue #2251 doesn't appear to affect my linux system, but does affect my mac. I have tested this fix on both system, and it appears to fix the PATH issue for the mac.

### DIFF
--- a/cli/src/cli.ts
+++ b/cli/src/cli.ts
@@ -1,5 +1,4 @@
 import * as path from "path"
-
 import { CLIArgs, parseCLIProcessArgv } from "./cli_args"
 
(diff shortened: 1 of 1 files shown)


jordan-arenstein:
This fixes the primary issue for me. $PATH in terminal nvim and oni are the same :)

bryphe:
Cool, makes sense to me! There are cases on OSX where we don't pick up the shell environment - so this ensures we always have it. 

Thanks for the contribution @psxpaul ! 💯 And thanks @jordan-arenstein for testing it out, much appreciated. Bringing this in now!

PR #2443, opened 2018-07-20
Refactoring: Phase 1

TalAmuyal:
# TL;DR 

This is a first PR in a series of refactors.

# Details

**What:** Trying to bring in the `oni-api` package to the same repo as the app.
**Reason:** So correlating changes could be made in an atomic (single) PR and without requiring a deployment of a new and untested version of the API.

Simply setting the API as a local-package yields errors that are usually aren't there (I guess it is a bug/feature on TS's side).
So this is a first in a series of PRs that incrementally fix those issues.


bryphe:
> What: Trying to bring in the oni-api package to the same repo as the app.

Great! Thanks for doing this refactoring, @TalAmuyal 💯

PR #2447, opened 2018-07-22
Feature/Color-highlight improvements

Akin909:
Update the color highlight's fontfamily if a user changes their font mid session currently the font is set in the constructor and doesn't update if the the users font family changes ~, update the regex to match only on words i.e not mid string~

Add a background to cover the underlying text so the color doesnt bleed through

This PR doesn't address the regex issue as it is seemingly more complicated to resolve than I imagined for example `white-space` currently gets matched because of the white the issue with this is that the regex fuzzy matches so that for example a react component like `<Component color={"white"} />` gets matched i.e. the quotes wouldnt stop the match or `:black` the colon wouldn't stop the match, I'm not sure just stopping the hyphen specifically would be enough as html entities can also match so its not entirely clear what the right boundaries should be. Looking at vscode highlight plugins they seem to employ a range of different regexes depending on the filetype so support is only for some filetypes that there is a strategy for. Our current solution is far more general but causes these edge cases 🤔 

Outstanding:
- [x] mid color input rarely the editor can crash whilst an incomplete rgba color was being input, somehow the incomplete color was passed to the styled component causing it throw an error -> prevent incomplete match, seems rare as its is not easily reproduc

### DIFF
--- a/browser/src/Editor/OniEditor/ColorHighlightLayer.tsx
+++ b/browser/src/Editor/OniEditor/ColorHighlightLayer.tsx
@@ -67,6 +67,7 @@ class Highlight extends React.PureComponent<IProps, IState> {
                     left={this.props.left}
                     height={this.props.height}
                     width={this.props.width}
+                    data-id="color-highlight"
                 >
                     <HighlightSpan
                         fontFamily={this.props.fontFamily}
@@ -318,7 +319,6 @@ export default class ColorHighlightLayer implements Oni.BufferLayer {
                                     fontSize={this._fontSize}
                                     fontFamily={this._fontFamily}
                                     color={location.color.toLowerCase()}
-                                    data-id="color-highlight"
                                 >
                                     {location.color}
                                 </Highlight>
--- a/browser/src/Editor/OniEditor/ColorHighlightLayer.tsx
+++ b/browser/src/Editor/OniEditor/ColorHighlightLayer.tsx
@@ -24,7 +24,7 @@ const ColorHighlight = withProps<IProps>(styled.div).attrs({
         width: pixel(props.width),
     }),
 })`
-    display: flex;
+    display: block;
     justify-content: center;
     align-items: center;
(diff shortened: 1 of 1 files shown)


CrossR:
I'm getting a slight misalignment on my end with this, so the cursor can start covering the next char along, as well as not being aligned at the top/bottom.

image

It's possibly something to do with scaling on my machine?
This was in the `onedark.vim` file, but also happened in the `onedark.json` file.
It might be anecdotal...but I felt like it was becoming more misaligned the further down the string I went.

CrossR:
Assuming @bryphe is okay with it, seems like its more an issue of the buffer layer and how it is aligned, rather than your colour highlight layer? So its probably worth fixing that in a separate PR since I expect it'll come up again if we do whitespace etc.

Akin909:
Yeah I think there are few things which need to be updated re. the layer rendering in general like pixel positions and visible lines that adapt to wrapping and folding, inputting lines programmatically as well as slight differences in positioning

although I think the last of those will be the hardest, that being said the cursor seems able to account for the way things are positioned so there might be some tweaks we need to replicate for renderer elements but in general positions are reliable as long as were not dealing with tiny mismatches so think white space would actually be fine

I plan on (when/if I figure out solutions) make a series of small prs for stuff like that as well as ongoing tweaks since both this and the indent lines still have issues to be resolved

Akin909:
@CrossR do you want to wait for @bryphe  on this PR as it currently just tackles on of a few different issues with this layer I'd ideally like to keep this PR scoped to the background bleeding issue and the crash which I noted which this fixes although there are definitely still some outstanding fixes I need to add (in other PRs).

Akin909:
@CrossR thanks for the approval I'll add the alignment issues to the project board for the layers re. solution I'm actually thinking of just not rendering the highlight for the current cursor line since it's the most straight forward fix and not sure how desirable/necessary it is to have the highlight on the current line

PR #2456, opened 2018-07-24
Explorer: Locate Current Buffer (II)

feltech:
Add "Explorer: Locate Current Buffer" command palette item.

Split from #2429 at the last point that CI tests were passing, plus a merge of master.

This means there is a missing fix for symlinked files that are detected as outside the workspace when they're actually within it, as well as a few missing additional tests.


feltech:
I hope that by merging in master I haven't ended up in the same back and forth with flaky integration tests.

feltech:
Closing this as the more advanced PR #2429 is now passing.

PR #2457, opened 2018-07-25
Feature/uncommit unstage functionality

Akin909:
This PR adds the ability to unstage a file, as well undo the *last* commit

Issues/Limitations:

Currently although the git logs is shown the last 25 commits (for performance sake only 25 as testing with a large repo with loads of commits rendering and fetching that log is much slower) they can only interact with the one that is the latest as the uncommit functionality uses `git reset --soft HEAD^` to remove the last commit.

with the log though other functionality like checkout etc. can eventually be added but if the user were able to select each commit atm they would have to scroll throw all 25 to reach the next section, in order to get round this the VimNavigator would have to be expanded to allow dynamically setting a selected ID and some functionality added to allow skipping sections so for now I have removed the option to do that

uncommit

Help Section
screen shot 2018-07-25 at 17 07 42

Outstanding:

- [x] Update commands in help section
- [x] Add tests for help section


vim/core/oni-plugin-git/src/index.tsx
```diff
@@ -120,6 +120,42 @@ export class GitVersionControlProvider implements VCS.VersionControlProvider {
...
+        }
+    }
+
+    public uncommit = async (sha?: string) => {
+        try {
+            await this._git(this._projectRoot).reset(["--soft", "HEAD^"])
+            this._onFileStatusChanged.dispatch()
+        } catch (error) {
+            this._oni.log.warn(`
+                Git Provider unable to get logs because of: ${error.message}`)
```
CrossR: Tiny change, but this should be `unable to undo most recent commit`.

CrossR:
Functionally seems to all work well! Code looks good too.

I tried both unstaging files, as well as undoing commits.

One comment I would make is that the UX for undoing commits potentially could do with being changed, since my first instinct on an item is usually to hit `enter`, but if I did that to a commit message it would undo it, with no confirmation. Though, that is more polish rather than functionality like this PR is adding.

I'll wait on approving till the CI passes.

Akin909:
@CrossR thanks for the review 👍 yeah think I should have stuck with `u` convention re uncommit was the simplest at the time to use enter but can definitely change that in a follow up PR

PR #2460, opened 2018-07-25
Fix + tidy descriptions of Explorer commands

feltech:
* In the command palette when the Explorer is selected there were two
options to create a new file.  One of them is actually the command to
create a new directory.
* So fix the incorrect name for that.
* Also harmonise the naming scheme of the other explorer actions.


bryphe:
Looks like the build is passing now so I'll bring this in. Change looks great! 

Thanks @feltech for the contribution and @Akin909 for reviewing! 💯

PR #2461, opened 2018-07-25
Bump oni-neovim-binaries version.

CrossR:
Pending on the PR in the oni-neovim-binaries repo being merged.


bryphe:
Thanks for upgrading us, @CrossR ! I just approved + brought in that PR (had to update the token too, but looks like it's published now: https://www.npmjs.com/package/oni-neovim-binaries

PR #2463, opened 2018-07-26
Add the CLI scripts back to the build command.

CrossR:
Looks like the building of the CLI script was removed at some point, which meant the CLI scripts weren't working if the user didn't explicitly call `yarn run build:cli`.

Should we do a bit of a shuffle with the `package.json` to rename some of the scripts so we can use `run-p` more?

### DIFF
--- a/package.json
+++ b/package.json
@@ -590,7 +590,7 @@
         "precommit": "pretty-quick --staged",
         "prepush": "yarn run build && yarn run lint",
         "build":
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for catching this, @CrossR ! Looks good to me 😄 

> Should we do a bit of a shuffle with the package.json to rename some of the scripts so we can use run-p more?

Sure, that would make sense.

PR #2465, opened 2018-07-26
Make Mac CI non-optional.

CrossR:
This swaps the Mac 10.11 CI to being non-optional.

I'm not sure if we do want to do this quite yet, since the Mac CI can be flaky already, never mind adding a second one to fail...

Also fixed an issue with the travis doc check, since it needed to be quoted to include new lines.


CrossR:
The fact both mac CIs didn't pass the first time is not a good start to this PR...

PR #2469, opened 2018-07-27
[WIP] Feature/git blame layer

Akin909:
Given the functionality was almost there I decided to try my hands at a git blame layer. The idea being to emulate a little the functionality of the vscode `GitLens` git blame functionality. What this adds is a git blame per line, when the user is resting on the line like the hover tool tip.

It renders inline if there is space otherwise it renders above the line in a tooltip like container.

blame_layer

Outstanding:

- [ ] Tests
- [ ] Configuration: experimental flag, off by default, enabled: auto i.e. per new line after a wait of a configurable amount of time, manual i.e. triggered manually per line via a key binding


bryphe:
Very cool! 😄

browser/src/Editor/NeovimEditor/BufferLayerManager.ts
```diff
@@ -69,6 +69,13 @@ export class BufferLayerManager {
     }
 }
 
+const getInstance = (() => {
```
bryphe: Just curious - why is this wrapped in an IIFE? Is there a functionaility reason for it?

browser/src/Services/VersionControl/VersionControlBlameLayer.tsx
```diff
@@ -0,0 +1,386 @@
...
+        clearTimeout(this._timeout)
+        this._timeout = setTimeout(() => {
+            if (this.props.currentLine) {
+                this.setState({ showBlame: true })
+            }
+        }, this.props.timeout)
+    }
+
+    public getLastEmptyLine() {
+        const { cursorLine, visibleLines, topBufferLine } = this.props
```
bryphe: Nice work getting all this info from the buffer layer API, it's another great example 💯

browser/src/Editor/NeovimEditor/BufferLayerManager.ts
Akin909: It's to encapsulate the instance inside of the function and just return a function that provides access to it,

So rather than
```ts
	let instance = new BufferLayerManager() // global variable
	export getInstance = () => instance

	const getInstance = (() => {
		const instance = new Manager() // scoped to the IIFE and accessed via the closure
		return () => {
			return instance
		}
```

Its a very small distinction I've just been bitten by global vars previously and try to always find ways around them.

vim/core/oni-plugin-git/src/vcs.ts
```diff
@@ -42,6 +63,7 @@ export interface VersionControlProvider {
     canHandleWorkspace(dir?: string): Promise<boolean>
     getStatus(): Promise<StatusResult | void>
     getRoot(): Promise<string | void>
+    getBlame(args: BlameArgs): Promise<Blame>
```
bryphe: Cool! I like that this is abstracted out and that the layer itself doesn't need to know about Git - just about the `Blame` / `BlameArgs`.

bryphe:
Change looks good to me! Great work @Akin909 - really like the way you architected this functionality.

Minor feedback on the look - I'm wondering if it makes sense to right-align the blame text?

image

But I'm fine if that comes in separately, we can always tweak the styling later 👍

feltech:
I enabled this a few days ago and it's working wonderfully.  I really like how it finds nearby whitespace to render into - much less invasive than a pop-up that obscures code, or a split that reduces available window width.

My one issue is that I use Oni with two vertical splits.  This doesn't leave much room, and most of the space available is taken up with the author and time, when for me the message is more useful.  

Perhaps a future tweak would be some configuration as to the ordering of the fields?  Maybe that can already be done with "experimental.vcs.blame.mode" config option?  I haven't had chance to dive into the code to see what that does, though.

Issue #2071, opened 2018-04-07
QuickOpen does not regain focus if oni loses focus then regains it


akinsho:
If the QuickOpen menu is open and a user navigates away from oni, when Oni, regains focus the quick open menu is no longer focused.

Whilst the quick open is open it should hold focus including on window losing then regaining focus, I briefly tried the `autofocus` property of input's on the `lightWeightText` component but seems there's a bit going on there re. focus so the input does get focus this way.

PR #2472, opened 2018-07-28
Refocus previously open menu on reactivating Oni

feltech:
* If the command menu is open and you alt-tab away from Oni, then
alt-tab back again, the menu remains open but no longer has cursor
focus.
* So ensure focus is not lost on reactivating Oni by not overwriting the
saved focus stack.

### DIFF
--- a/browser/src/Input/KeyboardInput.tsx
+++ b/browser/src/Input/KeyboardInput.tsx
@@ -87,7 +87,7 @@ export class KeyboardInputView extends React.PureComponent<
         if (this.props.onActivate) {
             this._removeExistingDisposables()
             const d1 = this.props.onActivate.subscribe(() => {
-                focusManager.setFocus(this._keyboardElement)
+                focusManager.pushFocus(this._keyboardElement)
             })
             this._disposables.push(d1)
         }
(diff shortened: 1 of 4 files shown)


feltech:
This bug was annoying me when I needed to copy-paste text into the menu, but when I hit alt-tab back to Oni after copying the text, the menu was no longer selected.  A minor annoyance, but it happened often enough to be worth looking into (and I figured it wouldn't be too difficult to fix).

As far as I can tell this little solution doesn't have any bad side-effects, but I wanted to PR this before starting on unit tests to get some feedback, since it has the potential to cause unexpected glitches if there's something I've missed, given that it's messing with cursor focus state.

Akin909:
@feltech this is amazing!! this is definitely one of my top 5 annoyances in Oni and it should fix #2071 will be sooooo good to get this in thanks 👍 💯

feltech:
Provided this passes checks/review I think it's done.  I've tried a few movements and no ill effects.

Akin909:
@feltech I've been running a build with this in today and only just realised that following this PR if you focus the explorer with `<c-w-h>` and then try and return to the buffer you are unable to enter any keys. I'm gonna make a PR to revert it I didn't see anything in the implementation that seemed like it would cause this 😢

PR #2473, opened 2018-07-29
Move API to oni-api and use it in preparetion for pluginization

TalAmuyal:
A follow-up for: https://github.com/onivim/oni/pull/2443
Paired with: https://github.com/onivim/oni-api/pull/47

The intended change: Move part of the API into the API package (`IWorkspace` assimilated into `"oni-api"`-'s `Workspace`)

Additional change: Some almost-plugin-compatible modules got a further push that way by using `"oni-api"` instead of the core's singletons.

Also, an oni-api-impl mock is forming with most of it currently unimplemented.

### DIFF
--- a/browser/src/App.ts
+++ b/browser/src/App.ts
@@ -182,6 +182,8 @@ export const start = async (args: string[]): Promise<void> => {
     pluginManager.discoverPlugins()
     Performance.endMeasure("Oni.Start.Plugins.Discover")
 
+    const oniApi = pluginManager.getApi()
+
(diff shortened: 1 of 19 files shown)


TalAmuyal:
@Akin909, I was afraid it would be that way...
I'll try making a compatible `MockOni` that aggregates the `ui-tests` mocks as you suggested.
Thanks for letting me know :)

bryphe:
Thanks for all the work on this, @TalAmuyal 💯

PR #2479, opened 2018-07-30
Feature/oni sessions

Akin909:
Finally had a go at implementing this after the conversation on discord a while ago re. suggested implementation @bryphe (hopefully this is in line with what you wanted) re. multiplexing and putting the methods on the oni editor.

Currently this adds a side bar pane with a list of available sessions these are saved to a default directory `HOME/.config/oni/sessions` under the name a user inputs as a vim file. Selecting the input option adds an input which is the name we use for the session.

The list is gotten by reading the dir and getting all the files saved and on selecting one it call `so ${session.file}`

Outstanding
- [x] Flags and configurability
- [x] Fix vim navigator focus bug
- [x] improve ui
- [x] delete sessions
- [x] set current session and auto update it


browser/src/Services/Sessions/SessionManager.ts
```diff
@@ -0,0 +1,102 @@
...
+            new SessionsPane({ store: this._store, commands: this._commands }),
+        )
+    }
+
+    public get sessions() {
+        return this._store.getState().sessions
+    }
+
+    public get sessionsDir() {
+        return path.join(getUserHome(), ".config", "oni", "sessions")
```
CrossR: I think you'll want to use `getUserConfigFolderPath` here, which gets the `oni` folder directly.

On Windows the `oni` folder is in `%APPDATA%\oni` versus `~/.config/oni`.
May also help testing.

https://github.com/onivim/oni/blob/e46c720fc82f1466347016bc0916a8d744c94793/browser/src/Services/Configuration/UserConfiguration.ts#L27-L40

TalAmuyal:
Looking forward to that feature!

Akin909:
@CrossR @TalAmuyal would appreciate a review on this as I think its ready to go now.
The functionality is intentionally pretty minimal at this stage, its essentially a wrapper around calling `mksession` and `source` session, session files default to being saved in `~/config/oni/sessions/` but can be configured. Its under an experimental flag by default also.

Akin909:
@CrossR thanks for the review 👍 ✨,
Re. functionality thats actually all it does for the time being 😆, it relies heavily on the `sessiopts` that a user has set so things like what tabs are where etc. all come from a users init.vim seemed the cleanest way for a vim user though for new users of vim it might eventually need some sort of default setting for sessionopts if neovim doesn't already have sensible defaults I've no idea what is set by default.

re saving all the time, its debounced 200ms and tbh listens to the bufenter event so shouldn't be calling `mksession` unless you are entering a buffer, when you say constantly save what was happening? calls to `mkession` in the console?

Akin909:
Ahh that's a result of the way that I render the time string which I guess can be confusing basically the time shown is the time *since* it was last modified so basically if its a very recent change say a completely new session then it will say last modified 10s then 15s as time goes on.

The idea was that you could see at a glance the last time you used a session so if one is months old you might think to delete it, wasn't strictly necessary just something I though would be nice to have, the actual time is static a prev. commit had it rendering as a full date string but it just seemed a bit less useful since the core info I was trying to convey was how recently it was used

Akin909:
Actually just re-read your comment if the file in the filesystem is changing its modified time every second thats not intended but if oni is getting and losing focus repeatedly that would trigger a bufEnter, I just had a look at the dir locally and all the session files have the right mtime's none are changing until I switch buffers around 😕

Akin909:
Going to merge this now and can keep an eye on how often the session is being saved and see if I can reproduce the constant save issue.

feltech:
I had a go at enabling this.  Previously I had been using [vim-session](https://github.com/xolox/vim-session), but that is rather glitchy in Oni.

Unfortunately I found with Oni's session support enabled I seemed to randomly lose unsaved work.  I would look away for a minute or two, then when I came back the text had reverted to what's saved on disk. 

I haven't had a chance to look into it further.  I'm using Oni as my main code editor at work (in python), so can't spend much time trying to debug it here.

Akin909:
@feltech thanks or trying out the functionality 👍, you are right I've noted the issue but havent had time to log it properly, it probably relates to what @CrossR was noticing re. the updating not sure what or how or whats going on but will debug once i get a chance

Akin909:
Hmm seems the bug here persists with this PR disabled so might actually be due to something else 😭, not sure what would cause a buffer to revert to an earlier version maybe a change thats affecting the prettier autoformatting on save 😕

PR #2483, opened 2018-07-31
[WIP] Fix east asian width issue

ackey:
Some characters (ex. ①: aka "Ambiguous" width characters) are cut in half.

ambiguous

This PR adds "editor.useEastAsianWidth" option. (The default value leaves false)
Enabling this option fixes the issue. (You also need "set ambiwidth=double" in init.vim)

### DIFF
--- a/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
+++ b/browser/src/Editor/NeovimEditor/NeovimEditor.tsx
@@ -207,7 +207,7 @@ export class NeovimEditor extends Editor implements Oni.Editor {
 
         this._neovimInstance = new NeovimInstance(100, 100, this._configuration)
         this._bufferManager = new BufferManager(this._neovimInstance, this._actions, this._store)
-        this._screen = new NeovimScreen()
+        this._screen = new NeovimScreen(this._configuration)
 
         this._screenWithPredictions = new ScreenWithPredictions(this._screen, this._configuration)
 
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -162,6 +162,8 @@ const BaseConfiguration: IConfigurationValues = {
 
     "editor.imageLayerExtensions": [".gif", ".jpg", ".jpeg", ".bmp", ".png"],
 
+    "editor.useEastAsianWidth": false,
+
     "explorer.persistDeletedFiles": true,
     "explorer.maxUndoFileSizeInBytes": 500_000,
 
--- a/browser/src/Services/Configuration/IConfigurationValues.ts
+++ b/browser/src/Services/Configuration/IConfigurationValues.ts
@@ -233,6 +233,10 @@ export interface IConfigurationValues {
     "editor.cursorColumn": boolean
     "editor.cursorColumnOpacity": number
 
+    // Use eaw (east asian width) to get the width of the character
+    // instead of wcwidth
+    "editor.useEastAsianWidth": boolean
+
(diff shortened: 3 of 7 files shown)


justinmk:
> This PR adds "editor.useEastAsianWidth" option

What's the purpose of that, instead of just letting `ambiwidth` control the decision?

ackey:
Even with `ambiwidth=double`, those ambiguous characters are cut in half.

Oni uses `wcwidth()` under the hood to get the character width. It returns `1` for those ambiguous characters.
This option uses `eaw.getWidth()` (https://github.com/mysticatea/eaw) instead of that. It returns `2` for those ones.

justinmk:
My question is why add another setting? In other words, let the fixes you made be controlled by `ambiwidth` instead of a new Oni setting?

I'm not quiet sure about Oni's position regarding new settings. Is it typical to add Oni settings even if a Nvim setting already exists?

ackey:
Oh, now I understand the meaning of your question.
Yes, the idea to use existing Nvim's setting seems to be good 🙂 
I don't know the method now, but I will try to find it.

Akin909:
@ackey I think there are few issues that currently surround rendering of ambiwidth characters including #2427 so definitely appreciate the PR.

Re. re-implementing `nvim` features @justinmk cant speak for everyone but I think we've avoided re-implementing anything nvim already provides and stuck to just nice to have gui things, or non-native features and is something I personally think is v. important since that will just bloat and overcomplicate oni needlessly.

@ackey do you know whether setting ambiwidth resolves the issue? (or improves it the canvas renderer is know to clip text due to a very involved issue I cant find), note though re this PR and that solution anyway that the canvas renderer is a strategy we are trying to eventually move away from in favour of the `webgl` renderer so this will only solve the problem there.

ackey:
@Akin909 setting ambiwidth doesn't resolve the issue.

The cursor is next to the character in the bellow.

Oni: ambiwidth=double
oni-double

Oni: ambiwidth=single
oni-single

Nvim: ambiwidth=double
nvim-double

Nvim: ambiwidth=single
nvim-single

These are all ok except that the character is clipped in Oni.

This is not the canvas rendering issue. The character isn't actually clipped but the program code fills the right-hand side of the character with the background color rectangle.

Akin909:
@ackey thanks for looking into it and the images, I wonder if the results are the same if you set `editor.renderer: "webgl"`, re. solution that wouldn't involve oni taking control of this setting, we can read via the nvim api the value of a setting such as `ambiwidth`

> nvim_get_option({name})                                    *nvim_get_option()*
>                Gets an option value string
>
>                Parameters: 
>                    {name}  Option name
>
>                Return: 
>                    Option value (global)

from the docs,

we could then use that as the boolean to check, currently at work but can flesh out when I'm free when and how you could get access to the variable in both of the renderers

justinmk:
See also the `option_set` redraw event in `:help ui-global`. Since Nvim 0.3.0, 'ambiwidth' and other UI-related options are published in the `option_set` event.

ackey:
@Akin909 @justinmk Thank you for the information 🙂 

@Akin909 webgl looks perfect without this solution 😍 

webgl: ambiwidth=double
webgl-double

webgl: ambiwidth=single
webgl-single

Akin909:
@ackey since the issue seems to relate to the canvas renderer vs. the webgl one its known issues is it alright to close this PR with the definitive fix being eventually fully migrating to the webgl renderer

ackey:
I'm using the webgl renderer now and it seems it's perfect to me. I close this PR. Thanks 🙂

PR #2488, opened 2018-08-03
Fix local building issue

TalAmuyal:
Some small changes to `oni-plugin-git`-'s `package.json`:
- Moved `oni-api` & `oni-types` from `dependencies` to `peerDependencies`, so it now only references to the root's `node_modules` rather than having its own copy.
- Added missing `rimraf` to `devDependencies` as it is needed (I think) for the build command.


TalAmuyal:
As mentioned in https://github.com/onivim/oni/pull/2493, I tested the strange error and it occurs when using `peerDependencies`.
I reverted that part since it is a nice-to-have and will likely face it later when I try pluginizing more parts.

@Akin909, will you re-review?

TalAmuyal:
Yes, it still works.
The important thing is the `rimraf` dependency.

PR #2495, opened 2018-08-06
Bugfix/laggy buffer scrollbar responsiveness

### LINKED ISSUE #2114: Scrollbar lags behind oni screen on resize
When the screen is resized if the scrollbar is set to active it lags behind resizing of the window briefly appearing in the center of the window. Not sure how easy this will be to solve but might be a css alignment issue, will post GIF shortyly

Akin909:
This PR fixes #2114 which was the fact the the buffer bar lagged behind if the window was resized. The specific change that fixes this is changing the position from `absolute` to `fixed`

Also took the opportunity to change the components to styled components with dynamic attrs.

### DIFF
--- a/browser/src/UI/components/BufferScrollBar.tsx
+++ b/browser/src/UI/components/BufferScrollBar.tsx
@@ -2,7 +2,7 @@ import * as React from "react"
 
 import * as uniqBy from "lodash/uniqBy"
 import styled from "styled-components"
-import { bufferScrollBarSize } from "./common"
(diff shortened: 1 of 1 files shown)


Akin909:
would appreciate a review @CrossR or @TalAmuyal, this fixes a UI issue which was really annoying me which is that there was a lag when resizing the window so the scrollbar would visibly be slow to move when resizing the original issue has a GIF also fixed an issue where error markers weren't being shown properly on the scrollbar

PR #2499, opened 2018-08-08
Feature/Add AppImage and archlinux support

barni2000:
I have configured the electron-builder for AppImage and archlinux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -95,6 +95,8 @@ deploy:
       file:
           - dist/*.deb
           - dist/*.rpm
+          - dist/*.pacman
+          - dist/*.AppImage
           - dist/*.tar.gz
       skip_cleanup: true
       on:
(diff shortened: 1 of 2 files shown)


Akin909:
Does anyone who uses linux want to give this a try @CrossR? it looks good to me but I won't be able to test this unfortunately

barni2000:
zenity, kdialog, Xdialog is dependency for a message "Would you like integrate ... ". It assumes that you have a fully installed X environment. With this application can appear in your menu. It registers the desktop file.

barni2000:
electron-builder create this script https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/templates/linux/AppRun.sh for run.

badosu:
@barni2000 What's the output for the pacman target? Seems like a good way to automate the package generation.

PR #2513, opened 2018-08-20
[WIP] Handle missing mouse events

### LINKED ISSUE #1068: Linux: Middleclick paste
It would be very nice to support Linux-like middleclick paste. I'm very used to middleclick paste the current clipboard contents.

Behavior in Terminals and GVIM:
- Command mode: Insert directly on mouse-position
- Insert mode: Insert on textcursor-position

badosu:
Handle the other mouse events:

-`<MouseRight>`: enters visual mode if not into it and sets the selection from the cursor to the mouse position
-`<MouseMiddle>`: If OS is linux and running X, paste the selection clipboard.

I ran into this because I was looking for a way to make middle click paste work. At first I looked into using `clipboard.readText('selection')` but decided to let vim handle it instead.

I'll remove WIP when I add some tests and receive feedback.

Fix #1068

### DIFF
--- a/browser/src/Input/Mouse.ts
+++ b/browser/src/Input/Mouse.ts
@@ -16,8 +16,20 @@ export class Mouse extends EventEmitter {
         this._editorElement.addEventListener("mousedown", (evt: MouseEvent) => {
             const { line, column } = this._convertEventToPosition(evt)
 
-            this.emit("mouse", `<LeftMouse><${line},${column}>`)
-            this._isDragging = true
+            switch (evt.which) {
+                case 1:
+                    this.emit("mouse", `<LeftMouse><${line},${column}>`)
+                    this._isDragging = true
+                    break
+                case 2:
+                    this.emit("mouse", `<MiddleMouse>`)
+                    break
+                case 3:
+                    this.emit("mouse", `<RightMouse><${line},${column}>`)
+                    break
+                default:
+                    break
+            }
         })
 
         this._editorElement.addEventListener("mousemove", (evt: MouseEvent) => {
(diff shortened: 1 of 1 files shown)


badosu:
@CrossR @Akin909 Any feedback on this?

I'd like to know if you think this is a good direction, keeping in mind we can also map a lot of other mouse events that we aren't currently.

It is improbable that a user has a particular mouse mapping, but it's good to cover this. Also we may in the future map mouse events directly from Oni, before passing to vim.

Akin909:
@badosu looks good to me, I actually work almost exclusively off a laptop with a monitor so can't actually test it out properly

badosu:
@daxid If on Linux and X11, yes

badosu:
Testing this has been a challenge, I exposed the raw sendInputEvents function and the clipboard from electron and it still does not work, see:

<details>
<pre>
export const test = async (oni: Oni.Plugin.Api) => {
    if (!isLinux()) {
      return
    }

    await oni.automation.waitForEditors()

    await createNewFile("js", oni)

    oni.automation.sendKeys("i")

    await oni.automation.waitFor(() => oni.editors.activeEditor.mode === "insert")

    const expectedResult = [
        "const test = {",
        "    window.setTimeout(() => {",
        '        let testString = "Oni"',
        "    })",
        "}",
    ]
    oni.automation.writeToClipboard(expectedResult.join(os.EOL), 'selection')

    const mouseDown:any = {type: 'mouseDown', x: 10, y: 10, button: 'middle'}
    oni.automation.sendInputEvent(mouseDown)

    const mouseUp:any = {type: 'mouseUp', x: 10, y: 10, button: 'middle'}
    oni.automation.sendInputEvent(mouseUp)

    oni.automation.sendKeys("<esc>")

    // Because the input is asynchronous, we need to use `waitFor` to wait
    // for them to complete.
    await oni.automation.waitFor(() => oni.editors.activeEditor.mode === "normal")

    const lines: string[] = await oni.editors.activeEditor.activeBuffer.getLines(0, 5)

    assert.deepEqual(lines, expectedResult, "Verify lines are as expected")
}
</pre>
</details>

Any help is welcome

omolteni:
@badosu Tell me how I can test it and I will try to help.
I am currently running ONI on Ubuntu bionic and this feature is really essential for me.

bryphe:
I think the issue is our `Automation` API doesn't have a `sendInputEvent` method: <https://github.com/onivim/oni/blob/74a4dc7f2240a1f5f7a799b2f3f9d01d69b01bac/browser/src/Services/Automation.ts> 

A couple of ways you can exercise this locally:
- In developer tools, you can use the automation API right from the console - ie `Oni.automation.sendKeys(":new test.txt")`

The `WebContents` object in Electron does expose this event: https://electronjs.org/docs/api/web-contents#contentssendinputeventevent - we could either get a reference to the webcontents from the test (using `require("electron").remote.getCurrentWebContents`), or add a method to the `Automation` API to proxy it.

Issue #2414, opened 2018-07-09
Truncated copy/paste


Piping:
**Oni Version: **
**Operating System: windows**

**Issue:**
Copy a large number of lines into Oni and only got paste partially.

CrossR:
Confirmed using the `csv` file we have in `test/collateral/`, copied the raw file from GitHub and pasted into Oni and got 975 lines and 1 partial line, rather than 1000 lines. Pasting into Notepad++ (to prove the file was fully copied) worked fine.

Copying/pasting directly inside Oni (with `y`/`p`) is fine, only copying from outside Oni and pasting in seems to be broken.

The way we read the clipboard (`clipboard.readText()`) seems to be the only way to do so, so I think the issue lies in the use of `nvim_input` rather than  `nvim_set_lines`?

Akin909:
Hopefully this is a good first issue the related code is [`here`](https://github.com/onivim/oni/blob/e0aa2a556b3cc287cb932058d92440e8311a0d0f/browser/src/Editor/NeovimEditor/NeovimEditorCommands.ts#L68) I believe the steps to resolve this would be first confirming that `clipboard.readText()` isn't clipping the text if not then as @CrossR suggested I think switching to using `neovimInstance.setLines` would be preferable docs [here](https://neovim.io/doc/user/api.html)

It would be interesting to know why the `input` function is truncating the input as this maybe an issue we ought to raise with `neovim` or we might be using the method incorrectly

psxpaul:
I took a quick look at this over the weekend, and here are a few things I found:

- `clipboard.readText()` is not clipping, but `neovimInstance.input(sanitizedText)` is. The documentation for `nvim_input()` [here](https://neovim.io/doc/user/api.html#nvim_input()) seems to indicate that clipping will occur if the buffer becomes full.
- Using \<c-v>/\<m-v> to paste even short text, then using the repeat operator (`.`) results in all pastes after the first one to have incorrect indentation. I believe this is because `set paste` and `set nopaste` are not invoked with the repeat operator.
- Using `nvim_buf_set_lines` like `BufferManager.setLines` would likely break the repeat operator in a different way. Issue #2501 has some more information on that.
- The Edit->Paste, Edit->Paste Line Before, and Edit->Paste Line After do not seem to suffer from the truncation issue. They are executing the `\\"+gP`, `[p`, `]p` commands respectively, instead of simulating user input.
[...]

PR #2515, opened 2018-08-20
Fix for #2414 - truncated copy/paste

psxpaul:
This is a fix for some of the issues raised in #2414. One thing to note - there was some question as to whether win32yank works for everyone. I did find a possible workaround for that, if it's still an issue:

```
const textToPaste = clipboard.readText()
const sanitizedTextLines = replaceAll(textToPaste, { "'": "''" })
await neovimInstance.command("let @+='" + sanitizedTextLines + "'")
```

I don't have a windows box to test this on, so perhaps we can add this code to `NeovimEditorCommands. pasteContents` later if needed.


browser/src/Input/KeyBindings.ts
```diff
@@ -79,6 +80,7 @@ export const applyDefaultKeyBindings = (oni: Oni.Plugin.Api, config: Configurati
         input.bind("<c-,>", "oni.config.openConfigJs")
 
         if (config.getValue("editor.clipboard.enabled")) {
+            input.bind("<c-x>", "editor.clipboard.cut", isVisualMode)
```
badosu: `<c-x>` in visual mode decrements an integer value under the cursor

psxpaul: Thanks for the heads up, I'll remove this part

badosu: Well, you're not required to remove it, `<c-v>` is also overriden.

`<c-x>` makes some sense, perhaps `<C-c-x>`?

Akin909:
@psxpaul seems the large paste test is repeatedly failing on windows 😭 we might need the solution in the snippet you pasted for windows

CrossR:
I've not had a chance to test this yet, but I've set a reminder to do it tonight, so should be able to test it then. I'll also give it a test on my normal dev box, see if its all Windows or just the CI / base config versions of Oni.

CrossR:
This is pretty odd....I ran the CI tests locally, it passed fine twice. I started Oni, pasted and copied a bunch, all good.

Then I tried it one last time, and now there is no clipboard provider registered, `:checkhealth` says its not hooked up...

psxpaul:
Thanks for trying it out @CrossR. I got a Windows VM setup and managed to get the win32yank workaround working. I wasn't sure if this workaround should only be applied for windows machines, but the approach seems to work fine on my Mac and Linux machines.

josemarluedke:
This PR might created a issue when pasting on command. Say I start a search with `/` and paste using `cmd + v`. The content will actually be pasted under the buffer that was focused instead of putting that content on the search.

Issue #2514, opened 2018-08-20
Provide binding to jump to next/previous error


psxpaul:
**Oni Version:** 0.3.6
**Neovim Version (Linux only):** n/a
**Operating System:** MacOS 10.12

**Issue:**
When editing a file that is supported by a configured language server, I would like to be able to jump to the next/previous highlighted error. There doesn't appear to be any way to do this currently, unless I'm missing something.

Looking into the code, I see 3 ways of implementing this:

- add all diagnostic errors to the Quickfix List, so they can be navigated with `[q` and `]q`
- add all diagnostic errors to the Location List, so they can be navigated with `[l` and `]l`
- add a new entry to `browser/src/Services/GlobalCommand`, so that custom bindings can be added by the user

The Quickfix and Location List approaches have the downside of potentially overwriting anything already in those lists. I know that workspace searches and find-all-references searches populate the Quickfix list, so that's probably not ideal for many users.

I'm also curious if there's a default binding we should set for this. JetBrains IDEs like WebStorm and IntelliJ use `F2/Shift-F2` for these bindings, while VSCode and VisualStudio use `F8/Shift-F8`.

**Expected behavior:**
Press a key combo (e.g. <F8>), and the cursor would jump to the next highlighted error.

**Actual behavior:**
No such binding is available or documented

**Steps to reproduce:**
- Open a Typescript file
- type in some gibberish
- verify the nonsensical line is highlighted
- jump to a different spot in the file
- how to quickly jump back to error?

psxpaul:
Thanks for the feedback @badosu. I'm also leaning toward the 3rd implementation I listed. I have a prototype locally just for jumping between errors. It should be possible to later add a GUI element for listing errors across the entire project.

I'll probably implement this without a default binding, since most IDEs seem to have different shortcuts for this.

badosu:
> I'll probably implement this without a default binding, since most IDEs seem to have different shortcuts for this.

Kudos for that :+1:, although default FX keybindings (also C-PgUp, C-PgDwn and similar) are usually not an issue for experienced vimmers (which tend to concentrate keybindings in the middle of the keyboard).

Akin909:
@psxpaul @badosu I agree re. avoiding the quickfix and loclists for this as I think that could open a can of worms but might be interesting to revisit once they are externalised I think the third option sound great 👍, there should be a list in the/a redux store of all the errors and the associated lines its currently being used to render markers on the buffer scroll bar but could be also be used for this

PR #2526, opened 2018-08-28
Add commands to jump to previous/next error

psxpaul:
This PR adds `oni.editor.nextError` and `oni.editor.previousError` commands, as discussed in #2514. I have mainly tested this with Typescript.

### DIFF
--- a/browser/src/Services/Commands/GlobalCommands.ts
+++ b/browser/src/Services/Commands/GlobalCommands.ts
@@ -9,17 +9,13 @@ import { remote } from "electron"
 
 import * as Oni from "oni-api"
 
-import {
-    getAllErrorsForFile,
-    getInstance as getDiagnosticsInstance,
-} from "./../../Services/Diagnostics"
+import { gotoNextError, gotoPreviousError } from "./../../Services/Diagnostics/navigateErrors"
 import { EditorManager } from "./../../Services/EditorManager"
 import { MenuManager } from "./../../Services/Menu"
 import { showAboutMessage } from "./../../Services/Metadata"
 import { multiProcess } from "./../../Services/MultiProcess"
 import { Tasks } from "./../../Services/Tasks"
 import { windowManager } from "./../../Services/WindowManager"
-import { isInRange } from "./../../Utility"
 
 // import * as UI from "./../UI/index"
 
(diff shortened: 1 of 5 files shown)


browser/src/Services/Commands/GlobalCommands.ts
```diff
@@ -45,6 +50,78 @@ export const activate = (
     const popupMenuPrevious = popupMenuCommand(() => menuManager.previousMenuItem())
     const popupMenuSelect = popupMenuCommand(() => menuManager.selectMenuItem())
 
+    const gotoNextError = async () => {
```
Akin909: Could these functions be move into their own files in the `Diagnostics` directory maybe `Diagnostics/navigateErrors.ts` so the global commands isn't where they are implemented and just import them here?

browser/src/Services/Commands/GlobalCommands.ts
```diff
@@ -45,6 +50,78 @@ export const activate = (
...
+        const errors = getDiagnosticsInstance().getErrors()
+        const activeBuffer = editorManager.activeEditor.activeBuffer
+        const currentFileErrors = getAllErrorsForFile(activeBuffer.filePath, errors)
+        const currentPosition = activeBuffer.cursor
+
+        if (!currentFileErrors || currentFileErrors.length === 0) {
```
Akin909: `currentFileErrors.length === 0` could be `!currentFileErrors`

browser/src/Services/Commands/GlobalCommands.ts
```diff
@@ -45,6 +50,78 @@ export const activate = (
...
+            ) {
+                await activeBuffer.setCursorPosition(
+                    error.range.start.line,
+                    error.range.start.character,
+                )
+                return
+            }
+        }
+
+        activeBuffer.setCursorPosition(
```
Akin909: This should also be awaited as above I think?

browser/src/Services/Commands/GlobalCommands.ts
```diff
@@ -45,6 +50,78 @@ export const activate = (
...
+                await activeBuffer.setCursorPosition(
+                    lastError.range.start.line,
+                    lastError.range.start.character,
+                )
+                return
+            }
+            lastError = error
+        }
+
+        activeBuffer.setCursorPosition(lastError.range.start.line, lastError.range.start.character)
```
Akin909: not sure if this is async but its likely to be in which case does it need to be awaited since its not being returned to be awaited elsewhere

test/ci/Editor.NextPreviousErrorTest.ts
```diff
@@ -0,0 +1,126 @@
+/**
+ * Test that you can jump to the next/previous error
+ */
+import * as assert from "assert"
+import * as Oni from "oni-api"
```
Akin909: Nice 💯thanks for adding this

browser/src/Services/Commands/GlobalCommands.ts
psxpaul: Thanks for the feedback! I moved this to `browser/src/Services/Diagnostics/navigateErrors.ts`, but that might be confusing since there's already a `browser/src/Services/Diagnostics.ts`. I can move that file inside the new `browser/src/Services/Diagnostics` directory if you want.

Akin909: we could make the already existing `Diagnostics.ts => browser/src/Services/Diagnostics/index.ts` that way if there are future helpers etc. they can go in the same dir plus you wont have to change any imports since as index all the imports of the file should work as is

Akin909:
going to merge this now if you're ready @psxpaul? Thanks again 🎉

badosu:
@psxpaul Do you mind adding an entry on the wiki for this?

I fear the lack of default keybinding may make this feature invisible to users.

badosu:
Seems good enough, thanks!

bryphe:
Awesome! Been wanting this for a while, very nice work - great test coverage too @psxpaul ! 💯

PR #2527, opened 2018-08-28
Feature/add sessions to welcome

Akin909:
Based off of #2522 (that should go in first) this PR adds a section with the available sessions to the welcome screen as well as the ability to restore a session.

This PR also has an updated welcome screen design (I begged a colleague who is a UI designer for some ideas for this hopefully we didn't got in the wrong direction)

screen shot 2018-08-28 at 20 54 00


badosu:
Wow, that looks amazing! :clap:

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
```diff
@@ -404,22 +507,38 @@ export class WelcomeView extends React.PureComponent<WelcomeViewProps, WelcomeVi
                         <HeroImage src="images/oni-icon-no-border.svg" />
                     </Column>
                     <Column alignment="flex-start">
-                        <SubtitleText>{`v${this.state.version}`}</SubtitleText>
+                        {version && <SubtitleText>{`v${version}`}</SubtitleText>}
```
Akin909: this change doesnt prevent the whole welcome layer from rendering if there is an issue with the `getMetadata` function not sure how it could fail but I don't think anything other than a full editor/app crash should prevent the initial welcome page from rendering

CrossR:
Functionally, loading sessions works great! I like the redesign too, but a few points:

* The header text ("Quick Commands"/ "Learn"/"Sessions") is a bit smaller than I'd expect, so it doesn't stand out from the contents below it that much.
* The gap between the two sections (the menu + sessions) is pretty big (at least on my monitor).

Image for reference:

image

Akin909:
@CrossR thanks for having a look much appreciated 👍 

re. the heading sizes I can increase those a bit for legibility
re. the spacing the idea was to have some spacing but between them and a visible divider, with your current theme it seems less obvious, but either way the gap may be too big there (as always im on a 15' mac so not the same effect, ill add a max width so that box is never too big so you dont get the large gap)

Akin909:
@CrossR just tweaked the styles to reduce the overall size of the box, and increase the size of the headers, can you give it a try and let me know if its improved, locally the change is less obvious, also if the text is still to small can you see if tweaking you `ui.fontSize` changes it.

Akin909:
@CrossR thanks for checking those I'll merge this and then open a smaller pr to fix up the other bits 👍

Issue #1874, opened 2018-03-21
Unable to scroll through autocomplete suggestions


ExpectationMax:
It is not possible to scroll down to completion suggestions outside of the initial view of the completion box.
The view does not scroll down when elements further downward are selected, neither does the additional description field update.

Regular view:
<img width="657" alt="screen shot 2018-03-21 at 11 14 49 am" src="https://user-images.githubusercontent.com/1303655/37704473-2f027b8e-2cf9-11e8-9dfb-aa514699c5d0.png">

Pressing \<ctrl-n\> to scroll down
<img width="676" alt="screen shot 2018-03-21 at 11 15 07 am" src="https://user-images.githubusercontent.com/1303655/37704483-340a717c-2cf9-11e8-918a-8f43900da6d4.png">

I experienced this problem on macOS high Sierra using Oni version 0.3.1 .

badosu:
Yes

Issue #2046, opened 2018-04-05
Autocomplete/Suggestions doesn't scroll


AUTplayed:
Hi, if I write code like this: ![code](https://i.imgur.com/0FBRMks.png)

and scroll down the suggestions with the down arrow key, the suggestion window doesn't scroll down, but the highlighting disappears.

I am on windows 10 using v0.3.2

AUTplayed:
bonus video: https://my.mixtape.moe/twlmzw.mp4

Toxiapo:
Same for Mac running on 10.13.4 with the latest Oni version.

badosu:
Duplicate of #1874

PR #2529, opened 2018-08-30
Fix for #2046 Autocomplete/Suggestions don't scroll

psxpaul:
Potential fix for the issue described in #2046


badosu:
Fixes #1874 as well

Issue #2365, opened 2018-06-27
Hide hidden files from sidebar


khamer:
**Oni Version:** 0.3.6
**Neovim Version (Linux only):** v0.3.1-113-gda6874a7b
**Operating System:** Linux

**Issue:** Currently, all hidden files, including ones hidden by my global gitignore that git ls-files doesn't show, are being shown in the Explorer sidebar. Any thoughts on how to change that?

keforbes:
I think this relates to the discussions in #2217 and #2149.  Not adding any value here, just linking issues. :slightly_smiling_face:

PR #2531, opened 2018-08-30
Hook up showing files in hidden folders in QuickOpen.

CrossR:
This adds a flag for showing hidden files, and files in hidden folders.

By default, you can't see any hidden files, nor any files in hidden folders which can be a pain, since for example in my dotfiles, most of the files are stored in "hidden" folders, such as `.config` folders.

I've left it to disabled for now, but we may want to just always show them?

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -145,6 +145,7 @@ const BaseConfiguration: IConfigurationValues = {
     "editor.quickOpen.filterStrategy": "vscode",
     "editor.quickOpen.defaultOpenMode": Oni.FileOpenMode.Edit,
     "editor.quickOpen.alternativeOpenMode": Oni.FileOpenMode.ExistingTab,
+    "editor.quickOpen.showHidden": false,
 
     "editor.split.mode": "native",
 
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -145,7 +145,7 @@ const BaseConfiguration: IConfigurationValues = {
     "editor.quickOpen.filterStrategy": "vscode",
     "editor.quickOpen.defaultOpenMode": Oni.FileOpenMode.Edit,
     "editor.quickOpen.alternativeOpenMode": Oni.FileOpenMode.ExistingTab,
-    "editor.quickOpen.showHidden": false,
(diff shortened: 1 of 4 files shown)


badosu:
I agree, for a coding tool it does not make sense to not be able to see dotfiles.

I would even argue that it should be on by default, but having the configuration is enough.

Akin909:
I agree with @badosu this should be on by default imho 🤷‍♂️ (its very confusing behaviour that expected files are hidden in a code editor)

CrossR:
I added it as a config option based on the fact there was a `TODO` there for it, but now that I look at it, that `TODO` was added as part of refactoring the Quick Open menu out, such that it could be used for the "Find in files" functionality.

Perhaps @TalAmuyal could explain what they think is the best approach. Currently, I've just set it as a flag for both the QuickOpen and find in files (since I'd want it for both), but potentially that isn't the best approach?

I also noticed that as part of that PR, the "--case-sensitive" flag was removed, which I didn't think much of at the time, since I thought the filtering would help on our end, but I've now realised that its causing some weird results with the quick open as well....Maybe that should be reverted, since the results are less relevant, and also it gives Oni more files to process when the processing is fairly expensive.

badosu:
> I also noticed that as part of that PR, the "--case-sensitive" flag was removed, which I didn't think much of at the time, since I thought the filtering would help on our end, but I've now realised that its causing some weird results with the quick open as well....Maybe that should be reverted, since the results are less relevant, and also it gives Oni more files to process when the processing is fairly expensive

See: https://github.com/onivim/oni/issues/825

> For clarity, the --hidden flag will allow hidden folders/files to be seen, but since ripgrep uses a projects .gitignore by default, I don't see any issue with including it always? It would only show files the user wants to my knowledge. The only issue I can see is it may attempt to search the .git folder, but thats easily fixed with --glob "!.git/*" to not glob the git folder

Yes, I think everyone agrees with having the hidden option, the proposal it being turned on by default.

CrossR:
I've swapped this over to enabled by default, but left the config option in "just in case".
I don't know of a reason why it wouldn't be needed..but still.

If @TalAmuyal is free anytime soon, it may be worth just waiting before merging to check I'm not being stupid here.

badosu:
https://github.com/onivim/oni/issues/2365 should also use this setting when/if it's addressed.

CrossR:
Should probably just get this merged since there has been no opposition, could I get a re-review?

badosu:
@CrossR Can you describe this option in the wiki?

PR #2533, opened 2018-08-31
Feature/fix remaining welcome commands

Akin909:
This tackles some of the left over issues re. the welcome as described by @CrossR as well as adds a command to open the welcome screen and I discovered a nice way to check if a key is remapped in vim which was necessary for the `:` command so I added it to the neovimInstance and editor for future use


Akin909:
@CrossR @TalAmuyal would appreciate if either of you could review as this I think (for now ) ties up the work I was doing on the welcome screen it hooks up some more commands and adds a command to the quick open that allows a user to reopen the welcome screen

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
```diff
@@ -553,74 +590,74 @@ export interface IWelcomeCommandsViewProps extends Partial<WelcomeViewProps> {
...
                         title="Open File / Folder"
-                        onClick={() => executeCommand(commands.openWorkspaceFolder.command)}
                         description="Control + O"
+                        onClick={() => commands.openWorkspaceFolder.execute()}
                         command={commands.openWorkspaceFolder.command}
-                        selected={isSelected(commands.openWorkspaceFolder.command)}
+                        selected={this.isSelected(commands.openWorkspaceFolder.command)}
                     />
                     <WelcomeButton
                         title="Command Palette"
```
CrossR: This and line 617 need updating.

"Command Palette" -> "Quick Open"
"Control + Shift + P" -> "Ctrl + P".

CrossR: Actually, It may be worth having both, the Command Palette is a little hard to find currently.

It also just made me realise, is it `Ctrl-Shift-P`/`Ctrl-P` on both platforms? We might need to show `cmd` on Mac.

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
```diff
@@ -553,74 +590,74 @@ export interface IWelcomeCommandsViewProps extends Partial<WelcomeViewProps> {
...
-                        selected={isSelected(commands.openDocs.command)}
+                        selected={this.isSelected(commands.openDocs.command)}
                     />
                 </AnimatedContainer>
                 <AnimatedContainer duration="0.25s">
                     <SectionHeader>Customize</SectionHeader>
                     <WelcomeButton
                         title="Configure"
-                        onClick={() => executeCommand(commands.openConfig.command)}
+                        onClick={() => commands.openConfig.execute}
```
CrossR: This doesn't seem to work for me?

From checking the menu, we use `"oni.config.openConfigJs"` there, would that work here? Its also possible something is going weird in my setup.

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
```diff
@@ -553,74 +590,74 @@ export interface IWelcomeCommandsViewProps extends Partial<WelcomeViewProps> {
...
+                        onClick={() => commands.commandline.execute()}
+                        selected={this.isSelected(commands.commandline.command)}
                     />
                 </AnimatedContainer>
                 <AnimatedContainer duration="0.25s">
                     <SectionHeader>Learn</SectionHeader>
                     <WelcomeButton
                         title="Tutor"
-                        onClick={() => executeCommand(commands.openTutor.command)}
+                        onClick={() => commands.openTutor.execute()}
```
CrossR: This also doesn't seem to be working for me?

Not sure why, I think it used to, but its possible the command has been renamed.

Akin909: @CrossR this I believe never worked I don't think the functionality exists when I started refactoring it I noted it didn't work but also didn't know what it was supposed to do. Given the stated goals of oni having something related to that seemed important to keep on the front page, but it wasn't clear what that should be and tbh if the command doesn't already exist I think implementing it from scratch is beyond the scope of what I wanted to do here

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
Akin909: woops I left off the calling parens on that function will push a fix 👍

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
Akin909: Its the one thats left but I couldn't find a quick solution and realllly didnt want to remove it, could always put a coming soon on it or disable the button but not sure we have a template for styling a disabled button

CrossR: Thats fair enough, we can look at fixing up the `experimental.tutorial.start` command I've found in a different PR. Might make a good "first issue" PR, just need to look into the scope of it.

Or, it can be swapped to be like the "Toggle Explorer" command, to pop out the tutorial side bar.

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
Akin909: This one is intended to be the `command palette` I've just stuck with @bryphe original list for the most part, and on mac `c-s-p` works just as well as `m-s-p`, I think its simpler to leave it as the control binding as its correct for both?

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
Akin909: Or open up your current/last tutorial? think those are all good things it could do, a good first issue sounds like a good idea since we can then just point the person to the example with the explorer and the other panes that have that functionality

browser/src/Editor/NeovimEditor/WelcomeBufferLayer.tsx
CrossR: Okay!

The command should be updated from `commands.quickOpenShow` to `commands.commandPalette` then I think, as well as changing the actual command being called from `quickOpen` to `command` (lines 355)? Otherwise its saying "Command Palette" and opening a different menu.

(It was like this in the original Welcome screen, so not 100% what @bryphe wanted here, but should probably make the description and action line up at least, can always swap / add what is here later.)

Akin909: LOL I think I only just realised what you meant re. the name changes at @CrossR  I didn't realise the command palette command was opening the quick open 🤣 I thought you meant the names should be different aka that the command palette is actually called the quick open menu, sorry about that 👋

Akin909:
@CrossR yeah was wondering whether or not it would be good to have on by default at this point.

The sessions bit though should stay as experimental atm, its *possibly* got a weird bug where on occasion WIP is disappearing, frankly dont know how thats possible since all we do is call `mksession` and `source ${session}` the latter of which is a very visible command and isnt being called, but somehow wip is still occasionally disappearing although this seems to happen with it disabled either way so not sure if its the cause 🤷‍♂️ 😕

PR #2542, opened 2018-09-01
Feature/convert menu context menu styled components

Akin909:
Based of @Cryza's initial styled components branch with some updates, basically removed references to `this._colors` and using colors in redux since this is now all passed via SC context.


CrossR:
Seems to be that the highlighted chars are just missing. I swapped back to the other master build I had (with the VCS changes) and that works fine, so its something in this branch...

The missing characters are `DetailHighlight`s:

image

Akin909:
I'll have a look a submit a fix asap (maybe tomorrow as its fairly late but will have a quick look)

PR #2551, opened 2018-09-04
OSX: reduce threshold (getting a much more fluid scrolling) 

jaanauati:
I'm honestly not sure how scrolling is behaving for other platforms/operative systems, but it's been working a little oddly these past months for OSX, after some digging in my local I found that having a small value like `10` improves things a lot getting me a more fluid scrolling experience.

Note: I didn't have so much time to poke with this, so I don't know how your building system works, when tested in my local I just made the change to the minified file (`/Applications/Oni.app/Contents/Resources/app/lib/browser/vendor.bundle.js`), so please let me know if i'm missing to push any bundle or something else.

Thanks.

### DIFF
--- a/browser/src/Input/Mouse.ts
+++ b/browser/src/Input/Mouse.ts
@@ -2,7 +2,7 @@ import { EventEmitter } from "events"
 
 import { IScreen } from "./../neovim"
 
-const SCROLL_THRESHOLD_IN_PIXELS = 100
+const SCROLL_THRESHOLD_IN_PIXELS = 10
 
 // TODO
 // Handle modifier keys


CrossR:
I'll give this a test when I'm home on my Windows Surface and Desktop, so we get both trackpad + trackpad events, since they are both treated differently.

Akin909:
@jaanauati sorry for the delay on trying this out been quite busy lately but will give this a go once I get a moment

badosu:
Tested on Arch, looks good to me :+1:

Cryza:
I tested this with both a mouse and a trackpad on Max OS X High Sierra and it worked nicely.

Just a thought on the side: If we want to make this even smoother and more accurate, could we directly use the `fontHeightInPixels` as the scrolling threshold and then, instead of `ScrollWheelDown/Up` keys, send `<C-E>`/`<C-Y>` for scrolling only one line? Could this possibly create issues when users have remapped `<C-E>` or `<C-Y>` to something else?

Maybe we could even decide based on the `_scrollDelta` to send one-line(`<C-E>`/`<C-Y>`), three-line (`<ScrollWheel*>`), half-page (`<C-D>`/`<C-U>`) or full-page (`<C-ScrollWheel*>`) keys?

In any case, what you built so far already seems like a great improvement. Let's merge it once @CrossR was able to verify that it works on Windows :)

CrossR:
Tested this on my Surface and with an MX Master and I didn't notice any difference, so if its helping on Mac, thats great!

PR #1915, opened 2018-03-24
Refactor some .less files to styled-components

Cryza:
Migrate some existing `.less` files to `styled-components`. The only `.less` files that are left are those for `QuickInfo` and `Tabs`, which are by far the most convoluted cases. I might tackle them in a different PR.

PR #2558, opened 2018-09-08
Bugfix/fix tabs focus state border regression

Akin909:
@CrossR seems there was one small-ish regression with the tabs the border top of the inactive tabs is supposed to be transparent so there isnt a resizing flicker this PR fixes it and adds some test cases for the functionality.

I'll avoid using cherry pick commits from #1915 in the future, since it was a WIP and its hard to know what bits were converted or how, makes it hard for me to find the regressions, but think this was it for that branch will create one from scratch for the menu components and later the other bits


Akin909:
Thanks for reviewing 👍

PR #2560, opened 2018-09-09
WebGL Renderer Ligatures

### LINKED ISSUE #2161: Feature request: editor.fontWeight
Hi!

VS-Code added a while back "editor.fontWeight" option. I'm used to run with this setting as "bold". It would be nice if Oni also supposed this feature.

On Linux I usually use "Monaco" , and set it to bold so I can see it from a bit further away. (And I think it renders better as bold on my machines)

Thanks

Cryza:
This introduces the feature to render ligatures when using the WebGL renderer.
It also makes the WebGL renderer use the `editor.fontWeight` setting and dynamically make bold text 300 units heavier based on that setting.

The ligatures were made working with the following strategy, visualized with the example of the `=>` ligature inside the snippet `const noop = () => {}` that many coding fonts provide:
1. We first group all cells inside each row based on their styling. That basically means we retroactively try to find tokens based on cell color and styling (background color, foreground color, bold, italic, underline). This would for example mean that we would create one group for a `const` token, one group for the following, differently colored function name, one for the `=` and `()` each and one for the `=>`. See `groupCells.ts`
2. We take each of these `ICellGroup`s and group the characters inside the group if they will produce a ligature. Simply speaking, we map `["=", ">"]` to `["=>"]` but keep `["c", "o", "n", "s", "t"]` unchanged. That is done like so inside `LigatureGrouper.ts`:
    - Read the `editor.fontFamily` setting and use `font-manager` to find the corresponding actual *font file* on the host system
    - Load the font file using `fs` and parse it using `fontkit`
    - Use `fontkit` to map all the characters to so-called glyphs, which is essentially an index into the list of different versions of all characters that the font can display. Our data for `=>` is now essentially something like `[{ glyphId: 69, contextGroup: A }, { glyphId: 82, contextGroup: B }]`
    - Use our own, adapted version of `fontkit`'s `applyFeatures` to apply ligature substitution on the glyphs. Namely, this means applying the OpenType features `calt`, `rclt`, `liga`, `dlig`, `clig`. See [Wikipedia](https://en.wikipedia.org/wiki/List_of_typographic_features) for a nice explanation of what these are. This changes our glyph data to something like `[{ glyphId: 103, contextGroup: A }, { glyphId: 107, contextGroup: A }]`

> We now have a completely processed array of glyphs containing ligatures and contextual alternatives (which are in fact much more common than real liagtures in coding fonts 😄) that we could in fact directly render to a canvas. We will not do that however because then we would lose important features like auto-hinting and subpixel antialiasing which the chrome text renderer does for us. That's why we now go all the way back again with the information we gathered.
    - Along the way, our adapted replacement algorithm kept track of which groups of glyphs created a replacement by assigning the same `contextGroup` to each of them. Using this information and the stored `codePoints` (essentially the original character(s) that the glyph came from), we can now map the glyphs back to actual string characters and keep them grouped in the right way so that Chrome can also apply ligatures to them. This means we now finally have our "=>" :)
3. Render each of the ligature groups inside each `ICellGroup` individually by sending them to the `GlyphAtlas` as a concatenated string.
4. Let the `TextRenderer` use the `GlyphAtlas` for rendering the actual text just like before, only now the glyphs also might be ligatures that represent several characters.

Since the OpenType processing is a very resource-heavy task, each distinct sequence of characters that has been processed is cached inside the `LigatureGrouper`. That means that each different token will only ever run through that algorithm once. 

Huge shoutout to @devongovett for his great work on `fontkit` and `font-manager` and the fact that he made this possible as easily as it was now! Devon, I hope you're ok with us copy-pasting a lot of the logic from your `OTProcessor` and your `GSUBProcessor`. I didn't open a PR on fontkit because I think that our requirements of tracking the `contextGroup`s are really specific and not relevant to other users of fontkit. If you think otherwise, I'll gladly open a PR!

Fixes #2161


Akin909:
@Cryza this is awesome 😍 🎉, thanks for what looks like a lot of work and the detailed explanation re how you went about implementing it.

Just gave it a try locally and I'm getting:

screen shot 2018-09-09 at 16 01 46

Which relates to this function:

screen shot 2018-09-09 at 16 02 35

I was trying it out using `FiraCode-Retina` and then switched over to `Hasklig` both of which threw the same error, lemme know if I can provide anymore info re. my setup

PS: sorry for almost immediately piling in there with an issue 😆

Akin909:
@Cryza using `Fira Code` works like a charm 😄 ✨ sooo nice 👍, I'm wondering whether the way we specify fonts might be at fault(although we cant really guarantee how user will specify it anyway), not sure how it works on linux and windows but on `MacOS` to find the font name that oni can use I use the `postscript name` there was a related issue a while ago.

Not sure how `font-manager` resolves the name, but it always seemed a bit odd we had to resort to that although not sure how it would work in vscode or atom might be the exact same thing 🤷‍♂️

Cryza:
@Akin909 I made the code for finding the right font file a bit smarter, so it should now also work with `FiraCode-Retina`!

browser/src/Renderer/WebGLRenderer/TextRenderer/LigatureGrouper/OpenTypeLigatureGrouper.ts
```diff
@@ -0,0 +1,112 @@
...
+        // Group GlyphInfo[] by contextGroup
+        const contextGroupDictionary = groupBy(glyphInfos, glyphInfo => glyphInfo.contextGroup)
+        const contextGroupSymbols = Object.getOwnPropertySymbols(contextGroupDictionary)
+        // TODO remove the "as any" once we upgraded to TS 3
+        const contextGroups = contextGroupSymbols.map(
+            contextGroupSymbol => contextGroupDictionary[contextGroupSymbol as any],
+        )
+
+        // Map GlyphInfo[][] to string[]
+        const ligatureGroups = contextGroups.map(glyphsInContextGroup =>
```
devongovett: Why do you need these contextGroups? fontkit already groups glyphs together and stores the grouped codePoints, so you should just be able to create your ligatureGroups based on that right? When would you have multiple glyphs grouped into a single contextGroup?

```javascript
const ligatureGroups = glyphInfos.map(g => String.fromCodePoint(...glyph.codePoints));
```

browser/src/Renderer/WebGLRenderer/TextRenderer/LigatureGrouper/OpenTypeLigatureGrouper.ts
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        const concatenatedCharacters = characters.join("")
+
+        const cachedLigatureGroups = this._cache.get(concatenatedCharacters)
+        if (cachedLigatureGroups) {
+            return cachedLigatureGroups
+        }
+
+        const fontGlyphs = this._font.glyphsForString(concatenatedCharacters)
```
devongovett: Note that fontkit's `layout` function does a lot more than just getting glyphs. See https://github.com/foliojs/fontkit/blob/master/src/layout/LayoutEngine.js. You might want to use that instead of doing it yourself manually.

browser/src/Renderer/WebGLRenderer/TextRenderer/TextRenderer.ts
```diff
@@ -285,43 +285,54 @@ export class WebGlTextRenderer {
...
-            for (let columnIndex = 0; columnIndex < columnCount; columnIndex++) {
-                const cell = getCell(columnIndex, rowIndex)
-                const char = cell.character
-                if (!isWhiteSpace(char)) {
+                const ligatureGroups = this._ligatureGrouper.getLigatureGroups(characters)
+
+                let offsetWithinCellGroup = 0
+                ligatureGroups.forEach(ligatureGroup => {
+                    const columnIndex = startColumnIndex + offsetWithinCellGroup
+                    const x = pixelRatioAdaptedFontWidth * columnIndex
```
devongovett: I'd be careful about this if you want to support non-monospace fonts.

devongovett:
Very cool work guys! One thing you might look at sometime would be a new project we've been building on top of fontkit and font-manager already: [textkit](https://github.com/foliojs/textkit). It's a full text layout engine in JS.

It does font loading and substitution, line wrapping, complex script itemization, and more. It handles run grouping similar to your `groupCells` for rich text support. The output is a container filled with lines of text, each containing runs of glyphs. There is a pluggable renderer architecture to allow rendering to any surface: currently there is a PDF renderer, but more could easily be added. The rest of it is also pluggable so you can swap out the implementation of any part.

It's already being used in [react-pdf](https://github.com/diegomura/react-pdf) for their text layout needs. Aside from that, there isn't much docs yet so you can check out a small example of how the API works here: https://github.com/foliojs/textkit/blob/master/temp.js

browser/src/Renderer/WebGLRenderer/TextRenderer/LigatureGrouper/OpenTypeLigatureGrouper.ts
Cryza: Thanks for the review!

The new `contextGroup` property is there mainly for tracking contextual alternates from `calt` etc. where the glyphs are replaced one by one without combining them. Whenever this happens, we set the context group of all involved glyphs to be the same so that we can track which glyphs are involved in such a replacement.

I realize that the approach that we take here is a bit unconventional, but let me explain the reasoning behind it:

We have a WebGL-based editor renderer that we use for performance reasons. It uses a texture atlas approach which means every glyph is rendered into a texture exactly once and then copied over into the editor view at the required positions using the GPU. Normally, this is done character by character. To support ligatures though, we need to group the characters that will produce ligatures and have chrome render them into the atlas as one string. This is why we need the `contextGroup` approach: we need to map the glyphs back to one string as a group but still keep all other characters that don't produce ligatures separate in their own string.

We decided not to use fontkit's own rendering algorithm because it does not support auto-hinting and also no subpixel antialiasing if I remember correctly. Since both features are very important for creating a crisp and well-readable font (even more so on low-DPI screens), we decided to stick to Chrome's harfbuzz-based renderer.

browser/src/Renderer/WebGLRenderer/TextRenderer/LigatureGrouper/OpenTypeLigatureGrouper.ts
Cryza: I had a thorough look at what all the parts of fontkit do and I tried to replicate and adapt only the really necessary parts. For us, that means only the GSUB replacements were relevant. Other features like kerning, fractional positioning or similar are not relevant in our grid-based editor (It is still using Neovim, after all).

browser/src/Renderer/WebGLRenderer/TextRenderer/TextRenderer.ts
Cryza: Thanks for the heads-up!

I don't think that non-monospace fonts will be relevant though, because we build the whole editor on the basis of Neovim, which is a grid-based editor that fundamentally works by using monospaced cells. If we decided to go away from this, not only will the keyboard navigation between lines potentially become hard to understand but also many existing features and plugins might not make sense anymore. That's why I think it's safe to assume that we can position all characters based on this monospaced grid without worrying about it.

Cryza:
Cool, thanks for the hint about textkit! I would never have found that repo on my own :)

I just had a quick look at it: If I understood it correctly, the PDF renderer takes the glyph paths from the glyph run and renders them to the context  just like they are described, right?
I'm asking so specifically because like stated in my other comment, we put a lot of focus on using all possible optimizations towards crispness and readability of our text. With Chrome's harfbuzz-based renderer, we get auto-hinting and subpixel antialiasing to optimize sharpness. That's the main reason we decided not to render the glyph paths directly to the canvas ourselves.
So my current assessment would be: If we somehow manage to do (auto)-hinting and subpixel-antialiased rendering ourselves with textkit, fontkit, or any other JS font library for that matter, this might be a really interesting option.

browser/src/Renderer/WebGLRenderer/TextRenderer/LigatureGrouper/OpenTypeLigatureGrouper.ts
devongovett: Ah I see: you need to preserve the context in order for Chrome to reproduce the correct glyphs. Yeah fontkit is more meant for layout rather than rendering so its rendering features are somewhat limited. It would be really nice if you could pass glyph ids to the canvas fillText command instead of characters... that would solve this problem and a few others I bet.

Theoretically you could re-encode the font using a different cmap that mapped characters directly to glyphs using the same ids, and then call fillText but using glyph ids as character codes. Not sure how much work that would be.

devongovett:
@Cryza the PDF renderer doesn't render the glyph paths using fontkit, it just places the glyphs in the PDF using native PDF text setting operators. This ensures that the text is selectable, copyable, etc. Unfortunately, as I mentioned above, canvas doesn't have a way to render text by glyph id instead of characters without some hacks, so that would be the main challenge.

browser/src/Renderer/WebGLRenderer/TextRenderer/LigatureGrouper/OpenTypeLigatureGrouper.ts
Cryza: I totally agree, being able to render by glyph ID would save us a lot of trouble :)

The re-encoding of the font file sounds interesting. I'll think about that a little more. Thanks!

Akin909:
It works! 🎉 no console errors, no crashes  ⭐️

Cryza:
Thanks for the thorough testing, @Akin909 ! You were right, for normal font weights the renderer tried to write a broken style string to the canvas which was blocked and resulted in the glyph re-using the previous style instead - which was the bold style for all pre-rendered glyphs. I was able to fix the style string calculation now :)

Also big thanks to @CrossR for testing this on Windows! Apparently the font matching algorithm within `font-manager` works differently on Windows than it does for Linux and Mac OS, which is why it didn't find the correct font file for Fira Code. I replaced the lookup with a platform-independent scan through all available fonts now and it seems to be working fine. :)

PR #2578, opened 2018-09-22
Fixed problem when some visual artifacts may appear after resize

lenenel:
I have menu hidden by default and observed visual artifacts after pressing Alt. Menu appears after I press Alt and editor area is resized, but not the whole area is redrawn, so I see 2 neovim's status bars. I also can reproduce similar behaviour by resizing Oni window.
[See video with this problem](https://youtu.be/3rmkU2MQlk8)

After investigation I've found that method `CanvasRenderer.redrawAll` is called before neovim is really resized so it draws state of neovim before resize. So I've changed code so it called only when Promise returned by `NeovimInstance.resize` method is resolved. Which is actually Promise returned by `Session.request` method, so should be resolved when neovim finished resizing itself.


CrossR:
Sorry for the slow delay! If nobody else has got this, I'll take a look at the Weekend.

PR #2579, opened 2018-09-24
Feature/switch reasonml textmate grammar to sane reasonml grammar

Akin909:
This PR switches the grammar currently being used to the [`sane reasonml grammar`](https://github.com/glennsl/sane-reason-grammar)

The reason being (excerpt from the new grammar's repo)
> This is based on the grammar included in OCaml and Reason IDE, which unfortunately is hand-tuned in a rather haphazard way to be "optimized" for a few of the original author's preferred themes, to the detriment of everything else. This fork attempts to provide a more coherent grammar that properly distinguishes different tokens so that they can be properly selected for highlighting. In effect, this should provide a better default for most themes (except, of course, the select few the original grammar was hand-tuned for) and moves the repsonsibility of optimizing highlighting for specific languages from the grammar to each theme, where it should be to remain maintainable (and quite frankly just plain sane).

Having tested it out locally this seems to provide more regular and consistent highlighting, open to discussion though if any reason users disagree 👍

### DIFF
--- a/extensions/reason/syntaxes/reason.json
+++ b/extensions/reason/syntaxes/reason.json
@@ -33,7 +33,7 @@
                 },
                 {
                     "match": "\\b([[:alpha:]][[:word:]]*)\\b",
-                    "name": "entity.other.attribute-name.css constant.language constant.numeric"
+                    "name": "constant.language"
                 }
             ]
         },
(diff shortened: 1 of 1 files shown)


bryphe:
Sounds good to me, @Akin909 ! Thank you! I had a few issues with the previous highlighting for Reason too 👍

PR #2594, opened 2018-09-28
Add Backer.

CrossR:
I've done this one partially to check the process and partially to get it done quicker.

From what I can tell its just a case of adding their Github account ID to the config file (grabbed via the API call ie https://api.github.com/users/CrossR), and their name. In some cases it looks like we might not have one or the other, but it should just be that. Assuming thats correct @bryphe, I'll stick it on the Wiki somewhere.


bryphe:
Thank you @CrossR for doing this! Yes, it's just the GitHub account ID - when a user signs in, we check either the ID against this file, or Patreon has an API we can go directly against.

It would be nice to have this be a more automatic process at some point!

PR #2603, opened 2018-10-02
Appveyor: remove artifact publish on master

bryphe:
We're hitting an appveyor error:

```
Uploading artifacts...
Maximum allowed artifact storage size of 90000 Mb will be exceeded
```

We actually only need the appveyor artifacts for releasing builds - so I've removed it so we don't hit this limit (and keep getting build errors).

These are still released on our `staging` and `stable` branches, so if we release from there, we'll still get those artifacts we need for release. Also, we'll still always upload to the azure storage for master builds.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,22 +1,22 @@
 # Test against the latest version of this Node.js version
 environment:
-  nodejs_version: "8"
+    nodejs_version: "8"
 
 os: unstable
 
 branches:
     only:
-    - master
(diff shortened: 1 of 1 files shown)


bryphe:
I'm also following up  with AppVeyor support, to see if we can clean out our existing artifacts to unblock PR / release: https://help.appveyor.com/discussions/problems/17235-hitting-upload-limit-for-builds

PR #2605, opened 2018-10-03
Feature/Fix incorrect window positions passed to buffer layers

Akin909:
This PR adds a call to the `remeasure` function to derive the window top and bottom lines as previously this was obtained from the various events and the buffer update function was passing incorrect

### DIFF
--- a/browser/src/neovim/NeovimWindowManager.ts
+++ b/browser/src/neovim/NeovimWindowManager.ts
@@ -50,6 +50,11 @@ export interface NeovimActiveWindowState {
     dimensions: Oni.Shapes.Rectangle
 }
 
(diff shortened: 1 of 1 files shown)


browser/src/neovim/NeovimWindowManager.ts
```diff
@@ -167,31 +172,33 @@ export class NeovimWindowManager extends Utility.Disposable {
         currentWinId: number,
         context: EventContext,
     ): Promise<NeovimActiveWindowState> {
+        // We query the top and bottom line positions again despite these being on the context
```
bryphe: Ah interesting - nice catch! 👍

PR #2606, opened 2018-10-03
Feature/add available plugins section to readme

Akin909:
Whilst plugin discoverability is still fairly manual I discussed with @CrossR re. maintaining a list of user-created plugins this PR adds that section


bryphe:
Great idea @Akin909 and @CrossR !

bryphe:
Looks like AppVeyor was blocked here - bringing it in 👍

PR #2613, opened 2018-10-04
Update for Nvim 0.3.2

CrossR:
This change moves the call to `_checkAndFixIfBlocked` for the latest Nvim version.

This is because the API has changed, such that it does not work any more. (Buffers are not setup until `nvim_ui_attach` is called, which broke the way it worked, and caused us to hang).

Currently, the call is just moved to after the editor is setup. However, in reality we want to remove the hacky code entirely, since it isn't the best way of dealing with it now nvim has given more support. In the current code state, if the code is just straight removed, the messages do not show at all, and the user just has to hit enter blindly, which is exactly what  `_checkAndFixIfBlocked` was added to stop.

If we can show the errors, that would be better. If we can show them in some Oni UI, that would be nicer again. I think that may be possible once we externalise the messages? Not 100% on that.

Do you have any opinion on this @bryphe ? I'm not fully sure why the errors aren't displayed, since https://github.com/neovim/neovim/wiki/Following-HEAD#20180922 seems to imply we should be getting that info now, so its likely that Oni just isn't able to show it yet due to how we init the editors.


browser/src/neovim/NeovimInstance.ts
```diff
@@ -718,7 +739,7 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
 
             this._onMessage.dispatch({
                 severity: "error",
-                title: "Problem loading `init.vim`:",
+                title: "Nvim startup warning:",
```
CrossR: I swapped this as my understanding is that this catches any startup warnings, which could be anything, not just init.vim errors.

browser/src/neovim/NeovimInstance.ts
```diff
@@ -482,27 +497,33 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
...
-                    Performance.endMeasure("NeovimInstance.Start")
-
-                    this._initComplete = true
-                },
-                (err: any) => {
-                    this._onError(err)
-                    this._initComplete = true
-                },
-            )
+            return this._attachUI(version, size.cols, size.rows)
```
Akin909: @CrossR this definitely pre-dates your change but noted a while ago we had a mixture of patterns here i.e. `async/await` and a `promise chain`, was wondering could the attach ui call not be
```ts
try	{
  await this._attachUI(args)
  if (startUpErrorsChecked === false) {
    this._checkAndFixIfBlocked()
  }
  Log.info("Attach success")
  Performance.endMeasure("NeovimInstance.Start.Attach")
  // TODO: #702 - Batch these calls via `nvim_call_atomic`
  // Override completeopt so Oni works correctly with external popupmenu
  // await this.command("set completeopt=longest,menu")
  // set title after attaching listeners so we can get the initial title
  await this.command("set title")
  Performance.endMeasure("NeovimInstance.Start")
  this._initComplete = true
} catch(error) {
 this._onError(err)
 this._initComplete = true
}
```
Its a lot more seemingly linear and more readable (its a minor point stylistic point 😄)

browser/src/neovim/NeovimInstance.ts
```diff
@@ -482,27 +497,33 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
...
+                .then(async () => {
+                    if (startUpErrorsChecked === false) {
+                        this._checkAndFixIfBlocked()
```
Akin909: it seems this call is no longer being `awaited` at @CrossR but the actual function is async could relate to the issues your having

browser/src/neovim/NeovimInstance.ts
CrossR: Makes sense!

browser/src/neovim/NeovimInstance.ts
CrossR: Whoops! I remembered enough to make the function async, then forgot it somehow.

I think the issues with the CI is actually because of me changing the notification text, so pushing up a fix for that now.

Akin909:
@CrossR great you picked up on the change 👍 re. needing to still use check if blocked, from the link you posted it doesn't seem like anything was added to help with the actually handling of `init.vim` errors, seems like we'd still face the same situation since theres no api for those sorts of errors (i.e. an event to inform us of them and an api to use to clear the blockage), although I wonder if the section

> For an UI that wants to do additional initialization after init.vim the following pattern can be used: Before nvim_ui_attach send a singlenvim_command request with the command "autocmd VimEnter * call rpcrequest(1, 'vimenter')". In the vimenter method handler the UI can then safely execute any requests, and nvim will only enter normal mode when this handler returns.

Might be of some use although I'm not sure how it differs from just calling check if blocked (other than it would occur before going into normal mode)

justinmk:
> it doesn't seem like anything was added to help with the actually handling of init.vim errors, seems like we'd still face the same situation since theres no api for those sorts of errors (i.e. an event to inform us of them and an api to use to clear the blockage)

After the UI is attached, the old situation is no longer possible: errors that occur will display like any other errors. That's the purpose of https://github.com/neovim/neovim/wiki/Following-HEAD#20180922 : to allow a UI to attach. Previously `nvim` would block waiting for user-input _before_ the UI could attach.

Akin909:
@justinmk thanks for the explanation 👍, that makes sense

browser/src/neovim/NeovimInstance.ts
```diff
@@ -473,7 +473,22 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
...
-            await this._checkAndFixIfBlocked()
+            const version = await this.getApiVersion()
+            let startUpErrorsChecked = false
+
+            if (
+                version.major > 0 ||
+                version.minor > 3 ||
+                (version.minor === 3 && version.patch >= 2)
+            ) {
+                // We can't call _checkAndFixIfBlocked here, since the buffers etc are not
```
bryphe: Awesome! I would love to get rid of this hacky call. 👍

bryphe:
@CrossR - the changes look good to me! Thanks for taking on this work 👍 

I'm glad to see that hacky code start to go away...

> I'm not fully sure why the errors aren't displayed, since https://github.com/neovim/neovim/wiki/Following-HEAD#20180922 seems to imply we should be getting that info now, so its likely that Oni just isn't able to show it yet due to how we init the editors.

With the latest changes - do we get `init.vim` errors shown correctly? Or are there still issues?

justinmk:
> With the latest changes - do we get init.vim errors shown correctly

The current PR still calls `_checkAndFixIfBlocked()` with Nvim 0.3.2, which defeats the purpose of the aforementioned [new `--embed` behavior](https://github.com/neovim/neovim/wiki/Following-HEAD#20180922).  IIUC `_checkAndFixIfBlocked` tries to cancel out of a "Press Enter" prompt, but that's no longer necessary, and furthermore it will prevent errors from displaying by the normal mechanisms.

Simply remove the call to `_checkAndFixIfBlocked` entirely for Nvim 0.3.2.

browser/src/neovim/NeovimInstance.ts
```diff
@@ -482,27 +497,29 @@ export class NeovimInstance extends EventEmitter implements INeovimInstance {
...
-            )
+            try {
+                await this._attachUI(version, size.cols, size.rows)
+
+                Log.info("Attach success")
+                Performance.endMeasure("NeovimInstance.Start.Attach")
+
+                if (startUpErrorsChecked === false) {
+                    await this._checkAndFixIfBlocked()
+                }
```
justinmk: Not needed for Nvim 0.3.2, because `--embed` (without `--headless`) waits for the UI to attach before showing a "Press Enter" prompt. Therefore `_checkAndFixIfBlocked` is not needed. Any "Press Enter" prompt is normal prompt that should have been displayed to the user by now.

CrossR:
>With the latest changes - do we get init.vim errors shown correctly? Or are there still issues?

With the latest update to nvim (0.3.2) and `_checkAndFixIfBlocked` removed entirely for 0.3.2, the result (for cases of init.vim issues) is as follows:

image

Which is why I had left the call in so far. I assume what is happening is that Oni hasn't initialised the editor in some fashion, or the errors are not being captured correctly on our end etc. Once I hit enter, the editor loads as expected. With the call left in, the behaviour is the same as the latest release and nvim 0.3.1.

I would prefer to remove the call entirely, so do you have any idea what would be waiting/causing the errors to not being shown?

justinmk:
> I assume what is happening is that Oni hasn't initialised the editor in some fashion, or the errors are not being captured correctly on our end etc

It must be the former, because, unless Oni usually "captures" errors during normal operation, there should be no need to capture anything on startup.

@CrossR in your screenshot I don't see anything that looks like "Oni UI started". Oni should draw the UI before doing anything else.

> Once I hit enter, the editor loads as expected.

That means that a prompt is waiting but wasn't displayed. Did Oni receive a `redraw` UI event with the screen contents? If so, it should draw the screen.

Issue #2612, opened 2018-10-04
Indent lines incorrectly aligned whilst searching in file


Akin909:
Currently when a user searches in buffer using the command line the user position in the file can change if the match is currently outside of the visible range, in this case the visible line passed to the `BufferLayerRenderContext` are not updated because the show commandline mode setting does not trigger a recalculation of the visible lines

PR #2614, opened 2018-10-04
Feature/remeasure window on commandline hide and show

Akin909:
This relates to #2612, the commandline can cause the window/viewport to be repositioned if `incsearch` is enabled and so the window will shift to the next available match if this is off screen the window will reposition but this isn't passed on to the buffer layer render context so things like the indent lines remain incorrectly placed.

This change causes the window to remeasure when this changes (the observable is essentially already throttled using `distinctUntil` so don't think this should cause a performance issue as it will remeasure if the context changes i.e. if a reposition occurred.

Unfortunately there isn't a more specific event to hook into if the window moves

### DIFF
--- a/browser/src/neovim/NeovimWindowManager.ts
+++ b/browser/src/neovim/NeovimWindowManager.ts
@@ -75,6 +75,12 @@ export class NeovimWindowManager extends Utility.Disposable {
         this._scrollObservable = new Subject<EventContext>()
 
         const updateScroll = (evt: EventContext) => this._scrollObservable.next(evt)
+
+        const handleCommandLineEvent = async () => {
+            const ctx = await this._neovimInstance.getContext()
+            updateScroll(ctx)
+        }
+
         // First element of the BufEnter event is the current buffer
         this.trackDisposable(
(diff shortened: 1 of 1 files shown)


browser/src/neovim/NeovimWindowManager.ts
```diff
@@ -172,12 +184,17 @@ export class NeovimWindowManager extends Utility.Disposable {
...
+            ["nvim_call_function", ["line", ["w0"]]],
+            ["nvim_call_function", ["line", ["w$"]]],
+        ]
         // We query the top and bottom line positions again despite these being on the context
         // as the values from the `BufferUpdate` event can be incorrect
-        const [topLine, bottomLine] = await Promise.all([
-            this._neovimInstance.callFunction("line", ["w0"]),
-            this._neovimInstance.callFunction("line", ["w$"]),
-        ])
+        const [[topLine, bottomLine]] = await this._neovimInstance.request<[[number, number], any]>(
```
Akin909: The logic is the same I just swapped this to use the atomic calls api so it works the same way as the call below (also I'm not sure if the atomic calls might be more performant vs. promise.all)

PR #2617, opened 2018-10-06
Feature/add workerize utility

Akin909:
cc @bryphe

I've been experimenting with worker support still, and discovered a very minimal way to execute a *pure* function in a separate thread, this is similar to the `greenlet` library just even smaller.

It takes a function, and its arguments and runs the code in a separate thread, unfortunately I can't find any expensive *pure* functions to use it on.

I'm working on the syntaxHighlightingPeriodicJob on another branch, unfortunately it has quite a few external dependencies, currently blocked by the fact that `vscode-textmate` references the process variable which is undefined in the worker thread so causes errors

### DIFF
--- a/browser/src/Workerize.ts
+++ b/browser/src/Workerize.ts
@@ -2,10 +2,6 @@ interface IArgs<T> {
     data: T
 }
 
-interface IContext {
-    [key: string]: any
-}
-
 type Callback<T> = (args?: T) => void
 
 /**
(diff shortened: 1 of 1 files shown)


bryphe:
Thanks for introducing this, @Akin909 !

> I'm working on the syntaxHighlightingPeriodicJob on another branch, unfortunately it has quite a few external dependencies, currently blocked by the fact that vscode-textmate references the process variable which is undefined in the worker thread so causes errors

That's a bummer. I wonder why they need process? Potentially, we could fork `vscode-textmate` and pass in whatever context it needs as part of the our 'pure function' we run in the worker, as opposed to it needing to query the `process` global.

Would it be possible to add a quick test case demonstrating its usage?

Even just a simple test case like:

let result = await workerize(({a, b}) => a + b, {a: 2, b: 3});
assert.strictEqual(result, 5);

(Essentially its documentation on how this can be used 😄 )

browser/src/Workerize.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @returns Promise<R>
+ */
+export function workerize<T, R>(work: Callback<T>, args?: T) {
+    const handleResult = ({ data }: IArgs<T>) => {
+        const result = work(data)
+        const webWorker: Worker = self as any
+        return webWorker.postMessage(result)
+    }
+
+    // courtesy of this gem -
```
bryphe: Nice find!

PR #2618, opened 2018-10-07
Feature/upgrade typescript

Akin909:
Upgrade typescript to v3.1 & tslint (was dependent on some deprecated methods of the old version of ts) also fix associated type errors

Next up Im gonna try to use the project references feature of ts 3.0 to see if building all the various oni sub-projects can be simplified


bryphe:
Awesome! Thank you @Akin909 !

Issue #1016, opened 2017-11-25
Integrated a new c/c++ language server: cquery


saibing:
I'm try a new c/c++ language server: [cquery](https://github.com/jacobdufault/cquery)

CQuery is a Low-latency language server for large C++ code-bases, powered by libclang. It is much more powerful than clangd.

But I ran into some trouble that cquery need add some initializationOptions:

```json
{
        "initializationOptions": {
                "cacheDirectory": "/tmp/cquery",
		"resourceDirectory": "yourCQueryDirectory/clang_resource_dir"
        }
}
```
[How to use cquery with LanguageClient-neovim](https://github.com/autozimu/LanguageClient-neovim/wiki/cquery)

May be Oni need to support config language server's initializationOptions?

bryphe:
Thanks for logging the issue, @saibing !

Yes, unfortunately our configuration doesn't support this currently... But it would make sense to have something like this, ie:
```
"language.cpp.languageServer.initializationOptions": {
                 "cacheDirectory": "/tmp/cquery",
		"resourceDirectory": "yourCQueryDirectory/clang_resource_dir"
}
```

Will use this to track setting the `initializationOptions`.

bryphe:
A good place to start looking is here:
https://github.com/onivim/oni/blob/45f51c59b40dab399f96f2476e9040ee75ee725e/browser/src/Services/Language/LanguageConfiguration.ts#L99

Which is the code responsible for taking the configuration settings and handing them off to the language client.

MaskRay:
See https://github.com/autozimu/LanguageClient-neovim/wiki/cquery 
`"resourceDirectory"` is auto detected and can be removed now.

MaskRay:
I'd like to ask how you would like to detect the project root. When there are subpackages, the root  `compile_commands.json` contains entries for subproject files and you do not want files in subprojects to be associated with subprojects, but rather the root project.

```
proj
 compile_commands.json
 .git
 third_party
  proj0
   .git
   a.cc       // you want this file to be associated with the root project
  proj1
   .git
```

Sorry for pasting the link of the Emacs lsp-mode discussion thread. But they have the same issue
https://github.com/emacs-lsp/lsp-mode/issues/293 and I think it is beneficial to draw on the wisdom of the masses

saibing:
@MaskRay 
Thank you very much

Now I can solve this problem by add follow config into config.tsx
```ts
[...]

PR #2624, opened 2018-10-09
Only send `completionItem/resolve` if supported

feltech:
* Oni will hang indefinitely if the language server simply chooses to
ignore unsupported requests.
* This can be avoided if the server provides correct capabilities and
we actually check them before making the request.


feltech:
This addresses one of the shortcomings I mentioned in https://github.com/onivim/oni/issues/1016#issuecomment-428380453

feltech:
There are still tests to write, so this is still WIP at the moment.

feltech:
Also note to self (and whoever else cares): update [wiki page](https://github.com/onivim/oni/wiki/Language-Support#cc) to remove warning about completion failing in cquery/ccls once this is merged.

feltech:
OK, added tests.  

I took the liberty of upgrading sinon (spys, stub, etc library).  I wasn't quite sure what to do about `yarn.lock`, since it seemed to include more than just sinon version changes, so I left that out...

PR #2632, opened 2018-10-14
 Highlight first item in menus upon filtering

feltech:
If the user moves through the menu then resumes typing, the
`selectedIndex` can end up larger than the contents of the menu.  This
can cause errors, or at least graphical glitches, and seems odd.

### DIFF
--- a/browser/src/Services/ContextMenu/ContextMenu.tsx
+++ b/browser/src/Services/ContextMenu/ContextMenu.tsx
@@ -160,10 +160,6 @@ export class ContextMenu {
         )
     }
 
-    public updateItem(item: any): void {
-        this._actions.setDetailedMenuItem(item)
-    }
-
     public hide(): void {
         this._actions.hidePopupMenu()
     }
--- a/browser/src/Services/Menu/MenuActionCreators.ts
+++ b/browser/src/Services/Menu/MenuActionCreators.ts
@@ -66,13 +66,6 @@ export const setMenuItems = (id: string, items: any[]) => ({
     },
 })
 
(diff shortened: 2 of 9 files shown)


feltech:
Unfiltered
Unfiltered

Then type to filter:
Filtered

You can see no option is highlighted, and the following exception occurs
```
vendor.bundle.js:925 Uncaught TypeError: Cannot read property 'rawCompletion' of undefined
    at EventEmitter.Ce._contextMenu.onSelectedItemChanged.subscribe.e (vendor.bundle.js:925)
    at emitOne (events.js:116)
    at EventEmitter.emit (events.js:211)
    at t.Event.dispatch (vendor.bundle.js:632)
    at Object.onSelectedItemChanged (vendor.bundle.js:987)
    at r (vendor.bundle.js:839)
    at vendor.bundle.js:839
    at vendor.bundle.js:824
    at n (vendor.bundle.js:632)
    at Object.filterMenu (vendor.bundle.js:632)
```

With the fix
Fixed

Akin909:
@feltech I noticed that this PR removes the details code, I'm wondering if this impacts rendering of the documentation section for the autocomplete menu?

feltech:
I believe the code responsible for processing a `completionItem/resolve`, which must be where the documentation comes from, was the bit I fixed in #2633.  I cannot find any references to the string `SET_DETAILED_MENU_ITEM` and the related methods that I removed.  Grepping the git history, it looks like the final reference to the top-level function `updateItem` was removed in e2dc9bcbfc11b2d73061deaada865fda26bed9ef as part of a refactor.

PR #2633, opened 2018-10-14
Fix updating detailed completion for overloaded methods

feltech:
* For languages that allow method overloads, where the same method name
can be defined with different arguments (e.g. C++), the `label` of a
completion may be duplicated in the completions list, but the `detail`
should be different.
* Hence use the `detail` field, if available, to disambiguate between
completion items when updating them with the result of a
`completionItem/resolve`.


feltech:
An example from cquery (C++) completions

Before:
oni_completion_detail_1
oni_completion_detail_2

With the fix the second "duplicate" item shows it's different signature:
oni_completion_detail_fixed

PR #2634, opened 2018-10-14
Retry symbol search whilst language server is still initialising

feltech:
* Langauge servers can take a while to initialise, and this has a
specific error code associated with it when querying.
* So detect that specific error response and retry symbol search every
second until it works, or until the search menu is closed.

### DIFF
--- a/browser/test/Editor/NeovimEditor/SymbolsTests.ts
+++ b/browser/test/Editor/NeovimEditor/SymbolsTests.ts
@@ -149,7 +149,9 @@ describe("Symbols", () => {
             languageManager.sendLanguageServerRequest.onCall(2).returns("mock result")
 
             // action
-            const request: Promise<any> = symbols["_requestSymbols"](buffer, "mock command", menu)
+            const request: Promise<any> = symbols["_requestSymbols"](buffer, "mock command", menu, {
+                mock: "option",
+            })
 
             // confirm
             sinon.assert.callCount(languageManager.sendLanguageServerRequest, 1)
@@ -167,6 +169,13 @@ describe("Symbols", () => {
             sinon.assert.callCount(languageManager.sendLanguageServerRequest, 3)
             clock.tick(1000)
             await waitForPromiseResolution()
+            sinon.assert.alwaysCalledWith(
+                languageManager.sendLanguageServerRequest,
+                "mocklang",
+                "/mock/path",
+                "mock command",
+                { mock: "option", textDocument: { uri: "file:///mock/path" } },
+            )
             assert.equal(await request, "mock result")
         })
 
--- a/browser/test/Editor/NeovimEditor/SymbolsTests.ts
+++ b/browser/test/Editor/NeovimEditor/SymbolsTests.ts
@@ -7,6 +7,7 @@ import { ErrorCodes } from "vscode-jsonrpc/lib/messages"
 
 import { Definition } from "../../../src/Editor/NeovimEditor/Definition"
 import { Symbols } from "../../../src/Editor/NeovimEditor/Symbols"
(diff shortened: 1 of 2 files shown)


feltech:
Specifically, I found this when trying to use the cquery (C++) language server.  Opening a new file then `<C-S-t>` too quickly would lead to a search box with a loading spinner that never stops loading.  This was caused by an exception like
```
Error: /path/to/file.cpp is being indexed.
    at new ResponseError (/mnt/2TB-Storage/Working/Programs/oni/node_modules/vscode-jsonrpc/lib/messages.js:46)
    at handleResponse (/mnt/2TB-Storage/Working/Programs/oni/node_modules/vscode-jsonrpc/lib/main.js:430)
    at processMessageQueue (/mnt/2TB-Storage/Working/Programs/oni/node_modules/vscode-jsonrpc/lib/main.js:258)
    at Immediate._onImmediate (/mnt/2TB-Storage/Working/Programs/oni/node_modules/vscode-jsonrpc/lib/main.js:242)
    at runCallback (timers.js:789)
    at tryOnImmediate (timers.js:751)
    at processImmediate [as _immediateCallback] (timers.js:722)
```
which I traced to a specific error code coming from the language server.

feltech:
Hmm, those "WARNING: head commit changed" commits are annoying, but harmless (they're empty).  I think I accidentally tried to amend a commit I'd already pushed.  Not sure how those messages got put there, I've not seen that before.  Something to do with eclipse (which I use as a history/conflict viewer), probably.

browser/test/Editor/NeovimEditor/SymbolsTests.ts
```diff
@@ -0,0 +1,206 @@
...
+    describe("_requestSymbols", () => {
+        let menu: any
+        let buffer: any
+
+        beforeEach(() => {
+            menu = sinon.createStubInstance(Menu)
+            menu.isOpen.returns(true)
+            buffer = sinon.stub()
+            buffer.language = "mocklang"
+            buffer.filePath = "/mock/path"
```
Akin909: @feltech since our CI runs on multiple platforms you might need to use `path.join` here to make the mock paths platform agnostic

Akin909:
@feltech thanks for the PR its nice to see the language server get some much needed love and tests as well 😍, re. the retry strategy I'm wondering if its possible to also delay queries till the server is initialised. There is an [`intialize request`](https://microsoft.github.io/language-server-protocol/specification#initialize) sent by the client in the spec and possibly we can await its result before taking other requests?

feltech:
Unfortunately, at least for cquery, the `initialize` request seems inadequate.  Oni logs "[LanguageClientManager]: Initialized" after the `initialize` request completes, but we still see the exception after that (see pic of debug console)

oni_language_client_initialize_not_ready

Akin909:
@feltech based on the protocol it seems you have to wait for the `InitializeResult` to be sent back from the lsp, does the `initialize` call return with the InitializeResult or is that sent back separately? I think based on [`this example`](https://github.com/Microsoft/vscode-languageserver-node/blob/master/server/src/main.ts) there might be separate notification event from the lsp when the initialization is complete

Akin909:
Btw don't let the initialisation issue be a blocker for now, I'm planning to have a look at bits of our lsp client at some point to facilitate things like #1717

feltech:
Good stuff!  I have a couple more PRs incoming, so I'll perhaps take a deeper looks at language server initialize later.

CrossR:
Can this be merged now then? (I assume it was waiting on some checks when you approved it @Akin909)

Akin909:
@CrossR was good to merge just forgot to hit the button

PR #2638, opened 2018-10-16
De-deduplicate completion items + improve sorting

feltech:
* Remove duplicate completion items.  This seems to happen a lot with `cquery` (C++) completions.
* Continue to sort abbreviations below full matches.
* Sort matching `filterText`/`labels` so they appear next to each other.
* Otherwise use `sortText` given by the language server.

### DIFF
--- a/browser/src/Services/Completion/CompletionSelectors.ts
+++ b/browser/src/Services/Completion/CompletionSelectors.ts
@@ -3,6 +3,7 @@
  *
  * Selectors are functions that take a state and derive a value from it.
  */
+import * as _ from "lodash"
 
 import { ICompletionState } from "./CompletionState"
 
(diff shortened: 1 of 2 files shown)


feltech:
The main motivation for this PR was to remove the duplicates coming back from `cquery`.  

However, I found that the sort order of results was odd (which makes filtering duplicates more expensive, and just looks strange).  Perhaps they make some sense straight from the language server, but Oni then filters as the user types, so retaining the sort order from the language server makes even less sense.  

Hence why I thought it reasonable to "fix" the sort order for filtered completions.

browser/src/Services/Completion/CompletionSelectors.ts
```diff
@@ -3,6 +3,7 @@
  *
  * Selectors are functions that take a state and derive a value from it.
  */
+import * as _ from "lodash"
```
Akin909: @feltech if you import just the specific functions you need rather than `_` from lodash this will help reduce our bundle size which is something we should aim for but probably don't get right in many situations

e.g.
```ts
import uniq from "lodash/uniq"
// or if type issues
import * as omit from "lodash/omit"
```

PR #2640, opened 2018-10-18
Bugfix/Fix dynamic import for shell env

Akin909:
The `shell env` package has been silently failing meaning that fixes that allow oni to read in a user's env var have regressed. This happened when we upgraded `webpack` because with the new version of webpack, dynamic imports now return objects with the shape `{ default: DefaultFnExport, aVar: var }`

Also upgraded the `shellenv` package as is since been upgraded.

### DIFF
--- a/browser/src/Plugins/Api/Process.ts
+++ b/browser/src/Plugins/Api/Process.ts
@@ -28,18 +28,18 @@ export class ShellEnvironmentFetcher implements IShellEnvironmentFetcher {
     public async getEnvironmentVariables(): Promise<NodeJS.ProcessEnv> {
         if (!this._shellEnv) {
             this._shellEnv = await this._shellEnvPromise
-            try {
-                const env = this._shellEnv.default.sync()
-                return env
-            } catch (error) {
-                Log.warn(
-                    `[Oni environment fetcher]: unable to get enviroment variables because: ${
-                        error.message
-                    }`,
-                )
-            }
         }
-        return this._shellEnv.default.sync()
+        try {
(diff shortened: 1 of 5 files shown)


Akin909:
Whilst the current change restores previous env reading functionality an issue which it raises (this was raised as a separate issue I currently can't find) is that if a user has a problematic script or command in their zsh or bashrc etc. then oni's initialisation will hang. As an example I previously used `oh-my-zsh` and had no issues sourcing my zshrc to read in my env vars, however since switching to `zplug` which adds a command to source some scripts in the rc file, booting oni now blocks.

This may have something to do with running the shell interactively or not, its not clear what the exact issue is, I think an interim solution is raising the issue with users that complex scripts might block the process and should have a guard around them (not sure what form this would take, would be specific to the users rc file) or perhaps forking shellenv or raising a PR/issue to have it not block in those cases

Akin909:
Following further investigation seems the issue only occurs as a result of [very unusual](https://github.com/zplug/zplug/issues/209) `rc` file settings in my case a very obscure bug in zplug, a work around I've added in as this issue also occurred (at another point, though I cant find it 💢) is a config option called `oni.userShell` this allows a user to have oni derive the `$PATH` and other shell variables from a different shell if another is causing an issue. Tbh this is more of a workaround that allows oni to run but I think this issue is v. v. unlikely to hit users and if it does is usually I think the result of an offending line in the `zshrc` or `bashrc` which means pulling vars out of it doesnt work which shouldnt be the case for a normal functioning rc files (history may prove me wrong 😆 )

PR #2644, opened 2018-10-22
Feature/fix buffer opening closing

### LINKED ISSUE #2639: Tabs no longer close correctly in buffer mode on latest
**Oni Version:** Built from latest commit on master,  commit hash 2ae1224adf0659003355a62582641f8b45d37399

**Neovim Version (Linux only):** 0.2.2.
**Operating System:** Ubuntu 18.04

**Issue:** `:close`, `:quit` no longer (visually) quit tabs. Instead the tab's top bar still remains but am unable  to switch to it with `gt/gT`. 

**Expected behavior:** The relevant tab disappears

**Actual behavior:** It remains

**Steps to reproduce:** This could also be related to my tabs mode, which is currently set to 'buffer'.

Akin909:
Fixes #2639, the issue here was that during the refactor of the quick open menu we now check against the `Oni.FileOpenMode` enum for the *legacy* setting `editor.quickOpen.fileOpenMode` which is set to `Edit` which in the enum is represented by `O` so the legacy check `if(legacy //which is 0)` coerces to false and this fails.

The fn then looks for the new setting `quickOpen.fileOpenMode` which isn't set in the default config so that fails and the default value it is passed is `Oni.FileOpenMode.NewTab` so it calls `tabnew!`

this PR fixes the legacy check so it works and also adds a default config for the new command and sets it to edit by default

### DIFF
--- a/browser/src/Services/Configuration/DefaultConfiguration.ts
+++ b/browser/src/Services/Configuration/DefaultConfiguration.ts
@@ -147,6 +147,8 @@ const BaseConfiguration: IConfigurationValues = {
     "editor.quickOpen.alternativeOpenMode": Oni.FileOpenMode.ExistingTab,
     "editor.quickOpen.showHidden": true,
 
+    "quickOpen.defaultOpenMode": Oni.FileOpenMode.Edit,
+
     "editor.split.mode": "native",
 
     "editor.typingPrediction": true,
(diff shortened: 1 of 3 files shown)


Akin909:
@CrossR when you get a chance can you have a look at this change I think its fairly minor and fixes the buffers mode opening/closing issue

Issue #2646, opened 2018-10-23
Unable to use LSP rename functionality with clangd


Siprj:
**Oni Version:** latest master b9a29925f88036a3eb089f61f48d30a220136948
**Neovim Version (Linux only):** v0.3.1
**Operating System:** Arch

**Issue:** Unable to use LSP rename functionality with `clangd`.

Rename always fails with error in development console.

**Expected behavior:** Variable is renamed.

**Actual behavior:** New empty file (with name `undefined`) is opened. And target of the rename is not renamed.

**Steps to reproduce:** Use clangd as LSP server for C++ and try to rename variable.

PR #2647, opened 2018-10-23
editsToUse are deeper in LSP WorkspaceEdit response

Siprj:
Edits are actually stored in `changes` field in WorkspaceEdit; see:
https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#workspaceedit

Resolves: #2646

### DIFF
--- a/browser/src/Services/Workspace/Workspace.ts
+++ b/browser/src/Services/Workspace/Workspace.ts
@@ -71,10 +71,9 @@ export class Workspace implements Oni.Workspace.Api {
     }
 
     public async applyEdits(edits: types.WorkspaceEdit): Promise<void> {
-        let editsToUse = edits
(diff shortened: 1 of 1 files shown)


Akin909:
@Siprj thanks for looking into this and raising a PR, I noticed this a part of #1717 	but got bogged down with other changes, I wonder re. the change if you could swap the variable to work as follows:

```suggestion
 const editsToUse = edits.documentChanges
            ? convertTextDocumentChangesToFileMap(edits.documentChanges)
            : edits.changes
```
It's a small change but the re-assignment using `let` is a little more verbose and this was the change I was gonna (eventually) propose in #1717 (which would help me out as it would reduce conflicts there

Siprj:
@Akin909 Done.

Issue #2584, opened 2018-09-26
Oni wont't open when passed a number as first parameter


Gahen:
**Oni Version:** 0.3.6
**Operating System:** macOS Sierra

**Issue:** Running oni from the command line with a number as first parameter won't open the editor

**Expected behavior:** I should see the filesystem tool and the editor area

**Actual behavior:** I only see the system menu bar and the window UX

**Steps to reproduce:** open a terminal; run `oni 123`

Maybe this is only happening to me? I would be available to follow any debug step needed.

This is what I get: 

![screen shot 2018-09-26 at 11 42 05 am](https://user-images.githubusercontent.com/1663320/46101663-60932c80-c181-11e8-9543-df1253d1b4ca.jpg)

CrossR:
Hey, sorry for the slow reply!

This looks to be that the path checking logic we have fails in some fashion for just numbers:

![image](https://user-images.githubusercontent.com/10038688/46164012-bef50380-c284-11e8-90bb-750b8e592d2f.png)

I think this is type of the input being swapped to an int when just numbers are passed. Any other string seems to work (`123.txt` etc). The fix here _should_ just be to force it always to a string, as well as not crashing if there is an error there.

The lines in question are these:

https://github.com/onivim/oni/blob/e3f65e694828f025186e3672f62d23a92f07833a/browser/src/App.ts#L110-L112

We can either swap the type just in there, or just update `minimist` to always treat everything as a string (except stuff like the debug flag etc).

I'll tag this as a good first issue since it seems easy enough, though we should try and get this sorted before the next release.

PR #2649, opened 2018-10-25
fix problem when Oni start with number passed as first parameter (#2584)

borjarus:
* add force to treat args always as string

### DIFF
--- a/browser/src/App.ts
+++ b/browser/src/App.ts
@@ -105,7 +105,7 @@ export const start = async (args: string[]): Promise<void> => {
     const cssPromise = import("./CSS")
     const completionProvidersPromise = import("./Services/Completion/CompletionProviders")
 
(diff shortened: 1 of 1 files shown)


Akin909:
Thanks for the PR @borjarus sorry for the delay looking at it (started a new job so been mega busy) this looks good :+1:

browser/src/App.ts
```diff
@@ -105,7 +105,7 @@ export const start = async (args: string[]): Promise<void> => {
     const cssPromise = import("./CSS")
     const completionProvidersPromise = import("./Services/Completion/CompletionProviders")
 
-    const parsedArgs = minimist(args)
+    const parsedArgs = minimist(args, { string: "_" })
```
Akin909: just to confirm the `_` forces minimist to treat all args as strings

PR #2650, opened 2018-10-26
React/Redux DevTools no longer working

borjarus:
add new dev dependency to solve problem

I think latest commit of 'electron-devtools-installer' solve problem. Here is link to [issue](https://github.com/MarshallOfSound/electron-devtools-installer/pull/88)


bryphe:
👍 , thank you very much for the contribution @borjarus ! And thanks @Akin909 for reviewing! Bringing it in now.

PR #2679, opened 2018-12-14
Fix "upath@1.0.4": The engine "node" is incompaible with this module.

### LINKED ISSUE #2430: yarn install fails with node v10.6.0
**Oni Version:** Master
**Neovim Version (Linux only):** 0.3.0
**Operating System:** Arch Linux

**Issue:**
I basically had this issue: https://github.com/facebook/create-react-app/issues/4367
running `yarn --ignore-engines install` did the trick but it might be worth updating the upath version since 
it seems the latest version doesn't have this problem.

**Steps to reproduce:**
Simply follow what the wiki says to build Oni with node v10.6.0

samvv:
This is a very small fix that prevents yarn from aborting the installation on
newer versions of NodeJS. It is merely a temporary fix; the real fix should be
to upgrade all of the dependencies.

Closes #2430.

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,0 +1,1 @@
+--add.ignore-engines true


bryphe:
Thanks for cleaning this up, @samvv ! Looks good to me 👍

bryphe:
> It is merely a temporary fix; the real fix should be to upgrade all of the dependencies.

Definitely!

PR #2683, opened 2018-12-15
Fix unable to open a new file via CLI

samvv:
See commit message for more information.


bryphe:
Thanks for the fix, @samvv ! I think this fixes an issue I have occasionally - I'll run `oni someNewFile.md` and it won't actually be created. Looks good to me! 👍

PR #2694, opened 2018-12-18
[WIP] Fix #2690 - Allow extra args to pass directly to Neovim

### LINKED ISSUE #2690: Support starting Oni with session files from command line
**Oni Version:**
0.3.10
**Neovim Version (Linux only):**
**Operating System:**
Windows
**Issue:**
The -S <session file> command line option doesn't seem to be supported.  Session files can be loaded after Oni starts, using the :so <session file> command, however...

**Expected behavior:**
Oni will open the specified session file and restore the buffers and view states that is typical of a normal Vim installation.

**Actual behavior:**
Running Oni with -S <session file> opens Oni with blank buffer

**Steps to reproduce:**

bryphe:
This is a fix proposed by @CrossR for #2690 - we can pass any additional arguments after a `--` directly to Neovim. For example, `oni some-file.js -- -S my-session.vim`

Still needs to be tested / validated.

### DIFF
--- a/browser/src/App.ts
+++ b/browser/src/App.ts
@@ -105,11 +105,13 @@ export const start = async (args: string[]): Promise<void> => {
     const cssPromise = import("./CSS")
     const completionProvidersPromise = import("./Services/Completion/CompletionProviders")
 
-    const parsedArgs = minimist(args, { string: "_" })
+    const parsedArgs = minimist(args, { "--": true, string: "_" })
     const currentWorkingDirectory = process.cwd()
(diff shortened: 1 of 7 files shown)


justinmk:
For what it's worth, my thoughts on this are at https://github.com/equalsraf/neovim-qt/issues/449 . 

Unless the UI plans to introduce a bunch of CLI options that potentially conflict with `nvim` options, it would be better to just forward all unrecognized options to `nvim`, rather than using this `--` workaround.

Issue #2688, opened 2018-12-17
Interactive tutorial broken in version 0.3.7 and 0.3.8


kifirkin:
**Oni Version:**
0.3.8
**Neovim Version (Linux only):**
0.3.1
**Operating System:**
Ubuntu 18.04 x64

**Issue:**
Interactive tutorial doesn't work. Unable to complete first task int first tutorial.

**Expected behavior:**
Interactive tutorial inline window goes into insert mode and input text goes to that window

**Actual behavior:**
When I press `i` key to switch into insert mode inside interactive tutorial inline window, whole Tutorial tab goes into insert mode and input text goes somewhere in background

**Steps to reproduce:**
Open first task in first tutorial, press `i`. Whole tab goes into insert mode instead of inline window

alexmironof:
Version 0.3.9 has the same issue. seems like inline windows cant get focus

alexmironof:
I've found place where this bug was introduced https://github.com/onivim/oni/commit/e612eb860224d9f36528501cfd7c3a221e635305#r31743213

gtlaserbeast:
I seem to be experiencing this same issue with the mac version. v0.3.9

Funeoz:
Same here on Kubuntu 18.10 : v 0.3.9

alexmironof:
I've tried to debug this thing a little bit: So I set `autoFocus = true` here https://github.com/onivim/oni/blob/c7b11e462cc97f56fce313106570e5bb1e02eee5/browser/src/Services/Learning/Tutorial/TutorialBufferLayer.tsx#L126 , but it seems that it helps once, if I open another tutorial, inline editor looses focus again.

Also after completing tutorial there are two problems:
1. Keystroke counter doesn't work
2. Completion layer doesn't handle input keys, so I can't restart with space or continue with enter key

danielo515:
I am experiencing this same problem in 0.3.9 for macos. It is a shame because I would love to complete all the tutorials

carderne:
Same here in 0.3.9 for Linux (Debian/Ubuntu).

yzlnew:
Having this issue on Windows.
Oni 0.3.9 / NeoVim 0.3.1

thefrontendwizard:
I'm experiencing this issue on 0.3.9 with neovim 0.3.4

goriunov:
Same :(

slipsnip:
Same issue on Oni 0.3.9 and Neovim 0.3.4

[...]

PR #2705, opened 2019-01-02
Fix typo in plugin name targets.vim

wellle:
Hey there!

As the author of targets.vim I'm honoured and want to thank you for including targets.vim as a default plugin! That really means a lot to me! ❤️

I just tried to check it out, but ran into the same issue reported in https://github.com/onivim/oni/issues/2688.

But I also noticed that the plugin was referred to `targets.vim` as `target.vim` in two places. This PR should fix that.

Thanks again, and good luck for the Oni project!


bryphe:
Hey @wellle ! 

Thanks a lot for the PR (apologies for the typo!) and the kind words! 

Big fan of `targets.vim` - one of my personal favorite must-have plugins, and feels like it should be in Vim by default 😄 

Looks like the build failure was an intermittent failure on the Windows machine - will bring it in now. Still need to investigate #2688

danielo515:
Please notify me when you make a release including this fix.

PR #2729, opened 2019-02-28
Fix upath conflict

TalAmuyal:
Had an error in Webpack when running `yarn run install`.
Copied solution from https://github.com/airbnb/enzyme/issues/1637.


bryphe:
Thanks for fixing this, @TalAmuyal !

PR #2757, opened 2019-09-07
Remove text about ongoing development

corollari:
Minor edit


bryphe:
Indeed, all our development focus is on [Onivim 2](https://github.com/onivim/oni2) now! Thank you for the edit, @corollari .

